PR #10, opened 2021-04-15
internal/apimodule/account: fix multiple imports

alrs:
This fixes two tests with multiple imports of `"github.com/superseriousbusiness/oauth2/v4/models"`.


tsmethurst:
Heya! Thanks for the PR! For now I'm going to just close it though, because I'm changing the whole test rig anyway (take a look at the branch I'm currently working on), in a way that will obviate making these changes, since tokens and such will be generated from the testrig so the imports will work in a different way :)

PR #11, opened 2021-04-19
Api/v1/statuses

tsmethurst:
This PR adds:

### Statuses

* New status creation.
* View existing status
* Delete a status
* Fave a status
* Unfave a status
* See who's faved a status

### Media

* Upload media attachment and store/retrieve it
* Upload custom emoji and store/retrieve it

### Fileserver

* Serve files from storage

### Testing

* Test models, testrig -- run a GTS test instance and play around with it.


tsmethurst:
Merging this in already because the PR is become huge.

PR #27, opened 2021-05-21
Follows and relationships

tsmethurst:


tsmethurst:
* Follows -- create and undo, both remote and local
* Statuses -- federate new posts, including media, attachments, CWs and image descriptions.

Issue #32, opened 2021-05-27
[feature] Landing page at /


tsmethurst:
There should be a nice landing page at the root of the instance that shows:

* Some explanation about the instance (description) with instance header and profile images.
* Basic server configuration/software version.
* Number of users/number of statuses.
* Links to apps for using the instance (Tusky, Pinafore.social, etc)
* Instructions for registering.
* Link to GoToSocial source code.
* Etc.

This should be created as a template in the /web/template directory of the repository, which allows instance owners to overwrite it as they wish.

tsmethurst:
Handled by @f0x52 in https://github.com/superseriousbusiness/gotosocial/pull/46

9cat:
where is the most  IMPORTANT  session :    registering ?

PR #46, opened 2021-06-18
new styling for frontpage, update login and authorize templates

f0x52:
Basic implementation for #32  and also updates the login and authorize templates to match the styling.

Source lives at `web/source/style.css`, which is transpiled to browser-usable css with `web/source/build.js` which depends on postcss. A pre-bundled version is included at `web/assets/bundle.css` so instance owners don't *have* to install node + the dependencies unless they want to change color variables


f0x52:
TODO: maybe CI thingie to automatically rebuild the bundle on changes?

internal/router/router.go
```diff
@@ -151,6 +151,22 @@ func New(config *config.Config, logger *logrus.Logger) (Router, error) {
 	logger.Debugf("loading templates from %s", tmPath)
 	engine.LoadHTMLGlob(tmPath)
 
+	// serve static files from /assets
+	// FIXME: why doesn't this config var work??
```
tsmethurst, line 155: Looks like you didn't yet add it as a CLI flag in `cmd/gotosocial/main.go`, which is where the parsing of config is initiated.

internal/router/router.go
```diff
@@ -151,6 +151,22 @@ func New(config *config.Config, logger *logrus.Logger) (Router, error) {
...
+	// assetPath := filepath.Join(cwd, config.TemplateConfig.AssetBaseDir);
+	engine.Static("/assets", "./web/assets")
+
+	// FIXME: actual variables
+	engine.GET("/", func(c *gin.Context) {
```
tsmethurst, line 160: imo this should be moved from the router to something like `/internal/web/base` to bring it in line with the way the client api and S2S api are structured, and it should be set up as a module so that a router is passed to it (take a look at any of the client API modules -- the statuses one is a good example). What do you think? Maybe having it as internal/web is a bit confusing though? Not sure.

web/template/authorize.tmpl
```diff
@@ -1,34 +1,15 @@
...
-
-  <body>
-    <div class="container">
-      <div class="jumbotron">
-        <form action="/oauth/authorize" method="POST">
-          <h1>Hi {{.user}}!</h1>
-          <p>Application <b>{{.appname}}</b> {{if len .appwebsite | eq 0 | not}}({{.appwebsite}}) {{end}}would like to perform actions on your behalf, with scope <em>{{.scope}}</em>.</p>
-          <p>The application will redirect to {{.redirect}} to continue.</p>
-          <p>
+{{ template "header.tmpl" .}}
```
tsmethurst, line 1: god, so much cleaner <3

web/template/footer.tmpl
```diff
@@ -0,0 +1,13 @@
+	<!-- footer.tmpl -->
+	<footer>
+		<div id="version">
+			Running GoToSocial version: <span class="accent">{{.version}}</span><br>
+			<a href="https://github.com/superseriousbusiness/gotosocial">Source Code</a>
+		</div>
+		<div id="contact">
+			Administrated by: <a href="/{{.adminUsername}}" class="nounderline">{{.adminUsername}}</a><br>
+			<a href="/moderation">Moderation team</a>
```
tsmethurst, line 9: I'm not sure yet what moderation system to actually put in place for GtS. I was even considering (although this adds complexity, obviously) giving the instance admin the ability to choose between having a traditional moderation team, or implementing some kind of community moderation option where instance users can vote on decisions.

For now this is good, it just got me thinking again about this :P

web/template/index.tmpl
```diff
@@ -0,0 +1,42 @@
+{{ template "header.tmpl" .}}
+<aside class="left logo">
+	<img src="/assets/sloth.png" alt="Clipart styled sloth logo">
+</aside>
+<section>
+	<h1>Home to <span class="count">{{.countUsers}}</span> users who posted <span class="count">{{.countStatuses}}</span> statuses.</h1>
+
+	<h3>This is the default landing page, you can edit it from <span class="accent">./web/template/index.tmpl</span></h1>
```
tsmethurst, line 8: I like this!

tsmethurst (changes requested):
Mostly looks good, just gotta figure out where to put the '/' route, and add the CLI flag

internal/web/base.go
```diff
@@ -0,0 +1,62 @@
...
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package base
```
tsmethurst, line 19: OK now I'm seeing this it occurs to me that since there are gonna be a bunch of web pages available (user profiles, statuses -- if they're public, moderation, maybe a list of blocked instances or something), we should actually move this up a level to package `web` and then have that Module attach a few different handlers through the `Route` function, that all have common cause in serving HTML via a web request. For helping to populate the templates, we should also pass a pointer to the `Processor` interface into the Module, as is done in most of the Client API and S2S modules. Then if Handlers need, say, a user count or some other information that's not just in config, they can call a function in the processor to retrieve that information before returning.

What do you think?

tsmethurst: This will also be useful in cases I foresee in the future, where users have set a specific template for a public status that they want to be available, and the Handler will need to know which template is used and fall back to a default if none has been manually selected.

tsmethurst (approved):

PR #70, opened 2021-06-27
Remote instance dereferencing

tsmethurst:
Remote instances are now dereferenced when they post to an inbox on a GtS instance.

* Dereferencing will be done first by checking the `/api/v1/instance` endpoint of an instance.
* If that doesn't work, `/.well-known/nodeinfo` will be checked.
* If that doesn't work, only a minimal representation of the instance will be stored.

A new field was added to the Instance database model. To create it:

`alter table instances add column contact_account_username text;`


decentral1se:
image

:partying_face:

tsmethurst:
@decentral1se nice! :)

PR #186, opened 2021-09-02
Struct validation

tsmethurst:
This PR adds validation tags for all database model structs, and adds a struct validator in `internal/validate`, as well as tests for all of the defined structs.

It also reworks some of the structure of the packages to make a bit more sense, and adds a stub for Bun migration.

`gtsmodel` is now the only place that contains structs that enter the database, and it *only* contains structs that enter the database. Reason for this: when we do migrations, we can basically copy-paste whatever the current state of gtsmodel is into a subfolder of migrations that pertains to the specific migration ID, so each migration has a copy of the gtsmodels it pertains to, as they were at the time the migration was created. This will make it easier to play back or forward from a specific migration.

**Note that this test doesn't actually hook the validation into the database logic yet, it only adds struct tags and validation tests to prepare for creating these db hooks in future**

Interesting stuff in the PR starts from here downwards:
https://github.com/superseriousbusiness/gotosocial/pull/186/files#diff-7c8fa43ec6d65ce61a4f0617935c0e2fae26f679ffb4ac43ac06f41437521fd3

Tests are here onwards: https://github.com/superseriousbusiness/gotosocial/pull/186/files#diff-52727fb181f9f77e0dc72951e43903d0fc5355ed355f71923fb91d2c60518c16


internal/gtsmodel/application.go
```diff
@@ -18,23 +18,18 @@
...
-	// a vapid key generated for this app when it was created
-	VapidKey string `bun:",nullzero"`
+	ID           string    `validate:"required,ulid" bun:"type:CHAR(26),pk,nullzero,notnull,unique"`      // id of this item in the database
+	CreatedAt    time.Time `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item created
+	UpdatedAt    time.Time `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item last updated
+	Name         string    `validate:"required" bun:",nullzero,notnull"`                                  // name of the application given when it was created (eg., 'tusky')
+	Website      string    `validate:"omitempty,url" bun:",nullzero"`                                     // website for the application given when it was created (eg., 'https://tusky.app')
+	RedirectURI  string    `validate:"required,uri" bun:",nullzero,notnull"`                              // redirect uri requested by the application for oauth2 flow
+	ClientID     string    `validate:"omitempty,ulid" bun:"type:CHAR(26),nullzero"`                       // id of the associated oauth client entity in the db
```
NyaaaWhatsUpDoc, line 32: can the clientID really be empty?

internal/gtsmodel/application.go
```diff
@@ -18,23 +18,18 @@
...
+	ClientSecret string    `validate:"required,uuid" bun:",nullzero,notnull"`                             // secret of the associated oauth client entity in the db
+	Scopes       string    `validate:"-" bun:",nullzero,notnull,default:'read'"`                          // scopes requested when this app was created
```
NyaaaWhatsUpDoc, line 34: i'd say scopes should always be provided imo as they're the application permissions

internal/gtsmodel/account.go
```diff
@@ -27,124 +27,73 @@ import (
...
+	Sensitive               bool             `validate:"-" bun:",nullzero,default:false"`                                                                            // Set posts from this account to sensitive by default?
+	Language                string           `validate:"omitempty,bcp47_language_tag" bun:",nullzero,notnull,default:'en'"`                                          // What language does this account post in?
+	URI                     string           `validate:"required,url" bun:",nullzero,notnull,unique"`                                                                // ActivityPub URI for this account.
+	URL                     string           `validate:"required_without=Domain,omitempty,url" bun:",nullzero,unique"`                                               // Web URL for this account's profile
+	LastWebfingeredAt       time.Time        `validate:"required_with=Domain" bun:"type:timestamp,nullzero"`                                                         // Last time this account was refreshed/located with webfinger.
```
NyaaaWhatsUpDoc, line 58: is this meant to be  `required_without` ? i might be wrong here, just double checking :p

internal/gtsmodel/domainblock.go
```diff
@@ -22,23 +22,14 @@ import "time"
...
+	CreatedAt          time.Time `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item created
+	UpdatedAt          time.Time `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item last updated
+	Domain             string    `validate:"required,fqdn" bun:",nullzero,notnull"`                             // domain to block. Eg. 'whatever.com'
+	CreatedByAccountID string    `validate:"required,ulid" bun:"type:CHAR(26),nullzero,notnull"`                // Account ID of the creator of this block
+	CreatedByAccount   *Account  `validate:"-" bun:"rel:belongs-to"`                                            // Account corresponding to createdByAccountID
+	PrivateComment     string    `validate:"-" bun:",nullzero"`                                                 // Private comment on this block, viewable to admins
+	PublicComment      string    `validate:"-" bun:",nullzero"`                                                 // Public comment on this block, viewable (optionally) by everyone
```
NyaaaWhatsUpDoc, line 32: should these comments have any validation checks in place? e.g. no HTML, length that kinda thing? genuinely asking as i'm not sure whether this goes in with the validator, bun field or we do it manually

internal/gtsmodel/instance.go
```diff
@@ -1,41 +1,43 @@
...
-	ContactAccount   *Account `bun:"rel:belongs-to"`
-	// Reputation score of this instance
-	Reputation int64 `bun:",notnull,default:0"`
-	// Version of the software used on this instance
-	Version string `bun:",nullzero"`
+	ID                     string       `validate:"required,ulid" bun:"type:CHAR(26),pk,nullzero,notnull,unique"`                     // id of this item in the database
+	CreatedAt              time.Time    `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"`                // when was item created
+	UpdatedAt              time.Time    `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"`                // when was item last updated
+	Domain                 string       `validate:"required,fqdn" bun:",nullzero,notnull,unique"`                                     // Instance domain eg example.org
+	Title                  string       `validate:"-" bun:",nullzero"`                                                                // Title of this instance as it would like to be displayed.
```
NyaaaWhatsUpDoc, line 29: Is it worth having string validation (no HTML etc), same goes for description etc below, any fields that might be displayed on page to show instance information. Or is this again, not likely to be performed here?

internal/gtsmodel/token.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Token is a translation of the gotosocial token with the ExpiresIn fields replaced with ExpiresAt.
+type Token struct {
+	ID                  string    `validate:"required,ulid" bun:"type:CHAR(26),pk,nullzero,notnull,unique"`      // id of this item in the database
+	CreatedAt           time.Time `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item created
+	UpdatedAt           time.Time `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item last updated
+	ClientID            string    `validate:"required,ulid" bun:"type:CHAR(26),nullzero,notnull"`                // ID of the client who owns this token
+	UserID              string    `validate:"required,ulid" bun:"type:CHAR(26),nullzero,notnull"`                // ID of the user who owns this token
+	RedirectURI         string    `validate:"required,uri" bun:",nullzero,notnull"`                              // Oauth redirect URI for this token
+	Scope               string    `validate:"omitempty" bun:",nullzero,notnull,default:'read'"`                  // Oauth scope
```
NyaaaWhatsUpDoc, line 31: Again, do we want to accept an empty scope given that this is application permissions?

internal/gtsmodel/token.go
```diff
@@ -0,0 +1,43 @@
...
+	CodeExpiresAt       time.Time `validate:"-" bun:"type:timestamp,nullzero"`                                   // Code expires at -- null means the code never expires
+	Access              string    `validate:"-" bun:",pk,nullzero,notnull,default:''"`                           // User level access token, if present
+	AccessCreateAt      time.Time `validate:"required_with=Access" bun:"type:timestamp,nullzero"`                // User level access token created time, if access present
+	AccessExpiresAt     time.Time `validate:"-" bun:"type:timestamp,nullzero"`                                   // User level access token expires at -- null means the token never expires
+	Refresh             string    `validate:"-" bun:",pk,nullzero,notnull,default:''"`                           // Refresh token, if present
+	RefreshCreateAt     time.Time `validate:"required_with=Refresh" bun:"type:timestamp,nullzero"`               // Refresh created at, if refresh present
+	RefreshExpiresAt    time.Time `validate:"-" bun:"type:timestamp,nullzero"`                                   // Refresh expires at -- null means the refresh token never expires
```
NyaaaWhatsUpDoc, line 42: do we want to make refreshExpiresAt, accessExpiresAt, codeExpiresAt  all with `required_with={Code,Access,Refresh}` ?

internal/gtsmodel/user.go
```diff
@@ -26,97 +26,34 @@ import (
...
-	OTPBackupCodes         []string
-	ConsumedTimestamp      int
-	RememberToken          string    `bun:",nullzero"`
-	SignInToken            string    `bun:",nullzero"`
-	SignInTokenSentAt      time.Time `bun:",nullzero"`
-	WebauthnID             string    `bun:",nullzero"`
+	ID                     string       `validate:"required,ulid" bun:"type:CHAR(26),pk,nullzero,notnull,unique"`      // id of this item in the database
+	CreatedAt              time.Time    `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item created
+	UpdatedAt              time.Time    `validate:"-" bun:"type:timestamp,nullzero,notnull,default:current_timestamp"` // when was item last updated
+	Email                  string       `validate:"required_with=ConfirmedAt" bun:",nullzero,unique"`                  // confirmed email address for this user, this should be unique -- only one email address registered per instance, multiple users per email are not supported
```
NyaaaWhatsUpDoc, line 32: email should always be required even before confirmed shouldn't it? otherwise it'll have nothing to confirm against

internal/gtsmodel/user.go
```diff
@@ -26,97 +26,34 @@ import (
...
+	SignUpIP               net.IP       `validate:"-" bun:",nullzero"`                                                 // From what IP was this user created?
+	CurrentSignInAt        time.Time    `validate:"-" bun:"type:timestamp,nullzero"`                                   // When did the user sign in with their current session.
+	CurrentSignInIP        net.IP       `validate:"-" bun:",nullzero"`                                                 // What's the most recent IP of this user
+	LastSignInAt           time.Time    `validate:"-" bun:"type:timestamp,nullzero"`                                   // When did this user last sign in?
+	LastSignInIP           net.IP       `validate:"-" bun:",nullzero"`                                                 // What's the previous IP of this user?
+	SignInCount            int          `validate:"-" bun:",nullzero,notnull,default:0"`                               // How many times has this user signed in?
```
NyaaaWhatsUpDoc, line 41: maybe set a min here of 0 just so it nenver goes negative :p

NyaaaWhatsUpDoc (changes requested):

NyaaaWhatsUpDoc:
Just ran this PR with SQLite and confirmed all tests are passing. Aside from mentioned issues, looks good!

internal/gtsmodel/account.go
tsmethurst: When domain is empty, it means it's a local account, so one that we don't need to webfinger. I think this is correct -- we only require it if it's a remote account, ie., if `Domain` is defined

internal/gtsmodel/domainblock.go
tsmethurst: We have some functionality in the `text` package for stripping HTML out of things (`sanitize.go`), so we should just make sure to pass it through that before we put it in the database. I think checking for HTML as we put something in the database is probably overkill.

internal/gtsmodel/instance.go
tsmethurst: Same story as above ^^

internal/gtsmodel/token.go
tsmethurst: IMO no, because most masto apps that we use (pinafore, tusky) need codes that don't expire, because they don't contain the logic for refreshing them or obtaining new tokens except on sign-in. So the `ExpiresAt` fields can be empty to indicate that there's no expiration date.

internal/gtsmodel/user.go
tsmethurst: We have `Email` and `UnconfirmedEmail` for that :)

internal/gtsmodel/user.go
tsmethurst: Good idea :+1:

internal/gtsmodel/application.go
tsmethurst: No, indeed not! Will change

internal/federation/dereferencing/account.go
```diff
@@ -165,19 +165,19 @@ func (d *deref) dereferenceAccountable(ctx context.Context, username string, rem
 	}
 
 	switch t.GetTypeName() {
-	case string(gtsmodel.ActivityStreamsPerson):
+	case string(ap.ActorPerson):
```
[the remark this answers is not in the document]
tsmethurst: nope, we don't, well spotted! this is a leftover from when those `ap` things were a separate type

internal/db/bundb/migrations/main.go
```diff
@@ -16,4 +16,13 @@
    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-package gtsmodel
+package migrations
```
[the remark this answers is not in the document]
tsmethurst: It's to do with how the bun migrations work. This file isn't used yet, but the migrations work by doing something like this:

```
func init() {
	up := func(ctx context.Context, db *bun.DB) error {
		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
			// your logic here
            return nil
		})
	}

	down := func(ctx context.Context, db *bun.DB) error {
		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
			// your logic here
            return nil
		})
	}

	if err := Migrations.Register(up, down); err != nil {
		panic(err)
	}
}
```

So afaik it's necessary to populate this var just because of the way migration files call on it. There might be a better way though, i'm not sure...

internal/validate/structvalidation.go
```diff
@@ -0,0 +1,77 @@
...
+func init() {
+	v = validator.New()
+	if err := v.RegisterValidation("ulid", ulidValidator); err != nil {
+		panic(err)
+	}
+}
+
+// Struct validates the passed struct, returning validator.ValidationErrors if invalid, or nil if OK.
+func Struct(s interface{}) error {
+	switch reflect.ValueOf(s).Kind() {
```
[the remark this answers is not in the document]
tsmethurst: Right! Changed :)

internal/federation/dereferencing/account.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: thought it might have been :3

internal/validate/structvalidation.go
```diff
@@ -53,11 +52,9 @@ func init() {
 
 // Struct validates the passed struct, returning validator.ValidationErrors if invalid, or nil if OK.
 func Struct(s interface{}) error {
-	switch reflect.ValueOf(s).Kind() {
+	switch reflect.TypeOf(s).Kind() {
```
NyaaaWhatsUpDoc, line 55: i just checked the source and i don't think we even need to check the type here considering they do it in the validate package :') sorry to keep requesting changes! i think we should just be safe with doing `v.Struct(s)` and passing error straight to our error converter

NyaaaWhatsUpDoc (changes requested):

internal/validate/structvalidation.go
tsmethurst: ahhh nice, even easier! I'll make the change :)

PR #191, opened 2021-09-03
rework media processing a little bit

### LINKED ISSUE #168: [bug] Image Descriptions no longer working
Since https://github.com/superseriousbusiness/gotosocial/pull/148, it seems that image descriptions aren't being parsed from federated messages and put in the database properly.

tsmethurst:
This PR reworks some of the media processing logic by taking a minimal MediaAttachment as a parameter and fleshing that out, rather than accepting a long list of parameters like content type, account ID, etc.

This also means that more fields are preserved on inserted images, including description!

Also adds some tests for dereferencing remote attachments, and some tests for extracting values from activitypub representations of attachments.

closes #168


internal/federation/dereferencing/attachment.go
```diff
@@ -28,17 +28,27 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 )
 
-func (d *deref) GetRemoteAttachment(ctx context.Context, requestingUsername string, remoteAttachmentURI *url.URL, ownerAccountID string, statusID string, expectedContentType string) (*gtsmodel.MediaAttachment, error) {
+func (d *deref) GetRemoteAttachment(ctx context.Context, requestingUsername string, minAttachment *gtsmodel.MediaAttachment) (*gtsmodel.MediaAttachment, error) {
+	if minAttachment == nil {
+		return nil, fmt.Errorf("GetRemoteAttachment: minAttachment was nil")
```
NyaaaWhatsUpDoc, line 33: I would say it's probably not worth doing this check. A nil supplied attachment is definitely bad behaviour going into this function and a panic is much more likely to attract our attention

NyaaaWhatsUpDoc (changes requested):
99% looks good, just the one point i mentioned :)

internal/federation/dereferencing/attachment.go
tsmethurst: Changed!

NyaaaWhatsUpDoc (approved):

Issue #175, opened 2021-08-30
[frontend] Thread views


f0x52:
Web-accessible views of threads. I'm currently working on this, with a template that's roughly modelled after Mastodon's layout.

In the future multiple layouts could be provided (something that shows sub-threads distinctly with coloured lines would be a nice experiment imo)

f0x52:
Current UI demo, looking for feedback!

https://gts-thread.dev.cthu.lu

f0x52:
updated UI demo, now with styling for media posts too!

https://gts-thread.dev.cthu.lu/media.html

PR #207, opened 2021-09-10
Thread views on the web

f0x52:
Implementation for #175. Also includes the new CSS bundling setup


web/template/status.tmpl
```diff
@@ -0,0 +1,36 @@
...
+	<div class="content">
+		{{.Content |noescape}}
+	</div>
+</div>
+{{with .MediaAttachments}}
+	<div class="media {{(len .) | oddOrEven }}{{if eq (len .) 1}} single{{end}}{{if eq (len .) 2}} double{{end}}">
+		{{range .}}
+			<a href="{{.URL}}" target="_blank" title="{{.Description}}">
+					{{if not .Description}}
+			 		<div class="no-image-desc" aria-hidden="true" >(!)<span>Missing image description</span></div>
```
tsmethurst, line 20: is it possible to use something from FA instead of the literal `(!)`?

internal/web/thread.go
```diff
@@ -0,0 +1,82 @@
...
+	l.Trace("rendering thread template")
+
+	var statusLink StatusLink
+
+	if err := c.ShouldBindUri(&statusLink); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "status not found"})
+		return
+	}
+
+	authed, err := oauth.Authed(c, false, false, false, false) // we don't really need an app here but we want everything else
```
tsmethurst, line 44: you might as well drop the line comment here, since it doesn't really apply :P

internal/web/thread.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil {
+		l.Errorf("error authing status GET request: %s", err)
+		c.JSON(http.StatusBadRequest, gin.H{"error": "status not found"})
+		return
+	}
+
+	instance, err := m.processor.InstanceGet(c.Request.Context(), m.config.Host)
```
tsmethurst, line 51: since you're using c.Request.Context() several times, you can just define it at the top as something like `ctx := c.Request.Context()` and then pass that in to the processor functions, it's a bit neater

internal/web/thread.go
```diff
@@ -0,0 +1,82 @@
...
+	ID   string `uri:"id"   binding:"required"`
+}
+
+func (m *Module) threadTemplateHandler(c *gin.Context) {
+	l := m.log.WithField("func", "threadTemplateGET")
+	l.Trace("rendering thread template")
+
+	var statusLink StatusLink
+
+	if err := c.ShouldBindUri(&statusLink); err != nil {
```
tsmethurst, line 39: COOL, I didn't know this was possible!

internal/web/thread.go
```diff
@@ -0,0 +1,82 @@
...
+package web
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/superseriousbusiness/gotosocial/internal/oauth"
+)
+
+type StatusLink struct {
```
tsmethurst, line 28: can you make this unexported if you're just using it here? (ie., use lower-camel instead of upper-camel.) Does it still work then?

testrig/testmodels.go
```diff
@@ -302,7 +302,7 @@ func NewTestAccounts() map[string]*gtsmodel.Account {
 			Username:                "admin",
 			AvatarMediaAttachmentID: "",
 			HeaderMediaAttachmentID: "",
-			DisplayName:             "",
+			DisplayName:             "Admin",
```
tsmethurst, line 305: in cases where no display name is set, we should just use the 'username' instead, does the templating allow for this?

testrig/testmodels.go
```diff
@@ -540,7 +540,7 @@ func NewTestAttachments() map[string]*gtsmodel.MediaAttachment {
 				},
 			},
 			AccountID:         "01F8MH1H7YV1Z7D2C8K2730QBF",
-			Description:       "90's Trent Reznor turning to the camera",
+			Description:       "",
```
tsmethurst, line 543: Were these changes a left over from testing maybe? :P

tsmethurst (changes requested):
Mostly looks fucking GREAT! <3 Just a few little comments/potential changes/questions. Really excited about this :D

internal/router/template.go
```diff
@@ -37,16 +39,56 @@ func loadTemplates(cfg *config.Config, engine *gin.Engine) error {
 
 	tmPath := filepath.Join(cwd, fmt.Sprintf("%s*", cfg.TemplateConfig.BaseDir))
 
+	println("loading html templates")
```
tsmethurst, line 42: best drop this or convert it to log with logrus

internal/web/thread.go
```diff
@@ -0,0 +1,84 @@
...
+		return
+	}
+
+	status, err := m.processor.StatusGet(ctx, authed, uriParts.ID)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "status not found"})
+		return
+	}
+
+	println(uriParts.User[:1], uriParts.User, status.Account.Username)
```
tsmethurst, line 66: same comment as above println call

testrig/testmodels.go
```diff
@@ -800,7 +800,7 @@ func NewTestStatuses() map[string]*gtsmodel.Status {
 			URI:                      "http://localhost:8080/users/admin/statuses/01F8MH75CBF9JFX4ZAD54N0W0R",
 			URL:                      "http://localhost:8080/@admin/statuses/01F8MH75CBF9JFX4ZAD54N0W0R",
 			Content:                  "hello world! #welcome ! first post on the instance :rainbow: !",
-			AttachmentIDs:            []string{"01F8MH6NEM8D7527KZAECTCR76"},
+			AttachmentIDs:            []string{"01F8MH6NEM8D7527KZAECTCR76", "01F8MH7TDVANYKWVE8VVKFPJTJ", "01F8MH8RMYQ6MSNY3JM2XT1CQ5"},
```
tsmethurst, line 803: do you still need the changes in this file?

f0x52: oh not really, but it would be good to have a test thread with a bunch of images populated

PR #312, opened 2021-11-18
anticipating / warning of / preventing cryptic sqlite errors

ForestJohnson:
I manually tested 4 cases on linux:

1. parent folder does not exist
1. folder not owned by UID / GID of process
1. folder not readable+writable by user/group/everyone
1. and of course the happy path where everything works

### DIFF
--- a/internal/db/bundb/bundb.go
+++ b/internal/db/bundb/bundb.go
@@ -140,20 +140,31 @@ func NewBunDBService(ctx context.Context, c *config.Config) (db.DB, error) {
 			return nil, fmt.Errorf("could not open sqlite db: could not determine parent folder permissions for '%s': %s", providedPath, err)
 		}
 		parentFolderStat := parentFolderInfo.Sys().(*syscall.Stat_t)
-		readWritableBecauseUser := ((parentFolderInfo.Mode() & 0600) == 0600) && os.Geteuid() == int(parentFolderStat.Uid)
-		readWritableBecauseGroup := ((parentFolderInfo.Mode() & 0060) == 0060) && os.Getegid() == int(parentFolderStat.Gid)
-		readWritableBecauseEveryone := ((parentFolderInfo.Mode() & 0006) == 0006)
+		processUserOwnsFolder := os.Geteuid() == int(parentFolderStat.Uid)
+		processIsRunningAsRoot := os.Geteuid() == 0
+		processHasGroupMatchForFolder := os.Getegid() == int(parentFolderStat.Gid)
+		folderAllowsOwnerToReadAndWrite := (parentFolderInfo.Mode() & 0600) == 0600
+		folderAllowsGroupMembersToReadAndWrite := (parentFolderInfo.Mode() & 0060) == 0060
+		folderAllowsAnyoneToReadAndWrite := (parentFolderInfo.Mode() & 0006) == 0006
+		readWritableBecauseUser := processUserOwnsFolder && folderAllowsOwnerToReadAndWrite
+		// could use os.Getgroups() here to determine if the process user has membership in the group for the folder...
+		// but does it really matter that much just for a warning message? 99% of the time the process UID should own the folder anyways
+		readWritableBecauseGroup := processHasGroupMatchForFolder && folderAllowsGroupMembersToReadAndWrite
+
+		logrus.Infof("attempting to open sqlite database at %s", providedPath)
+		if !(processIsRunningAsRoot || readWritableBecauseUser || readWritableBecauseGroup || folderAllowsAnyoneToReadAndWrite) {
+			logrus.Warnf("WARNING: parent folder permissions for '%s' do not seem to allow us to read and write", providedPath)
+			if !processUserOwnsFolder {
+				logrus.Warnf("GoToSocial is running as user id %d, however the folder is owned by user %d", os.Geteuid(), int(parentFolderStat.Uid))
+			} else if !folderAllowsOwnerToReadAndWrite {
+				logrus.Warnf("GoToSocial (user id %d) owns the folder, however the owner is not allowed to read and write", os.Geteuid())
+			}
+		}
 
 		// Append our own SQLite preferences
 		c.DBConfig.Address = "file:" + c.DBConfig.Address + "?cache=shared"
 
-		logrus.Infof("attempting to open sqlite database with address %s", c.DBConfig.Address)
-		if !(readWritableBecauseUser || readWritableBecauseGroup || readWritableBecauseEveryone) {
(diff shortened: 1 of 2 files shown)


ForestJohnson:
ah crap 

```
time="2021-11-18T03:29:51Z" level=info msg="attempting to open sqlite database at /drone/src/internal/timeline/:memory:"
```

ForestJohnson:
the build is failing on this test: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/timeline/manager_test.go#L123

I'm not sure why.... is this a spurious failure ?

ForestJohnson:
How frequently is NewBunDBService called? I assumed it would be called once on setup... does my permissions check logic need to be memoized ?

tsmethurst:
> How frequently is NewBunDBService called? I assumed it would be called once on setup... 

once on setup is right :)

tsmethurst:
>  is this a spurious failure ?

yes -- there's a race condition in the timeline part of the tests we still need to fix, i'll kick the test server

internal/db/bundb/bundb.go
```diff
@@ -123,16 +125,51 @@ func NewBunDBService(ctx context.Context, c *config.Config) (db.DB, error) {
 		conn = WrapDBConn(bun.NewDB(sqldb, pgdialect.New()))
 	case dbTypeSqlite:
 		// SQLITE
+		var err error
 
 		// Drop anything fancy from DB address
 		c.DBConfig.Address = strings.Split(c.DBConfig.Address, "?")[0]
 		c.DBConfig.Address = strings.TrimPrefix(c.DBConfig.Address, "file:")
 
+		// if we are about to try to create/open a sqlite database file, we should do some checking to ensure
```
tsmethurst, line 134: IMO you can put all this logic in the `util` package and just have it return an error if there's a problem, instead of printing warnings. Then the caller can choose to print the error as a warning and continue, or just return an error and fail. We'll probably need this kind of checking logic elsewhere so it makes sense to make it reusable.

ForestJohnson: :+1:

internal/util/filesystem.go
```diff
@@ -0,0 +1,61 @@
...
+package util
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+func FilePathExistsAndIsReadWritable(path string) (exists bool, err error) {
```
tsmethurst, line 28: I would refactor this to just return an error if we know something won't work, and nil if we can be confident it will work. Returning a boolean as well just complicates things IMO. I'd also rename it to something like 'FilePathOK' and annotate the method to explain what 'OK' means in this context (ie., it exists, it's readable and writeable, etc).

internal/db/bundb/bundb.go
```diff
@@ -123,16 +125,36 @@ func NewBunDBService(ctx context.Context, c *config.Config) (db.DB, error) {
...
+		// that the place this file is about to be opened even exists and allows us to read and write.
+		// if it doesn't, sqlite will return extremely cryptic errors... so we pre-empt this and warn the user ourselves
+		if c.DBConfig.Address != ":memory:" {
+			absPathForLog, err := filepath.Abs(c.DBConfig.Address)
+			if err != nil {
+				absPathForLog = c.DBConfig.Address
+			}
+			logrus.Infof("attempting to open sqlite database at %s", absPathForLog)
+
+			exists, err := util.FilePathExistsAndIsReadWritable(c.DBConfig.Address)
```
tsmethurst, line 144: The usual style (in GtS at least) is to check the error first, then check the other value. Also, I think if err isn't nil, we should return an error rather than proceeding, since we know it's going to fail anyway. So I'd replace this chunk with something more like:

```go
exists, err := util.FilePathExistsAndIsReadWritable(c.DBConfig.Address)
if err != nil {
	return nil, fmt.Errorf("could not open sqlite db: %s", err)
}

if !exists {
	return nil, errors.New("blah blah blah")
}
```

internal/util/filesystem.go
tsmethurst: I'd also refactor a lot of the logic inside the function itself:

1. Return earlier if we *know* we're OK or not OK (ie., return OK early if we're running as root, rather than doing the rest of the checks). That is, rather than gathering all the info and then doing an `or` check with four different options, just check them one by one further up.
2. Move the `0`, `0600`, `0060`, `0006` to `const`s with names that make clear what they actually are.
3. Rather than calling things like `os.Geteuid()` more than once, just call it once and assign it to a variable.
4. Don't swallow the possible `filepath.Abs` error.
5. Rather than setting a variable `extraInfo` just return an error. There's no point setting is a variable and then assigning it as one thing or another, if that's all that's being done with it.

internal/db/bundb/bundb.go
ForestJohnson: > we know it's going to fail anyway

This is not true. The FilePathExistsAndIsReadWritable code is just a heuristic guess, its not guaranteed to be 100% accurate.

Now that I think about it, essentially trying to re-implement the operating systems permission logic is  really silly, instead `FilePathExistsAndIsReadWritable` should just try to write a test file and then delete it.  That way if it fails, we know for sure sqlite will fail

ForestJohnson:
Here is the output from the new version: 

Database file is owned root (its readable, but not writable):

`time=2021-11-20T13:29:20-06:00 level=fatal msg=error creating dbservice: could not open sqlite db: checking read/write access failed: open /home/forest/Desktop/git/gotosocial/storage/sqlite.db: permission denied`


`storage` folder was deleted:

`time=2021-11-20T13:30:06-06:00 level=fatal msg=error creating dbservice: could not open sqlite db: checking read/write access failed: I cannot write a file inside '/home/forest/Desktop/git/gotosocial/storage': open /home/forest/Desktop/git/gotosocial/storage/sqlite.db.iotest: no such file or directory`


Database file is deleted and `storage` folder is owned by root w/ chmod 700:

`time=2021-11-20T13:33:17-06:00 level=fatal msg=error creating dbservice: could not open sqlite db: checking read/write access failed: open /home/forest/Desktop/git/gotosocial/storage/sqlite.db: permission denied`

happy path:

`time=2021-11-20T13:36:54-06:00 level=info msg=file or directory exists and permissions seem ok, attempting to open sqlite database at '/home/forest/Desktop/git/gotosocial/storage/sqlite.db'`

ForestJohnson:
By the way maybe this is a separate issue but it looks like right now `sqlite.db` is being created with `-rw-r--r--` aka `0644` permissions.  Maybe it should be 600 (only readable and writable by owner) for security hardening?  IDK :shrug:

ForestJohnson:
This PR was rejected by the GTS project

PR #314, opened 2021-11-19
Fix broken defaults and broken flags

ForestJohnson:
See here for explanation (very tall wall of text  warning ) https://picopublish.sequentialread.com/files/chatlog_2021_11_18.txt

It works like this (NEW) as of commit 4c5fa42565abd1988e1f5297a1ab74784e3885ca
```
forest@thingpad:~/Desktop/git/gotosocial$ GTS_HOST=localhost  GTS_ACCOUNTS_OPEN_REGISTRATION=true go run ./cmd/* server start 
INFO[0000] f.IsSet(fn.AccountsOpenRegistration): true   
time=2021-11-18T20:58:16-06:00 level=fatal msg=error creating dbservice: db connection error: failed to connect to `host=localhost user=postgres database=postgres`: dial error (dial tcp 127.0.0.1:5432: connect: connection refused)
exit status 1
forest@thingpad:~/Desktop/git/gotosocial$ GTS_HOST=localhost   go run ./cmd/* server start 
INFO[0000] f.IsSet(fn.AccountsOpenRegistration): false  
time=2021-11-18T20:59:12-06:00 level=fatal msg=error creating dbservice: db connection error: failed to connect to `host=localhost user=postgres database=postgres`: dial error (dial tcp 127.0.0.1:5432: connect: connection refused)
exit status 1
```

On the main branch (OLD) as of f3e36de78aa9d13e74fabbb115bd7a20be57abaf

the same print debugging log looks like this: 
```
forest@thingpad:~/Desktop/git/gotosocial$ GTS_HOST=localhost  GTS_ACCOUNTS_OPEN_REGISTRATION=true go run ./cmd/* server start 
INFO[0000] f.IsSet(fn.AccountsOpenRegistration): false  
time=2021-11-18T21:20:37-06:00 level=fatal msg=error creating dbservice: db connection error: failed to connect to `host=localhost user=postgres database=postgres`: dial error (dial tcp 127.0.0.1:5432: connect: connection refused)
exit status 1
forest@thingpad:~/Desktop/git/gotosocial$ GTS_HOST=localhost   go run ./cmd/* server start 
INFO[0000] f.IsSet(fn.AccountsOpenRegistration): false  
time=2021-11-18T21:20:41-06:00 level=fatal msg=error creating dbservice: db connection error: failed to connect to `host=localhost user=postgres database=postgres`: dial error (dial tcp 127.0.0.1:5432: connect: connection refused)
exit status 1
```

### DIFF
--- a/cmd/gotosocial/accountsflags.go
+++ b/cmd/gotosocial/accountsflags.go
@@ -24,7 +24,6 @@ import (
 )
 
 func accountsFlags(flagNames, envNames config.Flags, defaults config.Defaults) []cli.Flag {
-
 	return []cli.Flag{
 		&cli.BoolFlag{
 			Name:    flagNames.AccountsOpenRegistration,
--- a/cmd/gotosocial/accountsflags.go
+++ b/cmd/gotosocial/accountsflags.go
@@ -24,6 +24,7 @@ import (
 )
 
 func accountsFlags(flagNames, envNames config.Flags, defaults config.Defaults) []cli.Flag {
+
 	return []cli.Flag{
 		&cli.BoolFlag{
 			Name:    flagNames.AccountsOpenRegistration,
--- a/cmd/gotosocial/admincommands.go
+++ b/cmd/gotosocial/admincommands.go
@@ -25,7 +25,7 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
-func adminCommands() []*cli.Command {
+func adminCommands(allFlags []cli.Flag) []*cli.Command {
 	return []*cli.Command{
 		{
 			Name:  "admin",
@@ -56,7 +56,7 @@ func adminCommands() []*cli.Command {
 								},
 							},
 							Action: func(c *cli.Context) error {
-								return runAction(c, account.Create)
+								return runAction(c, allFlags, account.Create)
 							},
 						},
 						{
@@ -70,7 +70,7 @@ func adminCommands() []*cli.Command {
 								},
 							},
 							Action: func(c *cli.Context) error {
-								return runAction(c, account.Confirm)
+								return runAction(c, allFlags, account.Confirm)
 							},
 						},
 						{
@@ -84,7 +84,7 @@ func adminCommands() []*cli.Command {
 								},
 							},
 							Action: func(c *cli.Context) error {
-								return runAction(c, account.Promote)
+								return runAction(c, allFlags, account.Promote)
 							},
 						},
 						{
@@ -98,7 +98,7 @@ func adminCommands() []*cli.Command {
 								},
 							},
 							Action: func(c *cli.Context) error {
-								return runAction(c, account.Demote)
+								return runAction(c, allFlags, account.Demote)
 							},
 						},
 						{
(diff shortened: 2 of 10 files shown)


ForestJohnson:
actually I'm not sure, should the tests use the default config too, instead of continuing using the empty config ?

depending on how the tests are implemented, either one could be absolutely the right one to choose. However I wouldn't be surprised if it will work either way.

ForestJohnson:
i decided to go with having the test start from the default config instead of the empty config. the test passes just fine and I think its more consistent + less code overall which seems better to me

tsmethurst:
Heya! I had a read through all this and it looks good :+1: 

I think this will probably end up being a temporary fix until we move from urfave to cobra + viper for our CLI tooling needs, but that's fine.

Two things only:

1. I just updated the main branch with stricter linters for golangci-lint, to help us catch bugs + redundant code etc. Can you merge main into this branch and run golangci-lint with the stricter linters? They helped me pick up a bunch of stuff that I totally missed in my own code.
2. In this line: https://github.com/superseriousbusiness/gotosocial/pull/314/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR95 AFAIK it's best practice to unmarshal into an empty struct (pointer), not one that's already been populated with some values, so I'd change that back to using the empty config.

Apart from that, this is ready to go imo.

ForestJohnson:
>  it's best practice to unmarshal into an empty struct (pointer), not one that's already been populated with some values

Why is that?

I thought this was fine because of the way the unmarshaler works.  It [loops over the yaml tokens  that it parsed from the file ](https://github.com/go-yaml/yaml/blob/7649d4548cb53a614db133b2a8ac1f31859dda8c/decode.go#L382-L386)and then calls unmarshal(n *node, out reflect.Value)` for each one. `unmarshal(n *node, out reflect.Value)` [does different things depending on the type of yaml token being unmarshaled](https://github.com/go-yaml/yaml/blob/v2.4.0/decode.go#L351-L378)  but ultimately it [uses reflection to set properties on the struct it's unmarshalling into](https://github.com/go-yaml/yaml/blob/v2.4.0/decode.go#L437).

I think the stdlib JSON unmarshaller works the same way.  Essentially it "overlays" the values from the config file onto the existing struct regardless of whether the existing struct is empty or already populated with default values. 

Maybe we could write a test that demonstrates that unmarshalling the config like this overlays properly onto the defaults?

I think this is borderline necessary, because if we implement this config overlaying logic ourselves, we'll have to differentiate between a user-specified value that just so happens to be the default value for the type, and an un-set value.  (`OpenRegistration = false `because the user disabled it versus `OpenRegistration = false` because `false` is the default value for a boolean property on a struct)  That's the problem this PR was intended to fix. 

If we aren't going to use the default unmarshaller to overlay the config file onto the default values, we'll end-up re-creating the logic that's already written into the  default unmarshaller..  Iterating over the YAML tokens in the config and setting their values onto the  config struct, if any.

tsmethurst:
Alright, that makes sense :+1: Thanks for explaining. Scratch my second point then :)

ForestJohnson:
i rebased the branch onto main

PR #324, opened 2021-11-22
Use IPv6 doc prefix for docs

LittleFox94:
IPv6 has a dedicated prefix for documentation and examples, this commit makes use of it for `BindAddress` config flag docs and examples.


tsmethurst:
Didn't know about this, thanks!

PR #344, opened 2021-12-13
Remove unnecessary storage config variables

tsmethurst:
This PR removes the config variables `storage-serve-protocol`, `storage-serve-host`, and `storage-serve-base-path`, none of which worked properly or were really necessary.

It also renames the config variable `storage-base-path` to `storage-file-base-path` to make room for a `storage-s3-*` namespace in the future.

Finally, it tidies up some of the URI generation logic into its own `uris` package, and adds a few more functions for generating URIs for emoji and media attachments.


tsmethurst:
@LittleFox94 i'm mentioning you at last :D

example/config.yaml
```diff
@@ -213,27 +213,7 @@ storage-backend: "local"
...
-# Default: "localhost" -- you should absolutely change this.
-storage-serve-host: "localhost"
-
-# String. Base path for serving stored files. This will be added to serveHost and serveProtocol
-# to form the prefix url of your stored files. Eg., https://example.org/fileserver/.....
-# It's unlikely that you will need to change this.
-# Examples: ["/fileserver", "/media"]
-# Default: "/fileserver"
-storage-serve-base-path: "/fileserver"
+storage-file-base-path: "/gotosocial/storage"
```
LittleFox94, line 216: so I guess we have no easy chance to have something like this instead?
```yaml
storage:
  file:
    base-path: /gotosocial/storage
```

internal/config/defaults.go
```diff
@@ -54,11 +54,8 @@ var Defaults = Values{
 	MediaDescriptionMinChars: 0,
 	MediaDescriptionMaxChars: 500,
 
-	StorageBackend:       "local",
-	StorageBasePath:      "/gotosocial/storage",
-	StorageServeProtocol: "https",
-	StorageServeHost:     "localhost",
-	StorageServeBasePath: "/fileserver",
+	StorageBackend:      "local",
```
LittleFox94, line 57: I think we should decide on either `local` or `file`

internal/media/handler.go
```diff
@@ -28,36 +28,22 @@ import (
...
-	Header Type = "header"
-	// Avatar is the key for profile avatar requests
-	Avatar Type = "avatar"
-	// Emoji is the key for emoji type requests
-	Emoji Type = "emoji"
+	SizeSmall    string = "small"    // SizeSmall is the key for small/thumbnail versions of media
+	SizeOriginal string = "original" // SizeOriginal is the key for original/fullsize versions of media and emoji
+	SizeStatic   string = "static"   // SizeStatic is the key for static (non-animated) versions of emoji
+
+	TypeAttachment string = "attachment" // TypeAttachment is the key for media attachments
```
LittleFox94, line 43: If you put these in their own `const` block, `godoc` will see them as enum and group them together with the type

```go
type Size string

const (
  SizeSmall Size    = "small"
  SizeOriginal Size = "original"
  SizeStatic Size   = "static"
)

type Type string

const (
  TypeAttachment Type = "attachment"
  TypeHeader Type     = "header"
  TypeAvatar Type     = "avatar"
  TypeEmoji Type      = "emoji"
)

File: internal/media/handler.go [hunk 4]
```diff
@@ -28,36 +28,22 @@ import (
...
-	"github.com/superseriousbusiness/gotosocial/internal/config"
 	"github.com/superseriousbusiness/gotosocial/internal/db"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/id"
 	"github.com/superseriousbusiness/gotosocial/internal/transport"
+	"github.com/superseriousbusiness/gotosocial/internal/uris"
 )
 
-// Size describes the *size* of a piece of media
-type Size string
```
LittleFox94, line 40: any particular reason to remove these types? More type safety is often better

internal/media/util.go
```diff
@@ -294,29 +294,29 @@ type imageAndMeta struct {
...
 // ParseMediaSize converts s to a recognized MediaSize, or returns an error if unrecognized
-func ParseMediaSize(s string) (Size, error) {
-	switch Size(s) {
-	case Small:
-		return Small, nil
-	case Original:
-		return Original, nil
-	case Static:
-		return Static, nil
+func ParseMediaSize(s string) (string, error) {
```
LittleFox94, line 312: ... and without this being a type it's a bit funny - more like `if none of these, error - else return as is` ^^'

LittleFox94 (commented):
I like most of what I see in this one, have some questions and suggestions though :)

example/config.yaml
tsmethurst: unfortunately not really

LittleFox94: ok, sad but let's not block on this

internal/config/defaults.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/344/commits/f4ad9bf12defb1376a07c64724f16cffd9f8c9ee

internal/media/handler.go
tsmethurst: Got it :+1:

internal/media/handler.go
tsmethurst: No not really, I felt it was cleaner but now I look at it again it doesn't make sense, I've put the types back in :P

LittleFox94 (approved):

PR #347, opened 2021-12-15
Update codeberg.org/gruf libraries and fix go-store issue

NyaaaWhatsUpDoc:
Fixes a critical issue in the codeberg.org/gruf/go-store library in which path traversal would have been possible. Fortunately this doesn't work with our HTTP muxer but still something that's worth fixing!

I will leave a comment below in the relevant vendored code that's worth checking.

This has passed my own module's tests but it's still worth checking that this doesn't cause any fileserver path issues on a test instance.


vendor/codeberg.org/gruf/go-store/storage/disk.go
```diff
@@ -282,10 +280,10 @@ func (st *DiskStorage) filepath(key string) (string, error) {
 	pb.AppendString(st.path)
 	pb.AppendString(key)
 
-	// If path is dir traversal, and traverses FURTHER
-	// than store root, this is an error
-	if util.CountDotdots(pb.StringPtr()) > st.dots {
+	// Check for dir traversal outside of root
+	if util.IsDirTraversal(st.path, pb.StringPtr()) {
```
NyaaaWhatsUpDoc, line 284: So here is where the dir traversal check is being performed. Next comment will be the implementation of said function

vendor/codeberg.org/gruf/go-store/util/fs.go
```diff
@@ -9,14 +9,22 @@ import (
 	"codeberg.org/gruf/go-fastpath"
 )
 
-var dotdot = "../"
+// IsDirTraversal will check if rootPlusPath is a dir traversal outside of root,
+// assuming that both are cleaned and that rootPlusPath is path.Join(root, somePath)
+func IsDirTraversal(root string, rootPlusPath string) bool {
```
NyaaaWhatsUpDoc, line 14: This is the dir traversal check function.

PR #365, opened 2022-01-30
adding authorization to the already-existing authentication

ForestJohnson:
adding the following conditions:

1. Email has been confirmed
1. either the GTS instance is configured to not require admin approval for new accounts, or the admin has approved this account
1. the account has not been disabled / suspended

TO:

  -  the browser-based login handler and to
  -  the `TokenCheck` function that is used throughout the API.


internal/api/client/auth/authorize.go
```diff
@@ -216,3 +234,22 @@ func extractAuthForm(s sessions.Session, form *model.OAuthAuthorize) error {
 	s.Set(sessionState, uuid.NewString())
 	return s.Save()
 }
+
+func ensureUserIsAuthorizedOrRedirect(ctx *gin.Context, user *gtsmodel.User) bool {
+	if user.ConfirmedAt.IsZero() {
+		ctx.Redirect(http.StatusSeeOther, CheckYourEmailPath)
```
tsmethurst, line 240: these redirects don't land anywhere yet right? just for my understanding

ForestJohnson: exactly, I haven't finished all of those parts yet, at least not on my current branch.  I suppose its sort of disconnected like this because I was trying to do multiple smaller PRs

ForestJohnson: If you would rather, I could make it 1 big one.

tsmethurst: no no it's good as is, no worries

internal/api/client/auth/authorize.go
```diff
@@ -44,7 +44,7 @@ func (m *Module) AuthorizeGETHandler(c *gin.Context) {
 	s := sessions.Default(c)
 
 	if _, err := api.NegotiateAccept(c, api.HTMLAcceptHeaders...); err != nil {
-		c.JSON(http.StatusNotAcceptable, gin.H{"error": err.Error()})
+		c.HTML(http.StatusNotAcceptable, "error.tmpl", gin.H{"error": err.Error()})
 		return
```
tsmethurst, line 48: Can you add a `m.clearSession(s)` here? I forgot it last time when I added this error path; might as well fix it in this pr

tsmethurst (commented):
Can you write some tests (in the same style as elsewhere in the codebase) to trigger `!ensureUserIsAuthorizedOrRedirect`? Once it's tested nicely I think this PR is good to go

internal/api/client/auth/authorize.go
ForestJohnson: why does it need to clear the session in that case? That would only happen if someone was trying to hit the endpoint with a really weird client that probably doesn't implement sessions anyways

PR #366, opened 2022-01-30
Plaintext email to fix "message refused: Message is not RFC 2822 compliant"

ForestJohnson:


internal/email/util.go
```diff
@@ -19,15 +19,12 @@
 package email
 
 import (
+	"errors"
 	"fmt"
 	"html/template"
```
tsmethurst, line 24: we should use text/template (or whatever it's called) instead of html/template now, since we're not sending html emails anymore

internal/email/util.go
```diff
@@ -19,15 +19,12 @@
...
 	"os"
 	"path/filepath"
-)
-
-const (
-	mime = `MIME-version: 1.0;
```
tsmethurst, line 29: shouldn't we change this to text/plain instead of removing it altogether?

internal/email/util.go
```diff
@@ -41,16 +38,34 @@ func loadTemplates(templateBaseDir string) (*template.Template, error) {
...
+	if strings.Contains(mailFrom, "\r") || strings.Contains(mailFrom, "\n") {
+		return nil, errors.New("email from address must not contain newline characters")
+	}
+
+	if strings.Contains(mailTo, "\r") || strings.Contains(mailTo, "\n") {
+		return nil, errors.New("email to address must not contain newline characters")
+	}
+
+	// normalize the message body to use CRLF line endings
+	mailBody = strings.ReplaceAll(mailBody, "\r\n", "\n")
```
tsmethurst, line 60: wait... doesn't this just convert `\r\n` to `\n` and then back again? am i missing something

internal/email/util.go
ForestJohnson: I just now tried adding it back and I got this:

```
time=2022-01-31T06:12:54Z level=error msg=SendConfirmEmail: error sending to email address q6qjo3+2eksherhhgf0w@sharklasers.com belonging to user test7: 550 message rejected func=github.com/superseriousbusiness/gotosocial/internal/processing.(*processor).Start.func1.1 file=github.com/superseriousbusiness/gotosocial/internal/processing/processor.go:313
```

Like I said before I have no interest in studying how to serialize email by hand, I think it's something like JSON where we should **not** try to serialize it by hand, we should use a library.  TBH there might be something in the standard library which would provide what we need but it sounded like you wanted to make the smallest change possible to get the emails sending, so I didn't start looking into that yet. 

Instead I simply copied and pasted the "by hand" email sending code that was in the golang documentation, and it seems to work so far.

internal/email/util.go
ForestJohnson: Notice it also converts `\n` to `\r\n`.  I didn't know if this was required or not but the [plain text email protocol example that I was working from](https://pkg.go.dev/net/smtp#example-SendMail) had `\r\n` so I made sure we always use `\r\n` regardless of what's in the template files.

ForestJohnson: However I will say that I tested it just now without converting the body to CRLF and it did  work with guerrillamail.com. So maybe it doesn't matter either way.

internal/email/util.go
tsmethurst: okay, no problem :+1: thanks for trying it, let's just leave it out then

internal/email/util.go
tsmethurst: `\r\n` is the standard way in the protocol for breaking between the important lines at the beginning (subject, to, from etc) if i remember rightly -- it's been a couple months since i read the spec; beyond that I think the message body can contain newlines without the carriage return

tsmethurst: Oh no look, you're right: https://datatracker.ietf.org/doc/html/rfc2822#section-2.3

Let's keep the conversion of newlines to CRLF then, good one

PR #372, opened 2022-02-01
Fix start command invocation

pwoolcoc:
When I tried to start the server using the command `./gotosocial server start --config-file ./config.yaml`, I got the following error:

```
root@gotosocial-experiment:/gotosocial# ./gotosocial server start --config-file ./config.yaml
Incorrect Usage: flag provided but not defined: -config-file

NAME:
   gotosocial server start - start the gotosocial server

USAGE:
   gotosocial server start [command options] [arguments...]

OPTIONS:
   --help, -h  show help (default: false)

FATA[0000] flag provided but not defined: -config-file
```

Putting the flag after the `./gotosocial` but before the `server start` seems to work properly

### DIFF
--- a/docs/installation_guide/binary.md
+++ b/docs/installation_guide/binary.md
@@ -63,7 +63,7 @@ You can now run the binary.
 Start the GoToSocial server with the following command:
 
 ```bash
-./gotosocial server start --config-path ./config.yaml
+./gotosocial --config-path ./config.yaml server start
 ```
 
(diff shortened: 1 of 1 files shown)


tsmethurst:
The docs are ahead of the release unfortunately: on main, we already redid a bunch of the CLI stuff to make it so you can put the flags in practically any order, so in the next release the docs will be correct. But i'll pull this in anyway, because for the current release it's indeed bum that the flags are incorrectly documented.

Thanks for the PR!

PR #398, opened 2022-02-15
[bugfix] Only mark cookies as Secure on https

shadowfacts:
Fixes cookies not being stored/sent by Safari when serving over plain http


tsmethurst:
Thanks!

Issue #401, opened 2022-02-16
[feature] Stream re-cached media to the user


f0x52:
on recache of media that's been uncached / removed from cache, we should tee it from the io.reader of the remote instance into storage and also back to the user that requested the recache, finishing the request seamlessly.

Issue #393, opened 2022-02-13
[feature] Clean up remote media older than x days to save storage space


tsmethurst:
Right now we keep media downloads in storage indefinitely.

To avoid ballooning storage unnecessarily, we should:

- implement a mechanism in the media manager for removing stored attachments older than x days (make this configurable with a sensible default: 30 days?)
- make sure these can be seamlessly downloaded again later if someone wants to view the attachment

Considerations:

If we don't delete the attachment from the database entirely, we can keep the blurhash around and serve it while we download media in the background (masto does this iirc).

If we delete the attachment from the database entirely, we need to make sure that when it's recreated, it either uses the same ID as before, or the status it belonged to is updated: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/gtsmodel/status.go#L33

The functionality for cleaning media should be able to be called from the CLI so that admins can trigger it manually if they want to, although adding it to the CLI is not necessarily in scope of this issue and can be done separately.

PR #407, opened 2022-02-21
[feature] Clean up/uncache remote media

tsmethurst:
This PR adds automatic cleanup of cached remote media, which runs every night at midnight unless configured otherwise. The length of time to keep media in the cache is configurable, and defaults to 30 days. In other words, media older than 30 days will be removed from storage every night, to help save drive space.

Importantly, the underlying database entries for the attachments are not deleted on cleanup, which allows the media to be fetched easily again if a caller triggers a recache (by looking through old media, for example).

When media is recached, it is streamed from the remote instance simultaneously into storage and to the caller who triggered the request, using tee. This means that callers don't have to wait for media to be processed by our instance, it will be streamed to them live as it's processed.

closes https://github.com/superseriousbusiness/gotosocial/issues/393

closes https://github.com/superseriousbusiness/gotosocial/issues/401

### Database Migration

In order to make uncacheing and recacheing work, another field needed to be added to the `gtsmodel.MediaAttachment` struct, to indicate whether an attachment is currently cached or not. A database migration is included in this pull request, which has been tested on both sqlite and postgres, and seems to work OK. Instance admins shouldn't have to take any manual steps to update their database.

### Extra Config

An extra variable has been added to configuration: `media-remote-cache-days`. This allows the user to customize how long remote media should be cached for. The variable has been added to documentation and tests, and has a sensible default so most admins won't need to touch it.


tsmethurst:
From testingtesting123.xyz:

```
time=2022-03-02T00:00:00Z level=info msg=media manager cron logger: wake[now 2022-03-02 00:00:00.000282959 +0000 UTC]
time=2022-03-02T00:00:00Z level=info msg=media manager cron logger: run[now 2022-03-02 00:00:00.000282959 +0000 UTC entry 1 next 2022-03-03 00:00:00 +0000 UTC]
time=2022-03-02T00:00:00Z level=info msg=PruneRemote: pruning media older than 2022-02-23 00:00:00.000553085 +0000 UTC m=-569899.870261660
time=2022-03-02T00:00:01Z level=info msg=PruneRemote: finished pruning remote media: pruned 80 entries
time=2022-03-02T00:00:01Z level=info msg=media manager: pruned 80 remote cache entries in 1.732744963s
```

Storage size before: 131M
Storage size after: 106M

PR #414, opened 2022-02-27
Multi arch with buildx

0hlov3:
So after doing some more tests, it looks like you can leave the Dockerfile as it is.
`goreleaser release --rm-dist --snapshot --debug` = "release succeeded after 47.95s" :)


0hlov3:
I see, there is an error in .drone.yaml.

PR #416, opened 2022-02-27
Creates Docker documentation and docker-compose.yaml

### LINKED ISSUE #190: Add docker deployment documentation

0hlov3:
This is the first Version of a Docker Documentation. Suggestions for improvement and questions are welcome. :)
Fixes #190

### DIFF
--- a/docs/installation_guide/docker.md
+++ b/docs/installation_guide/docker.md
@@ -1,3 +1,181 @@
 # Docker
 
-TODO
+The Official GoToSocial docker images are provided through [docker hub](https://hub.docker.com/r/superseriousbusiness/gotosocial "docker hub gotosocial").
+
+Currently the docker images only support the amd64 architechture. In the future we will add support for more architechtures. For now if you wish to run GoToSocial on, for example, a Raspberry Pi, you may [build the docker image yourself](https://github.com/ForestJohnson/gotosocial/compare/main...forest-jank-multi-arch-docker-build)
+
+GoToSocial can be configured using [Environment Variables](../configuration/index.md#environment-variables) if you wish, allowing your GoToSocial configuration to be embedded inside your docker container configuration.
+
+## Run with Docker Compose (recommended)
+Assuming you will be using something like [docker compose](https://docs.docker.com/compose/ "Docker Compose Docs") to configure your GoToSocial docker container, you might want to follow the next Steps.
+
+### Create a Working Dir
+You need a Working Directory in which the data of the PostgreSQL and the GoToSocial container will be located, so create this directory for example with the following command. 
+The directory can be located where you want it to be later.
+
+```shell
+mkdir -p /docker/gotosocial
+cd /docker/gotosocial
+```
+### Get the latest docker-compose.yaml and config.yaml
+You can get an example [docker-compose.yaml](../../example/docker-compose/docker-compose.yaml "Example docker-compose.yaml") and [config.yaml](../../example/config.yaml "Example config.yaml") here, which you can download with wget.
+
+```shell
+wget https://raw.githubusercontent.com/superseriousbusiness/gotosocial/main/example/docker-compose/docker-compose.yaml
+wget https://raw.githubusercontent.com/superseriousbusiness/gotosocial/main/example/config.yaml
+```
+
+### Edit the docker-compose.yaml
+You can modify the docker-compose.yaml to our needs, but in any case you should generate a Postgres password and bind it as environment variable into the postgreSQL container. For this we can write the password directly into the docker-compose.yaml like in the example or we create an [.env file](https://docs.docker.com/compose/environment-variables/#the-env-file "Docker Docs") with which we load the environment variables into the container.
+
+```shell
+$EDITOR docker-compose.yaml
+```
+### Edit the config.yaml
+When we want to use the config.yaml, we should make the following changes to config.yaml.
+| Config Option   | Value  |
+| --------------- | ------ |
+| host            | Hostname of your Inctanse e.g. gts.example.com |
+| account-domain  | Domain to use when federating profiles e.g. gts.example.com |
+| trusted-proxies | We need to trust our hostmashine and the Docker Network e.g.<br>- "127.0.0.1/32"<br>- "10.0.0.0/8"<br>- "172.16.0.0/12"<br>- "192.168.0.0/16" |
+| db-address      | gotosocial_postgres |
+| db-user         | gotosocial |
+| db-password     | same password as postgres environment $POSTGRES_PASSWORD |
+
+```shell
+$EDITOR config.yaml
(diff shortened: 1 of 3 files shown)


example/docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
+version: "3.3"
+
+services:
+  gotosocial:
+    image: superseriousbusiness/gotosocial:0.2.0
+    networks:
+      - gotosocial
+    ports:
+     - "127.0.0.1:8080:8080" # If you don't want to use the live feature, you can comment this line
```
tsmethurst, line 9: imo the comment is not necessary here, there are v. few cases where people wouldn't want to expose the API

example/docker-compose/docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
+version: "3.3"
+
+services:
```
tsmethurst, line 3: You can also specify user 1000:1000 here I think; that's something docker-compose can do right?

docs/installation_guide/docker.md
```diff
@@ -1,3 +1,181 @@
...
+
+```
+GTS_PORT=8080
+GTS_PROTOCOL=https
+GTS_TRUSTED_PROXIES=0.0.0.0/0
+GTS_HOST=gotosocial.example.com
+GTS_ACCOUNT_DOMAIN=gotosocial.example.com
+GTS_DB_TYPE=sqlite
+GTS_DB_ADDRESS=/gotosocial/database/sqlite.db
+GTS_STORAGE_SERVE_PROTOCOL=https
```
tsmethurst, line 112: check whether these STORAGE vars are still necessary: we removed some of them but I can't remember which ones

docs/installation_guide/docker.md
```diff
@@ -1,3 +1,181 @@
...
+  <summary>Example .env File</summary>
+
+```shell
+$EDITOR .env
+```
+
+```
+GTS_PORT=8080
+GTS_PROTOCOL=https
+GTS_TRUSTED_PROXIES=0.0.0.0/0
```
tsmethurst, line 107: careful... this is not a good setting to recommend

docs/installation_guide/docker.md
```diff
@@ -1,3 +1,181 @@
...
+```shell
+docker-compose up -d
+```
+
+### Create your first User
+First of all, we have to find the right Container ID so we will check the Container ID with `docker ps` or if we got tons of containers running `docker ps -f NAME=gotosocial`.
+If we execute the command like this, we will get an output similar to the following:
+
+```shell
+CONTAINER ID   IMAGE                                      COMMAND                  CREATED          STATUS          PORTS                      NAMES
```
tsmethurst, line 59: If we name the containers explicitly in the docker compose, then the user won't even have to look for the container ID first; they'll just be able to run docker exec -ti gotosocial etc

docs/installation_guide/docker.md
```diff
@@ -1,3 +1,181 @@
...
+```shell
+$EDITOR docker-compose.yaml
+```
+### Edit the config.yaml
+When we want to use the config.yaml, we should make the following changes to config.yaml.
+| Config Option   | Value  |
+| --------------- | ------ |
+| host            | Hostname of your Inctanse e.g. gts.example.com |
+| account-domain  | Domain to use when federating profiles e.g. gts.example.com |
+| trusted-proxies | We need to trust our hostmashine and the Docker Network e.g.<br>- "127.0.0.1/32"<br>- "10.0.0.0/8"<br>- "172.16.0.0/12"<br>- "192.168.0.0/16" |
```
tsmethurst, line 40: *host machine

docs/installation_guide/docker.md
```diff
@@ -1,3 +1,181 @@
 # Docker
 
-TODO
+The Official GoToSocial docker images are provided through [docker hub](https://hub.docker.com/r/superseriousbusiness/gotosocial "docker hub gotosocial").
+
+Currently the docker images only support the amd64 architechture. In the future we will add support for more architechtures. For now if you wish to run GoToSocial on, for example, a Raspberry Pi, you may [build the docker image yourself](https://github.com/ForestJohnson/gotosocial/compare/main...forest-jank-multi-arch-docker-build)
```
tsmethurst, line 5: Maybe we should already remove this line, since the next release will have multi-arch docker builds

docs/installation_guide/docker.md
0hlov3: Checked the storage options and was able to remove some.

docs/installation_guide/docker.md
0hlov3: Edited the Trusted Proxies to the same from before (Docker-Compose) Documentation.

docs/installation_guide/docker.md
0hlov3: The Problem is, that docker compose does something like a Kubernetes deployment and adds stuff to the name like ` docker-gotosocial-1` if someone had tried to start a container with the name `docker-gotosocial-1` before, compose will rename the container to something like `docker2-gotosocial-1`

```shell
❯ docker ps
CONTAINER ID   IMAGE                                   COMMAND                  CREATED          STATUS          PORTS                      NAMES
0639ddf9d278   superseriousbusiness/gotosocial:0.2.0   "/gotosocial/gotosoc…"   6 seconds ago    Up 5 seconds    127.0.0.1:8082->8080/tcp   docker2-gotosocial-1
0d07527eb59a   postgres:14-alpine                      "docker-entrypoint.s…"   29 seconds ago   Up 28 seconds   5432/tcp                   docker2-gotosocial_postgres-1
a77238866cf0   superseriousbusiness/gotosocial:0.2.0   "/gotosocial/gotosoc…"   15 minutes ago   Up 12 minutes   127.0.0.1:8080->8080/tcp   docker-gotosocial-1
3400147dcc55   postgres:14-alpine                      "docker-entrypoint.s…"   15 minutes ago   Up 13 minutes   5432/tcp                   docker-gotosocial_postgres-1
```

I'll check how to describe that another way. :)

docs/installation_guide/docker.md
0hlov3: Fixed Typo.

docs/installation_guide/docker.md
0hlov3: You are right... removed the line.

example/docker-compose/docker-compose.yaml
0hlov3: Right, I tested that for now. 
Should be working.

docs/installation_guide/docker.md
0hlov3: So I did a workaround in the description. Please tell me what you think. :O

PR #424, opened 2022-03-12
Adds two new test images for media processing

0hlov3:
@tsmethurst Here are the two test Images.


0hlov3:
Images for reproduce the Bug in #425

tsmethurst:
Thanks! I'm gonna leave the PR open rather than pulling it in, and include the images as a test case in a fix.

PR #441, opened 2022-03-26
Better nginx config

OniriCorpe:
Hi!

The [nginx config in the doc](https://docs.gotosocial.org/en/latest/installation_guide/nginx/) is not good 😬 

I suggest this new configuration file which has good security practices and all basic stuff like TLS support and http to https redirect

### DIFF
--- a/example/nginx.conf
+++ b/example/nginx.conf
@@ -22,18 +22,23 @@ server {
 
     location / {
 
+        # EDITME: please edit '8080' with your choosen port for gotosocial
+        proxy_pass http://localhost:8080;
+
+        # you can change this if you want to upload larger files (Default: 50M)
+        client_max_body_size 50M;
+
+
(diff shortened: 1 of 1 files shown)


tsmethurst (changes requested):
For this one, we should firstly make sure that none of the config here interferes with the running of GoToSocial (for example, GoToSocial uses gzip as configured in the gin router -- will this prevent that from working?), and also update the documentation here in the same pull request: https://github.com/superseriousbusiness/gotosocial/blob/main/docs/installation_guide/nginx.md

Can you take a look at those things?

tsmethurst:
Heya, since the docs have been updated with some links to documents about hardening nginx, i'm gonna close this one

OniriCorpe:
no problem! :)

I didn't have the time to go back into it

Issue #453, opened 2022-04-14
[bug] can't follow mastodon accounts: Unauthorized


igalic:
trying to follow `@meena@glitch.social` from [`@me@soc.eena.me`](https://soc.eena.me/@me):

```
Apr 14 20:02:20 social gotosocial[28943]: time=2022-04-14T20:02:20Z level=info msg=authentication not passed for public key owner https://glitch.social/actor; signature value was 'keyId="https://glitch.social/actor#main-key",algorithm="rsa-sha256",headers="(request-target) host date accept",signature="I3v2a1AuO0hVB3fib83VZARhUL3BO9HXbdvtHaq8L2vqu5gYWUy1z64Y7JNJ2CX1WQauZ+J+49fx3eK5BnU3O+bEJo1B6r4VDW7mLVzRemzHvQOLMgySLXtncn16P2PriimqQ1Ejn/q1B8O//POcs0n7Ot7z2pc4Uu3BMTY9eek7u6zv1hfXEoMavROwdyvJ2vBTxc6+1If185OVIJeVkOMivtufYRNkl1w29EaQVVsB0aRvL+oHm16+a5cCHV7MHnuEtyvOk0Ognt1n4/LtRPfp/VKpDKznzqmrGRA45UdzQncB8DpbB6OEQtl+Cd+kigbIQ6Ja+9NzAZyFR86Abw=="' func=AuthenticateFederatedRequest
Apr 14 20:02:20 social gotosocial[28943]: time=2022-04-14T20:02:20Z level=info msg=not authorized func=UsersGETHandler url=/users/me
Apr 14 20:02:20 social gotosocial[28943]: time=2022-04-14T20:02:20Z level=info msg=[8.08787ms] Unauthorized: wrote 54 bytes clientIP=192.168.17.10 latency=8.08787ms method=GET path=/users/me statusCode=401 userAgent=http.rb/4.4.1 (Mastodon/3.4.6; +https://glitch.social/)
Apr 14 20:02:20 websrv2-hel1 /usr/local/bin/gotosocial[28944]: time=2022-04-14T20:02:20Z level=error msg=batch deliver had at least one failure: POST request to https://glitch.social/users/meena/inbox failed (401): 401 Unauthorized 
Apr 14 20:02:20 social gotosocial[28943]: time=2022-04-14T20:02:20Z level=error msg=batch deliver had at least one failure: POST request to https://glitch.social/users/meena/inbox failed (401): 401 Unauthorized
```

[...]

PR #455, opened 2022-04-15
docs: unify nginx explainers and add apache httpd

igalic:
there are two places where nginx + certbot is explained, unify that into one place.
Add apache httpd, following the same steps, but using mod_md for LetsEncrypt
add a note about #453 in both guides.
Link to both, and call the section reverse proxy, instead of NGINX

### DIFF
--- a/docs/installation_guide/apache-httpd.md
+++ b/docs/installation_guide/apache-httpd.md
@@ -1,0 +1,129 @@
+# Reverse proxy with Apache httpd
+
+## Requirements
+
+For this you will need Apache httpd server.
+That is a fairly popular package so your distro will probably have it.
+
+### Ubuntu
+
+```bash
+sudo apt install apache2
+```
+
+### Arch
+
+```bash
+sudo pacman -S apache
+```
+
+### OpenSuse
+
+```bash
+sudo zypper install apache2
+```
+
+## Configure GoToSocial
+
+In your GoToSocial config turn off letsencrypt.
+First open the file in your text editor.
+
+```bash
+sudoedit /gotosocial/config.yaml
+```
+
+Then set `letsencrypt-enabled: false`.
+
+If GoToSocial is already running, restart it.
+
+```bash
+sudo systemctl restart gotosocial.service
+```
+
+Or if you don't have a systemd service just restart it manually.
+
+## Set up Apache httpd
+
+First we will set up Apache httpd to serve GoToSocial as unsecured http and then later use certbot to automatically upgrade to https.
+Please do not try to use it until that's done or you'll be transmitting passwords over clear text.
+
+First we'll write a configuration for Apache httpd and put it in `/etc/apache2/sites-available`.
+
+```bash
+sudo mkdir /etc/apache2/sites-available/
+sudoedit /etc/apache2/sites-available/yourgotosocial.url.conf
+```
+
+The file you're about to create should look a bit like this:
+
+```apache
+<VirtualHost *:80>
+  ServerName example.com
+  ProxyPreserveHost On
+  ProxyPass / http://localhost:8080/
+  ProxyPassReverse / http://localhost:8080/
+</VirtualHost>
+```
+
+**Note***: `ProxyPreserveHost On` is essential: It guarantees that the proxy and the gotosocial speak of the same Server name. If not, gotosocial will build the wrong authentication headers, and all attempts at federation will be rejected with 401.
+
+Change `ProxyPass` to the ip and port that you're actually serving GoToSocial on and change `ServerName` to your own domain name.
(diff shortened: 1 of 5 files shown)


docs/installation_guide/nginx.md
```diff
@@ -109,7 +111,7 @@ sudo certbot --nginx
 ```
 
 After you do, it should have automatically edited your configuration file to enable https.
-Just reload it one last time and after that you should be good to go!
+Reload it one last time and after that you should be good to go!
```
igalic, line 114: tuning language here, removing "just"

docs/installation_guide/binary.md
```diff
@@ -138,6 +138,6 @@ sudo systemctl enable --now gotosocial.service
 
 At some point you'll likely want to do things like change instance information, and block domains you don't want to interact with. See the [admin panel](../admin/admin_panel.md) instructions for this.
 
-## 9. Reverse proxy with nginx (optional)
+## 9. Reverse proxy (optional)
 
-If you want to run other webservers on port 433 or simply want to add an additional layer of security you might want to [use nginx as a reverse proxy](./nginx.md).
+If you want to run other webservers on port 433 or want to add an additional layer of security you might want to use [nginx](./nginx.md) or [Apache httpd](./apache-httpd.md) as reverse proxy
```
igalic, line 143: tuning language here: removing simply

nothing in software is simple

docs/installation_guide/nginx.md
```diff
@@ -68,7 +68,9 @@ server {
 }
 ```
 
-Note: You can remove the line `listen [::]:80;` if your server is not ipv6 capable or you'd rather not use ipv6.
+**Note***: You can remove the line `listen [::]:80;` if your server is not ipv6 capable.
```
igalic, line 71: removing the bit about not wanting to bother with IPv6 👀

docs/installation_guide/docker.md
```diff
@@ -125,66 +125,6 @@ GTS_LETSENCRYPT_ENABLED=false
...
-
-  ssl_protocols             TLSv1.2 TLSv1.3;
-  ssl_prefer_server_ciphers on;
-  ssl_ciphers "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305";
-  ssl_session_timeout       1d; # defaults to 5m
-  ssl_session_cache         shared:SSL:10m; # estimated to 40k sessions
-  ssl_session_tickets       off;
-  ssl_stapling              on;
-  ssl_stapling_verify       on;
-  ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem;
```
igalic, line 174: this whole section needs to be a link to documentation somewhere on how to harden nginx' SSL config
and it shouldn't be in the docker part.

docs/installation_guide/apache-httpd.md
```diff
@@ -0,0 +1,129 @@
+# Reverse proxy with Apache httpd
```
igalic, line 1: this whole document is basically copy-paste from your nginx document.

docs/installation_guide/apache-httpd.md
```diff
@@ -0,0 +1,129 @@
...
+
+Everything working? Great! Then restart Apache httpd to load your new config file.
+
+```bash
+sudo systemctl restart apache2
+```
+
+## Setting up SSL with mod_md
+
+To setup Apache httpd with mod_md, we'll need to load the module and then modify our vhost files:
```
igalic, line 101: except for this section.

igalic:
i don't know why the build is failing
(it wasn't me)

docs/installation_guide/docker.md
tsmethurst: Agreed it shouldn't be in the docker section, however I think instead of totally deleting it we should merge it with the existing nginx example in the nginx.md document. Can you do that?

tsmethurst (changes requested):
Looks good, just the comment I added, and also you still need to add this to the mkdocs.yml in the root of the repo so it appears in the documentation properly.

igalic:
i think i can, and i can use #441 for picking out some details and pointers
might take a while *\*points at toddler\**

tsmethurst:
No rush :)

igalic:
added mkdocs entry
and restored big nginx.conf and i have absolutely nothing to add to it, because well, i still don't know how nginx works.

PR #465, opened 2022-04-18
[feature] Add `log-db-queries` config option

tsmethurst:
The added option allows users to enable/disable verbose logging of database queries. Perfect for situations where you want to enable debug or trace logging, but you're not looking for database issues and don't want to be distracted by wicked verbose logs. Defaults to 'false'.


igalic:
nice!

PR #466, opened 2022-04-18
[bugfix] Use our own (Batch)Deliver implementation for federated messages

tsmethurst:
We had a code path for this before, but it wasn't being properly used. This PR makes sure it is, which *also* makes sure that we don't do silly things like attempt http delivers to our own addresses, which is a total waste of time.

### DIFF
--- a/internal/transport/deliver.go
+++ b/internal/transport/deliver.go
@@ -20,7 +20,10 @@ package transport
 
 import (
 	"context"
+	"fmt"
 	"net/url"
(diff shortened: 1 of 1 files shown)


tsmethurst:
Update on this: fixing this also closed the above issues. Explanation: when the instance was trying to deliver messages to itself, it was holding http connections and locks open. Preventing the instance from trying to deliver to itself fixes that.

PR #470, opened 2022-04-19
[documentation] Add WebSocket protocol documentation

### LINKED ISSUE #463: [documentation] Docs don't say whether gotosocial use websocket
but the error logs heavily imply it:
```
Apr 16 21:53:22 social gotosocial[36272]: time=2022-04-16T21:53:22Z level=info msg=error upgrading websocket connection: websocket: the client is not using the websocket protocol: 'upgrade' token not found in 'Connection' header func=github.com/superseriousbusiness/gotosocial/internal/api/client/streaming.(*Module).StreamGETHandler file=github.com/superseriousbusiness/gotosocial/internal/api/client/streaming/stream.go:145 fields.func=StreamGETHandler func=github.com/superseriousbusiness/gotosocial/internal/api/client/streaming.(*Module).StreamGETHandler
```

tsmethurst:
This PR adds a document explaining which port, protocol, and headers are used by GoToSocial for making WebSocket connections.

closes #463

### DIFF
--- a/docs/installation_guide/websocket.md
+++ b/docs/installation_guide/websocket.md
@@ -1,0 +1,43 @@
+# WebSocket
+
+GoToSocial uses the secure [WebSocket protocol](https://en.wikipedia.org/wiki/WebSocket) (aka `wss`) to allow for streaming updates of statuses and notifications via client apps like Pinafore.
+
+In order to use this functionality, you need to ensure that whatever proxy you've configured GoToSocial to run behind allows WebSocket connections through.
+
+The WebSocket endpoint is located at `wss://example.org/api/v1/streaming` where `example.org` is the hostname of your GoToSocial instance.
+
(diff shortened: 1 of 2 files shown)


tsmethurst:
@igalic Should anything else be added?

igalic (commented):
this is a good start
i think we need to extended the proxy docs as well

this is the lounge configuration,  https://thelounge.chat/docs/guides/reverse-proxies

does the Websocket thing also rely on the same Hostname?

tsmethurst:
I'm on it with extending the proxy docs :) working on nginx now, i'll tag you in that PR too when I'm done with it, then we can merge them at the same time

> does the Websocket thing also rely on the same Hostname?

yes I believe so

tsmethurst:
Here's the nginx pr: https://github.com/superseriousbusiness/gotosocial/pull/471

igalic (approved):
looks good

PR #471, opened 2022-04-19
[documentation] NGINX docs updates

tsmethurst:
Some updates to the NGINX documentation.

* Add websockets options.
* Replace explicit security hardening with a recommendation.
* Stop GoToSocial while the procedure is happening.

### DIFF
--- a/docs/installation_guide/nginx.md
+++ b/docs/installation_guide/nginx.md
@@ -81,7 +81,7 @@ If you're running GoToSocial on another machine with the local ip of 192.168.178
 
 **Note**: `proxy_set_header Host $host;` is essential. It guarantees that the proxy and GoToSocial use the same server name. If not, GoToSocial will build the wrong authentication headers, and all attempts at federation will be rejected with 401.
 
-**Note**: The `Connection` and `Upgrade` headers are used for WebSocket connections.
+**Note**: The `Connection` and `Upgrade` headers are used for WebSocket connections. See the [WebSocket docs](./websocket.md).
 
 Next we'll need to link the file we just created to the folder that nginx reads configurations for active sites from.
 
--- a/docs/installation_guide/nginx.md
+++ b/docs/installation_guide/nginx.md
@@ -26,32 +26,34 @@ sudo zypper install nginx python3-certbot python3-certbot-nginx
 
 ## Configure GoToSocial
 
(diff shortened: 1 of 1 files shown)


docs/installation_guide/nginx.md
```diff
@@ -25,64 +26,72 @@ sudo zypper install nginx python3-certbot python3-certbot-nginx
...
-First open the file in your text editor.
+If GoToSocial is already running, stop it.
 
 ```bash
-sudoedit /gotosocial/config.yaml
+sudo systemctl stop gotosocial
 ```
 
-Then set `letsencrypt-enabled: false`.
+Or if you don't have a systemd service just stop it manually.
```
igalic, line 35: just?
how?
`Ctrl+C` ? `pkill`?

docs/installation_guide/nginx.md
```diff
@@ -25,64 +26,72 @@ sudo zypper install nginx python3-certbot python3-certbot-nginx
...
 
-Change `proxy_pass` to the ip and port that you're actually serving GoToSocial on and change `server_name` to your own domain name.
-If your domain name is `gotosocial.example.com` then `server_name gotosocial.example.com;` would be the correct value.
 If you're running GoToSocial on another machine with the local ip of 192.168.178.69 and on port 8080 then `proxy_pass http://192.168.178.69:8080;` would be the correct value.
 
+**Note**: You can remove the line `listen [::]:80;` if your server is not ipv6 capable.
+
+**Note**: `proxy_set_header Host $host;` is essential. It guarantees that the proxy and GoToSocial use the same server name. If not, GoToSocial will build the wrong authentication headers, and all attempts at federation will be rejected with 401.
+
+**Note**: The `Connection` and `Upgrade` headers are used for WebSocket connections.
```
igalic, line 84: We can later link to the websocket doc here.

igalic (changes requested):
looks good.
just minor change.

docs/installation_guide/nginx.md
tsmethurst: I'll do that before I pull it in

docs/installation_guide/nginx.md
tsmethurst: Hmm, I guess telling them how to stop it is out of scope here if we don't know how they started it :P What do you think?

PR #480, opened 2022-04-24
[feature] Implement media v2 endpoint to accommodate Tusky 17

### LINKED ISSUE #462: [bug] can't upload images
when trying to upload a 1 MB image to my instance, it fails after about 600 bytes:

```
Apr 16 16:24:52 social gotosocial[36272]: time=2022-04-16T16:24:52Z level=info msg=[5.80325ms] Not Found: wrote 632 bytes func=github.com/superseriousbusiness/gotosocial/internal/router.loggingMiddleware.func1 file=github.com/superseriousbusiness/gotosocial/internal/router/logger.go:65 clientIP=192.168.17.10 latency=5.80325ms method=POST path=/api/v2/media statusCode=404 userAgent=Tusky/17.0 Android/10 OkHttp/4.9.3
Apr 16 16:24:53 webserver httpd[37201]: [proxy:error] [pid 37201:tid 35124299520] (32)Broken pipe: [client 95.223.76.65:46035] AH01084: pass request body failed to 192.168.17.40:8080 (social)
Apr 16 16:24:53 webserver httpd[37201]: [proxy_http:error] [pid 37201:tid 35124299520] [client 95.223.76.65:46035] AH01097: pass request body failed to 192.168.17.40:8080 (social) from 95.223.76.65 ()
```

i don't think it's the proxy, all values here https://httpd.apache.org/docs/2.4/mod/mod_proxy.html like it should fit (time and space)

the image in question:
steveklabnik-FQaWvpKXsAEGdr--01

i've had similar trouble with other jpegs before, tho they might have been a lot bigger

tsmethurst:
This PR 'implements' the /api/v2/media endpoints to accommodate their use by Tusky version 17 (see https://github.com/tuskyapp/Tusky/issues/2409). The implementation is really just a copy: the same handlers serve both v1 and v2 of the API because it's prettymuch identical.

Closes #462 (tested and working with tusky 17)

### DIFF
--- a/internal/api/client/media/mediacreate_test.go
+++ b/internal/api/client/media/mediacreate_test.go
@@ -150,7 +150,7 @@ func (suite *MediaCreateTestSuite) TestMediaCreateSuccessful() {
 	if err != nil {
 		panic(err)
 	}
-	ctx.Request = httptest.NewRequest(http.MethodPost, fmt.Sprintf("http://localhost:8080/%s", mediamodule.BasePath), bytes.NewReader(buf.Bytes())) // the endpoint we're hitting
+	ctx.Request = httptest.NewRequest(http.MethodPost, fmt.Sprintf("http://localhost:8080/%s", mediamodule.BasePathV1), bytes.NewReader(buf.Bytes())) // the endpoint we're hitting
 	ctx.Request.Header.Set("Content-Type", w.FormDataContentType())
 	ctx.Request.Header.Set("accept", "application/json")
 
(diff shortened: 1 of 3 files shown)


internal/api/client/media/mediacreate_test.go
```diff
@@ -234,7 +234,7 @@ func (suite *MediaCreateTestSuite) TestMediaCreateLongDescription() {
 	if err != nil {
 		panic(err)
 	}
-	ctx.Request = httptest.NewRequest(http.MethodPost, fmt.Sprintf("http://localhost:8080/%s", mediamodule.BasePath), bytes.NewReader(buf.Bytes())) // the endpoint we're hitting
+	ctx.Request = httptest.NewRequest(http.MethodPost, fmt.Sprintf("http://localhost:8080/%s", mediamodule.BasePathV1), bytes.NewReader(buf.Bytes())) // the endpoint we're hitting
 	ctx.Request.Header.Set("Content-Type", w.FormDataContentType())
```
igalic, line 238: what about tests for V2?

igalic (commented):
❓

internal/api/client/media/mediacreate_test.go
tsmethurst: They just call the same handlers underneath :P So it's just duplication to do that

PR #491, opened 2022-04-25
Add support for running profiling when debug build-tags provided

NyaaaWhatsUpDoc:
wrap root HTTP handler in debug.WithPprof(), rearrange router.Start() to support this


NyaaaWhatsUpDoc:
If you run `script/build.sh` with `DEBUG=1` or any non-empty value then it will compile gotosocial with an extra build-tag debugenv.

What this will mean is that if you run gotosocial with the environment variable `DEBUG=1` or again, any non-empty value, profiling will be active and available under "/debug/pprof".

tsmethurst (approved):
Looks good to me! Don't forget to add your library dependencies to the README.md if you want to :) If you'd rather not, then go ahead and merge

NyaaaWhatsUpDoc:
> Looks good to me! Don't forget to add your library dependencies to the README.md if you want to :) If you'd rather not, then go ahead and merge

ah yes i shall do! is it alright if i just use this PR to add all the libraries of mine that are missing?

PR #493, opened 2022-04-26
[bugfix] Trim log entries to 1700 chars before they enter syslog

### LINKED ISSUE #454: [bug] some [trace] log messages are too long for syslog
turned up logging to trace, and now i'm seeing a few of these:
```
Apr 15 14:15:49 social gotosocial[36272]: Failed to fire hook: write unixgram ->/isolates/social/root/var/run/log: write: message too long
```
maybe this is a logrus bug, and should be handled there?
as in: messages > syslog RFC limit should be split.

tsmethurst:
This PR adds a wrapper around the syslog logrus hook which trims the `Message` part of incoming `*logrus.Entry`s to no more than 1700 bytes, to avoid overloading syslog with v. long trace messages.

closes #454 (i hope)


igalic:
i think in 2022, we can be more generous, and go with the SHOULD length of 2048:
https://www.rfc-editor.org/rfc/rfc5424.html#section-6.1

tsmethurst:
Hmm that's fair. The reason I took 512 was because the trimmer only shortens the actual message; the other metadata that logrus prints is still attached, including any custom fields. So the actual message would end up being 512 + metadata... How about meeting near the middle at something like 1700, so it still leaves room for ~ 300 bytes of metadata?

tsmethurst:
^^ trimmed to 1700. Whaddya think?

igalic (approved):
This looks sensible

but note: i haven't had time to test this yet.

PR #494, opened 2022-04-26
[chore] Signature check updates

tsmethurst:
This PR updates some of the signature checking/authentication functionality in two ways:

1. Make the signature for the authorization function simpler (drop the bool), and have it return an error with code instead so that codes logged and returned to callers are more informative.
2. Change the order in which http signature algorithms are checked, to put the most common -- rsa_sha256 -- first. This should same some time/cpu since it's more likely to pass first without having to check more than one algorithm :muscle:


igalic:
cool!

PR #497, opened 2022-04-27
replace async client API / federator msg processing with worker pools

NyaaaWhatsUpDoc:
By replacing the existing asynchronous client API / federator message passing with worker pools, it allows for fine-tuning performance when it comes to processing. for the moment it just follows GOMAXPROCS for worker count (and determines max queued messages from this also). this is to help stop instances getting bogged down if they're say dealing with a deluge of incoming federated AP messages

### DIFF
--- a/cmd/gotosocial/action/server/server.go
+++ b/cmd/gotosocial/action/server/server.go
@@ -63,6 +63,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/federation/federatingdb"
 	"github.com/superseriousbusiness/gotosocial/internal/gotosocial"
 	"github.com/superseriousbusiness/gotosocial/internal/media"
+	"github.com/superseriousbusiness/gotosocial/internal/messages"
 	"github.com/superseriousbusiness/gotosocial/internal/oauth"
 	"github.com/superseriousbusiness/gotosocial/internal/oidc"
 	"github.com/superseriousbusiness/gotosocial/internal/processing"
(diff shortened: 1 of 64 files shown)


igalic:
i'd like to see a bit more elaboration in the pr / commit message as to what the benefits are going to be?

just looking at the code it looks like better encapsulation.

NyaaaWhatsUpDoc:
> i'd like to see a bit more elaboration in the pr / commit message as to what the benefits are going to be?
> 
> just looking at the code it looks like better encapsulation.

totally fair! it's been something I've been discussing heavily in the matrix so maybe left my commentary here a little bare bones...

it encapsulates things better and allows for fine-tuning performance when it comes to federator and client API message processing. for the moment it just follows GOMAXPROCS for worker count (and determines max queued messages from this also). this is to help stop instances getting bogged down if they're say dealing with a deluge of incoming federated AP messages

igalic:
sounds really good!
could you amend the PR &  commit message?

NyaaaWhatsUpDoc:
> sounds really good! could you amend the PR & commit message?

done :)

tsmethurst (commented):
Overall this looks awesome.

One question, should we change the logic here: https://github.com/NyaaaWhatsUpDoc/gotosocial/blob/processor-worker-pools/internal/media/manager.go#L111 to also use the new internal/worker package?

NyaaaWhatsUpDoc:
> Overall this looks awesome.
> 
> One question, should we change the logic here: https://github.com/NyaaaWhatsUpDoc/gotosocial/blob/processor-worker-pools/internal/media/manager.go#L111 to also use the new internal/worker package?

i was thinking we could maybe make that another issue -- it may take a little refactoring the way it's been setup (i think..) so would increase the scope of this PR quite a bit.

tsmethurst:
> i was thinking we could maybe make that another issue -- it may take a little refactoring the way it's been setup (i think..) so would increase the scope of this PR quite a bit.

okay no problem! feel free to merge this as-is then :) <3

PR #507, opened 2022-04-29
[feature] allow absolute path to templates

### LINKED ISSUE #411: [tests/packaging] Router testrig has no way to set absolute directory for templates
There's always a path.Join with the `projectRoot` (https://github.com/superseriousbusiness/gotosocial/blob/main/testrig/router.go#L51), but there's no way to override that when that projectRoot is detected wrong.

It's an issue for the NixOS build, where that somehow gets set to `/build/source/internal/api/client/auth/github.com/superseriousbusiness/gotosocial` while running `go test ./internal/api/client/auth`. While the correct `web` directory is `/build/source/web`

terinjokes:
Only attempt to join the `WebTemplateBaseDir` with the current working
directory if the user has not configured an absolute path to the
template directory.

This changeset also makes a similar change to the testrig, allowing
tests to configure an alternative location for the templates directory.

Fixes #411

Signed-off-by: Terin Stock <terinjokes@gmail.com>

### DIFF
--- a/internal/email/util.go
+++ b/internal/email/util.go
@@ -28,14 +28,16 @@ import (
 )
 
 func loadTemplates(templateBaseDir string) (*template.Template, error) {
-	cwd, err := os.Getwd()
-	if err != nil {
-		return nil, fmt.Errorf("error getting current working directory: %s", err)
+	if !filepath.IsAbs(templateBaseDir) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return nil, fmt.Errorf("error getting current working directory: %s", err)
+		}
+		templateBaseDir = filepath.Join(cwd, templateBaseDir)
 	}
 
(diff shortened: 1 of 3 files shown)


tsmethurst:
Thanks!

tsmethurst:
Oh no! This looks like it introduced a regression with our docker-compose. Could you have a look at https://github.com/superseriousbusiness/gotosocial/issues/511? I should have tested it myself with docker-compose but I didn't :upside_down_face:

terinjokes:
Oh weird, I tested locally with and without absolute paths, but I must have missed something. My apologies!

I can take a look this weekend, fill free to revert in the meantime.

tsmethurst:
This weekend is fine! Things are allowed to be a little bit broken on main, as a treat

Issue #520, opened 2022-05-01
[performance] Avoid using `LOWER()` in database queries


tsmethurst:
We currently use LOWER() quite liberally to make case-insensitive selects, but this prevents the database from using indexes to select on, leading to slow queries. We should:

1. Remove LOWER() from hot codepaths.
2. Write a database migration script to just lowercase all rows of columns where we need to do case-insensitive selects.

tsmethurst:
It turns out that step 2 is unnecessary, as long as we're careful when we do selects.

PR #523, opened 2022-05-01
[performance] Speed up some of the slower db queries

tsmethurst:
This PR:

1. Removes use of the `LOWER()` db function where it wasn't necessary (this function prevented the db indexes from working properly).
2. Swaps the db exists function from using `COUNT(*)` to using `EXISTS`, which should be faster.
3. Reworks the blocked function to always present params in the order that's been indexed.
4. Adds a warning level log when db queries take longer than 1s to help debugging.

closes https://github.com/superseriousbusiness/gotosocial/issues/520

### DIFF
--- a/internal/db/bundb/status.go
+++ b/internal/db/bundb/status.go
@@ -70,7 +70,7 @@ func (s *statusDB) GetStatusByID(ctx context.Context, id string) (*gtsmodel.Stat
 			return s.cache.GetByID(id)
 		},
 		func(status *gtsmodel.Status) error {
-			return s.newStatusQ(status).Where("LOWER(status.id) = LOWER(?)", id).Scan(ctx)
+			return s.newStatusQ(status).Where("status.id = ?", id).Scan(ctx)
 		},
 	)
 }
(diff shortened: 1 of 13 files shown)


igalic:
doesn't this still need migrations for usernames?

tsmethurst:
Aha see the addition I made to the issue. I just realized that all the usernames we use for accounts on a GtS instance are lowercase, so it should be fine.

internal/db/bundb/status.go
```diff
@@ -70,7 +70,7 @@ func (s *statusDB) GetStatusByID(ctx context.Context, id string) (*gtsmodel.Stat
 			return s.cache.GetByID(id)
 		},
 		func(status *gtsmodel.Status) error {
-			return s.newStatusQ(status).Where("LOWER(status.id) = LOWER(?)", id).Scan(ctx)
+			return s.newStatusQ(status).Where("status.id = ?", id).Scan(ctx)
```
NyaaaWhatsUpDoc, line 73: do you need to do a `strings.ToLower(id)` etc on all of these changes in this file for the query argument?

internal/db/bundb/status.go
```diff
@@ -82,7 +82,7 @@ func (s *statusDB) GetStatusByURI(ctx context.Context, uri string) (*gtsmodel.St
 			return s.cache.GetByURI(uri)
 		},
 		func(status *gtsmodel.Status) error {
-			return s.newStatusQ(status).Where("LOWER(status.uri) = LOWER(?)", uri).Scan(ctx)
+			return s.newStatusQ(status).Where("status.uri = ?", uri).Scan(ctx)
```
NyaaaWhatsUpDoc, line 85: Where statuses are put into the database, we should probably force a `strings.ToLower(...)` on their URI/URL, and do the same for the incoming parameter coming in here (unless you've already included that somewhere else)

PR #528, opened 2022-05-02
[bugfix] Fix panic when relative tmpl path given to router

tsmethurst:
Fixes a regression from https://github.com/superseriousbusiness/gotosocial/pull/507


tsmethurst:
No worries at all :)

PR #531, opened 2022-05-02
fix existing bio text showing as HTML

NyaaaWhatsUpDoc:
- add new gtsmodel.Account{}.NoteRaw field (and appropriate migration) to store raw bio text upon `/api/v1/accounts/update`
- when serving a sensitive account (i.e. requester = target) set the API model's source note field to Account.NoteRaw.
- update testrig models and existing tests to set and check for this new .NoteRaw field
- some small fixes to regex patterns
- add a `regexes.ReplaceAllStringFunc(rgx *regexp.Regexp, ...)` function that provides the same standard library function of this name, but provides you with a memory-pooled byte buffer for performing replacements. for improved performance
- updates to mention, link and tag replacement logic in the `internal/text` package to use this new function as described above

note: the last two bullet points were the result of earlier attempts to fix this bio-text issue without adding a new database column.


internal/regexes/regexes.go
```diff
@@ -47,6 +51,16 @@ const (
 )
 
 var (
+	schemes = `(((http|https))://)`
```
igalic, line 54: why are there so many capture groups in this regex?

igalic: this should be enough:
```suggestion
	schemes = `(http|https)://`
```

igalic (changes requested):
some regex nonsense spotted

internal/regexes/regexes.go
NyaaaWhatsUpDoc: this was literally just a drag and drop from another area of the code. you're right :sweat_smile:

NyaaaWhatsUpDoc:
This still needs a specific test writing for the AccountVerify method, and there's an existing replacementions test that needs fixing

igalic:
btw, is this https://github.com/superseriousbusiness/gotosocial-admin/issues/11 related?

NyaaaWhatsUpDoc:
> btw, is this [superseriousbusiness/gotosocial-admin#11](https://github.com/superseriousbusiness/gotosocial-admin/issues/11) related?

not directly i don't think, as the code with the logic mainly being changed here is the mentions code. but certainly an issue in a similar area of the codebase i'm sure

NyaaaWhatsUpDoc:
Todo:
- update internal/processing/status tests
- update internal/api/client/accountverify test to also check for a plaintext note in Account.Source

tsmethurst (approved):

PR #548, opened 2022-05-08
[bugfix] Fix remote media pruning failing if media already gone

tsmethurst:
The storage implementation only returns `storage.ErrNotFound` for in-memory stores, and returns a different error for the disk implementation when a file isn't found in the store.

This PR updates the pruning logic to not fail if it gets an error containing `no such file or directory` from the store when doing deletes, since it means that media has already gone.


NyaaaWhatsUpDoc:
I'll also fix this upstream in my library :)

internal/api/client/auth/authorize_test.go
```diff
@@ -80,7 +78,7 @@ func (suite *AuthAuthorizeTestSuite) TestAccountAuthorizeHandler() {
 		testSession.Set(sessionUserID, user.ID)
 		testSession.Set(sessionClientID, suite.testApplications["application_1"].ClientID)
 		if err := testSession.Save(); err != nil {
-			panic(errors.WrapMsgf(err, "failed on case: %s", testCase.description))
+			panic(fmt.Errorf("failed on case %s: %w", testCase.description, err))
```
NyaaaWhatsUpDoc, line 81: this was just dropping an unnecessary dependency :)

PR #550, opened 2022-05-09
Include password strength in error message when password strength is too low

### LINKED ISSUE #549: Password validator error does not report how close you are to having enough entropy
When you create a new password, for example when doing `admin account create` and you don't have enough entropy in your password you get an error

> insecure password, try including more special characters, using uppercase letters, using numbers or using a longer password

However, there is no guidance on how *much* you need to change the password. Do you need to just add one or two characters, or do I need to add many more characters of different types?

eobrain:
When the password validator fails because the attempted password is not strong enough, change the error message to include the percentage strength of the attempted password.

This makes it easier for the user to judge how much they need to change the password.

Fixes #549


tsmethurst:
This is neat! Thanks :)

Issue #489, opened 2022-04-25
[bug] Mastonaut App fails to look up instance and log in 


f2k1de:
My favorite Client-App [Mastonaut](https://mastonaut.app/) doesn't work with gotosocial. 

When i try to add an instance it doesn't allow me to do it:
![Bildschirmfoto 2022-04-25 um 19 04 59](https://user-images.githubusercontent.com/11199213/165142072-79dd4e1f-4c4d-45ab-b33e-3b0548020bc4.png)

I found out that it fails because the `thumbnail` key in the `/api/v1/instance` endpoint is empty. If i change that to something (it doesn't even have to be a working image, lol)... It can look the instance up:
![Bildschirmfoto 2022-04-25 um 19 04 12](https://user-images.githubusercontent.com/11199213/165142652-c6d84ee6-4b4c-4fce-b9b0-033f40e312cb.png)

If I try to connect the usual oauth-flow starts and it opens a browser with the gts login form, when i log in there i get:
![Bildschirmfoto 2022-04-25 um 19 08 26](https://user-images.githubusercontent.com/11199213/165143125-57c65bf7-484e-4285-8364-6f3c0fab036a.png)

If i click the Allow button it brings me back to the Mastonaut-App and gives me the following error:
![Bildschirmfoto 2022-04-25 um 19 38 21](https://user-images.githubusercontent.com/11199213/165143634-97587074-7c21-4856-93b5-7d2ff9783876.png)

Here is a capture of the auth-request:
![Bildschirmfoto 2022-04-25 um 19 09 34](https://user-images.githubusercontent.com/11199213/165144406-c0f129a9-2f47-4f8c-9e8a-673ff48b9955.png)

tsmethurst:
Heya, just a little update on this:

I just spent like a day setting up a MacOS installation in qemu to try to replicate the bug. Got the virtual machine working, but then I get errors trying to download Mastonaut from the app store (even after giving it my phone number and an email address :grimacing: ). So unless someone gives me a Macbook or something, I won't be able to test fixes for this one. HOWEVER, I will take a look at the GtS code and the Mastonaut code (which is now open source apparently, as of 25 days ago) and try to figure out a solution.

tsmethurst:
[...]

PR #553, opened 2022-05-10
[bugfix] Oauth token allow scope slice

tsmethurst:
Some Mastodon apps (Mastonaut specifically) submit Scope as a string array when doing token requests, instead of submitting Scope as a space-separated string.

This PR allows for either comma-separated strings *or* string arrays to be submitted, and adds tests for both.

Related to #489

### DIFF
--- a/internal/api/client/auth/token.go
+++ b/internal/api/client/auth/token.go
@@ -21,6 +21,7 @@ package auth
 import (
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 	"github.com/superseriousbusiness/gotosocial/internal/api"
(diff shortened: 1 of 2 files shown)


NyaaaWhatsUpDoc:
I'm a little hesitant on this, as space delimited is the actual oauth spec https://tools.ietf.org/html/rfc6749#section-3.3 . This seems like something the application author should handle as opposed to us

tsmethurst:
> I'm a little hesitant on this, as space delimited is the actual oauth spec https://tools.ietf.org/html/rfc6749#section-3.3 . This seems like something the application author should handle as opposed to us

Indeed, normally I'd agree that we shouldn't fix it on our side, but in this case it looks like the Mastonaut app is no longer maintained, but a lot of people still use it (judging by its position in the app store). Since it's just a small change to our logic, I don't mind making the adjustment, but if you'd rather not include it I very much understand and we can give it a miss.

igalic:
how come this works on other AP servers?

tsmethurst:
> how come this works on other AP servers?

Honestly I'm not sure... I assume Mastodon also accepts scope as an array :thinking: Although that's not documented afaik

tsmethurst:
Mmm okay, that's fair. I don't think we can open issues on their github since it's archived, but we can at least link to this discussion in #489

PR #560, opened 2022-05-13
[feature] Media cleanup endpoint

### LINKED ISSUE #348: [feature] method to trigger clean up of old media
as admin api call?
clean up media from disk, and remove references so it doesn't error

zethra:
Adds an admin endpoint to trigger a remote media cleanup.

Fixed #348

### DIFF
--- a/internal/api/client/admin/mediacleanup.go
+++ b/internal/api/client/admin/mediacleanup.go
@@ -1,0 +1,110 @@
+/*
+   GoToSocial
+   Copyright (C) 2021-2022 GoToSocial Authors admin@gotosocial.org
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package admin
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	"github.com/superseriousbusiness/gotosocial/internal/api/model"
+	"github.com/superseriousbusiness/gotosocial/internal/config"
+	"github.com/superseriousbusiness/gotosocial/internal/oauth"
+)
+
+// MediaCleanupPOSTHandler swagger:operation POST /api/v1/admin/media_cleanup mediaCleanup
+//
+// Clean up remote media older than the specified number of days.
+//
+// ---
+// tags:
+// - admin
+//
+// consumes:
+// - application/json
+// - application/xml
+// - application/x-www-form-urlencoded
+//
+// produces:
+// - application/json
+//
+// security:
+// - OAuth2 Bearer:
+//   - admin
+//
+// responses:
+//   '200':
+//     description: |-
+//      Echos the number of days requested. The cleanup is performed asynchronously after the request completes.
+//   '403':
+//      description: forbidden
+//   '400':
+//      description: bad request
+func (m *Module) MediaCleanupPOSTHandler(c *gin.Context) {
+	l := logrus.WithFields(logrus.Fields{
+		"func":        "MediaCleanupPOSTHandler",
+		"request_uri": c.Request.RequestURI,
+		"user_agent":  c.Request.UserAgent(),
+		"origin_ip":   c.ClientIP(),
+	})
+
+	// make sure we're authed...
+	authed, err := oauth.Authed(c, true, true, true, true)
+	if err != nil {
+		l.Debugf("couldn't auth: %s", err)
+		c.JSON(http.StatusUnauthorized, gin.H{"error": err.Error()})
+		return
+	}
+
+	// with an admin account
+	if !authed.User.Admin {
+		l.Debugf("user %s not an admin", authed.User.ID)
+		c.JSON(http.StatusForbidden, gin.H{"error": "not an admin"})
+		return
+	}
+
+	// extract the form from the request context
+	l.Tracef("parsing request form: %+v", c.Request.Form)
+	form := &model.MediaCleanupRequest{}
+	if err := c.ShouldBind(form); err != nil {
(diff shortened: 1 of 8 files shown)


internal/api/client/admin/mediacleanup.go
```diff
@@ -0,0 +1,116 @@
...
+
+// MediaCleanupPOSTHandler swagger:operation POST /api/v1/admin/media_cleanup mediaCleanup
+//
+// Clean up remote media older than the specified number of days.
+//
+// ---
+// tags:
+// - admin
+//
+// consumes:
```
tsmethurst, line 40: Since it can also consume json or xml, this should be:

```
// consumes:
// - application/json
// - application/xml
// - application/x-www-form-urlencoded
```

internal/api/model/admin.go
```diff
@@ -91,3 +91,7 @@ type AdminAccountActionRequest struct {
 	// ID of the account to be acted on.
 	TargetAccountID string `form:"-" json:"-" xml:"-"`
 }
+
+type MediaCleanupRequest struct {
```
tsmethurst, line 95: You should annotate this as pertaining to a specific request endpoint, like we do here: https://github.com/superseriousbusiness/gotosocial/blob/898d256511fdddbc6b5e783116b07fd240b28819/internal/api/model/account.go#L95-L97

In this case, the annotation should be:

```
// MediaCleanupRequest models admin media cleanup parameters
//
// swagger:parameters mediaCleanup
```

internal/api/model/admin.go
```diff
@@ -91,3 +91,7 @@ type AdminAccountActionRequest struct {
...
+	RemoteCacheDays string `form:"remote_cache_days" json:"remote_cache_days" xml:"remote_cache_days"`
```
tsmethurst, line 96: You can make this an int :)

internal/api/client/admin/mediacleanup.go
```diff
@@ -0,0 +1,116 @@
...
+// - multipart/form-data
+//
+// produces:
+// - application/json
+//
+// parameters:
```
tsmethurst, line 46: If you annotate the struct consumed by this endpoint (see below) you can delete this 'parameters' section

tsmethurst (changes requested):
Almost there, just a few swagger things

internal/api/model/admin.go
zethra: Go question. If I do make this an int and no value or an empty value is provided, will parsing produce and error or a zero value?

tsmethurst: good question; I think it will only fail if you have the 'required' annotation in the struct tags, so you should be OK and empty values will just parse as the zero value for an int (which is zero); if they provide no value, or zero, better take whatever the admin has set as their media-remote-cache-days config value (you can get this with `viper.GetInt(config.Keys.MediaRemoteCacheDays)`

igalic: in general, in go, ***everything*** is zero initialised, unless you tell it to do otherwise

sometimes this isn't useful or sensible, but it's marginally better than having leftover garbage (or for security… researchers: treasure) in a new variable.

zethra: That would mean it's impossible to request clearing all remote media by explicitly requesting 0 days. This is acceptable?

igalic: you could have -1 be that magic number
or you could set a different default in the "constructor"

tsmethurst: Yeah I'd also say -1 can mean 'clear everything', that seems OK

zethra: This is why I used a string. I can differentiate between no value given and 0/

zethra: Making -1 clear everything and 0 use the default time is unintuitive. Having 0 clear everything and having no value use the default time makes much more sense.

tsmethurst: i think it's OK if the api is like:

- a positive number: uncache media older than this number
- `0` or not set: uncache media according to the setting `media-remote-cache-days`
- `-1` or less: uncache everything

zethra: Alright, if that's you decision. But I personally dislike that design.

tsmethurst: Hmm okay thinking about it more maybe it makes sense to use an int pointer here then, and if it's nil take the default, and if it's zero or negative, clear everything

internal/api/client/admin/mediacleanup.go
```diff
@@ -0,0 +1,111 @@
...
+	if form.RemoteCacheDays == nil {
+		remoteCacheDays = viper.GetInt(config.Keys.MediaRemoteCacheDays)
+	} else {
+		remoteCacheDays = *form.RemoteCacheDays
+	}
+	if remoteCacheDays < 0 {
+		remoteCacheDays = 0
+	}
+
+	err = m.processor.AdminMediaRemotePrune(c.Request.Context(), remoteCacheDays)
```
tsmethurst, line 103: Forgot to mention earlier, this function returns a gtserror which should have a code and a message built in, so you can do:

```
if errWithCode := m.processor.AdminMediaRemotePrune(c.Request.Context(), remoteCacheDays); errWithCode != nil {
  		l.Debugf("error starting prune of remote media: %s", errWithCode.Error())
		c.JSON(errWithCode.Code(), gin.H{"error": errWithCode.Safe()})
		return
}

```

We do something similar here: https://github.com/superseriousbusiness/gotosocial/blob/9f2a2abe42a6211df6970ba7625d245298778a38/internal/api/client/status/statusboost.go#L69-L104

Eventually we will refactor all the processor functions that are called by the API packages to return this kind of error, so might as well do this now to save time later :)

internal/api/model/admin.go
igalic: @zethra what would your string based API have been?

tsmethurst:
This looks great :) Do you feel like also writing a small test suite for this, in the style of the other test suites in the admin package? No worries if not, I'm happy to merge as-is and write a test myself later, just thought I'd check :)

zethra:
I can take a crack it. Though due to the timezone difference I probably won't get to it until tomorrow for you.

zethra:
Fixed the formattting

internal/api/client/admin/mediacleanup_test.go
```diff
@@ -0,0 +1,96 @@
+package admin_test
```
tsmethurst, line 1: can you add the license comment in here?

tsmethurst (commented):
Looks great now! Good job on the tests. If you can just add the license text to that one file and then i will merge this :)

internal/api/client/admin/mediacleanup_test.go
zethra: Sure

tsmethurst (approved):

PR #399, opened 2022-02-15
introduce development settings and config validation

ForestJohnson:

PR #562, opened 2022-05-13
[feature] Basic config validation

### LINKED ISSUE #495: [bug] Confusing error message when `host` is unset
When running `0.3.0` of the docker image without a `host` variable set but sqlite configured, the app crashes with the following output :

```
time=2022-04-26T16:20:43Z level=info msg=connected to SQLITE database
time=2022-04-26T16:20:43Z level=info msg=there are no new migrations to run func=doMigration
time=2022-04-26T16:20:43Z level=fatal msg=error executing command: error creating instance account: constraint failed: NOT NULL constraint failed: accounts.username (1299)
```

This is a confusing error message as it is a clean database and looks to be creating a record with an empty username.

Either the username should have a default value if the host is unset, or the app should bail with a nicer error message leading people to set the host env/config.

tsmethurst:
This PR adds basic config validation for `host` and `protocol`, and opens the door for more config validation in future.

The validation is performed only for CLI commands, in the `preRun` function, and it can be disabled for commands that shouldn't validate config (like printed collated command to stdout as in `gotosocial debug config`).

Also adds test for some of the validation logic.

This was inspired by @ForestJohnson's earlier PR here: https://github.com/superseriousbusiness/gotosocial/pull/399

closes #495


igalic:
once the full Config is validated, we can put it into the systemd unit's reload step, and fail with an error there, instead of reloading a broken Config and then failing to start

PR #563, opened 2022-05-13
implement custom_emojis endpoint

i-am-a-paull:
This is a stab at implementing /api/v1/custom_emojis. I made some guesses about where things went, so if they need to go elsewhere I can move them when I get the chance. I'm also not completely satisfied with the tests but I can get around to improving those too at some point :sweat_smile:


internal/processing/media/getemoji.go
```diff
@@ -0,0 +1,49 @@
...
+
+func (p *processor) GetCustomEmojis(ctx context.Context) ([]*apimodel.Emoji, gtserror.WithCode) {
+	emojis, err := p.db.GetCustomEmojis(ctx)
+	if err != nil {
+		if err != db.ErrNoEntries {
+			return nil, gtserror.NewErrorNotFound(fmt.Errorf("db error retrieving custom emojis: %s", err))
+		}
+	}
+
+	apiEmojis := []*apimodel.Emoji{}
```
NyaaaWhatsUpDoc, line 39: Since you already know the number of DB emoji entries you can initialize this with `make([]*apimodel.Emoji, 0, len(emojis))` which for a very large slice will be more performant :)

i-am-a-paull: Ooh thanks!

internal/processing/media/getemoji.go
```diff
@@ -0,0 +1,49 @@
...
+		if err != db.ErrNoEntries {
+			return nil, gtserror.NewErrorNotFound(fmt.Errorf("db error retrieving custom emojis: %s", err))
+		}
+	}
+
+	apiEmojis := make([]*apimodel.Emoji, 0, len(emojis))
+	for _, gtsEmoji := range emojis {
+		apiEmoji, err := p.tc.EmojiToAPIEmoji(ctx, gtsEmoji)
+		if err != nil {
+			logrus.Errorf("error converting emoji with id %s: %s", gtsEmoji.ID, err)
```
tsmethurst, line 43: if you error here you should `continue` after logging it so you don't risk putting a malformed emoji struct in the apiEmojis slice

tsmethurst (changes requested):
Looks brilliant to me, just one small fix and then I think this can be merged :) Great job

internal/processing/media/getemoji.go
i-am-a-paull: thanks! I've been mostly afk for a few days but I finally got it in :sweat_smile:

tsmethurst (approved):

PR #564, opened 2022-05-13
transport.Controller{} and transport.Transport{} security and performance improvements

NyaaaWhatsUpDoc:
Among this PR, is a new HTTP client implementation found under `internal/httpclient`, and used within the `internal/transport/controller{}` which provides the following:
- setting a maximum received request body size, returning error on large content lengths, and using a limited reader in all other cases to protect against forged / unknown content-lengths
- protection from server side request forgery (SSRF) by only dialing out to known public IP prefixes, configurable with allows/blocks
- limit number of concurrent requests, else blocking until a slot is available (context channels still respected)

The `internal/transport/transport{}` logic for performing HTTP requests has been refactored and updated to now provided a simplistic retry logic to account for temporary / network errors.

The `internal/transport/controller{}` has been updated to provided a cache of transports, stored under their computed public key (from private key) for security, instead of initializing each time one is requested.

This PR initially started with just the retry logic, adding in limiting of open connections to counter the fact that retries could very quickly stack up. Given the required codechanges for that it made adding the SSRF protection a breeze.


NyaaaWhatsUpDoc:
The tests are broken currently as it turns out I broke the "sneaky sneaky" behaviour of the testrig `GetSignatureFor___()` functions :p a fix is in the works that also makes them less sneaky :D

internal/transport/controller.go
```diff
@@ -37,109 +42,107 @@ import (
...
+	build, _ := debug.ReadBuildInfo()
 
-		return json.Marshal(i)
+	c := &controller{
+		db:        db,
+		fedDB:     federatingDB,
+		clock:     clock,
+		client:    client,
+		cache:     cache.New[string, *transport](),
+		userAgent: fmt.Sprintf("%s %s gofed/activity gotosocial-%s", applicationName, host, build.Main.Version),
```
tsmethurst, line 75: nice, was wondering a while ago about the best way to do this

internal/transport/controller.go
```diff
@@ -37,109 +42,107 @@ import (
...
+	postHeaders = []string{httpsig.RequestTarget, "host", "date", "digest"}
+)
 
+func (c *controller) NewTransport(pubKeyID string, privkey *rsa.PrivateKey) (Transport, error) {
+	// Generate public key string for cache key
+	//
+	// NOTE: it is safe to use the public key as the cache
+	// key here as we are generating it ourselves from the
+	// private key. If we were simply using a public key
+	// provided as argument that would absolutely NOT be safe.
```
tsmethurst, line 101: noted :+1:

internal/transport/controller.go
```diff
@@ -164,3 +167,45 @@ func (c *controller) NewTransportForUsername(ctx context.Context, username strin
 	}
 	return transport, nil
 }
+
+// dereferenceFollowersShortcut is a shortcut to dereference followers of an
```
tsmethurst, line 171: this comment doesn't match the name of the function anymore

internal/transport/controller.go
```diff
@@ -164,3 +167,45 @@ func (c *controller) NewTransportForUsername(ctx context.Context, username strin
...
+
+	i, err := streams.Serialize(followers)
+	if err != nil {
+		return nil, err
+	}
+
+	return json.Marshal(i)
+}
+
+// dereferenceUserShortcut is a shortcut to dereference followers an account on
```
tsmethurst, line 189: same as above

internal/transport/controller.go
```diff
@@ -164,3 +167,45 @@ func (c *controller) NewTransportForUsername(ctx context.Context, username strin
...
+		return nil, err
+	}
+
+	return json.Marshal(i)
+}
+
+// privkeyToPublicStr will create a string representation of RSA public key from private.
+func privkeyToPublicStr(privkey *rsa.PrivateKey) string {
+	b := x509.MarshalPKCS1PublicKey(&privkey.PublicKey)
+	return *(*string)(unsafe.Pointer(&b))
```
tsmethurst, line 210: This syntax is beyond me. I'm not saying don't do it, but could you explain what's happening here so I can learn? :P

internal/transport/transport.go
```diff
@@ -43,28 +50,111 @@ type Transport interface {
...
+	getSigner  httpsig.Signer
+	postSigner httpsig.Signer
+	signerMu   sync.Mutex
 }
 
-func (t *transport) SigTransport() pub.Transport {
-	return t.sigTransport
+// GET will perform given http request using transport client, retrying on certain preset errors, or if status code is among retryOn.
+func (t *transport) GET(r *http.Request, retryOn ...int) (*http.Response, error) {
+	if r.Method != "GET" {
```
tsmethurst, line 67: Might as well use http.MethodGet here right? just a quibble though, feel free to leave as is

internal/transport/transport.go
```diff
@@ -43,28 +50,111 @@ type Transport interface {
...
+		t.signerMu.Lock()
+		defer t.signerMu.Unlock()
+		err := t.getSigner.SignRequest(t.privkey, t.pubKeyID, r, nil)
+		return err
+	}, retryOn...)
+}
+
+// POST will perform given http request using transport client, retrying on certain preset errors, or if status code is among retryOn.
+func (t *transport) POST(r *http.Request, body []byte, retryOn ...int) (*http.Response, error) {
+	if r.Method != "POST" {
```
tsmethurst, line 80: same comment as above

internal/transport/transport.go
```diff
@@ -43,28 +50,111 @@ type Transport interface {
...
+			return nil, err
+		} else if errors.As(err, &x509.UnknownAuthorityError{}) {
+			// Unknown authority errors we do NOT recover from
+			return nil, err
+		}
+
+		l.Errorf("backing off for %s after http request error: %v", backoff.String(), err)
+
+		// Backoff for some time
+		time.Sleep(backoff)
```
tsmethurst, line 144: is sleeping the best way to handle this? genuine question because i don't know

tsmethurst (commented):
Looks *GREAT*, just some comments + questions (and ofc you gotta get the tests working again :stuck_out_tongue_winking_eye: ). This looks much much tidier than the scrappy code I'd written before, nice one!!

testrig/testmodels.go
```diff
@@ -1905,22 +1904,21 @@ func GetSignatureForActivity(activity pub.Activity, pubKeyID string, privkey cry
...
 	// create a client that basically just pulls the signature out of the request and sets it
 	client := &mockHTTPClient{
 		do: func(req *http.Request) (*http.Response, error) {
 			signatureHeader = req.Header.Get("Signature")
 			dateHeader = req.Header.Get("Date")
-			r := ioutil.NopCloser(bytes.NewReader([]byte{})) // we only need this so the 'close' func doesn't nil out
 			return &http.Response{
 				StatusCode: 200,
-				Body:       r,
+				Body:       http.NoBody,
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: yeah it's helpful :3 i think they only tend to use it in requests but i figure why not also use it in situations like these too

internal/httpclient/client.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, ErrBodyTooLarge
+	}
+
+	// Seperate the body implementers
+	rbody := (io.Reader)(rsp.Body)
+	cbody := (io.Closer)(rsp.Body)
+
+	var limit int64
+
+	if limit = rsp.ContentLength; limit < 0 {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: Yeah it's pretty nice, just another way of writing things to increase code prettiness at times :p

internal/httpclient/sanitizer.go
```diff
@@ -0,0 +1,46 @@
...
+	ipport, err := netip.ParseAddrPort(addr)
+	if err != nil {
+		return err
+	}
+
+	// Seperate the IP
+	ip := ipport.Addr()
+
+	// Check if this is explicitly allowed
+	for i := 0; i < len(s.allow); i++ {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: So this is a habit I've picked up when micro-optimizing things. One of the optimizations the compiler can do is inlining, so taking the logic of a function and placing it inline where it's being called instead of it having the overhead of calling into it. But there's certain things that can prevent the compiler from inlining a function e.g. `defer` and relevant here, `range`. I may not know when the compiler will find it fit to do this optmization, but I make it possible in case it can. Though probably unnecessary here, just a habit I've picked up

internal/transport/controller.go
NyaaaWhatsUpDoc: I was taking a look at the goblin.technology logs to see how it looked receiving this UA from my instance, and I was thinking of maybe adding semicolons between each of the sections? Also just wanted to check, is the `build.Main.Version` the best place to grab the current build or did you use another method for when you print the build version on the page footers?

internal/transport/controller.go
NyaaaWhatsUpDoc: on it

internal/transport/controller.go
NyaaaWhatsUpDoc: on it

internal/transport/transport.go
NyaaaWhatsUpDoc: Agreed, I was being lazy :p

internal/transport/transport.go
NyaaaWhatsUpDoc: same as mine above :D

internal/transport/transport.go
NyaaaWhatsUpDoc: That's a good point, using `time.After()` we can respect the request context. I just tend to use sleep normally if I'm not actually using the channel of of a timer for anything

internal/transport/controller.go
NyaaaWhatsUpDoc: So as you know, according to the Go spec strings are immutable and byte slices are not. The way this gets implemented is strings under the hood are:
```go
type StringHeader struct {
    Data uintptr
    Len  int
}
```

while slices are:
```
type SliceHeader struct {
    Data uintptr
    Len  int
    Cap  int
}
```

Because being mutable, you can change the data, length etc, but the string has to promise to be read-only to all who get copies of it.

So in practice when you do a conversion of `string(byteSlice)` it actually performs an allocation because it has to copy the contents of the byte slice into a safe read-only state.

Being essentially the same under the hood, except for the single extra `Cap` field in the slice, means that if you have a byte slice you can "forcibly" cast it to a string. Which in a lot of situations can be risky, because then it means you have a string that is NOT immutable, as if someone changes the data in the originating byte slice then the string will reflect that change! Now while this does seem hacky, and it _kind_ of is, it is something that you see performed in the standard library. If you look at the definition for `strings.Builder{}.String()` you'll see this :)

In this case, I knew that the x509 marshaler function freshly allocated that byte slice, and we're not passing it to anyone literally just converting to string, so there's no risk of the contents of it changing. Again this is maybe more of a habit that I've picked up from micro-optimizing so if you don't want this in here I totally understand :D. I just figured it's a safe case for this particular usage and it helps trim some allocations off a hot code path since it's generating the key used for caching

NyaaaWhatsUpDoc: Actually a nice way of seeing this is through the inverse of this, a conversion from string to bytes without allocation. You can see the 2 implementations as `B2S()` and `S2B()` here: https://codeberg.org/gruf/go-byteutil/src/branch/main/bytes.go

tsmethurst: Aha! Thank you for the great explanation! I'm wondering if there's a little writeup of that somewhere that we can add in a code comment, just so other people also know what's up without having to go digging... do you know of anything?

internal/transport/transport.go
tsmethurst: I think respecting the request context would be nice :)

NyaaaWhatsUpDoc: Agreed!

internal/transport/controller.go
NyaaaWhatsUpDoc: I could add this write-up to my repository, and then instead use `byteutil.B2S()` so if they hover over they can see the explanation?

tsmethurst: Sure, that sounds good :)

NyaaaWhatsUpDoc:
@tsmethurst you might want to take another brief look at the transport useragent (i updated it to hopefully look a little more pleasing), and i updated the transport to update the `httpsig.Signer` whenever the expiry time is reached (since i realised via caching of transports we'd then be signing with out of date signers)

internal/httpclient/sanitizer.go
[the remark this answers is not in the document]
tsmethurst: Thanks for explaining! That makes sense

internal/transport/controller.go
tsmethurst: build.Main.Version is also used in main.go in the cmd package to build the version number, but its concatenated with stuff like build time and go version to make the longer version string you can see on the page footers. for a short version number like we need here, this is perfect :)

tsmethurst:
Looks good to me! :)

Issue #535, opened 2022-05-03
[feature] Clean up unused headers + avatars


tsmethurst:
Right now the media cache only cleans up old attachments.

We should also clean up headers + avis that are no longer used anymore.

## To clean up existing leftover headers + avis

Something like the following (can probably be optimized but this is just a rough idea):

For each account with a header attachment ID set, select every attachment that 1. is a header, 2. belongs to that account id, and 3. *isn't* the currently-set header. Prune these.

Then do the same for avatars. 

## To avoid keeping unused headers + avis around in storage

When a new avatar or header is set on an account by an update from a remote instance, prune the old one immediately. Do the same thing for headers + avatars from accounts on our instance when they're updated via the client API.

PR #574, opened 2022-05-15
[feature] Unused avatar and header cleanup

tsmethurst:
This PR adds logic for cleaning up unused avatars and headers (ie., avatars and headers that have been replaced by a profile update, but are still sitting around in storage).

The new functionality looks like this:

1. Select all headers + avatars (in batches on 20).
2. For each avatar + header, check if it's currently in use.
3. If it's not, delete it from storage, then from the database.

This has been added to the processor's MediaRemotePrune function, which has been renamed to MediaPrune to reflect that it's not just remote media being pruned anymore.

This cleanup will be run on instance startup, and every night @ midnight along with the remote media cache prune job.

My tests on goblin.technology and gts.superseriousbusiness.org showed a saving of 1gb-2gb after running this cleanup :)

closes https://github.com/superseriousbusiness/gotosocial/issues/535


internal/media/manager.go
```diff
@@ -224,5 +189,66 @@ func (m *manager) Stop() error {
...
 	return cronErr
 }
+
+func scheduleCleanupJobs(m *manager) error {
+	// create a new cron instance for scheduling cleanup jobs
+	c := cron.New(cron.WithLogger(&logrusWrapper{}))
+	cancels := []context.CancelFunc{}
+
+	pruneMetaCtx, pruneMetaCancel := context.WithCancel(context.Background())
+	cancels = append(cancels, pruneMetaCancel)
```
NyaaaWhatsUpDoc, line 202: if you rename pruneMetaCtx to just pruneCtx you can reuse this cancellable context for both set of the pruning jobs :)

NyaaaWhatsUpDoc:
Aside from single comment this looks great :+1:

internal/media/manager.go
tsmethurst: Ack! Updated it :)

igalic:
question for @NyaaaWhatsUpDoc, now that my brain has booted  (17:35 local time)

> if you rename pruneMetaCtx to just pruneCtx you can reuse this cancellable context for both set of the pruning jobs :)

won't reusing a variable in a different context make it harder for the GC to know if or when to clean it up?

Issue #517, opened 2022-04-30
[performance] Perform viper.Get__() functions only at init


NyaaaWhatsUpDoc:
Parse config values into struct at init and from there we should just access the values through the initialized struct. Running `viper.Get___()` in HTTP handlers for example, or other (relatively) hot codepaths is unnecessary given it's not a super simple implementation.

It should be a very easy to change to make, just possibly quite a bit of code to change :'). I'm happy to do it myself assuming this is something we agree on here!

igalic:
(how) will this affect our plans re https://github.com/superseriousbusiness/gotosocial/issues/270 ?

tsmethurst:
Indeed, that would be my concern too: one of the reasons we switched to viper was to allow config to be reloaded from file or (eventually) from other config storage systems like etcd. If it's only a small performance hit to call `viper.Get()` then I think I'd rather we kept it so we can allow hot reload later on

NyaaaWhatsUpDoc:
> Indeed, that would be my concern too: one of the reasons we switched to viper was to allow config to be reloaded from file or (eventually) from other config storage systems like etcd. If it's only a small performance hit to call `viper.Get()` then I think I'd rather we kept it so we can allow hot reload later on

Ah I was not aware of this. It definitely would be possible to manage this still implementing the initial viper calls at init, but maybe reduces the priority of this by a little.

PR #575, opened 2022-05-15
Global server configuration overhaul

NyaaaWhatsUpDoc:
Overhauls server configuration to be found under a single struct instance in the `internal/config` package. Flag and usage names are moved to struct tags of this, and the global instance is protected during concurrency with a mutex.

Benefits:
- handles what's discussed here https://github.com/superseriousbusiness/gotosocial/issues/517 while still supporting reloading (in fact makes it very easy!)
- consolidating code, at the moment we have config key names in the config package, usage in the cmd package and values as a separate struct to the keys themselves
- code "sugar" so instead of fetching runtime config via `viper.GetString(config.Keys.Host)` you can now just do `config.GetHost()`, which also actually removes the risk of miscasting a config value type
- moves us closer to having a singular `ConfigState{}` struct that can be used in over-arching state that is passed round everywhere


tsmethurst (commented):
Had a look through this and it looks really cool and exciting!

NyaaaWhatsUpDoc:
@tsmethurst this has been updated to fix merge issues and should be ready to review now :)

NyaaaWhatsUpDoc:
I think the failed test is one of the flaky ones, if I could get a rerun of them please :')

test/cliparsing.sh
```diff
@@ -5,7 +5,7 @@ set -e
...
...:"","db-tls-mode":"disable","db-type":"postgres","db-user":"","email":"","host":"","letsencrypt-cert-dir":"/gotosocial/storage/certs","letsencrypt-email-address":"","letsencrypt-enabled":false,"letsencrypt-port":80,"log-db-queries":false,"log-level":"info","media-description-max-chars":500,"media-description-min-chars":0,"media-image-max-size":2097152,"media-remote-cache-days":30,"media-video-max-size":10485760,"oidc-client-id":"","oidc-client-secret":"","oidc-enabled":false,"oidc-idp-name":"","oidc-issuer":"","oidc-scopes":["openid","profile","email","groups"],"oidc-skip-verification":false,"password":"","path":"","port":8080,"protocol":"https","smtp-from":"GoToSocial","smtp-host":"","smtp-password":"","smtp-port":0,"smtp-username":"","software-version":"","statuses-cw-max-chars":100,"statuses-max-chars":5000,"statuses-media-max-files":6,"statuses-poll-max-options":6,"statuses-poll-option-max-chars":50,"storage-backend":"local","storage-local-base-path":"/gotosocial/storage","syslog-address":"localhost:514","syslog-enabled":false,"syslog-protocol":"udp","trusted-proxies":["127.0.0.1/32"],"username":"","web-asset-base-dir":"./web/assets/","web-template-base-dir":"./web/template/"}'
```
tsmethurst, line 8: so, my understanding here is that the config fields from other more specific commands are now included as well (email, password, username etc), is that right? do we want that behavior? for me it doesn't matter too much, but i wanted to check what your thoughts are on it

testrig/config.go
```diff
@@ -19,45 +19,19 @@
...
-			panic("could not convert config.Keys value to string")
-		}
-
-		// the value should be the test default corresponding to the given fieldName
-		value := testDefaults.FieldByName(field.Name).Interface()
-
-		// actually set the value in viper -- this will override everything
-		viper.Set(key, value)
-	}
+	config.Config(func(cfg *config.Configuration) {
```
tsmethurst, line 28: much cleaner than the mess I was making

tsmethurst (commented):
Overall looks absolutely brilliant and much cleaner :) Just a few comments/questions but nothing major. This is really good stuff, thank you for all your hard work on it!

test/cliparsing.sh
NyaaaWhatsUpDoc: Yes that is unfortunately the behaviour. I don't think it matters too much imo, as the config printing is more a debug feature. Is that what you were thinking also?

internal/config/helpers.gen.go
```diff
@@ -0,0 +1,1477 @@
+// THIS IS A GENERATED FILE, DO NOT EDIT BY HAND
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: Yes we can I'll add that in :)

internal/config/defaults.go
```diff
@@ -22,7 +22,7 @@ import "github.com/coreos/go-oidc/v3/oidc"
 
 // Defaults returns a populated Values struct with most of the values set to reasonable defaults.
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: What would you change this too?

internal/config/flags.go
```diff
@@ -0,0 +1,137 @@
...
+		cmd.Flags().String(SMTPPasswordFlag(), cfg.SMTPPassword, fieldtag("SMTPPassword", "usage"))
+		cmd.Flags().String(SMTPFromFlag(), cfg.SMTPFrom, fieldtag("SMTPFrom", "usage"))
+
+		// Syslog
+		cmd.Flags().Bool(SyslogEnabledFlag(), cfg.SyslogEnabled, fieldtag("SyslogEnabled", "usage"))
+		cmd.Flags().String(SyslogProtocolFlag(), cfg.SyslogProtocol, fieldtag("SyslogProtocol", "usage"))
+		cmd.Flags().String(SyslogAddressFlag(), cfg.SyslogAddress, fieldtag("SyslogAddress", "usage"))
+	})
+}
+
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: What parts is it you think are missing?

test/cliparsing.sh
tsmethurst: Yeah I was thinking prettymuch the same thing, just wanted to check :) This is fine as-is then

PR #576, opened 2022-05-15
Mention `AmbientCapabilities=CAP_NET_BIND_SERVICE` in example systemd service

illfygli:
I don't know systemd so it took me a while to figure out how to run on a privileged port as non-root.
I added it to the example service, commented out.


igalic (approved):

tsmethurst:
Thanks!

PR #593, opened 2022-05-21
[security] Check all involved IRIs during block checking

tsmethurst:
This PR updates the logic in the Blocked function of the federator to make more aggressive checks against accounts involved in the Activity in some way.

In short, we now make sure that:

* the account receiving an Activity doesn't block an account either To'd, CC'ed, or RepliedTo
* an account either To'd, CC'ed, or RepliedTo doesn't block the account that created an Activity

This means that in the following scenario --

* local1 follows remote
* local2 blocks remote
* remote delivers an activity replyingTo/CCing local2, to the inbox of local1, who is in their Followers collection

-- the request will be refused because our instance knows that local2 blocks remote. So even though local1 technically can see the status, the instance won't process it because of the block. This makes it much less likely that local2 will bump into the status somehow, and see a message from someone who they blocked.

The PR also tightens up some of the domain block logic in the same way. Now, if an Activity is To'd, CC'ed, or RepliedTo someone from a blocked domain, the request will be refused, even if the inbox account follows the account who created the Activity.

In order to make these extra checks work without totally tanking the database doing lots of selects, some new lighter database functions have been added to select just the account ID of an account or status by its URI.

### DIFF
--- a/internal/db/bundb/account.go
+++ b/internal/db/bundb/account.go
@@ -70,22 +70,6 @@ func (a *accountDB) GetAccountByURI(ctx context.Context, uri string) (*gtsmodel.
 	)
 }
 
-func (a *accountDB) GetAccountIDForAccountURI(ctx context.Context, uri string) (string, db.Error) {
-	if cached, ok := a.cache.GetByURI(uri); ok {
-		return cached.ID, nil
-	}
-
-	account := &gtsmodel.Account{}
-	if err := a.conn.NewSelect().
-		Model(account).
-		Column("id").
-		Where("uri = ?", uri).Scan(ctx); err != nil {
-		return "", a.conn.ProcessError(err)
-	}
-
-	return account.ID, nil
-}
-
 func (a *accountDB) GetAccountByURL(ctx context.Context, url string) (*gtsmodel.Account, db.Error) {
 	return a.getAccount(
 		ctx,
--- a/internal/db/bundb/account.go
+++ b/internal/db/bundb/account.go
@@ -70,6 +70,22 @@ func (a *accountDB) GetAccountByURI(ctx context.Context, uri string) (*gtsmodel.
 	)
 }
 
+func (a *accountDB) GetAccountIDForAccountURI(ctx context.Context, uri string) (string, db.Error) {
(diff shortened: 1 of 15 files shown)


internal/db/bundb/account.go
```diff
@@ -70,6 +70,22 @@ func (a *accountDB) GetAccountByURI(ctx context.Context, uri string) (*gtsmodel.
 	)
 }
 
+func (a *accountDB) GetAccountIDForAccountURI(ctx context.Context, uri string) (string, db.Error) {
+	if cached, ok := a.cache.GetByURI(uri); ok {
+		return cached.ID, nil
+	}
+
+	account := &gtsmodel.Account{}
```
NyaaaWhatsUpDoc, line 78: Is it worth doing this separate db query? Because in the case this account isn't cached the result will never be cached, increasing database access. Alternatively it would be just doing `a.GetAccountByURI(...)` and returning the ID from that

internal/db/bundb/status.go
```diff
@@ -87,6 +87,22 @@ func (s *statusDB) GetStatusByURI(ctx context.Context, uri string) (*gtsmodel.St
 	)
 }
 
+func (s *statusDB) GetAccountIDForStatusURI(ctx context.Context, uri string) (string, db.Error) {
+	if cached, ok := s.cache.GetByURI(uri); ok {
+		return cached.AccountID, nil
+	}
+
+	status := &gtsmodel.Status{}
```
NyaaaWhatsUpDoc, line 95: Same as above, but using relevant status function

internal/db/bundb/account.go
tsmethurst: Yeah I also wasn't sure about these. The problem is that we really just need that one field, whereas a full-blown GetAccountByURI does a massive join with all the relations we have set on it. So this is a really cheap query, but it will indeed mean more queries in total. I'm not really sure which way ends up faster overall

tsmethurst: Replaced with cacheable functions :+1:

internal/db/bundb/status.go
tsmethurst: Replaced with cacheable functions :+1:

PR #594, opened 2022-05-22
Add Caddy instructions to the documentation

sexybiggetje:
This PR adds Caddy instructions to the documentation. It borrows some text from the NGINX document for the configuration of GoToSocial. It doesn't require much configuration, so the installation instruction is actually a little bit easier.

### DIFF
--- a/docs/installation_guide/caddy.md
+++ b/docs/installation_guide/caddy.md
@@ -33,6 +33,10 @@ dnf install caddy
 pacman -Syu caddy
 ```
 
+### FreeBSD
+```bash
+sudo pkg install caddy
+```
 
 ## Configure GoToSocial
 
@@ -41,9 +45,6 @@ If GoToSocial is already running, stop it.
 ```bash
 sudo systemctl stop gotosocial
 ```
-
-Or if you don't have a systemd service just stop it manually.
-
 In your GoToSocial config turn off Lets Encrypt by setting `letsencrypt-enabled` to `false`.
 
 If you we running GoToSocial on port 443, change the `port` value back to the default `8080`.
@@ -90,7 +91,7 @@ Now check for configuration errors.
 sudo caddy validate
 ```
 
-If everything is fine, you should get some info lines as output. Unless there are lines marked with [err] in front of them, you are all set.
+If everything is fine, you should get some info lines as output. Unless there are lines marked with *[err]* in front of them, you are all set.
 
 Everything working? Great! Then restart caddy to load your new config file.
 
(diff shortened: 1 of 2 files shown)


docs/installation_guide/caddy.md
```diff
@@ -0,0 +1,109 @@
...
+
+For advanced configuration check the [reverse_proxy directive](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy) at the Caddy documentation.
+
+Now check for configuration errors.
+
+```bash
+sudo caddy validate
+```
+
+If everything is fine, you should get some info lines as output. Unless there are lines marked with [err] in front of them, you are all set.
```
igalic, line 93: i would put markup `[err]`

docs/installation_guide/caddy.md
```diff
@@ -0,0 +1,109 @@
...
+
+If everything went right, you're now all set to enjoy your GoToSocial instance, so we are going to start it again.
+
+```bash
+sudo systemctl start gotosocial
+```
+
+## Results
+
+You should now be able to open the splash page for your instance in your web browser, and will see that it runs under HTTPS!
```
igalic, line 109: how long does caddy take to retrieve the certificate?
is it necessary to restart it, when it got new ones?

docs/installation_guide/caddy.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Configure GoToSocial
+
+If GoToSocial is already running, stop it.
+
+```bash
+sudo systemctl stop gotosocial
+```
+
+Or if you don't have a systemd service just stop it manually.
```
igalic, line 45: I think we can get rid of this line, here and in the nginx document

docs/installation_guide/caddy.md
```diff
@@ -0,0 +1,109 @@
...
+dnf copr enable @caddy/caddy
+dnf install caddy
+```
+
+### Arch
+
+```bash
+pacman -Syu caddy
+```
+
```
igalic, line 35: it's even in FreeBSD!

but most people who use a BSD can follow this document just fine ;)

igalic (commented):
some feedback

docs/installation_guide/caddy.md
sexybiggetje: Caddy fully manages it using a challenge. It's near instant. Whenever the certificate is set to expire it will renew it before that happens, which gives you zero downtime

sexybiggetje:
I've added the FreeBSD installation command, and resolved your comments @igalic thanks for the feedback!

tsmethurst:
Thanks for this! :)

Issue #565, opened 2022-05-14
[bug] `error extracting status content: no content found`


tsmethurst:
This occurs sometimes on posts that are just an image and no text:

```
time=2022-05-13T22:16:51+02:00 level=info msg=ASStatusToStatus: error extracting status content: no content found statusURI=https://mastodon.social/users/yogthos/statuses/108296425480940485
```

I thought I'd fixed this but apparently not quite!

NyaaaWhatsUpDoc:
I just realised if you look at the log entry, it's at info level so it's part of that code block that logs the error but doesn't actually stop parsing because of it. I think we might want to update those log messages to be warnings and not look like errors :')

PR #598, opened 2022-05-23
[bugfix] Fix `error extracting status content: no content found`

tsmethurst:
This PR closes https://github.com/superseriousbusiness/gotosocial/issues/565 by updating ExtractContent to just return an empty string if no Content is found in an Activity, since this isn't technically an error, it just means that no text content was set and most likely the post contains a poll or media or something. It also updates the extract content function to check if there's an IRI in the content field, because that's a thing that can sometimes happen.

### DIFF
--- a/internal/ap/extract.go
+++ b/internal/ap/extract.go
@@ -349,7 +349,7 @@ func ExtractContent(i WithContent) string {
 		}
 		if iter.IsIRI() && iter.GetIRI() != nil {
 			return iter.GetIRI().String()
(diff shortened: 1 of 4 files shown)


igalic:
that makes this PR then a prerequisite for https://github.com/superseriousbusiness/gotosocial/pull/597 ?

tsmethurst:
> that makes this PR then a prerequisite for #597 ?

mmm not sure, i don't think the order really matters

PR #605, opened 2022-05-24
Don't double escape instance title

zethra:
Currently the instance title is escaped when it's put in the DB and when it's displayed, leading to errors. The short description is not escaped when displayed, so this change bring the title in line with that.


f0x52 (approved):

zethra:
We doing a different thing

Issue #554, opened 2022-05-11
[bug] Link account handle inside bio doesn't always work


jugendhacker:
When I'm trying to put the handle of an account into the bio of a GoToSocial account other people will not be able to click on it. In contrast mastodon seems to make a clickable link out of it. Would be cool if GoToSocial could also do this.

tsmethurst:
Heya! You should be able to do this if you format it like so: `@user@example.org`.

jugendhacker:
Seems not to work. My friend over at https://social.jugendhacker.de/@krea does exactly this and it does not seem to work.

tsmethurst:
AH weird! Okay, thanks for raising this, we'll take a look :)

tsmethurst:
Changed the title to reflect that it's something we've gotta fix, hope that's OK with you

jugendhacker:
Yeah it's okay for me, just thought it's not implemented yet and therefore I wanted to request.

igalic:
I did say i need to take a look at all our regrets, i mean Regexes

Issue #503, opened 2022-04-28
[bug] Search account by `@account@account_domain` doesn't work but search by `@account@host` does


tsmethurst:
From the matrix chat:

> if someone's instance domain is different than their handle domain, you have to search for the instance domain rather than the handle one

GtS should be able to follow webfinger redirects etc and dereference an account by their account domain, so we need to check this and fix it up.

PR #627, opened 2022-06-02
[chore] Webfinger rework

tsmethurst:
This PR reworks several areas of account looking up, webfingering, and searching.

In summary:

1. Rework the GetRemoteAccount function in the dereferencer to perform webfinger requests, and use the accountDomain of the target instance as account.Domain instead of the Host part of the instance URI.
2. Also in GetRemoteAccount: do webfinger lookups of accounts if they haven't been fingered within the last 2 days.
3. Rework search functionality to make it more permissive: you can now search for accounts at account domains, hosts, or even search for accounts without including the first `@` -- or, if it's a local account, search just by entering the name.
4. To help test all of this: move mock http client logic out of individual tests and into the testrig. Now, all tests can use the mock http client defined there, which already stubs out some routes and responses. This saves a lot of error-prone repetition within test packages.

An unrelated change is also to increase the default number of workers in a concurrent worker pool to 4*cpu instead of 2*cpu. If desired I'll pull this out and put it in a separate PR!

closes https://github.com/superseriousbusiness/gotosocial/issues/503
closes https://github.com/superseriousbusiness/gotosocial/issues/554 (I believe)


tsmethurst:
Note for reviewers:

the most spicy and consequential changes are right here: https://github.com/superseriousbusiness/gotosocial/pull/627/files#diff-784aaa02a974b99c75baecaebd8152c761ecce9414e2dd9639964bfd4b3b093d

NyaaaWhatsUpDoc:
Was there a reason for a worker pool number change?

tsmethurst:
> Was there a reason for a worker pool number change? 

Yeah sort of, but I think it was ill informed. My thinking was that since we'll be making more outgoing requests tied in with webfingering etc, we should bump the workers (especially since I noticed that the cpu usage on goblin tech was still pretty low, so we have some headroom to work with). But I think maybe this belongs in a separate PR after, so we can have a proper discussion about it first. If you like, I can take it out of this one :) 

> If I get the chance later today I'll take a look on my laptop, if not then when I get back from EMF :)

Hell yeah, enjoy your field!

tsmethurst:
I believe this also closes https://github.com/superseriousbusiness/gotosocial/issues/554, but I need to investigate just a little more first

PR #634, opened 2022-06-06
🐸Restructure Frontend Sources

f0x52:
Include admin and future user panel in main repo, properly deduplicate bundles for css+js across uses

Mostly ready, but needs more testing to prevent regressions, and some documentation updates


f0x52:
- [ ] documentation
- [ ] check release flow
- [ ] update dev server to work better in tandem with testrig

.goreleaser.yml
```diff
@@ -11,8 +11,8 @@ before:
   - sed -i "s/REPLACE_ME/{{ incpatch .Version }}/" docs/api/swagger.yaml
   - cp docs/api/swagger.yaml web/assets/swagger.yaml
   # install and bundle the web assets and styling
```
f0x52, line 13: goreleaser untested

Dockerfile
```diff
@@ -5,17 +5,18 @@ FROM --platform=${BUILDPLATFORM} node:17.6.0-alpine3.15 AS admin_builder
...
-
-RUN npm install
-RUN node index.js
-
 FROM --platform=${TARGETPLATFORM} alpine:3.15.0 AS executor
 
 # copy over the binary from the first stage
 COPY --chown=1000:1000 gotosocial /gotosocial/gotosocial
 
+# bundle frontend stuff
```
f0x52, line 13: docker changes are also untested

internal/web/base.go
```diff
@@ -136,16 +136,74 @@ func (m *Module) NotFoundHandler(c *gin.Context) {
 	})
 }
 
+// TODO: abstract the {admin, user}panel handlers in some way
```
f0x52, line 139: good way to abstract this? maybe save it for a later PR

web/template/header.tmpl
```diff
@@ -9,21 +9,23 @@
 	<meta name="og:title" content="GoToSocial Testing Instance">
```
sexybiggetje, line 9: The meta og:title should match the title tag here I think?

Dockerfile
tsmethurst: Unfortunately I think we'll still need to do this as a two-step build where the admin panel is built in the first step (using `--platform=${BUILDPLATFORM}`). Reason for this is that with our current multi-architecture setting, we can't run commands in the second docker container while building.

tsmethurst (commented):
Looking good so far :) Since the javascript files from the admin panel etc are now in the main repo, you should probably add the license text that's in the go files to the top of the javascript files as well.

Dockerfile
tsmethurst: Oh wait scratch that.... I think we can just have GoReleaser do the bundling and then copy the bundled files directly into the Docker container, which would be even neater, then we don't need a two-stage build at all

web/template/header.tmpl
f0x52: yep https://github.com/superseriousbusiness/gotosocial/pull/634/commits/66ec12728ad7c682e44131a8b2c38e2b76ef133b

.goreleaser.yml
tsmethurst: tested it :)

Dockerfile
tsmethurst: Fixed

internal/web/base.go
```diff
@@ -120,16 +120,74 @@ func (m *Module) baseHandler(c *gin.Context) {
...
+func (m *Module) AdminPanelHandler(c *gin.Context) {
+	l := logrus.WithField("func", "admin-panel")
+	l.Trace("serving admin panel")
+
+	host := config.GetHost()
+	instance, err := m.processor.InstanceGet(c.Request.Context(), host)
+	if err != nil {
+		l.Debugf("error getting instance from processor: %s", err)
```
tsmethurst, line 131: i'll update these new functions to use the new api error handling stuff tomorrow

Issue #587, opened 2022-05-18
[bug] Multiple dereferences of same media attachment leads to weird UI issues


tsmethurst:
When you're browsing through posts by a specific account, sometimes you'll see some absolutely bonkers shit like this:

![Screenshot from 2022-05-18 16-24-26](https://user-images.githubusercontent.com/31960611/169085567-ce497009-f9a6-4882-a3eb-ce1ef11866cc.png)

What's happening here is (from matrix)

> if something gets boosted a lot, every time it gets boosted we (for some reason) store it again
> normally you don't notice this because the status the media is attached to is updated to use the latest attachment id(s)
> however, when you're doing stuff like looking through a user's timeline, NOT looking through the home timeline, the database query does some funky thing where it selects all attachments where media_attachment.status_id IN 'whatever_status_id'
> this means for statuses that have had their media dereferenced a couple times, you might see a bunch of copies of the same attachment on that status

In the case of the above image (again from matrix, hah!):

> the original status here only had one attachment
> but because it's a cool image, it got boosted lots of times
> and we ended up in a situation where we dereferenced the media multiple times, created a new media_attachment in the database each time, and updated the status.media_attachments[] to the latest attachment we just saved
> now in the database you can run...
> select * from media_attachments as media_attachment where media_attachment.status_id in ('01EZEEQP59MWVR2AZEHVYCNXYR');
> and it comes out like this:

![Screenshot from 2022-05-18 16-30-33](https://user-images.githubusercontent.com/31960611/169086004-9e0fc93a-9877-4e5e-b804-74d877c61e46.png)

> a shitload of copies of the same image, each with a different attachment id in our database and (therefore) a different path in storage, and a different serve path, but all with the same remote_url

We should fix this! So here's what we should do (are you tired of matrix yet? i'm not):

> 1. fix that weird fucking database query
> 2. make sure the media manager or dereferencer or wherever the logic is (can't remember) doesn't put more than one copy of a piece of media in the database -- we can use remote_url for this, and maybe even make it unique in the db
> 3. write a job to clean up all those annoying multiples
>
> 3 is kind of optional because actually they'll be cleaned up eventually anyways by the normal cleanup operation, i believe

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/pull/589

PR #649, opened 2022-06-12
[chore] Duplicated media cleanup

### LINKED ISSUE #599: [feature] Add task for cleaning up duplicated media
Since https://github.com/superseriousbusiness/gotosocial/pull/589 we shouldn't be pulling any duplicated media attachments in, however we still need to create a task or something (via API, cli, migration, whatever) to clean up media that was already duplicated.

tsmethurst:
This PR adds a database migration for cleaning up duplicated media caused by https://github.com/superseriousbusiness/gotosocial/issues/587

To do this, it needs to instantiate a storage interface so that the stored media files can be removed; for testing, this fails if the default test storage path of `/gotosocial/storage` is taken, because the gts process doesn't have permissions on that path.

To work around this, this value was changed to `/tmp/gotosocial` (or os-equivalent). The testrig doesn't actually use this directory for anything aside from instantiating the store during that one migration.

Closes #599

### DIFF
--- a/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
+++ b/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
@@ -35,19 +35,23 @@ func init() {
 	deleteAttachment := func(ctx context.Context, l *logrus.Entry, a *gtsmodel.MediaAttachment, s *kv.KVStore, tx bun.Tx) {
 		if err := s.Delete(a.File.Path); err != nil && err != storage.ErrNotFound {
 			l.Errorf("error removing file %s: %s", a.File.Path, err)
+		} else {
+			l.Debugf("deleted %s", a.File.Path)
 		}
-		l.Debugf("deleted %s", a.File.Path)
 
 		if err := s.Delete(a.Thumbnail.Path); err != nil && err != storage.ErrNotFound {
 			l.Errorf("error removing file %s: %s", a.Thumbnail.Path, err)
+		} else {
+			l.Debugf("deleted %s", a.Thumbnail.Path)
 		}
-		l.Debugf("deleted %s", a.Thumbnail.Path)
 
 		if _, err := tx.NewDelete().
 			Model(a).
 			WherePK().
 			Exec(ctx); err != nil {
 			l.Errorf("error deleting attachment with id %s: %s", a.ID, err)
+		} else {
+			l.Debugf("deleted attachment with id %s", a.ID)
 		}
 	}
 
--- a/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
+++ b/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
@@ -103,11 +103,18 @@ func init() {
 				}
 				l.Debugf("found %d duplicates of attachment with remote url %s", len(dupedAttachments), dupedRemoteURL.RemoteURL)
 
-				statusID := dupedAttachments[0].StatusID
+				var statusID string
+				for _, dupe := range dupedAttachments {
+					if statusID == "" && dupe.StatusID != "" {
+						statusID = dupe.StatusID
+					}
+				}
+
 				if statusID == "" {
 					l.Debugf("%s not associated with a status, moving on", dupedRemoteURL.RemoteURL)
 					continue
 				}
+				l.Debugf("%s is associated with status %s", dupedRemoteURL.RemoteURL, statusID)
 
 				// step 3: get the status that these attachments are supposedly associated with
 				status := &gtsmodel.Status{}
--- a/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
+++ b/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
@@ -116,13 +116,15 @@ func init() {
 				}
 				l.Debugf("%s is associated with status %s", dupedRemoteURL.RemoteURL, statusID)
 
-				// step 3: get the status that these attachments are supposedly associated with
+				// step 3: get the status that these attachments are supposedly associated with, bail if we can't get it
 				status := &gtsmodel.Status{}
 				if err := tx.NewSelect().
 					Model(status).
 					Where("id = ?", statusID).
 					Scan(ctx); err != nil {
-					l.Errorf("error selecting status with id %s: %s", statusID, err)
+					if err != sql.ErrNoRows {
+						l.Errorf("error selecting status with id %s: %s", statusID, err)
+					}
 					continue
 				}
 
--- a/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
+++ b/internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
@@ -20,7 +20,6 @@ package migrations
 
 import (
 	"context"
-	"database/sql"
 	"fmt"
 	"path"
 
@@ -49,12 +48,6 @@ func init() {
 			WherePK().
 			Exec(ctx); err != nil {
 			l.Infof("error deleting attachment with id %s: %s", a.ID, err)
-		}
-	}
-
-	deleteAttachments := func(ctx context.Context, l *logrus.Entry, as []*gtsmodel.MediaAttachment, s *kv.KVStore, tx bun.Tx) {
-		for _, a := range as {
-			deleteAttachment(ctx, l, a, s, tx)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,144 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/uptrace/bun"
+)
+
+func init() {
+	deleteAttachment := func(ctx context.Context, l *logrus.Entry, a *gtsmodel.MediaAttachment, s *kv.KVStore, tx bun.Tx) {
+		if err := s.Delete(a.File.Path); err != nil && err != storage.ErrNotFound {
+			l.Errorf("error removing file %s: %s", a.File.Path, err)
+		}
+		l.Debugf("deleted %s", a.File.Path)
```
NyaaaWhatsUpDoc, line 39: i'd add this as the `else` clause so it doesn't claim as deleted when it failed, same as below

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,144 @@
...
+			Exec(ctx); err != nil {
+			l.Infof("error deleting attachment with id %s: %s", a.ID, err)
+		}
+	}
+
+	up := func(ctx context.Context, db *bun.DB) error {
+		l := logrus.WithField("migration", "20220612091800_duplicated_media_cleanup")
+
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			storageBasePath := config.GetStorageLocalBasePath()
```
NyaaaWhatsUpDoc, line 58: Is it worth doing a `basePath == ""` check here? or will these checks already have been performed by the time we reach the migrations :thinking:

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,144 @@
...
+					continue
+				}
+
+				// step 3: get the status that these attachments are supposedly associated with
+				status := &gtsmodel.Status{}
+				if err := tx.NewSelect().
+					Model(status).
+					Where("id = ?", statusID).
+					Scan(ctx); err != nil {
+					l.Infof("error selecting status with id %s: %s", dupedAttachments[0].StatusID, err)
```
NyaaaWhatsUpDoc, line 110: inconsequential but you can use `statusID` as you already declared this in line 98 :)

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,144 @@
...
+
+					deleteAttachment(ctx, l, a, s, tx)
+				}
+			}
+			return nil
+		})
+	}
+
+	down := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
```
NyaaaWhatsUpDoc, line 136: is it worth just returning nil here instead of the empty tx?

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,144 @@
...
+					l.Infof("cleaning %d of %d", i, dupes)
+				}
+
+				// step 2: select all media attachments associated with this url
+				dupedAttachments := []*gtsmodel.MediaAttachment{}
+				if err := tx.NewSelect().
+					Model(&dupedAttachments).
+					Where("remote_url = ?", dupedRemoteURL.RemoteURL).
+					Scan(ctx); err != nil {
+					l.Infof("error running same attachments query: %s", err)
```
NyaaaWhatsUpDoc, line 93: I'd make these error log entries to error level imo (same in cases below)

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,144 @@
...
+				ColumnExpr("remote_url", "count(*)").
+				Where("remote_url IS NOT NULL").
+				Group("remote_url").
+				Having("count(*) > 1").
+				Scan(ctx); err != nil {
+				return err
+			}
+			dupes = len(dupedRemoteURLs)
+			l.Infof("found %d attachments with duplicate remote URLs", dupes)
+
```
NyaaaWhatsUpDoc, line 81: At the moment I think you're removing the whole set of media attachments for those with duplicates, instead of all except 1. I think it'd be worth doing a `sort.Sort(...)` here to sort by creation time, and then deleting either all by the newest or all but the oldest, but either way doing a reslice like `dupedRemoteURLs = dupedRemoteURLs[1:]`

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/67c524f568ba0f1079af20d658429d79942a2c8d

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/d5d4b741c1fc35c281f8d4346bda32423e63a04d

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/67c524f568ba0f1079af20d658429d79942a2c8d

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/1f25de14638e95fa892866cbc7a9f5176d8d0a5c

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: Hmm I'm not sure if that's what I'm doing. I tried to rephrase the logic a little bit here to make it clearer: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/7bf3007959b1d6a11adf29c6b149441f8f2275f4

By the time we reach step 4, we have a slice of `dupedAttachments`, which are a bunch of media attachments that all have the same `remote_url` value. We know *one* of them is the correct one, so we just need to check through the status's `attachment_ids` to make sure it's in there. When we do a delete of the others, we're not deleting all the other attachments on that status (unless I made a mistake somewhere), we should be deleting all other instances of that specific attachment. The other three (or however many) attachments on the status should remain untouched. Again, unless I made a mistake, which is quite possible :')

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/b014aaa8e030654aff1fa661ad54a2bceaa099b7

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,160 @@
...
+					Scan(ctx); err != nil {
+					l.Errorf("error running same attachments query: %s", err)
+					continue
+				}
+				l.Debugf("found %d duplicates of attachment with remote url %s", len(dupedAttachments), dupedRemoteURL.RemoteURL)
+
+				var statusID string
+				for _, dupe := range dupedAttachments {
+					if statusID == "" && dupe.StatusID != "" {
+						statusID = dupe.StatusID
```
NyaaaWhatsUpDoc, line 110: It would probably make more sense to do `if dupe.StatusID != "" { statusID = dupe.StatusID; break }` just to stop further unnecessary iterations

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -0,0 +1,160 @@
...
+					}
+					continue
+				}
+
+				// step 4: for each attachment, check if it's actually one that the status is currently set to use, and delete if not
+				for _, dupe := range dupedAttachments {
+					var currentlyUsed bool
+					for _, attachmentID := range status.AttachmentIDs {
+						if attachmentID == dupe.ID {
+							currentlyUsed = true
```
NyaaaWhatsUpDoc, line 137: Same as the above point about iterations. When you find a value for currentlyUsed, break out of the loop.

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/002ca07adfb633d4c8ed65e7b6fcb02961e33baf

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/649/commits/b93329c6ca6b0b2c02fbf12709e74d2bba0738c1

PR #651, opened 2022-06-15
postgres locale: fix accounts.note_raw migration

LittleFox94:
Database migration `20220506110822_add_account_raw_note.go` has [some error
handling code](https://github.com/superseriousbusiness/gotosocial/blob/13e4bbdbfa104a2384834b634285dce2f4dafe2e/internal/db/bundb/migrations/20220506110822_add_account_raw_note.go#L31) to detect some error messages as "ok", but only done for english error messages.

This adds a check for the specific error code, which should be locale agnostic.


tsmethurst:
Nice, thanks! ❤️

LittleFox94:
I'm afraid I'll have to open another one for the other migration for that column x)

No idea why that worked for tests, but not for migrating my old dev instance

PR #653, opened 2022-06-16
fix for the second accounts.note_raw migration

LittleFox94:
As mentioned in the comment below #651, here is the fix for the second migration


tsmethurst:
Sorry for all the false positives on the failing tests -_- We've got some race conditions in the testrig we need to hunt down

Issue #654, opened 2022-06-17
[bug] Release artefact doesn't contain built assets


igalic:
### Describe the bug with a clear and concise description of what the bug is.

When downloading the supposed release tarball (not the source!) from https://github.com/superseriousbusiness/gotosocial/releases/tag/v0.3.5, I'm faced with two issues:

1. it's a [tar bomb](http://www.linfo.org/tarbomb.html)
2. it doesn't contain the built CSS assets, so deploying it leads to lots of 404s

### What's your GoToSocial Version?

v0.3.5

### GoToSocial Arch

nah.

### What happened?

I tried to update the [gotosocial packaging for FreeBSD](https://github.com/freebsd/freebsd-ports/tree/main/net-im/gotosocial)
I switched the built from using the raw tag, to using the released tar ball: https://github.com/superseriousbusiness/gotosocial/releases/download/v0.3.5/gotosocial-0.3.5.tar.gz

That thing does not contain the web/assets/dist

### What you expected to happen?

I expected it to contain web/assets/dist

(i also expected it to not explode all over my filesystem when untaring it)

### How to reproduce it?

Download https://github.com/superseriousbusiness/gotosocial/releases/download/v0.3.5/gotosocial-0.3.5.tar.gz
untar it (into a subdirectory)

look into web/assets/dist
it's not there.

### Anything else we need to know?

Please don't make me use node and npm or yarn in a build system.

And please don't tell me to just use the readily built release for FreeBSD.
The point of having native packages is that they are built within the parameters the build chain or the user supplies.

(and we have some ways to go there, like, for instance right now there's no way to build WITH_DEBUG rn, that is: Have debug symbols preserved, without hackery that i didn't want to expend too much energy on.)

f0x52:
i did not know this either at first, but [gotosocial-0.3.5.tar.gz](https://github.com/superseriousbusiness/gotosocial/releases/download/v0.3.5/gotosocial-0.3.5.tar.gz) is a tar of the entire repo at time of release (so no builds), for bundled assets you need to download one of the platform specific tars, like [gotosocial_0.3.5_freebsd_amd64.tar.gz](https://github.com/superseriousbusiness/gotosocial/releases/download/v0.3.5/gotosocial_0.3.5_freebsd_amd64.tar.gz)

[...]

PR #655, opened 2022-06-19
[packaging] Packaging updates

tsmethurst:
This PR updates some of our packaging in the following ways:

1. Bump version of golang to 1.18.3.
2. Bump gotosocial-drone-build to 0.0.5 (which includes bumps to GoReleaser and Docker buildx).
3. Remove the bundled source tarball from releases, since it was misleading and unhelpful.
4. Add a release archive that just contains the bundled web/assets and web/templates folders, to make life easier for people doing ports.

Should close https://github.com/superseriousbusiness/gotosocial/issues/654

However, it doesn't really address the 'tarbomb' point raised in that issue, because while you can instruct goreleaser to embed a root directory inside a tar, it would cause our installation instructions to change, and people would have to run different installation commands for different releases. This is not something that I like the idea of doing, so I left it out of this PR. However, if there's enough impetus and interest in changing this, we can still do this in another PR.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -28,7 +28,7 @@ steps:
           - pull_request
 
   - name: test
-    image: golang:1.18-alpine
+    image: golang:1.18.3-alpine
     volumes:
       - name: go-build-cache
         path: /root/.cache/go-build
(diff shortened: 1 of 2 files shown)


igalic:
does this build the assets, too?

tsmethurst:
> does this build the assets, too?

Yes :) essentially it replaces that weird source tarball with a tar that just contains web/assets/dist and web/assets/template (so everything in the web/assets/ dir aside from the source dir). So when porting you can clone the latest tag, grab the compiled assets tar and untar it, then do whatever you need to do without having to install node/yarn

PR #659, opened 2022-06-20
[testrig] configure instance.Version for testrig

f0x52:
sets the instance version to 0.0.0-testrig when launched through there. mostly to just have the field populated in the Instance get api + page footers


tsmethurst:
Good idea! Just fix up the tests in `internal/api/client/instance` :P

Issue #650, opened 2022-06-13
[feature] `/api/v1/instance/peers` with filter parameters


tsmethurst:
Mastodon exposes an endpoint at `/api/v1/instance/peers` which returns a json array of domains that an instance federates with (see eg https://mastodon.social/api/v1/instance/peers):

```json
[
  "instance1.com",
  "instance2.com"
]
```

However, Mastodon does not show any json for instances that have been suspended/silenced, etc. In order to implement https://github.com/superseriousbusiness/gotosocial/issues/378, we should consider implementing this 'peers' endpoint but also allowing callers to submit query parameters to also show instances that have been suspended/silenced.

From the `gotosocial-code` Matrix chat:

---

`GET /api/v1/instance/peers` that also takes a query param `filter`

Possible values for `filter` are `friends,silenced,suspended` and you can provide any combination of these to filter the list of instances. When the value is empty, whether through no query or through `filter=` (empty), a default `filter` of `friends` is assumed.

Usage examples:

```text
GET /api/v1/instance/peers
- no filter value, default of `filter=friends` is assumed, returns only federated instances

GET /api/v1/instance/peers?filter=
- no filter value, default of `filter=friends` is assumed, returns only federated instances

GET /api/v1/instance/peers?filter=friends
- returns only federated instances

GET /api/v1/instance/peers?filter=suspended
- only returns suspended instances

GET /api/v1/instance/peers?filter=silenced
- only returns silenced instancces

GET /api/v1/instance/peers?filter=silenced,suspended,friends
- returns all known instances
```

---

We should also allow admins to limit permissions on this endpoint to authorized users only if they so desire; not everyone wants to share their blocklist/peer list since it has obvious safety implications.

When we implement https://github.com/superseriousbusiness/gotosocial/issues/3, we can consider allowing admins to also supply another instance's `/api/v1/instance/peers?filter=suspended` endpoint as a blocklist to subscribe to. This is convenient in cases where a known instance has a great block list, and a new instance wants to use the same block list without manually having to import the entries.

dariusk:
I like this API extension. My first issue is really minor: why `friends`? Something like `unrestricted` might be more appropriate, mostly because federation does not necessarily equal a friendship relationship.

[...]

PR #660, opened 2022-06-21
[feature] Implement `/api/v1/instance/peers` endpoint

tsmethurst:
This PR implements the `api/v1/instance/peers` endpoint, and closes https://github.com/superseriousbusiness/gotosocial/issues/650

By default, neither open nor suspended peers are exposed to unauthenticated users via the API. This PR also adds some new config options to allow admins to expose these endpoints if they desire.

If no query parameters are provided to the endpoint, then a mastodon-api compatible 'flat' list of open peers will be returned, to maintain compatibility.

If query parameters are provided, then a more informative, structured object will be returned.

---

Example responses:

`/api/v1/instance/peers`

```json
[
  "example.org",
  "fossbros-anonymous.io"
]
```

`/api/v1/instance/peers?filter=suspended`

```json
[
  {
    "domain": "replyguys.com",
    "suspended_at": "2020-05-13T13:29:12.000Z",
    "public_comment": "reply-guying to tech posts"
  }
]
```

`/api/v1/instance/peers?filter=suspended,open` -- with more domains and obfuscation

```json
[
  {
    "domain": "example.org"
  },
  {
    "domain": "fossbros-anonymous.io"
  },
  {
    "domain": "o*g.*u**.t**.*or*t.*r**ev**",
    "suspended_at": "2021-06-09T10:34:55.000Z",
    "public_comment": "just absolutely the worst, wowza"
  },
  {
    "domain": "replyguys.com",
    "suspended_at": "2020-05-13T13:29:12.000Z",
    "public_comment": "reply-guying to tech posts"
  }
]
```


ClearlyClaire:
It's incompatible with what other implementations have been using for this output, right?

tsmethurst:
> It's incompatible with what other implementations have been using for this API endpoint, right?

Unfortunately yes, because we provide a list of objects instead of a list of strings, but I also think that makes it more useful overall. If this was an integral part of the API we'd stick more closely to what mastodon does so as not to break compatibility, but I'm not sure what services or clients actually use that endpoint when talking to mastodon. Do you know?

ClearlyClaire:
I don't have a definite list, but as far as I know, pretty much all fediverse crawlers providing stats on number of users and software usage rely on this endpoint.

ClearlyClaire:
IIRC some projects use some nodeinfo endpoint to expose that kind of information.

tsmethurst:
> IIRC some projects use some nodeinfo endpoint to expose that kind of information.

Ah okay! Well we do still have the standard nodeinfo endpoint (discoverable via well-known), and the /api/v1/instance endpoint. As long as no client apps rely on /api/v1/instance/peers specifically for core functionality then I think we should be okay :)

tsmethurst:
Updated this PR following @f0x52 's suggestions. Now, for no query params, a mastodon-compatible 'flat' string array will be returned. When query params are provided, a domain object is returned instead. This allows us to provide more info to callers while still maintaining masto api compatibility.

Issue #664, opened 2022-06-23
[bug] Disallow following/blocking yourself


autumnull:
### Describe the bug with a clear and concise description of what the bug is.

It's currently possible for an authenticated user to follow/block themself. I feel like this is weird behavior and shouldn't be possible.

### What's your GoToSocial Version?

v0.3.6

### GoToSocial Arch

armv7 Binary

### What happened?

submitting an api request to `/api/v1/accounts/{id}/follow` (or `.../block`) with the currently authenticated user's id returns a `relationship` saying that you have followed/blocked yourself.

### What you expected to happen?

should probably return a `Error 406 (not acceptable)` ?

### How to reproduce it?

get an oauth token for an account, submit an api request as described above to the same account you auth'd with.

tsmethurst:
Haha, yes, this is definitely a bug :') Thank you!

PR #667, opened 2022-06-24
[bugfix] disallow following or blocking yourself (#664)

illfygli:
My first Go patch.
If you can tell me where to add/update tests I can see if it works too. :D

### DIFF
--- a/internal/api/client/account/block_test.go
+++ b/internal/api/client/account/block_test.go
@@ -1,0 +1,73 @@
+/*
+   GoToSocial
+   Copyright (C) 2021-2022 GoToSocial Authors admin@gotosocial.org
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package account_test
+
+import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


tsmethurst:
NICE, thank you!

You can add tests in this package: https://github.com/superseriousbusiness/gotosocial/tree/main/internal/api/client/account

Just copy one of the existing test files and adjust as necessary. If you need any help with it, or have questions about the test package we use etc, feel free to ask in the gotosocial code channel :)

internal/api/client/account/follow_test.go
```diff
@@ -0,0 +1,66 @@
...
+	suite.Equal(http.StatusNotAcceptable, recorder.Code)
+
+	result := recorder.Result()
+	defer result.Body.Close()
+
+	// check the response
+	b, err := ioutil.ReadAll(result.Body)
+	_ = b
+	suite.NoError(err)
+}
```
tsmethurst, line 66: You gotta add a bit more to the bottom of these two new test files, to actually run the tests: check the bottom of the other test files in this package to see what i mean

internal/api/client/account/block_test.go
```diff
@@ -0,0 +1,65 @@
...
+		"", "",
+		map[string]string{
+			"id": testAcct.ID,
+		})
+	if err != nil {
+		panic(err)
+	}
+	bodyBytes := requestBody.Bytes()
+	recorder := httptest.NewRecorder()
+	ctx := suite.newContext(recorder, http.MethodPost, bodyBytes, account.BlockPath, w.FormDataContentType())
```
tsmethurst, line 47: The params in blocks/follows are set from the path rather than from a form, so you don't need to call testrig.CreateMultipartFormData or provide bodyBytes, but instead just make sure you're calling the correct path, and inject the query param right into the test gin context. You can check this snippet for how to do that: https://github.com/superseriousbusiness/gotosocial/blob/7eacbd064b15d2a9ed2f6006c943e0ee65a31317/internal/api/client/status/statusfave_test.go#L57-L67

illfygli:
Thanks!
Now I think they're running and checking the right thing. :)

tsmethurst:
Thank you!!! :)

PR #674, opened 2022-06-26
[feature] S3 support

### LINKED ISSUE #308: [feature] S3 storage for media
I found some comments about planned S3 support but nothing seems to be here yet, so I thought creating an issue might be a good idea :)

Also, I'm working on it and just had my first image uploaded and successfully served again - but more discussion for that is probably better in the PR to follow :)

theSuess:
Resolves #308 

This adds S3 support by abstracting the storage driver in a new package `internal/storage`. When accessing an asset via S3, a signed resource URL is generated and returned as an HTTP Redirect. This helps to prevent hotlinking while still reducing load on the server.

I've tested it against MinIO and will continue tests on other providers tomorrow (Backblaze and Scaleway as they offer free tiers).

I also might want to refactor the `Content` type into an interface to have more type safety but wanted to create this PR to get early feedback if this approach is usable


internal/storage/storage.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/config"
+)
+
+var (
+	ErrNotSupported = errors.New("driver does not suppport functionality")
+)
+
+// Driver implements the functionality to store and retrieve blobs
+// (images,video,audio)
+type Driver interface {
```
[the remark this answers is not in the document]
theSuess: I've thought about this, but IMHO it does not fit well for two reasons:

1. The interface contains some functions which do not map cleanly to the minio SDK (which would mean extra implementation effort) and we do not use them
2. We need an interface providing the `URL` function anyway

tsmethurst:
Gonna review this in the coming days, thank you for all your hard work!

theSuess:
FYI: I've now also verified functionality with scaleway and backblaze, so I feel confident that this supports the majority of S3 providers

internal/storage/s3.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func NewS3(mc *minio.Client, bucket string) *S3 {
+	return &S3{
+		mc:     mc,
+		bucket: bucket,
+	}
+}
+
+func (s *S3) Get(key string) ([]byte, error) {
```
tsmethurst, line 28: For the contexts in the below functions: if we must implement a new interface for storage, perhaps it's worth making the interface functions accept a context. For local storage, we can just ignore the passed in context since it's not used, but then for the s3 implementation we would have a way of passing a context. This would require extra code changes in places which call the storage interface, but I think it would be worthwhile (and makes a good argument for wrapping everything in another interface). What do you think @NyaaaWhatsUpDoc ?

internal/storage/storage.go
```diff
@@ -0,0 +1,59 @@
+package storage
```
tsmethurst, line 1: license text

example/config.yaml
```diff
@@ -253,16 +253,25 @@ media-remote-cache-days: 30
...
-# NOTE: s3 storage is not yet supported!
 storage-backend: "local"
 
 # String. Directory to use as a base path for storing files.
 # Make sure whatever user/group gotosocial is running as has permission to access
 # this directory, and create new subdirectories and files within it.
+# Only required when running with the local storage backend
 # Examples: ["/home/gotosocial/storage", "/opt/gotosocial/datastorage"]
 # Default: "/gotosocial/storage"
 storage-local-base-path: "/gotosocial/storage"
```
tsmethurst, line 265: can you expose and document these variables properly in the style used elsewhere in the config file? and just note that they're only used if storage-backend is set to s3. I think it's important that we make the config file consistent

docs/configuration/storage.md
```diff
@@ -18,7 +18,18 @@ storage-backend: "local"
 # String. Directory to use as a base path for storing files.
 # Make sure whatever user/group gotosocial is running as has permission to access
 # this directly, and create new subdirectories and files with in.
+# Only required when running with the local storage backend
 # Examples: ["/home/gotosocial/storage", "/opt/gotosocial/datastorage"]
 # Default: "/gotosocial/storage"
 storage-local-base-path: "/gotosocial/storage"
+
+
```
tsmethurst, line 26: see comment below about exposing these

example/config.yaml
tsmethurst: (by expose them I mean uncomment then and fill them in with their default values -- empty strings in this case)

tsmethurst (commented):
Looks really good so far! Thank you for working on this, and for testing it :) It was a very pleasant surprise! Still have a few issues we need to discuss/resolve though, particularly around the interfacing and the contexts being passable-innable.

internal/api/model/content.go
```diff
@@ -28,6 +31,8 @@ type Content struct {
 	ContentLength int64
 	// Actual content
 	Content io.Reader
+	// Resource URL if file can be served directly
```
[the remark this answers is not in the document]
theSuess: I'm not too fond of the whole optional URL field myself

I was thinking about refactoring `Content` as an interface like
```
type Content interface {
  Serve(rw http.ResponseWriter)
}
```
and having two structs implement this (`LocalContent` and `RemoteContent`)

That way we resolve the ambiguity at a type level, and we do not accidentally access expect a reader if only the URL is set


What do you think of this approach?

internal/storage/s3.go
theSuess: I think it makes sense to pass a context here. I omitted it to keep it as simple as possible and get some feedback first.
During development, I checked the uses of these functions and I think we have a context available at every point we call them.

internal/api/model/content.go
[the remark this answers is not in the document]
tsmethurst: Oh, I see what you mean but I actually quite liked the optional url field, it feels fairly clean to me. I don't really like the idea of putting an http.ResponseWriter on one of the api models if we can avoid it. It also means we end up writing content to responses in places other than the api function with the gin context in it, which I think will end up confusing in the long run (we already do this in a few places because of the libraries we use, and I hate it :')). So I *think* this is fine as-is, just with a slightly more explanatory comment.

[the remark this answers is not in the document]
theSuess: Okay, then I'll use better wording when I get around to implementing the other changes

internal/storage/s3.go
theSuess: I've added the context parameter now in commit https://github.com/superseriousbusiness/gotosocial/pull/674/commits/1f9b27002fb2b16ea70704bbc08b7d7a99c740fb and kept it separate from the other topics, so we can easily revert it if we decide against it

tsmethurst: alright thank you! I'll consider it more + discuss it with the other devs when they have bandwidth, good thinking putting it in a separate commit :+1:

tsmethurst:
Alright, i discussed it with the other devs, and this looks good! So I'm gonna go ahead and merge it. Thank you so much for all your hard work! :)

LittleFox94:
Wow, this works great (running it on Ceph) and just the way I wanted to build it :D
Good job and many thanks!

PR #692, opened 2022-07-04
docs: document the migration between local and s3

theSuess:
Migrated my instance to the new backend and documented the process. Hope it's in the right place - if not, I'm happy to move it.


tsmethurst:
Thanks for this! We might want to move it somewhere else in the future (some kind of 'guides' section maybe?) but since that's not set in stone yet, I'm gonna merge it as is :) Again, thank you!

Issue #691, opened 2022-07-04
[bug] Page overflows to the right on mobile


zethra:
### Describe the bug with a clear and concise description of what the bug is.

On mobile, various page overflow to the right of the view port and introduce a right scroll.

![image](https://user-images.githubusercontent.com/6034238/177204271-876c613b-273e-43ec-bf6a-f872a5ddca7c.png)

### What's your GoToSocial Version?

0.3.7

### GoToSocial Arch

amd64

### What happened?

See above

### What you expected to happen?

No horizontal scroll

### How to reproduce it?

Load a GtS page on mobile, possible with enough content. Testrig triggers it.

tsmethurst:
Fixed with recent changes :)

PR #694, opened 2022-07-05
[bugfix] Fix footer info fields not wrapping

tsmethurst:
This PR makes the footer flexbox wrap contained items, which mitigates some of the issues we were seeing on mobile.

To test this, it adds an entry for the testrig instance to the test models.

relates to #691 but doesn't close it yet because there's a bit more to fix first


tsmethurst:
Screenshot from 2022-07-05 14-37-13
Screenshot from 2022-07-05 14-36-57
Screenshot from 2022-07-05 14-36-45

PR #697, opened 2022-07-07
Update config.yaml

bwoff11:
Updated examples for storage-s3-endpoint to include GCP example.


bwoff11:
For me, it'd be useful since I had difficulty determining what exactly to put for that field when utilizing GCP, but I was able to figure it out so that's understandable. Thanks for the review either way.

PR #700, opened 2022-07-08
Improve panic handling

NyaaaWhatsUpDoc:
Currently we're dropping a lot of information when a panic occurs, this rectifies that.


NyaaaWhatsUpDoc:
I have no idea what's happening with this commit situation. This is even with current status of main?? Was there a rebase of main that I missed or something?

I suppose the important part here is the changes which are as expected.

tsmethurst (approved):

PR #696, opened 2022-07-06
[bugfix] Fix profile card display issues on smaller screens w/responsive query

tsmethurst:
https://user-images.githubusercontent.com/31960611/177579013-a80f1d21-9c19-42dd-af4c-dc36dfabb031.mp4

PR #702, opened 2022-07-08
more consistent profile scaling accross firefox and chrome

f0x52:
still not quite ideal, but better I think


tsmethurst:
merged into https://github.com/superseriousbusiness/gotosocial/pull/696

PR #704, opened 2022-07-10
[chore] consolidate caching libraries

NyaaaWhatsUpDoc:
- consolidates multiple caching libraries to use only a single caching library
- updates DB to use caches for multi selects (accounts, statuses, notifs, mentions)
- improvements to the delete account statuses logic: doesn't fail-out early, extra error logging
- update DB to always add query hook, logs slow queries on production
- updates codeberg.org/gruf dependencies
- remove typeutil.Converter activitystreams cache as we have no means of invalidating


NyaaaWhatsUpDoc:
Again I don't know why all the extra commits here... Was there a rebase on main or something?

PR #705, opened 2022-07-10
[chore] improved router logging, recovery and error handling

NyaaaWhatsUpDoc:
- simplifies ErrorHandler
- updates disablequoting to only be enabled for >= debug log level
- update logging middleware to handle panic recovery (and response on failure) + reformats code
- remove generic Gin recovery middleware
- remove skip logging path code (it is useful to log the api/v1/streaming requests)


NyaaaWhatsUpDoc:
It seems I revealed an issue with our tests by removing that ErrorHandler panic recovery :')

NyaaaWhatsUpDoc:
instance tests fixed, i'm not even sure which ones are failing now as i can't tell from the logs... this is a later kim problem :p

internal/log/log.go
```diff
@@ -60,6 +54,17 @@ func Initialize() error {
 		}
 	}
 
+	// set our custom formatter options
+	logrus.SetFormatter(&logrus.TextFormatter{
+		DisableColors: true,
+		FullTimestamp: true,
+
+		// we disable quoting on debug builds, as this
```
tsmethurst, line 62: Hmm this is a bit misleading because it's not a debug build in this case, it's just log level >= debug. Was this intentional?

internal/router/logger.go
```diff
@@ -19,55 +19,72 @@
...
 	"time"
 
+	"codeberg.org/gruf/go-bytesize"
+	"codeberg.org/gruf/go-errors/v2"
 	"github.com/gin-gonic/gin"
 	"github.com/sirupsen/logrus"
 )
 
-var skipPaths = map[string]interface{}{
-	"/api/v1/streaming": nil,
```
tsmethurst, line 30: i'd be careful here; one reason we skipped this path was to avoid logging users' tokens, since for the streaming endpoint they're contained in the path... so make sure any changes you make don't end up leaking tokens!

tsmethurst (commented):
Looks good, just a couple little comments :)

internal/log/log.go
NyaaaWhatsUpDoc: debug >= logrus.DebugLevel means that it logs for debug and anything more verbose than that, e.g. trace. should i be a bit more clear about that?

internal/router/logger.go
NyaaaWhatsUpDoc: Maybe I should a comment explaining why it's important not to log the query params

testrig/gin.go
```diff
@@ -0,0 +1,19 @@
+package testrig
```
tsmethurst, line 1: license text ;)

NyaaaWhatsUpDoc: sadlfkjhadsf

PR #711, opened 2022-07-15
[feature] Implement `cache-control` and etags for static assets

tsmethurst:
This PR adds a new middleware used specifically on the static file server at `/assets`.

The middleware adds smarter use of the `cache-control` header to prevent clients from having to refetch assets over and over, or indeed to never fetch assets again (heuristic caching).

Each asset now has an etag generated for it, and the client's If-None-Match header is properly checked against the etag.

This means that when frontend styling stuff changes, clients will intelligently fetch the changed versions, but if nothing changes then they will receive a 304.

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -68,12 +68,9 @@ func (r *router) AttachStaticFS(relativePath string, fs http.FileSystem) {
 	// group will consiste of endpoints under relativePath, so
 	// something like "/assets"
 	group := r.engine.Group(relativePath)
-	group.Use(func(c *gin.Context) {
-		// file system will already set 'last-modified' headers,
-		// but we want to set cache-control too to make sure callers
-		// understand they can cache static assets
-		c.Header("Cache-Control", "max-age=604800")
-	})
+
+	// use the cache middleware on all handlers in this group
+	group.Use(cacheMiddleware(fs))
 
 	// serve static file system in the root of this group,
 	// will end up being something like "/assets/"
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -64,7 +64,20 @@ type router struct {
 
 // Add Gin StaticFS handler
 func (r *router) AttachStaticFS(relativePath string, fs http.FileSystem) {
-	r.engine.StaticFS(relativePath, fs)
+	// create a group so that we can attach a middleware to it
+	// group will consiste of endpoints under relativePath, so
+	// something like "/assets"
+	group := r.engine.Group(relativePath)
+	group.Use(func(c *gin.Context) {
+		// file system will already set 'last-modified' headers,
+		// but we want to set cache-control too to make sure callers
+		// understand they can cache static assets
+		c.Header("Cache-Control", "max-age=604800")
+	})
+
+	// serve static file system in the root of this group,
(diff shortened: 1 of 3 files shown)


internal/router/router.go
```diff
@@ -64,7 +64,17 @@ type router struct {
...
+	// group will consiste of endpoints under relativePath, so
+	// something like "/assets"
+	group := r.engine.Group(relativePath)
+
+	// use the cache middleware on all handlers in this group
+	group.Use(cacheMiddleware(fs))
+
+	// serve static file system in the root of this group,
+	// will end up being something like "/assets/"
+	group.StaticFS("/", fs)
```
NyaaaWhatsUpDoc, line 77: Have you checked if this group provides any caching by default? If so that could cause issues with this approach. Also how do you feel about caching the determined etags? And you could have an admin endpoint which allows clearing this cache for when you want to modify files on disk without restart (also handled by reload when we implement that)

internal/web/assetscache.go
```diff
@@ -0,0 +1,153 @@
...
+
+// generateEtag generates a weak etag by concatenating the given filePath string
+// with the unix timestamp of the given time, then doing a sha1 hash on the result.
+func generateEtag(filePath string, lastModified time.Time) (string, error) {
+	b := []byte(fmt.Sprintf("%s%d", filePath, lastModified.Unix()))
+
+	// nolint:gosec
+	hash := sha1.New()
+
+	if _, err := hash.Write(b); err != nil {
```
NyaaaWhatsUpDoc, line 43: instead of instantiating a sha1 hash, might be easier to replace that with just `sha1.Sum(b)`. The hash write will never return error anywho :)

internal/web/assetscache.go
```diff
@@ -0,0 +1,153 @@
...
+		return "", err
+	}
+
+	return `/W"` + hex.EncodeToString(hash.Sum(nil)) + `"`, nil
```
NyaaaWhatsUpDoc, line 47: actually, saying the above, i'm guessing you meant to do that but somehow ended up with nil here instead of `b` :D

tsmethurst: oh! i'll change it :D

internal/router/router.go
tsmethurst: As far as I can tell, it doesn't provide caching by default because it never sets the cache-control header.... that said, I should check what happens if the header is set... I'll get back to you!

tsmethurst: Hah! OK I ran the testrig with just the `cache-control: no-cache` header set on `/assets` endpoints, and indeed the http filesystem serving does return 304 if `if-modified-since` is set by the caller is newer than the `last modified` time of the asset.

So this suggests that actually we don't really need to do the etag thing at all, and we can probably just make do with using last-modified and if-modified-since..... :clown_face:

tsmethurst: OK further update! @f0x52 informs me that in nixos, all the files have a last-modified of 0 to help with reproducibility... so we still need the etag thing!

This means that we should also be wary of creating our etags using just last modified and filename, because this will break things on nixos. So we need to figure out a way to generate useful etags on the assumption that we can't trust last-modified to be present/correct.

PR #715, opened 2022-07-18
[chore] move dialer inside new, use default resolver

tsmethurst:
Fixes the very minor incorrect pointer copying bug we had by just moving instantiation of the dialer inside the New function.


NyaaaWhatsUpDoc:
I think my only critique is I'd prefer if we did instantiate the resolver, simply because it means we don't have to worry about weird errors ever caused by libraries modifying the global default resolver and us not being aware. Otherwise all looks good :)

Issue #632, opened 2022-06-05
[bug] Can't Clear Notifications in Tusky


zethra:
### Describe the bug with a clear and concise description of what the bug is.

In Tusky, the clear button under the notifications tab doesn't do anything. On Mastodon it works. 

### What's your GoToSocial Version?

v0. 3.4

### GoToSocial Arch

amd64 Docker

### What happened?

I click the clear button in the notifications tab in Tusky and nothing happens. 

### What you expected to happen?

The notifications are cleared. 

### How to reproduce it?

Click the clear button in Tusky when notifications are present.

tsmethurst:
Heya! This is normal: we haven't implemented that endpoint yet. I'll try to get it in soon because it's a pretty easy fix :)

tsmethurst:
@Arteneko has volunteered to take a look at this one :)

Here's a quick little summary of the work that needs to be done:

1. create a 'delete notifications' function in the interface here: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/db/notification.go and implement it here: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/db/bundb/notification.go
2. add a test for the function here: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/db/bundb/notification_test.go
3. Add a function to the processor here: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/processing/notification.go
4. Add an api endpoint for deleting notifications here: https://github.com/superseriousbusiness/gotosocial/tree/main/internal/api/client/notification. This should follow the mastodon api for `/api/v1/notifications/clear` as described here: https://docs.joinmastodon.org/methods/notifications/

Any questions or issues or anything, please feel free to ask! :sloth:

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/issues/632

PR #720, opened 2022-07-19
Implemented notification clear

Arteneko:
Per #632

### DIFF
--- a/internal/db/bundb/notification.go
+++ b/internal/db/bundb/notification.go
@@ -118,5 +118,7 @@ func (n *notificationDB) ClearNotifications(ctx context.Context, accountID strin
 		return n.conn.ProcessError(err)
 	}
 
+	n.cache.Clear()
+
 	return nil
 }
--- a/internal/db/bundb/notification.go
+++ b/internal/db/bundb/notification.go
@@ -108,3 +108,15 @@ func (n *notificationDB) GetNotifications(ctx context.Context, accountID string,
 
 	return notifs, nil
 }
+
+func (n *notificationDB) ClearNotifications(ctx context.Context, accountID string) db.Error {
+	if _, err := n.conn.
+		NewDelete().
+		Table("notifications").
+		Where("target_account_id = ?", accountID).
+		Exec(ctx); err != nil {
+		return n.conn.ProcessError(err)
+	}
+
+	return nil
+}
--- a/internal/db/bundb/notification_test.go
+++ b/internal/db/bundb/notification_test.go
@@ -118,6 +118,18 @@ func (suite *NotificationTestSuite) TestGetNotificationsWithoutSpam() {
 	}
 }
 
+func (suite *NotificationTestSuite) TestClearNotificationsWithSpam() {
+	suite.spamNotifs()
+	testAccount := suite.testAccounts["local_account_1"]
+	err := suite.db.ClearNotifications(context.Background(), testAccount.ID)
+	suite.NoError(err)
+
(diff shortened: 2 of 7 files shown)


internal/db/bundb/notification.go
```diff
@@ -108,3 +108,15 @@ func (n *notificationDB) GetNotifications(ctx context.Context, accountID string,
...
+
+func (n *notificationDB) ClearNotifications(ctx context.Context, accountID string) db.Error {
+	if _, err := n.conn.
+		NewDelete().
+		Table("notifications").
+		Where("target_account_id = ?", accountID).
+		Exec(ctx); err != nil {
+		return n.conn.ProcessError(err)
+	}
+
```
NyaaaWhatsUpDoc, line 120: I'm wondering whether we should clear the notifications cache here (i.e. `n.cache.Clear()`). The notifications won't be visible anymore to `.GetNotifications()`, but if the client has IDs cached then individual calls to `.GetNotification()` would still return deleted notifications.

From a performance perspective, clearing the entire notification cache because one user cleared their notifications isn't great, but I think it should be rare enough that it shoulldn't be a big issue, and it's something we can address at a later date if necessary.

internal/db/bundb/notification_test.go
```diff
@@ -118,6 +118,18 @@ func (suite *NotificationTestSuite) TestGetNotificationsWithoutSpam() {
...
 
+func (suite *NotificationTestSuite) TestClearNotificationsWithSpam() {
+	suite.spamNotifs()
+	testAccount := suite.testAccounts["local_account_1"]
+	err := suite.db.ClearNotifications(context.Background(), testAccount.ID)
+	suite.NoError(err)
+
+	notifications, err := suite.db.GetNotifications(context.Background(), testAccount.ID, 20, "ZZZZZZZZZZZZZZZZZZZZZZZZZZ", "00000000000000000000000000")
+	suite.NoError(err)
+	suite.NotNil(notifications)
```
NyaaaWhatsUpDoc, line 129: I wouldn't worry about doing a not-nil check here. A zero length slice and a nil slice are fairly interchangable in uses except for the nil check.

NyaaaWhatsUpDoc:
Aside from the few comments, this looks great :)

internal/db/bundb/notification_test.go
```diff
@@ -118,6 +118,18 @@ func (suite *NotificationTestSuite) TestGetNotificationsWithoutSpam() {
...
+	suite.Empty(notifications)
+}
+
```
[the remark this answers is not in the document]
tsmethurst: More tests certainly don't hurt! Are you up for it @Arteneko ? This can be definitely considered an optional extra, if you've still got time/energy you wanna spend on this PR :)

tsmethurst (commented):
Agree with @NyaaaWhatsUpDoc, this PR looks great, thank you for doing this! :)

Once the notifications cache clear is implemented, and (if you feel like doing it) the additional test is written, this looks ready to merge :+1:

internal/db/bundb/notification.go
NyaaaWhatsUpDoc: It's as simple as adding `n.cache.Clear()` on line 121 :)

internal/db/bundb/notification_test.go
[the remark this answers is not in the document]
Arteneko: sure, i'll do it once i have some downtime

tsmethurst:
Thanks! :) :)

PR #724, opened 2022-07-21
[bugfix] Make `/api/v2/media` more compatible with masto API

tsmethurst:
This PR tweaks the api model for a media attachment a little bit, to make some fields return null instead of empty strings.

In addition, it implements a workaround which nulls the URL field returned by `/api/v2/media`, in keeping with the mastodon API, which expects callers to just use the returned ID and preview URL rather than having the actual URL filled in already.


mirabilos:
Will this fix POST to /api/v2/media timing out in Pinafore?

Issue #686, opened 2022-07-03
[bug] Posts from toot cli don't show images on posts


sexybiggetje:
### Describe the bug with a clear and concise description of what the bug is.

When posting from the [toot cli](https://toot.readthedocs.io/en/latest/index.html) with attached media, the media is being uploaded and processed by GtS but when requesting the post it's not actually attached.

I'm seeing in the GtS debug log that it says **"object":{"attachment":[]"** maybe the problem lies there? Shouldn't that contain a reference to the actual attachment?

Invocation of toot. Note that I don't specify an image alt, since i'm using toot 0.27.0 but this was added in 0.28.0. 
```
 /usr/bin/toot post --debug -u dutchbeercollective@plantsand.coffee -l nld -m data/latestthumb-thomasopent.jpg -v public "Test toot"
```

Relevant toot
```
https://plantsand.coffee/@dutchbeercollective/statuses/01G7276A6VEFC93F5P64QX0VCE
```

Relevant debug output from toot
```
Uploading media: /thedutchbeercollective/data/latestthumb-thomasopent.jpg
DEBUG:toot:>>> POST https://plantsand.coffee/api/v1/media
DEBUG:toot:>>> HEADERS: {('Authorization', '***CENSORED***'), ('User-Agent', 'toot/0.27.0')}
DEBUG:toot:>>> FILES:   {'file': <_io.BufferedReader name='/thedutchbeercollective/data/latestthumb-thomasopent.jpg'>}
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): plantsand.coffee:443
DEBUG:urllib3.connectionpool:https://plantsand.coffee:443 "POST /api/v1/media HTTP/1.1" 200 322
DEBUG:toot:<<< <Response [200]>
[...]

PR #726, opened 2022-07-22
[bugfix] Fix Toot CLI media attachments not working properly

tsmethurst:
This PR fixes the issue of the Toot cli not attaching media correctly to new statuses.

The problem was that when it comes to POST form data, the Mastodon API expects clients to use the key `media_ids[]` for their list of media attachment IDs. However, GtS was just accepting `media_ids`, which works great for json, but not so much for clients that use POST forms like Toot does.

This PR implements a simple fix for this by parsing `media_ids[]` in forms, but still using `media_ids` for json and xml. Tested and working with Toot, and therefore should work for other mastodon clients as well.

closes https://github.com/superseriousbusiness/gotosocial/issues/686

### DIFF
--- a/docs/api/swagger.yaml
+++ b/docs/api/swagger.yaml
@@ -161,6 +161,10 @@ definitions:
         description: |-
           Array of Attachment ids to be attached as media.
           If provided, status becomes optional, and poll cannot be used.
+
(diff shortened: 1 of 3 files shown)


tsmethurst:
Closes https://github.com/superseriousbusiness/gotosocial/issues/686

PR #734, opened 2022-07-29
Update client list and apply some minor fixes

sexybiggetje:
Adds a two popular clients (Fedilab for Android and Toot! for iPhone+iPad) and a link to the joinmastodon/apps page which houses a) an elaborate client list and b) the official Mastodon-branded Mastodon clients.

Adds logos for those to the assets folder. They are grabbed from their own resources.
Adds periods after all client descriptions to make them uniform in writing.
Adds target = blank to all external links, adds rel = noopener to all external links.
Changes the Pinafore link to point to the homepage instead of the add instance settings. (Pointing to the instance add page causes problems if you already added it to your session, and use the button to get to Pinafore.).

Note that Toot! for iOS doesn't really have a public facing homepage, so it links directly to the app store page. I assume that's becuase that's the main way of discovering "apps" on iOS nowadays.

Thanks for looking at this PR. It's mainly meant to help users find a good client and the experience they want.


sexybiggetje:
A small screenshot of the changes as previewed on my staging html.
plantsand coffee_

sexybiggetje:
Removed the added clients Fedilab and Toot as per discussion on the matrix channel. Kept the More clients and other punctual changes.

tsmethurst:
Thank you for this! :heart:

PR #736, opened 2022-08-03
[feature] Add first iteration of a user panel at `/user`

tsmethurst:
This PR adds a first iteration of a user control / settings panel at the `/user` endpoint.

The panel allows users to change their profile + header pics, update their bio and display name, change their post settings, and change their password.

The PR also updates some of the Go code to support these changes and fix a couple bugs.

Screenshot of the user panel

Another screenshot of the user panel


f0x52:
Both the user and admin panel don't handle expired/wrong oauth tokens, resulting in some broken UI state.
Cosmetics wise I'll look at styling the dropdowns+file selectors to be more consistent with the existing buttons etc, and at deduplicating more code with the admin panel.
For now though I'll fix the inconsistent indentation, and then we can merge it for some user feedback I think.

f0x52 (approved):

f0x52 (approved):

f0x52:
I think it's good to merge now, although I will work on it again soon(tm)

PR #741, opened 2022-08-06
feat: add rate limit middleware

### LINKED ISSUE #23: [feature] Rate limiting of API requests
There's no rate-limiting for inbound requests currently, and this should be implemented to avoid the server being spammed/scraped by bad eggs.

There are a couple of middlewares available for Gin to do rate limiting:

* https://github.com/axiaoxin-com/ratelimiter
* https://github.com/yangxikun/gin-limit-by-key/blob/master/limit.go
* Probably some more that I didn't see yet.

These should be evaluated to see if they fit the needs of GoToSocial.

Ideally, whatever rate limiting implementation is used should have the following characteristics:

1. Use the same rate-limit response headers as Mastodon for compatibility with Mastodon applications.
2. Use the client IP as a rate-limiting key.
3. In-memory implementation (good-enough for now).
4. Sliding window limits (might be a bit complicated though -- fixed limits per minute would already be a fine start).

The rate limit implementation should be added to the security package here: https://github.com/superseriousbusiness/gotosocial/tree/main/internal/api/security

nya1:
Closes #23

Added rate limiter by IP address (in memory cache), by default a maximum of 300 requests are allowed with a time period of 5 minutes.

New module added: `github.com/ulule/limiter`
New response headers returned to the caller (same keys used by [mastodon](https://docs.joinmastodon.org/api/rate-limits/)):
- `x-ratelimit-limit` maximum number of requests allowed per time period (fixed)
- `x-ratelimit-remaining` number of remaining requests that can still be performed
- `x-ratelimit-reset` unix timestamp when the rate limit will reset

When the rate limit is exceeded an error with HTTP status code of 429 is returned to the caller, example:

```
curl -i http://localhost:8080/api/v1/instance
HTTP/1.1 429 Too Many Requests
Content-Type: application/json; charset=utf-8
X-Ratelimit-Limit: 300
X-Ratelimit-Remaining: 0
X-Ratelimit-Reset: 1659782950
Date: Sat, 06 Aug 2022 10:48:15 GMT
Content-Length: 30

{"error":"rate limit reached"}
```


tsmethurst (commented):
Looks good to me! I like how simple this implementation is :) Thank you for doing this!!

Just two things:

1. Could you add the dependency on the rate limiter to the README.md, in the same way the other Go dependencies are listed?
2. Could you add a bit of text at the top of [this documentation file](https://github.com/superseriousbusiness/gotosocial/blob/main/docs/api/swagger.md) to describe the rate limit headers?

Also as @f0x52 said in the matrix chat, we'll have to do some work in the future to limit ipv6 by block (perhaps this is something we can contribute upstream to this library so that more people can benefit from it). That is beyond the scope of this PR though, just something we need to bear in mind.

internal/api/security/security.go
```diff
@@ -45,6 +46,11 @@ func New(db db.DB, server oauth.Server) api.ClientModule {
 
 // Route attaches security middleware to the given router
 func (m *Module) Route(s router.Router) error {
+	s.AttachMiddleware(m.RateLimit(RateLimitOptions{
+		// accept a maximum of 300 requests in 5 minutes window
```
NyaaaWhatsUpDoc, line 50: At some point we should make this customizable. And until that point we should "err" on the side of caution. For an instance undergoing a lot of interactions from another big instance I'm worried that 300 requests could quite easily be reached in 5 minutes. Maybe increase the limit to be 500 / 1000?

tsmethurst: Perhaps we should leave out rate limiting for requests crossing the s2s/federation API entirely, or just attach a different (more relaxed) rate limiter there. Probably out of scope rn though, so maybe just upping the cap for now is a good move until we implement more granular limiting?

tsmethurst:
(took the liberty of merging main into this to resolve the conflict for ya)

internal/api/security/security.go
NyaaaWhatsUpDoc: Agreed :)

nya1: I see, I've now increased the request limit to 1000

tsmethurst:
Thank you! :) Once we've cut release 0.4.0 we'll merge this into main :)

PR #744, opened 2022-08-07
[chore] Update js deps

f0x52:
updates the postcss stack to version 8, including the postcss-custom-prop-vars library which was printing an upgrade message. Also removes some postcss plugins we weren't actually using


f0x52:
picked some css files from dist at random and they had 0 diffs, so seems consistent :)

Issue #486, opened 2022-04-25
[chore] Tidy up Bun annotations to allow upstep to bun v1.1.4


tsmethurst:
Stepping from Bun v1.1.3 => v1.1.4 produces lots of failing tests, because of the nullzero change in bun: https://github.com/uptrace/bun/blob/master/CHANGELOG.md#114-2022-04-20

To make GtS work properly with the new version of bun we need to carefully go through our structs in the `gtsmodel` package and make sure we're using the combination of `default`, `nullzero` and `notnull` tags correctly on our fields. See: https://github.com/uptrace/bun/issues/523#issuecomment-1107833377

igalic:
> In v1.1.4 we made change to automatically apply `nullzero` when you have `default:something` option, because otherwise the default is just ignored.

unless i completely misunderstanding what `nullzero` does, that sounds like a really weird change.
it also sounds like a total butt move to do that on a patch upgrade

tsmethurst:
Eh, it's ok :) we'll make it work

PR #754, opened 2022-08-13
[chore] Update bun / sqlite versions; update gtsmodels

tsmethurst:
This PR upsteps our bun and sqlite dependencies to the latest versions.

In order to make this work, `bool` values on many database models had to be changed from `bool` to `*bool`, due to a change in bun.

The PR also adds some logic into the database to update db entries more selectively when using `UpdateByID` -- callers can now pass in the specific columns they wish to update.

For reviewers:

- These are the commits where bools are changed to bool pointers: https://github.com/superseriousbusiness/gotosocial/pull/754/commits/672b3133bc8b9a0f2b3ffaae8ff96589dbb4553b https://github.com/superseriousbusiness/gotosocial/pull/754/commits/93db2ad1beaa025c4df17030ce1ee5092cca380f (arguably the most crucial part of the PR)
- Everything in the `vendor` subdirectories doesn't need to be reviewed, it's just updating dependencies.

Closes https://github.com/superseriousbusiness/gotosocial/issues/486

Deployed this on both gts.superseriousbusiness.org and goblin.technology (docker/postgres and binary/sqlite respectively) and it seems stable so far.


internal/ap/extract.go
```diff
@@ -529,6 +529,12 @@ func ExtractEmoji(i Emojiable) (*gtsmodel.Emoji, error) {
 	}
 	emoji.ImageRemoteURL = imageURL.String()
 
+	disabled := false
+	visibleInPicker := false
+
+	emoji.Disabled = &disabled
```
NyaaaWhatsUpDoc, line 535: not worth changing now, but a quicker way of initializing a zero value bool ptr is `new(bool)` :)

internal/cache/account.go
```diff
@@ -131,7 +126,38 @@ func copyAccount(account *gtsmodel.Account) *gtsmodel.Account {
 		SensitizedAt:            account.SensitizedAt,
 		SilencedAt:              account.SilencedAt,
 		SuspendedAt:             account.SuspendedAt,
-		HideCollections:         account.HideCollections,
 		SuspensionOrigin:        account.SuspensionOrigin,
 	}
+
+	if account.Memorial != nil {
```
NyaaaWhatsUpDoc, line 132: might be neater to make a `copyBoolPtr()` function under say `cache/util.go` and then you can keep the original code but just wrap the bools in `copyBoolPtr()`? Since it's just for neatness this is very much a subjective thing though so obviously let me know if you disagree :)

internal/cache/status.go
```diff
@@ -119,15 +118,46 @@ func copyStatus(status *gtsmodel.Status) *gtsmodel.Status {
...
-		Federated:                status.Federated,
-		Boostable:                status.Boostable,
-		Replyable:                status.Replyable,
-		Likeable:                 status.Likeable,
 		ActivityStreamsType:      status.ActivityStreamsType,
 		Text:                     status.Text,
-		Pinned:                   status.Pinned,
 	}
+
+	if status.Local != nil {
```
NyaaaWhatsUpDoc, line 127: same as cache/account.go

NyaaaWhatsUpDoc:
> Deployed this on both gts.superseriousbusiness.org and goblin.technology (docker/postgres and binary/sqlite respectively) and it seems stable so far.

Just to double check, have you grepped your logs for any nil-ptr related panics?

tsmethurst:
> Just to double check, have you grepped your logs for any nil-ptr related panics?

Yeah :) Only one cropped up, which led to https://github.com/superseriousbusiness/gotosocial/pull/754/commits/7ab81819fcc56de658eac9268adb8498acd89ab6, and since then I haven't seen any more

internal/cache/account.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/754/commits/ac1e20c678bc7f0609ed94198140ee1b726e329e

internal/ap/extract.go
tsmethurst: I will change it anyways :D Thanks for pointing it out!

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/754/commits/e731b3385454605b524d2d4196f79687340d1de6 done!

PR #760, opened 2022-08-21
[bugfix] Relax outgoing http request queue

tsmethurst:
This PR updates the http request queue to implement per-host and per-method queues.

This will ensure that a bunch of POST requests held open to an instance don't prevent outgoing GETs to main-keys from processing.

This is definitely not the only way we can fix this problem, but it seems like a decent fix for now until we can implement something better.

Writeup from matrix:

---

> you export your instance and import it again: all the accounts that you follow and that follow you (plus some others) are transferred from the old database to the new -- mastodon instance actors are not (yet) transferred
you make a post and it goes out to deliver to x amount of inboxes (let's say 100 if you have 100 followers) - then while your requests to these inboxes are still open:
> a number of requests come in from Mastodon instances, via their instance Actor, to dereference your main key (/users/some_user/main-key), and your AP representation (/users/some_user) to validate your signature/do whatever mastodon does
Your instance doesn't have the key of the instance actor that is trying to dereference your user, so --
> Suddenly, your instance has to make like 100 requests at once across the fediverse for these different instance actor keys, to verify the http signatures on the dereference requests from step 3
> These outgoing GET requests to remote instances can't go through because the request queue is full of waiting, outgoing POST requests.
> All the incoming requests from Mastodon for your /users/some_user time out after 10 seconds
> All the outgoing requests your instance was making to the instance actors time out with context-canceled (because they use the inbound request context to make the outbound request)
> Mastodon instances return 401s from all the inbox POST requests your making, because they couldn't validate your http signature


internal/httpclient/client.go
```diff
@@ -143,13 +148,24 @@ func New(cfg Config) *Client {
...
+	host := req.Host
+
+	l := log.WithFields(kv.Fields{
+		{"host", host},
+		{"method", method},
+	}...)
+
+	wait, done := c.rc.getWaitSpot(host, method)
+
+	l.Trace("client waiting for slot in request queue")
```
NyaaaWhatsUpDoc, line 161: I'd be a bit concerned with adding logging into the httpclient package. We already have logging one step up, but adding it here will add *a lot* of extra trace logging. And could add a lot of extra warn logging

tsmethurst: removed :)

tsmethurst:
Note: This will definitely need further tuning + fiddling in future to lower the overall number of potential outbound requests, but I will merge it now so we can test it out for a while.

PR #766, opened 2022-08-26
validateCreateStatus: Check the presence of form.MediaIDs more thoroughly, instead of just checking for null

blackle:
awesome software! just set up a test instance tonight and ran into this snag. It seems that polls aren't really supported yet, but to get it to work eventually this change will probably be necessary.

When sending a post on tusky, it appears that "form.MediaIDs" isn't nil but instead an empty array. If you try to attach a poll, the post is rejected because this code thinks there's both a poll and media on the status.

On pinafore, this doesn't occur (though the post is still rejected because the expires_in field of the payload is a string and not an int. different problem that probably needs to be fixed in pinafore)

### DIFF
--- a/internal/api/client/status/statuscreate.go
+++ b/internal/api/client/status/statuscreate.go
@@ -105,11 +105,15 @@ func (m *Module) StatusCreatePOSTHandler(c *gin.Context) {
 }
 
 func validateCreateStatus(form *model.AdvancedStatusCreateForm) error {
-	if form.Status == "" && form.MediaIDs == nil && form.Poll == nil {
+	hasStatus := form.Status != ""
+	hasMedia := len(form.MediaIDs) != 0
+	hasPoll := form.Poll != nil
+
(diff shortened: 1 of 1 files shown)


tsmethurst:
Thanks for the PR! Did you test this on your instance yet? Reason I'm asking is because the nil value for a `[]string` should (iirc) just be an empty slice anyway, since it's not a pointer, so I don't think the length check is necessary. Though I could well be wrong, I very often get these things mixed up :P

NyaaaWhatsUpDoc:
Yes @tsmethurst is correct you don't need a nil check on the media IDs. A nil slice returns a length of zero (no panic).

blackle:
ok, just updated it to check the length only. tested on my instance with both pinafore and tusky. it still rejects polls from pinafore, but that is because the expires_in field is an unexpected type (string instead of int)

tsmethurst:
Merged, thank you! :)

PR #770, opened 2022-08-27
Fix link to open collective

chmac:
I presumed that you want to link directly to your actual collective and not to the open collective homepage, hence this PR. Please disregard if I was mistaken.


tsmethurst:
Thanks! :)

PR #774, opened 2022-08-27
Sort follow requests, followers, and following by updated_at

blackle:
so it is consistent with mastodon and other AP implementations, where the most recent followers are shown first.


blackle:
added the migration. `explain query plan` seems not to use a temporary b-tree anymore!

PR #779, opened 2022-08-30
Fix issue where the 'favourited' field of a notification's status is always false

blackle:
since the requesting account was nil in the call to StatusToAPIStatus, it couldn't set the favourited flag correctly. so if you favourite a post from a notification then reload notifications, the favourite disappears (even though the favourite was successful)


tsmethurst:
Good catch :+1:

PR #782, opened 2022-08-30
Fix subway tooter problems

### LINKED ISSUE #304: [bug] Subway Tooter issues
Subway Tooter (https://github.com/tateisu/SubwayTooter) gets a 500 error when logging in.

Relevant logs:

```
time=2021-11-16T15:10:11Z level=info msg=[1.458397ms] Internal Server Error: wrote 154 bytes clientIP=REDACTED latency=1.458397ms method=POST path=/oauth/token statusCode=500 userAgent=SubwayTooter/4.7.3 Android/10
```

blackle:
fixes #304, but in a bad way so draft PR

it seems that subway tooter doesn't send along the redirect_uri in the token request. [the RFC for oauth says this is required](https://www.rfc-editor.org/rfc/rfc6749.html#section-3.1.2:~:text=redirect_uri%0A%20%20%20%20%20%20%20%20%20REQUIRED%2C%20if%20the%20%22redirect_uri%22%20parameter%20was%20included%20in%20the%0A%20%20%20%20%20%20%20%20%20authorization%20request%20as%20described%20in%20Section%204.1.1%2C%20and%20their%0A%20%20%20%20%20%20%20%20%20values%20MUST%20be%20identical.) (optional for the /authorize request). I guess mastodon uses the redirect_uri that the application originally registered with if it's absent. this change adopts that behaviour and allows me to login to subway tooter.

kinda a bad fix because we're going against the oauth spec, but if this is what mastodon also does maybe it's necessary. the alternative is to make a PR to subway tooter

### DIFF
--- a/internal/api/client/auth/token.go
+++ b/internal/api/client/auth/token.go
@@ -76,10 +76,9 @@ func (m *Module) TokenPOSTHandler(c *gin.Context) {
 		help = append(help, "client_secret was not set in the token request form")
 	}
 
+	// redirect_uri is optional
 	if form.RedirectURI != nil {
 		c.Request.Form.Set("redirect_uri", *form.RedirectURI)
-	} else {
-		help = append(help, "redirect_uri was not set in the token request form")
 	}
 
 	var code string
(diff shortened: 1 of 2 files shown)


tsmethurst:
Hmmm nice that you identified what the issue is, well spotted :sunglasses: However I don't think changing the code is a good idea. Even leaving aside the fact that it's against the oauth spec, the mastodon api docs also specify that this is a required field: https://docs.joinmastodon.org/methods/apps/oauth/. So to me this looks v. much like something that should be fixed in Subway Tooter

blackle:
ok, I'll make a subway tooter PR. seems like a quick fix, I've already identified the problematic code there

PR #785, opened 2022-08-30
Emojify spoiler and content in web templates

### LINKED ISSUE #648: [bug] Custom emojis aren't rendered properly in web views
In the web view of a mastodon status, custom emojis are rendered with a bunch of fancy image tags:

```
<p>can&apos;t get your ass kicked if you don&apos;t have an ass <img draggable="false" class="emojione" alt=":blobcatthinkOwO:" title=":blobcatthinkOwO:" src="https://ondergrond.org/system/custom_emojis/images/000/000/824/original/e29540d4e94b3dff.png" /></p>
```

We should do something similar for statuses rendered in our web view, so that custom emojis can be shown via the web.

Note that this should be done for frontend views of statuses only, the status should still be federated (and stored internally) without the image tags.

blackle:
fixes #648

funny enough, emojification bugs were one of the first things I worked on for mastodon 5 years ago

probably my most complicated commit so far, the creation of the regex and template can maybe be optimized?

emojify takes a template.HTML parameter for the text because it assumes that everything is already escaped, and its output can't be escaped again or it will destroy the img tags. I noticed that .Content gets sent through noescape, but not .SpoilerText. Should SpoilerText be html escaped?

the "emoji" css class still needs to be made, not sure where to put it

### DIFF
--- a/internal/router/template.go
+++ b/internal/router/template.go
@@ -21,6 +21,7 @@ package router
 import (
 	"bytes"
 	"fmt"
+	"html"
 	"html/template"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


blackle:
added an emoji class, reused the EmojiFinder regex, and escaped the spoiler text before emojifying it, since escaping the spoiler text would have been the original behaviour of the template

tsmethurst (commented):
Looks pretty cool and I'm excited to try it out :) just a few small comments

internal/router/template.go
```diff
@@ -97,12 +104,43 @@ func visibilityIcon(visibility model.Visibility) template.HTML {
...
+
+	emojiTemplate, _ := template.New("emojiTemplate").Parse(`{{define "Emoji"}}<img src="{{.URL}}" title=":{{.Shortcode}}:" alt=":{{.Shortcode}}:" class="emoji"/>{{end}}`)
+
+	emojifyer := func(shortcode string) string {
+		emoji, ok := emojisMap[shortcode]
+		if ok == false {
+			return shortcode
+		}
+		var buff bytes.Buffer
+		err := emojiTemplate.ExecuteTemplate(&buff, "Emoji", emoji)
```
NyaaaWhatsUpDoc, line 124: Is there a reason we're using a template here and not building the string using `fmt` or even manually? Using templates will incur quite an overhead compared to the former.

NyaaaWhatsUpDoc: Could we do something like this instead?

```golang
// replaces shortcodes in `text` with the emoji in `emojis`
// text is a template.HTML to affirm that the input of this function is already escaped
func emojify(emojis []model.Emoji, text template.HTML) template.HTML {
	emojisMap := make(map[string]model.Emoji, len(emojis))

	for _, emoji := range emojis {
		shortcode := ":" + emoji.Shortcode + ":"
		emojisMap[shortcode] = emoji
	}

	out := regexes.ReplaceAllStringFunc(
		regexes.EmojiFinder,
		string(text),
		func(shortcode string, buf *bytes.Buffer) string {
			// Look for emoji according to this shortcode
			emoji, ok := emojisMap[shortcode]
			if !ok {
				return shortcode
			}

			// Escape raw emoji content
			safeURL := html.EscapeString(emoji.URL)
			safeCode := html.EscapeString(emoji.Shortcode)

			// Write HTML emoji repr to buffer
			buf.WriteString(`<img src="`)
			buf.WriteString(safeURL)
			buf.WriteString(`" title=":`)
			buf.WriteString(safeCode)
			buf.WriteString(`:" alt=":`)
			buf.WriteString(safeCode)
			buf.WriteString(`:" class="emoji"/>`)

			return buf.String()
		},
	)

	/* #nosec G203 */
	// (this is escaped above)
	return template.HTML(out)
}
```

blackle: looks good! I was using templates to be overly cautious, but thinking about it more, all the stuff we're using as attributes for the image tag should already be safe (shortcodes only contain alphanumeric and underscores, urls should be url encoded and not contain a double quote)

I'll use this code

blackle: oh I just noticed you also call EscapeString, which shouldn't hurt to be extra careful. just pushed the change.

tsmethurst:
nice, thank you!! :)

tsmethurst:
Nice one :) very excited for getting a step closer to proper emoji support, thanks for all your hard work! I'll have a look over this tomorrow

Sent from ProtonMail mobile

-------- Original Message --------

PR #786, opened 2022-08-31
Fix tusky search issue by returning empty if offset is greater than zero

### LINKED ISSUE #68: [bug] Tusky: searching for @user@instance goes into a UI discovery loop
I don't have a nice way to take a  screencast from my phone right now but when I search for @user@instance in the tusky interface, when the account is found, the interface starts to endlessly list them in a loop. You can still choose the first user found and the follow works out, so this is not blocking me in any way. Hope that is clear :+1:

blackle:
fixes #68

basically the issue is not using limit/offset, but because the code only ever returns one item, we can underhandedly-implement offset to fix this bug


tsmethurst:
Nice catch :)

PR #787, opened 2022-08-31
Use custom blackfriday renderer to only add mention/hashtag links in normal text

### LINKED ISSUE #783: [bug] Hashtags inside markdown pre-blocks

Posting this:
````markdown
```
hej hej #hej #hej
```
````

Results in this:
````
hej hej <a href="https://doma.in/tags/hej" class="mention hashtag" rel="tag">#<span>hej</span></a> <a href="https://doma.in/tags/hej" class="mention hashtag" rel="tag">#<span>hej</span></a>
````

0.4.0-rc2

### GoToSocial Arch

amd64 binary

### What happened?

The comment character in the code block turned into links, but not actual links, just HTML within the code block.

No links should be there.

### How to reproduce it?

Post
````markdown
```
Anything #thisWillBecomeALink
```
````

You are nice and good.

### PR DESCRIPTION
**Pull request by blackle**
fixes #783. blackfriday lets you specify your own renderer, which gives us more control over where we insert the hashtag/mention links. see: https://github.com/russross/blackfriday/blob/v2/html.go#L508

basically, `blackfriday.Text` signals raw text output, and `blackfriday.Code` signals code block output, so we can avoid making the conversion there.

caveat: posting something like:
```
_#hashtag_
```
doesn't turn #hashtag into a hashtag, but
```
_#hashtag #hashtag #hashtag_
```
converts all three, so I think the code that generates the tag list needs some work in the markdown case.

### DIFF
--- a/internal/text/markdown.go
+++ b/internal/text/markdown.go
@@ -19,7 +19,9 @@
 package text
 
 import (
+	"bytes"
 	"context"
+	"io"
 
 	"github.com/russross/blackfriday/v2"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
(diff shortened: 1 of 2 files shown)


tsmethurst:
Nice fix! Can you add a couple test cases for this in the `text` package? 

> I think the code that generates the tag list needs some work in the markdown case

mm, sounds like it :thinking: we should open another issue to figure out these cases, will you do it or should I?

blackle:
added more tests, also opened #789 for the other issue

PR #791, opened 2022-09-01
Federate custom emoji (outbound)

blackle:
only the outgoing statuses for now. also adds a test that reads the full status from the DB and then serializes it, so we can compare the "tags" field (just using the raw test suite status doesn't have those fields populated)

I ran gofmt on the internaltoas.go file which is why there are more changes than just my code..

caveat: for some reason, gif emoji don't federate to some mastodon servers, even though the AP objects look basically identical (though we're missing the "updated" field in the emoji object, but static emoji still works in spite of that...)

### DIFF
--- a/internal/typeutils/converter.go
+++ b/internal/typeutils/converter.go
@@ -144,6 +144,8 @@ type TypeConverter interface {
 	FollowToAS(ctx context.Context, f *gtsmodel.Follow, originAccount *gtsmodel.Account, targetAccount *gtsmodel.Account) (vocab.ActivityStreamsFollow, error)
 	// MentionToAS converts a gts model mention into an activity streams Mention, suitable for federation
 	MentionToAS(ctx context.Context, m *gtsmodel.Mention) (vocab.ActivityStreamsMention, error)
+	// EmojiToAS converts a gts emoji into a mastodon ns Emoji, suitable for federation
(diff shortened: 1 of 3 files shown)


tsmethurst:
i saw your post about this and replied here: https://ondergrond.org/@dumpsterqueer/108922590917081621

just so it's here as well: I think it's because vanilla Mastodon has a size limit of 50kb for emojis, and those animated ones (or at least the one I checked) are bigger :(

tsmethurst:
To be as AP spec compliant as possible, we should also add an endpoint to the s2s API to serve the AP representations of emojis. For example, if you do (random example):

```
curl -H 'accept: application/activity+json' https://mstdn.social/emojis/42450
```

you get back:

```
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    {
      "toot": "http://joinmastodon.org/ns#",
      "Emoji": "toot:Emoji",
      "focalPoint": {
        "@container": "@list",
        "@id": "toot:focalPoint"
      }
    }
  ],
  "id": "https://mstdn.social/emojis/42450",
  "type": "Emoji",
  "name": ":blobcatnight:",
  "updated": "2020-06-13T18:33:25Z",
  "icon": {
    "type": "Image",
    "mediaType": "image/png",
    "url": "https://media.mstdn.social/custom_emojis/images/000/042/450/original/b9c906626b96a378.png"
  }
}
```

This doesn't have to be done in this PR necessarily, but we should do it before we cut a release including this.

PR #793, opened 2022-09-01
[performance] use cache for local account lookups

NyaaaWhatsUpDoc:
Removes GetLocalAccountByUsername() to instead rely on GetAccountByUsernameDomain()


tsmethurst:
Nice :)

PR #794, opened 2022-09-01
[performance] cache recently allowed/denied domains to cut down on db calls

NyaaaWhatsUpDoc:
the situation with `nil` sentinel values being handled in the cache isn't ideal (it's a bit messy imo), but i have future plans that should improve this :)


tsmethurst:
Love this PR :heart_eyes:

Issue #773, opened 2022-08-27
[bug] Toot! app fails when trying to log in with an account with no avatar/header


tsmethurst:
See screenshot:

![f77c0bb1a65059b8](https://user-images.githubusercontent.com/31960611/187032792-9a416564-26bd-44bb-b7f8-60053fa5cacf.png)

To fix this, I believe we need to update account serialization to give a `null` value instead of `""` for unset avatars + headers, to prevent clients from trying to parse the empty string as a URL. Should double check first what mastodon does in such a case, but they tend to use `null` rather than zero value fields so this would make sense.

thanks @DysphoricUnicorn

tsmethurst:
Update: looking at this now, seems like mastodon doesn't null this field, but rather it fills in a url (for example https://ondergrond.org/headers/original/missing.png) when the field is not set.

PR #799, opened 2022-09-02
[feature] Set default header and avatar for API accounts to GtS ones

tsmethurst:
This feature moves logic for default avatars out of the `web` package and into the type converter.

Now, every `apimodel.Account` that doesn't have a header or avatar URL set on it will receive a default header + avatar URL during the conversion process.

This is in-line with Mastodon API behavior, which returns default avis + headers if nothing is set.

fixes https://github.com/superseriousbusiness/gotosocial/issues/773


tsmethurst:
I checked this out and it looks beaut in pinafore, but unfortunately tusky doesn't display svgs, so we might want to consider making these default icons .pngs instead.

PR #801, opened 2022-09-03
[bugfix] Use 'Image' instead of unrecognized 'Gif' type for media attachments

tsmethurst:
We made a mistake earlier with our implementation of the media attachment api: instead of storing gifs with type `Image` we stored them with type `Gif`, which isn't part of the Mastodon API (`Gifv` is but that's for short looping soundless video files, not gifs). This meant that frontends like Tusky always showed gifs as unrecognized file types.

This PR corrects this behavior by:

1. Storing and presenting new gifs as type Image
2. Migrating already-stored Gif types to Image.


tsmethurst:
Tested this on goblin tech and it fixes the issue ^_^ Now gifs show as images normally in tusky

PR #807, opened 2022-09-07
[feature] Fetch + display custom emoji in statuses from remote instances

### LINKED ISSUE #295: [feature] Emoji grabber
Very low priority, but is it possible for GtS to pull the emoji icon from the remote server if it isn't locally stored\referenced and store it?

Possibly toggleable, as I guess some people may prefer :lotsof: :colon: :enclosed: :text:

tsmethurst:
This PR:

1. Adds `GetRemoteEmoji` function to the dereferencer, to put emojis in the db + in storage.
2. Adds `populateStatusEmojis` for statuses from remote instances.
3. Adds the UpdateStatus db function so that when we've pinned emojis to a status, we can put it in the cache.
4. Adds config entries for `media-emoji-local-max-size` and `media-emoji-remote-max-size`

Together, these functions allow fetching + display of remote emojis in statuses.

Screenshot from 2022-09-07 10-42-10

In a future PR, we should extend this remote emoji fetching to also fetch emojis in content warnings, display names, and account notes/bios, but this is already a good start :)

Closes #295


internal/api/client/admin/emojicreate.go
```diff
@@ -126,5 +130,9 @@ func validateCreateEmoji(form *model.EmojiCreateRequest) error {
 		return errors.New("no emoji given")
 	}
 
+	if form.Image.Size > emojiLocalMaxSize {
```
NyaaaWhatsUpDoc, line 133: Is it worth adding emoji size as a configurable and then using e.g. `if form.Image.Size > config.EmojiLocalMaxSize()`?

internal/media/processingemoji.go
```diff
@@ -170,6 +172,10 @@ func (p *ProcessingEmoji) store(ctx context.Context) error {
 		return fmt.Errorf("store: error executing data function: %s", err)
 	}
 
+	if fileSize > emojiRemoteMaxSize {
```
NyaaaWhatsUpDoc, line 175: Same goes for this as the local max size. What do you think?

internal/federation/dereferencing/emoji.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/media"
+)
+
+func (d *deref) GetRemoteEmoji(ctx context.Context, requestingUsername string, remoteURL string, shortcode string, id string, emojiURI string, ai *media.AdditionalEmojiInfo) (*media.ProcessingEmoji, error) {
+	t, err := d.transportController.NewTransportForUsername(ctx, requestingUsername)
+	if err != nil {
+		return nil, fmt.Errorf("GetRemoteEmoji: error creating transport: %s", err)
+	}
+
+	derefURI, err := url.Parse(remoteURL)
```
NyaaaWhatsUpDoc, line 36: I don't think you need to do a parse here. The transport already does URL parsing when creating the outgoing requests, so you should just be able to pass in `remoteURL` to the `dataFunc` :)

internal/media/processingemoji.go
tsmethurst: good idea, done in :) https://github.com/superseriousbusiness/gotosocial/pull/807/commits/c0a179ffc02d8a7d0287462d5383e160719f3433

internal/api/client/admin/emojicreate.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/807/commits/c0a179ffc02d8a7d0287462d5383e160719f3433

internal/federation/dereferencing/emoji.go
tsmethurst: Hmm, this is a bit fiddly because the `DereferenceMedia` function needs both a string iri and the parsed iri... If it's alright with you, I'll leave it for now in this PR, and we can sort it out in a separate PR, because changing the function signature in this PR means changing other parts of the code that aren't related to emojis

internal/db/bundb/status.go
```diff
@@ -175,6 +176,59 @@ func (s *statusDB) PutStatus(ctx context.Context, status *gtsmodel.Status) db.Er
...
+		}
+
+		// create links between this status and any tags it uses
+		for _, i := range status.TagIDs {
+			if _, err := tx.NewInsert().Model(&gtsmodel.StatusToTag{
+				StatusID: status.ID,
+				TagID:    i,
+			}).Exec(ctx); err != nil {
+				err = s.conn.errProc(err)
+				var alreadyExistsError *db.ErrAlreadyExists
```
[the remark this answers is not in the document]
tsmethurst: Like this? https://github.com/superseriousbusiness/gotosocial/pull/807/commits/b00c0ef5fe017418aaef8ef39942661786aa53bc

tsmethurst:
kim said, and i quote 'go nuts' so i am going nuts :)

Issue #765, opened 2022-08-24
[feature] Show link to unsupported media rather than dropping it completely


tsmethurst:
Right now we totally drop unsupported media from statuses (videos for instance) instead of processing them, so the user gets presented with a status that doesn't appear to have any media attached to it at all.

Often, though, the text of a status pertains to the media attached to it. Example: a video of a cat with the status text 'she's cute'. In GtS this would just look like a status that says 'she's cute' with no other context. This is confusing!

Instead of doing this, we should provide a link in the timelined status to the original URL. We should investigate how Mastodon does this so that we can ensure that clients using GtS will also be able to handle this.

This isn't just an interim measure until we implement video support: it'll be useful in future as well for other file types that GtS might never support.

tsmethurst:
![Screenshot from 2022-09-08 18-19-41](https://user-images.githubusercontent.com/31960611/189174339-019d5b4b-a834-4c6d-a8b7-af49b99bfda7.png)

for future reference, this is how mastodon serves unknown attachments via the api ^

Follpvosten:
I can only say that this would be really nice. I've been seeing empty posts where people just uploaded a video and that was also really confusing until now (when I tried to search the issues). Makes sense that support for some stuff just isn't implemented yet, and would be great to know about it as a user.

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/pull/2331

PR #810, opened 2022-09-08
[feature] Allow unknown media attachments

tsmethurst:
This PR updates media processing logic to no longer fail in the case of attachments that aren't supported by GtS yet.

Instead, an entry for unknown attachment types will still be created in the database and serialized to the frontend as an unknown type, but the attachment won't be stored in the database, and the URL will just redirect to the remote instance that owns the media.

This should circumvent the issue where statuses with currently unsupported attachment types appear to have no attachments at all. Instead, in clients like Pinafore and Tusky, the user will see that there are attachments, but will have to go to the remote instance to view them.

For statuses with unknown content, a note is added at the end of the status with links to the original media. So these sorts of statuses now look like this (as rendered in pinafore):

Screenshot from 2022-09-09 12-36-09

closes https://github.com/superseriousbusiness/gotosocial/issues/765
closes https://github.com/superseriousbusiness/gotosocial/issues/1013


tsmethurst:
Hmm best not merge this yet, I've seen an issue where if you have a mix of known and unknown media types, something goes wrong. Will polish it up this weekend.

NyaaaWhatsUpDoc:
How will this work out going forwards when we _do_ support these attachment types? Will this cause any issues with migration?

tsmethurst:
> How will this work out going forwards when we do support these attachment types? Will this cause any issues with migration?

I don't think it should cause issues. In future when we add more media types we can add some logic to try to lazily refetch `unknown` types to see if we can process them yet. At least, that's the approach that comes to mind :P But maybe down the road a better way will become apparent.

Also having tested this on goblin tech for a bit I'm not sure anymore about the sheer length of the note that's added to statuses :thinking: might need to refine that a bit. It looks very silly when a post is like 3 words and a gifv, and then there's this massive warning beneath it.

tsmethurst:
> I've seen an issue where if you have a mix of known and unknown media types, something goes wrong.

I checked a bit more and this looks like a client issue: Tusky and Pinafore both handle this case differently. For Tusky, it seems to ignore previews for all attachments if one is unknown. Links can still be clicked on and open OK, just no nice image previews for image attachments. For Pinafore, it just the known media types as normal, and the unknown type is shown as a blank frame with the description inside it. Both behaviors are a bit strange imo, so I think this is a case that just hasn't come up very much :P

tsmethurst:
Still to do:

- mark plurals properly
- mark media attachments as sensitive if post is sensitive

technomancy:
Would it help to have people testing this? really looking forward to this feature.

tsmethurst:
> Would it help to have people testing this? really looking forward to this feature.

Not yet, I gotta basically rebase the entire thing because it's a bit stale by now. One for after 0.6.0 i think.

PR #811, opened 2022-09-08
[frontend] colorscheme redesign, some other small css tweaks

f0x52:
Still kind of a work in progress, trying to pick a better color palette, and to document how those can be combined to maintain predictable contrast ratios.

image
image
image
image
 
Needs some more checking, especially with HTML/Markdown formatted posts I think

### DIFF
--- a/web/source/css/_colors.css
+++ b/web/source/css/_colors.css
@@ -23,57 +23,70 @@
 
 /* Color definitions */
 
-$near_white: #fafaff;
+/* Foreground */
+$white1: #fafaff; /* default text color, contrast >= 5.0 with all $grays */
+$white2: #b3b5c6; /* less important text, can be used with $gray1 (6.8), $gray2 (5.5), $gray3 (4.9), $gray4 (4.5) */
 
-$sloth_gray1: #b0b0b5;
-$sloth_gray2: #4d4e56;
+/* Background shades, contrast >= 5.0 with $white1 (#fafaff) */
+$gray1: #2a2b2f;
+$gray2: #35363b;
+$gray3: #3a3b41;
+$gray4: #45464e;
+$gray5: #4d4e56;
+$gray6: #575861;
+$gray7: #5d5e67;
+$gray8: #696a75;
 
-$sloth_orange1: #e78e5a;
-$sloth_orange2: #D87841;
-$blue: #63b1de; // complementary color to $sloth_orange1
+$orange1: #fd6a00; /* Used for non-text accent colors, can be used as background: $gray1 for text color (contrast 4.6)*/
+$orange2: #ff853e; /* hover/selected accent to $orange1, can be used with $gray1 (5.7), $gray2 (4.6) */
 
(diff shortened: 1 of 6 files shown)


tsmethurst:
Excited about this one :eyes:

tsmethurst:
I had a read through! Looks good so far. Some things that come to mind:

With the new unexpanded status layout, what happens if someone has a really long display name? does it smush into the date? And what do unexpanded statuses with show more / show less buttons look like?

For color variable names, instead of `white`, `gray`, `orange`, and `blue`, perhaps they could be foreground, background, accent, and trim, or something like that? Then it'll be less confusing if someone wants to change their foreground color to black and background to white, etc. Whaddya think?

f0x52:
> With the new unexpanded status layout, what happens if someone has a really long display name? does it smush into the date? And what do unexpanded statuses with show more / show less buttons look like?

Needs a bit of a fix still, but we can cut off the displayname with an ellipsis. Was also thinking it might be nice to have a smaller date format (with relative times), so then it takes up less space too.

> For color variable names, instead of `white`, `gray`, `orange`, and `blue`, perhaps they could be foreground, background, accent, and trim, or something like that? Then it'll be less confusing if someone wants to change their foreground color to black and background to white, etc. Whaddya think?

I'm considering adding a bit more proxy variables for that, we kind of already do, with colors, names for colors, specific locations that use those names.
so
```
$orange1: blah;
$accent: $orange1;
$avatar-border: $accent;
```

tsmethurst:
> it might be nice to have a smaller date format (with relative times), so then it takes up less space too.

I think this works better in clients than it does via a web UI, but we could definitely shorten the date format a bit in other ways (maybe just show like `10/09/22 - 13:36` or something?).

I'd say more proxy variables aren't necessarily required (though it's fine if so); imo the important thing is to just not name variables after certain colors, because it makes it more confusing when users/admins wanna write custom css in the future, if you get me

f0x52:
> I think relative time works better in clients than it does via a web UI, but we could definitely shorten the date format a bit in other ways (maybe just show like `10/09/22 - 13:36` or something?).
Yeah that'd help, maybe also shortening it to just the time if it's the same day?

> I'd say more proxy variables aren't necessarily required (though it's fine if so); imo the important thing is to just not name variables after certain colors, because it makes it more confusing when users/admins wanna write custom css in the future, if you get me
Fair, that does make it easier/more logical to change a shade everywhere in one change

tsmethurst:
> Yeah that'd help, maybe also shortening it to just the time if it's the same day?

Sure :) That's a good idea

f0x52:
This PR is kind of tangled with https://github.com/superseriousbusiness/gotosocial/pull/812, so I'd rather get that merged first then adapt the discussion mentioned here in a new PR

tsmethurst:
Alright, sounds good! 😊

Sent from ProtonMail mobile

-------- Original Message --------

PR #812, opened 2022-09-08
[frontend] Unified panels

f0x52:
Reworking the currently separate user and admin settings panels into a consistent whole. Hopefully fix some other papercuts with them along the way as well.

**very work in progress**
image


web/source/settings-panel/lib/get-views.js
```diff
@@ -0,0 +1,115 @@
...
+			let url = `${base}/${urlSafe(name)}`;
+
+			if (firstRoute == undefined) {
+				firstRoute = url;
+			}
+
+			panelRouterEl.push((
+				<Route path={`${url}/:page?`} key={url}>
+					<ErrorBoundary FallbackComponent={ErrorFallback} onReset={() => { }}>
+						{/* FIXME: implement onReset */}
```
tsmethurst, line 73: What should this do in future, just out of interest?

web/source/settings-panel/lib/get-views.js
```diff
@@ -0,0 +1,115 @@
...
+			);
+		});
+
+		panelRouterEl.push(
+			<Route key={base} path={base}>
+				<Redirect to={firstRoute} />
+			</Route>
+		);
+
+		// let childrenPath = `${base}/:section`;
```
tsmethurst, line 90: not sure what this is doing here

web/source/settings-panel/style.css
```diff
@@ -0,0 +1,493 @@
...
+			}
+
+			&:last-child {
+				margin-bottom: 0;
+			}
+		}
+	}
+
+	.sidebar {
+		align-self: start; // vertical
```
tsmethurst, line 71: are `vertical` and `horizontal` meant to be comments, or is this some syntax i didn't know about?

web/source/settings-panel/style.css
```diff
@@ -0,0 +1,493 @@
...
+}
+
+.error {
+	color: $error-fg;
+	background: $error-bg;
+	border: 0.02rem solid $error-fg;
+	border-radius: $br;
+	font-weight: bold;
+	padding: 0.5rem;
+	white-space: pre-wrap; // to show \n in json errors
```
tsmethurst, line 165: same comment as above

web/source/settings-panel/style.css
```diff
@@ -0,0 +1,493 @@
...
+		flex-direction: column;
+	}
+	
+	.file-input.button {
+		display: inline-block;
+		font-size: 1rem;
+		font-weight: normal;
+		padding: 0.3rem 0.3rem;
+		align-self: flex-start;
+		/* background: $border_accent; */
```
tsmethurst, line 265: delete unused

web/source/settings-panel/style.css
```diff
@@ -0,0 +1,493 @@
...
+			margin-right: 0;
+			display: flex;
+			flex-direction: column;
+			justify-content: center;
+
+			div.form-field {
+				width: 100%;
+				display: flex;
+
+				span {
```
tsmethurst, line 337: I've noticed that when running in mobile, the overflow on the file name isn't hidden, and ends up extending the page to the right, even though text-overflow etc aren't changed for the small view. This feels like a bug we had before, which I fixed *somehow* but I can't remember which bug or how I fixed it...

web/source/settings-panel/style.css
```diff
@@ -0,0 +1,493 @@
...
+					flex: 1 1 auto;
+					overflow: hidden;
+					text-overflow: ellipsis;
+					white-space: nowrap;
+					padding: 0.3rem 0;
```
tsmethurst, line 342: this feels a bit hacky; is there a way to align the text nicely without putting in this tiny pad?

tsmethurst (commented):
Overall this looks great and it's a big step up from what we had before :)

Just some comments about style, and some comments about commented-out code that perhaps should be deleted -- I may have been too zealous here though, because I saw you also put code in that we can just uncomment out later when more features are implemented, so take my recommendations here with a pinch of salt!

Love the experience of the emoji upload page, that's absolutely brilliant <3

Oh yeah, also please check through for `console.log` calls and make sure you're not logging things unnecessarily -- when running this locally i see quite a few logs in the console that you probably put in there for debugging

web/source/settings-panel/style.css
f0x52: they're comments, but parsed and stripped out by postcss. kinda redundant here though, will delete

web/source/settings-panel/style.css
f0x52: it's not that hacky imo, it's the same amount of padding as the .button before it has to extend the blue background
`line-height: 2.1rem` (1.5 + 0.3 + 0.3) would also work but is (even) hackier i'd say

tsmethurst: alright :)

PR #828, opened 2022-09-13
[feature] update config types to use bytesize.Size

NyaaaWhatsUpDoc:
Allows passing `40MiB`, `40MB`, `1kiB` etc instead of literal sizes in bytes. Also updates some of the file sizes returned in the codebase to be `int64` and not `int`.


NyaaaWhatsUpDoc:
Okay turns out mapstructure is angery and doesn't handle the decoding properly. Grrr.

NyaaaWhatsUpDoc:
@tsmethurst I am so confused as to why the pipeline is failing. running `test/envparsing.sh` locally passes for me. would you be able to checkout this branch and see if it works locally for you?

tsmethurst:
> I am so confused as to why the pipeline is failing. running test/envparsing.sh locally passes for me

You're running go 1.19 now right? This PR is still building/testing/etc on 1.18

NyaaaWhatsUpDoc:
> > I am so confused as to why the pipeline is failing. running test/envparsing.sh locally passes for me
> 
> You're running go 1.19 now right? This PR is still building/testing/etc on 1.18

This is true! Though I think I just managed to catch and fix the error now, related to a small change I made in envparsing :)

NyaaaWhatsUpDoc:
^ this is what I get for attempting to edit and fix things from my phone using the GitHub text editor lmao

NyaaaWhatsUpDoc:
@tsmethurst how do we feel about merging this one? :)

tsmethurst:
I feel good about it, BUT we should probably fix the broken snapshot build first, looks like swagger is not in the path for some reason? https://drone.superseriousbusiness.org/superseriousbusiness/gotosocial/2834/1/2

NyaaaWhatsUpDoc:
> I feel good about it, BUT we should probably fix the broken snapshot build first, looks like swagger is not in the path for some reason? https://drone.superseriousbusiness.org/superseriousbusiness/gotosocial/2834/1/2

how on earth :thinking: i'll take a look into this tomorrow

tsmethurst:
One thing I forgot to ask earlier: will this require users to update their configs? or is it backwards compatible?

PR #843, opened 2022-09-19
Add --config-path to example CLI commands where needed.

technomancy:
Previously we had a few examples referring to --config-file (which is not accepted) but most were missing it altogether, so the commands listed did not work.

Also replaced "./example.json" with just "example.json" in the import/export examples because the "./" was unnecessary.


tsmethurst:
Thank you! <3

PR #847, opened 2022-09-22
[feature] Allow delivery to sharedInboxes where possible

tsmethurst:
This PR adds logic for dereferencing, storing, and delivering to shared inboxes: https://www.w3.org/TR/activitypub/#shared-inbox-delivery

To do this, it adds a migration to add the field `shared_inbox_uri` to the gtsmodel.Account. This field is a string pointer, with three possible settings:

- null -- we don't know yet whether this account has a SharedInboxURI, we need to look it up
- a url string -- the discovered SharedInboxURI of the account
- an empty string -- we looked it up but the account doesn't have a shared inbox

Logic for looking up the shared inbox of an account has been added to the `GetRemoteAccount` function in the dereferencer. When that function is called, and we have a null value for SharedInboxURI, the account will be dereferenced (if it hasn't already) to discover the shared inbox address.

This means that shared inboxes will gradually be discovered by your instance as you use it, rather than a bunch of calls happening all at once after updating, potentially breaking things.

If admins want to preserve the old delivery behavior, they can set the new config option `instance-deliver-to-shared-inboxes` to false :) Then, shared inboxes will still be discovered, they just won't be used. This defaults to `true`.


internal/federation/dereferencing/account.go
```diff
@@ -279,6 +279,30 @@ func (d *deref) GetRemoteAccount(ctx context.Context, params GetRemoteAccountPar
 		}
 	}
 
+	// check if we need to look for a sharedInbox for the account
+	var sharedInboxChanged bool
+	if foundAccount.SharedInboxURI == nil {
+		// we need the accountable for this, so get it if we don't have it yet
+		if accountable == nil {
+			accountable, err = d.dereferenceAccountable(ctx, params.RequestingUsername, params.RemoteAccountID)
```
blackle, line 287: if you call this through searchAccountByMention, then params.RemoteAccountID is nil, and dereferenceAccountable and other functions in its call stack don't check for that

was getting a delayed segfault due to this when searching for a remote account. startHandshake would add the nil id to the handshakes map, then stopHandshake would fail later while iterating over it

blackle: as a workaround I changed the earlier check to webfinger to:

```
params.RemoteAccountID == nil || foundAccount == nil || (foundAccount.LastWebfingeredAt.Before(time.Now().Add(webfingerInterval)) && !instanceAccount(foundAccount))
```

this way I can also force the instance to re-fetch an account by searching for it, which is nice

tsmethurst: aha thanks blackle, I'll fix it up in the PR

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/847/commits/21b76f7192f8aa881c0b0f4be81708ea1aa78848 !

internal/federation/federatingdb/inbox.go
```diff
@@ -108,7 +108,15 @@ func (f *federatingDB) InboxesForIRI(c context.Context, iri *url.URL) (inboxIRIs
 				follow.Account = followingAccount
 			}
 
-			inboxIRI, err := url.Parse(follow.Account.InboxURI)
+			// deliver to a shared inbox if we have that option
+			var inbox string
+			if config.GetInstanceDeliverToSharedInboxes() && follow.Account.SharedInboxURI != nil && *follow.Account.SharedInboxURI != "" {
```
NyaaaWhatsUpDoc, line 113: Do we need to do the extra `!= ""` check at the end? Considering we don't do the same for `.InboxURI` and they both get passed down to `url.Parse()` ?

NyaaaWhatsUpDoc: (same goes for below)

internal/typeutils/internaltoas.go
```diff
@@ -85,6 +85,21 @@ func (c *converter) AccountToAS(ctx context.Context, a *gtsmodel.Account) (vocab
 	inboxProp.SetIRI(inboxURI)
 	person.SetActivityStreamsInbox(inboxProp)
 
+	// shared inbox
+	if a.SharedInboxURI != nil && *a.SharedInboxURI != "" {
```
NyaaaWhatsUpDoc, line 89: Same again with this check. I think it would be better to cach the error caused by a non-nil empty URI when passing to `url.Parse()` than to allow these cases to go unchecked

internal/typeutils/astointernal.go
```diff
@@ -156,6 +157,18 @@ func (c *converter) ASRepresentationToAccount(ctx context.Context, accountable a
 		acct.InboxURI = accountable.GetActivityStreamsInbox().GetIRI().String()
 	}
 
+	// SharedInboxURI
+	if sharedInboxURI := ap.ExtractSharedInbox(accountable); sharedInboxURI != nil {
+		sharedInboxHost := sharedInboxURI.Host
+
+		var sharedInbox string
+		if dns.IsSubDomain(acct.Domain, sharedInboxHost) || dns.IsSubDomain(sharedInboxHost, acct.Domain) {
```
NyaaaWhatsUpDoc, line 165: Would be simpler here to do `dns.CompareDomainName(acct.Domain, sharedInboxHost) > 2` (or maybe it would be `>=` :thinking:)

NyaaaWhatsUpDoc:
Have made a few comments, but otherwise looks good :)

internal/federation/federatingdb/inbox.go
tsmethurst: Yeah so basically I don't want an empty shared inbox URI to be a fail necessarily. For SharedInboxURI I figured if we use a string pointer, it's either set with an actual url string, set with an empty string, or null, and these correspond to 'we have a shared inbox for this account', 'we definitely don't have a shared inbox for an account,' and 'we don't know / haven't checked yet'. So that's basically why I added these empty checks here. This is different from InboxURI, which should always be set and not be an empty string (if it is an empty string, that means something has gone wrong).

I am aware while explaining it that this doesn't necessarily make a lot of sense though, perhaps there's a better way :thinking:

internal/typeutils/astointernal.go
tsmethurst: oh cool, I didn't know that was a thing! will check it out :)

internal/federation/federatingdb/inbox.go
NyaaaWhatsUpDoc: I think if this is the case then you should add a long comment explanation under the account model (unless I happened to miss this in which case I am sorry :') )

internal/typeutils/internaltoas.go
NyaaaWhatsUpDoc: Not necessary as discussed above

internal/federation/federatingdb/inbox.go
tsmethurst: I did add a comment but I will make it more explicit! :) https://github.com/superseriousbusiness/gotosocial/pull/847/files#diff-fc87d545d1f83b6155c6e9f11a3dcd21ace2a99ffe6e38dcc485815a15b22d50

tsmethurst: Bam! https://github.com/superseriousbusiness/gotosocial/pull/847/commits/a82564d433f35f52935f3fb8abc0d727eb57e2d0

internal/typeutils/astointernal.go
tsmethurst: I checked and it's at least two! Changed here: https://github.com/superseriousbusiness/gotosocial/pull/847/commits/f00bb5a2a4cb9cdca623e40ee09bca43b2bd5331

PR #851, opened 2022-09-24
[bugfix] update thread iterators to not use recursion

NyaaaWhatsUpDoc:
Rewrote both the `iterateAncestors` and `iterateDescendants` functions to both use loops instead of recursion, and rewrote both including their logging to hopefully provide more useful output.

This also updates the `DereferenceThread()` function signature to accept the originating status as arguments, instead of making a second call to `GetRemoteStatus()` within the function, and no longer returns an error as we don't want statuses further up/down the thread to cause us to error-out for this particular status.


NyaaaWhatsUpDoc:
I'll figure out the tests later. I still have some code commenting to finish up and cleanup before this is ready also.

internal/federation/dereferencing/thread.go
```diff
@@ -38,204 +44,262 @@ import (
...
+			}
+
+			// Set the fetched status
+			status = localStatus
+
+		} else {
+			l.Tracef("following remote status ancestors: %s", status.InReplyToURI)
+
+			// Fetch the remote status found at this IRI
+			remoteStatus, _, err := d.GetRemoteStatus(ctx, username, replyIRI, true, false)
```
tsmethurst, line 123: I think we can probs set refetch to false here, since we're not doing anything with the returned statusable AP representation.

internal/federation/dereferencing/thread.go
```diff
@@ -38,204 +44,262 @@ import (
...
-		l.Trace("iri belongs to us, bailing")
-		return nil
+	// Log function start
+	l.Trace("beginning")
+
+	// frame represents a single stack frame when iteratively
+	// dereferencing status descendants. where iri is the IRI
+	// of the status in question, status is the raw AP object,
+	// and page is the current child iterator.
+	type frame struct {
```
tsmethurst, line 153: Is there a clever compiler reason for declaring types inside functions? Or is it just for scoping since this particular struct isn't needed elsewhere? Not a critique or anything, legit just a question because I don't know and you probably do :D

internal/federation/dereferencing/thread.go
```diff
@@ -38,204 +44,262 @@ import (
...
-	for {
-		l.Tracef("dereferencing page %s", currentPageIRI)
-		collectionPage, err := d.DereferenceCollectionPage(ctx, username, currentPageIRI)
-		if err != nil {
-			l.Debugf("couldn't get remote collection page %s: %s; breaking pageLoop", currentPageIRI, err)
-			break pageLoop
+stackLoop:
+	for i := 0; i < maxIter; i++ {
+		// Pop next frame, nil means we are at end
+		if current = popStack(); current == nil {
```
tsmethurst, line 192: won't popStack always return nil at for the first iteration, because there's nothing in `stack` yet? :thinking: Or am I missing something? As far as I can tell, you haven't done anything with `current` yet

internal/federation/dereferencing/thread.go
```diff
@@ -38,204 +44,262 @@ import (
...
-				itemURI = iter.GetIRI()
-			case iter.IsActivityStreamsNote():
-				// note, get the id from it to use as iri
-				note := iter.GetActivityStreamsNote()
-				noteID := note.GetJSONLDId()
-				if noteID != nil && noteID.IsIRI() {
-					itemURI = noteID.GetIRI()
+		for /* page loop */ {
+			if current.itemIter == nil {
+				// Get the collection page "next" property
```
tsmethurst, line 225: One thing I've realised recently is that often the 'first' page *also* has items on it, example:

```
  "replies": {
    "id": "https://mastodon.social/users/trwnh/statuses/109051391335194698/replies",
    "type": "Collection",
    "first": {
      "type": "CollectionPage",
      "next": "https://mastodon.social/users/trwnh/statuses/109051391335194698/replies?min_id=109051976439984742&page=true",
      "partOf": "https://mastodon.social/users/trwnh/statuses/109051391335194698/replies",
      "items": [
        "https://mastodon.social/users/trwnh/statuses/109051976439984742"
      ]
    }
  }
```

So we should also check for items on the `first` page as well, and not just on `next` ones.

tsmethurst (commented):
Looks beaut so far. Few little comments, nothing major. Lmk if you want me to fiddle with the tests or anything :)

tsmethurst:
Just to make life a little easier, I extracted the reason for the test fail from the massive log file:

```
--- FAIL: TestFromFederatorTestSuite (13.31s)
    --- FAIL: TestFromFederatorTestSuite/TestProcessFederationAnnounce (0.89s)
        fromfederator_test.go:65: 
            	Error Trace:	/drone/src/internal/processing/fromfederator_test.go:65
            	Error:      	Received unexpected error:
            	            	error dereferencing announce from federator: DereferenceAnnounce: error dereferencing thread of boosted status: DereferenceThread: error getting initial status with id http://localhost:8080/users/the_mighty_zork/statuses/01F8MHAMCHF6Y650WCRSCP4WMY: GetRemoteStatus: error dereferencing statusable: DereferenceStatusable: error deferencing http://localhost:8080/users/the_mighty_zork/statuses/01F8MHAMCHF6Y650WCRSCP4WMY: GET request to http://localhost:8080/users/the_mighty_zork/statuses/01F8MHAMCHF6Y650WCRSCP4WMY failed (404): 
```

internal/federation/dereferencing/thread.go
NyaaaWhatsUpDoc: Just because it's not needed elsewhere, stops us cluttering up the rest of the package's private namespace :)

internal/federation/dereferencing/thread.go
NyaaaWhatsUpDoc: You are correct! I meant to move the first frame to be set in the stack instead of current :')

internal/federation/dereferencing/thread.go
NyaaaWhatsUpDoc: Good idea!

tsmethurst:
great stuff, thanks kimb :)

PR #863, opened 2022-09-28
[chore] simplify generating log entry caller information

NyaaaWhatsUpDoc:
Turns out that tweaking how we fetched the calling function info and other bits was faster than caching it! Much simpler code too :)


NyaaaWhatsUpDoc:
Failing test here is unrelated to this PR. It was a result of some changes in the version bump PR causing `main` to start failing. The test was incorrectly calling `processor.Start()` twice which caused a panic

PR #873, opened 2022-10-01
[chore] Normalize domain blocks to punycode

tsmethurst:
This PR makes it so that domain blocks are always normalized to punycode no matter how you create or check the domain block.

This should make life a bit easier for admins, since now they don't need to worry about whether to block the punycode domain or the unicode domain, both will work.

### DIFF
--- a/internal/db/bundb/domain.go
+++ b/internal/db/bundb/domain.go
@@ -28,6 +28,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/config"
 	"github.com/superseriousbusiness/gotosocial/internal/db"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"golang.org/x/net/idna"
 )
 
 type domainDB struct {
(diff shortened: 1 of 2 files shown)


internal/db/bundb/domain.go
```diff
@@ -28,22 +28,35 @@ import (
...
 type domainDB struct {
 	conn  *DBConn
 	cache *cache.DomainBlockCache
 }
 
+// normalize converts the given domain to lowercase
+// then to punycode (for international domain names).
+// Returns the resulting domain or an error if the
+// punycode conversion fails.
+func (d *domainDB) normalize(domain string) (out string, err error) {
```
NyaaaWhatsUpDoc, line 43: my only comment here is we don't need to make it a receiver method, could just have it as a separate function `normalizeDomain()` :)

tsmethurst: ah sure, will do that rn

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/873/commits/8cf09efb8eef17cfbb209cdf7fae5003675dbc38

PR #875, opened 2022-10-02
[chore] Use shorter timestamps in frontend for replies

tsmethurst:
This PR updates some of the timestamp string generation logic in web templates. Now, replies use timestamp strings that are shorter, to avoid obscuring the display name + username as much. These timestamps become more vague the longer ago the status was created. A status created today gets `Today, 15:04`; a status created this year gets `Jan 02, 15:04`, and a status created longer ago than this year gets `Jan 02, 2006`.

examples:

Screenshot from 2022-10-02 11-10-33

Screenshot from 2022-10-02 11-03-55

Screenshot from 2022-10-02 11-03-38

### DIFF
--- a/internal/router/template.go
+++ b/internal/router/template.go
@@ -36,8 +36,10 @@ import (
 
 const (
 	justTime     = "15:04"
-	dateAndTime  = "Jan 02, 15:04"
-	yearDateTime = "Jan 02, 2006, 15:04"
+	dateYear     = "Jan 02, 2006"
+	dateTime     = "Jan 02, 15:04"
+	dateYearTime = "Jan 02, 2006, 15:04"
+	monthYear    = "Jan, 2006"
 )
 
 // LoadTemplates loads html templates for use by the given engine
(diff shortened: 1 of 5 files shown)


tsmethurst:
Oh yeah, and also just some small CSS tweaks I snuck in, but I can make these in a separate PR instead if desired.

internal/router/template.go
```diff
@@ -76,13 +85,31 @@ func noescapeAttr(str string) template.HTMLAttr {
 }
 
 func timestamp(stamp string) string {
-	t, _ := time.Parse(time.RFC3339, stamp)
-	return t.Format("January 2, 2006, 15:04:05")
+	t, _ := util.ParseISO8601(stamp)
```
NyaaaWhatsUpDoc, line 88: Is it worth catching the parser error here? and in the below functions. to just return something like `bad timestamp` instead of defaulting to unix epoch. Which could also end up happening due to other unrelated errors passing in data to the template

tsmethurst: yeah sure :) I'll update it

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/875/commits/bbd6f993e0b34545115f552f5baadec28c77aeaa

NyaaaWhatsUpDoc:
Backend stuff looks good to me :)

PR #884, opened 2022-10-03
[performance] cache domains after max retries in transport

NyaaaWhatsUpDoc:
Max retries are only reached if on HTTP server, DNS resolution or connection type errors. By caching these hostname failures we don't clog up the outoging request queue with likely failing requests.

### DIFF
--- a/internal/transport/controller.go
+++ b/internal/transport/controller.go
@@ -51,7 +51,8 @@ type controller struct {
 	fedDB     federatingdb.DB
 	clock     pub.Clock
 	client    pub.HttpClient
-	cache     cache.Cache[string, *transport]
+	trspCache cache.Cache[string, *transport]
+	badHosts  cache.Cache[string, struct{}]
 	userAgent string
 }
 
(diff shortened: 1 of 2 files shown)


internal/transport/controller.go
```diff
@@ -102,13 +110,13 @@ func (c *controller) NewTransport(pubKeyID string, privkey *rsa.PrivateKey) (Tra
 	}
 
 	// Cache this transport under pubkey
-	if !c.cache.Put(pubStr, transp) {
+	if !c.trspCache.Put(pubStr, transp) {
```
tsmethurst, line 113: Can you remind me under what conditions Put would return false?

internal/transport/transport.go
```diff
@@ -88,7 +88,23 @@ func (t *transport) POST(r *http.Request, body []byte, retryOn ...int) (*http.Re
...
+		host = r.URL.Hostname()
+	)
+
+	// Check if recently reached max retries for this host
+	// so we don't need to bother reattempting it. The only
+	// errors that are retried upon are server failure and
+	// domain resolution type errors, so this cached result
+	// indicates this server is likely having issues.
+	if t.controller.badHosts.Has(host) {
+		return nil, errors.New("too many failed attempts")
```
tsmethurst, line 106: is it worth making an error type for this so we can check for it? or nah?

internal/transport/transport.go
```diff
@@ -88,7 +88,23 @@ func (t *transport) POST(r *http.Request, body []byte, retryOn ...int) (*http.Re
...
+		// Get request hostname
+		host = r.URL.Hostname()
+	)
+
+	// Check if recently reached max retries for this host
+	// so we don't need to bother reattempting it. The only
+	// errors that are retried upon are server failure and
+	// domain resolution type errors, so this cached result
+	// indicates this server is likely having issues.
+	if t.controller.badHosts.Has(host) {
```
tsmethurst, line 105: The `Has` call doesn't reset the ttl i assume?

PR #885, opened 2022-10-03
[frontend] Scroll to highlighted toot, improve highlight

f0x52:
Scrolls to highlighted toot on page load, as long as it's not the first one (because then it's not really needed). Also makes unfocused toots one shade darker


tsmethurst (approved):
lgtm!

PR #908, opened 2022-10-12
[frontend] Use new GET custom_emoji admin api

f0x52:
Fetches custom emoji from `/api/v1/admin/custom_emojis` (https://github.com/superseriousbusiness/gotosocial/pull/902), also enables emoji detail pages although they don't provide any functionality yet (missing api)
image

### DIFF
--- a/web/source/settings/admin/emoji.js
+++ b/web/source/settings/admin/emoji.js
@@ -179,10 +179,10 @@ function EmojiCategory({category, entries}) {
 			<div className="emoji-group">
 				{entries.map((e) => {
 					return (
-						<Link key={e.static_url} to={`${base}/${e.id}`}>
+						<Link key={e.id} to={`${base}/${e.id}`}>
 							{/* <Link key={e.static_url} to={`${base}`}> */}
 							<a>
-								<img src={e.static_url} alt={e.shortcode} title={`:${e.shortcode}:`}/>
+								<img src={e.url} alt={e.shortcode} title={`:${e.shortcode}:`}/>
 							</a>
 						</Link>
 					);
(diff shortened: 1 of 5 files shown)


web/source/settings/lib/api/admin.js
```diff
@@ -164,7 +164,7 @@ module.exports = function ({ apiCall, getChanges }) {
 		fetchCustomEmoji: function fetchCustomEmoji() {
 			return function (dispatch, _getState) {
 				return Promise.try(() => {
-					return dispatch(apiCall("GET", "/api/v1/custom_emojis"));
+					return dispatch(apiCall("GET", "/api/v1/admin/custom_emojis?filter=domain:local"));
```
tsmethurst, line 167: This will return only the default `limit` of 50 emojis. If you wanna show all local emojis set limit to 0 or less :)

OR you can already implement pagination logic using the Link header returned by requests, if you wanna be fancy.

web/source/settings/admin/emoji.js
```diff
@@ -195,18 +198,29 @@ function EmojiDetailWrapped() {
...
 }
 
-function EmojiDetail({id, Form}) {
+function EmojiDetail({emoji, Form}) {
 	return (
-		"Not implemented yet"
+		<div>
+			<h1><BackButton to={base}/> Custom Emoji: {emoji.shortcode}</h1>
+			<p>Editing custom emoji isn&apos;t implemented yet</p>
+			<img src={emoji.static_url} alt={emoji.shortcode} title={`:${emoji.shortcode}:`}/>
```
tsmethurst, line 223: for the full size view you can use the url rather than the static url imo :)

and also probably render it a little bit smaller, cuz right now it's absolutely *massive* :')

PR #922, opened 2022-10-25
Allow user to show instead of landing page on /

i5heu:


internal/web/profile.go
```diff
@@ -50,7 +50,15 @@ func (m *Module) profileGETHandler(c *gin.Context) {
 	}
 
 	// usernames on our instance will always be lowercase
-	username := strings.ToLower(c.Param(usernameKey))
+	var username = ""
+	config.Config(func(cfg *config.Configuration) {
```
tsmethurst, line 54: I think you can do this a bit more simply:

```go
var username string
if landingPageUser := config.GetLandingPageUser(); landingPageUser != "" {
  username = landingPageUser
} else {
  username = strings.ToLower(c.Param(usernameKey))
}
```

internal/web/web.go
```diff
@@ -94,7 +96,15 @@ func (m *Module) Route(s router.Router) error {
 	})
 
 	// serve front-page
-	s.AttachHandler(http.MethodGet, "/", m.baseHandler)
+	var username = ""
+	config.Config(func(cfg *config.Configuration) {
```
tsmethurst, line 100: you can simplify this one too, using something similar to what i wrote above :)

tsmethurst (changes requested):
Cool PR :)

You'll have to fiddle with `test/cliparsing.sh` and `test/envparsing.sh` to get it passing tests. Basically you just gotta add the new parameters to the expected output there. Let me know if you need any help with it :)

internal/web/web.go
```diff
@@ -94,7 +95,11 @@ func (m *Module) Route(s router.Router) error {
 	})
 
 	// serve front-page
-	s.AttachHandler(http.MethodGet, "/", m.baseHandler)
+	if landingPageUser := config.GetLandingPageUser(); landingPageUser == "" {
```
tsmethurst, line 98: Rather than serving a whole different handler here if `config.GetLandingPageUser()` is set, I would rather we do a 302 redirect to the set profile from within the baseHandler. Could you make that change? Then I think this will be good to merge :)

tsmethurst:
Heya, sorry for responding slowly again. I'm only just getting back to GtS properly now after time away and then being ill. I added just one more review comment :) After that this should be good to go

internal/web/web.go
i5heu: Ya this is a better way to do that :3

internal/web/profile.go
```diff
@@ -50,7 +55,13 @@ func (m *Module) profileGETHandler(c *gin.Context) {
 	}
 
 	// usernames on our instance will always be lowercase
-	username := strings.ToLower(c.Param(usernameKey))
+	var username string
+	if landingPageUser := config.GetLandingPageUser(); landingPageUser != "" {
```
NyaaaWhatsUpDoc, line 59: Will this code not always return the landing page user for any profile GET request? surely we can just leave this to the logic it was previously since we now have the redirect in place now? :thinking:

i5heu: true!
i change this in a minute ^^

internal/web/web.go
```diff
@@ -94,7 +95,11 @@ func (m *Module) Route(s router.Router) error {
...
+		s.AttachHandler(http.MethodGet, "/", m.baseHandler)
+	} else {
```
tsmethurst, line 100: rather than doing this if/else here, i would just serve m.baseHandler as before, but put the redirect logic right at the top of that function instead, and return early with the redirect if landingPageUser is set

this means that in the future if we have some way of changing landingPageUser while the instance is still running (hotloading config) then the instance won't need to be restarted to make it work :)

internal/web/profile.go
```diff
@@ -40,6 +40,11 @@ const (
 	MaxStatusIDKey = "max_id"
 )
 
+func (m *Module) profileGETRedirect(c *gin.Context) {
+	landingPageUser := config.GetLandingPageUser()
+	c.Redirect(http.StatusMovedPermanently, "/@"+c.Param(strings.ToLower(landingPageUser)))
```
tsmethurst, line 45: I would make this a 302 rather than 301 -- we don't want browsers to cache the redirect, in case the setting changes at some point

tsmethurst:
Looks good now! Thank you for this, and for making the changes :)

PR #925, opened 2022-10-25
Fix #923 and markdown css

### LINKED ISSUE #923: [bug] username is copied with break

Bildschirmfoto 2022-10-25 um 10 56 35

if I copy a username with the instance domain, a break will appear between the `span` content.
For continuance reasons this should be copied as one line without space.

Directly copied:
```
@mia
@gts.miau.pub
```

0.5.2 git-c31f219

i5heu:
Fixing #923 and this issue
Bildschirmfoto 2022-10-25 um 12 26 13

I am not entirely sure if the solution for centered imgs (alt needs to contain `!center`) is a good one since it will be read by screen readers.
Although maybe this is even a good solution to indicate that nothing on the same height of the img exists.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -146,7 +146,7 @@ yarn run dev
 
 The Pinafore instance will start running on `localhost:4002`.
 
-To connect to the testrig, navigate to `https://localhost:4002` and enter your instance name as `localhost:8080`.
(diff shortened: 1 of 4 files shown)


tsmethurst:
Quick question: emojis are also `` tags right? Will this affect how emojis are displayed within posts?

f0x52 (approved):

Issue #797, opened 2022-09-02
[feature] Allow admins to upload / delete custom emojis via admin panel


tsmethurst:
The admin section of the settings panel needs a subsection for emojis.

Here, admins should be able to:

- [x] View custom emojis present on the instance
- [x] (`GET /api/v1/admin/custom_emojis[?filter=local]`, returns an array of emojis like from `/api/v1/custom_emojis` but including domain, date uploaded, last updated, id, etc).
- [x] Upload a new custom emoji, with a shortcode and category (`POST /api/v1/admin/custom_emojis`) -- ideally, admins can drag and drop an emoji file into a box and see how it'll actually look when rendered on the instance, but this is probably FIDDLY so basic implementation of selecting a file + uploading it is fine to start with.
- [ ] Delete a custom emoji, either local or remote (`DELETE /api/v1/admin/custom_emojis/{id}`)
- [ ] Use `PUT` or `PATCH`? to `/api/v1/admin/custom_emojis/{id}`:
  - [ ] Modify a local custom emoji -- set a new image for it, put it in a different category, etc
  - [ ] Disable/suspend a remote custom emoji
  - [ ] Copy a remote emoji to local

tsmethurst:
Gonna work on this one now: https://github.com/superseriousbusiness/gotosocial/issues/821

And then work on the PATCH endpoint mentioned above :)

xplosionmind:
When visiting `https://gts.superseriousbusiness.org/settings/admin/custom-emoji` I get this.

```
An error occured, please report this on the [GoToSocial issue tracker](https://github.com/superseriousbusiness/gotosocial/issues) or [Matrix support room](https://matrix.to/#/#gotosocial-help:superseriousbusiness.org).
Include the details below:

TypeError: t is undefined

[112]</module.exports/r<@https://gts.superseriousbusiness.org/assets/dist/settings.js:2831:1567
useMemo@https://gts.superseriousbusiness.org/assets/dist/settings.js:2268:70076
[75]</exports.useMemo@https://gts.superseriousbusiness.org/assets/dist/settings.js:1785:6231
[112]</module.exports@https://gts.superseriousbusiness.org/assets/dist/settings.js:2831:1548
Xh@https://gts.superseriousbusiness.org/assets/dist/settings.js:2268:63826
Wk@https://gts.superseriousbusiness.org/assets/dist/settings.js:2268:120234
[...]

PR #939, opened 2022-11-01
[feature] Allow newly uploaded emojis to be placed in categories

tsmethurst:
- Added database model for emoji categories.
- Updated select for emojis to also select emoji category.
- Serialize emoji category properly to the frontend.
- Added endpoint `GET /api/v1/admin/custom_emojis/categories` to list all existing emoji categories.
- Added `category` to emoji upload. Emoji will be put in given category. If given `category` doesn't yet exist, it will be created.
- Limit category name to 64 characters (arbitrary number, should be long enough I think)

Part of https://github.com/superseriousbusiness/gotosocial/issues/797

Work still to do in another PR: allow deleting categories, allow moving emojis between categories.


docs/api/swagger.yaml
```diff
@@ -2745,6 +2764,10 @@ paths:
                   name: image
                   required: true
                   type: file
+                - description: Category in which to place the new emoji. 64 characters or less. If left blank, emoji will be uncategorized, but this can be updated later.
```
f0x52, line 2767: minor change here, but as we don't currently have the api routes to change them later, maybe note that?

f0x52 (changes requested):

docs/api/swagger.yaml
tsmethurst: Done! :) https://github.com/superseriousbusiness/gotosocial/pull/939/commits/eb72766e6f8666fe99ba670f514ee06b48d0ae52

NyaaaWhatsUpDoc:
Looks good to me :)

PR #940, opened 2022-11-01
[bugfix] create admin_account_actions table in tx

terinjokes:
The migration that adds the `admin_account_actions` table did so at the same time as adding indexes onto the new table. This code was ran inside a `RunInTx` function, but the table creation did not use the transaction reference, while the creation of the indexes did. This could cause a race between the table and index creations, depending on the scheduling order. If the table creation did not win the race, then the migration would fail.

This changeset corrects the table creation to also be done inside the same transaction as the index creation.

Signed-off-by: Terin Stock <terinjokes@gmail.com>


tsmethurst:
nice, thanks!

PR #941, opened 2022-11-01
[bugfix] Wrap media in read closer

tsmethurst:
- User ReadClosers instead of Readers where possible.
- Rejig close deferrals in some places.

### DIFF
--- a/internal/media/processingmedia.go
+++ b/internal/media/processingmedia.go
@@ -263,14 +263,18 @@ func (p *ProcessingMedia) store(ctx context.Context) error {
 		return nil
 	}
 
-	// execute the data function to get the reader out of it
-	reader, fileSize, err := p.data(ctx)
+	// execute the data function to get the readcloser out of it
+	rc, fileSize, err := p.data(ctx)
 	if err != nil {
 		return fmt.Errorf("store: error executing data function: %s", err)
 	}
 
 	// defer closing the reader when we're done with it
 	defer func() {
+		if err := rc.Close(); err != nil {
+			log.Errorf("store: error closing readcloser: %s", err)
+		}
+	}()
 
 	// execute the postData function no matter what happens
 	defer func() {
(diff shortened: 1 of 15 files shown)


internal/media/processingmedia.go
```diff
@@ -303,7 +310,7 @@ func (p *ProcessingMedia) store(ctx context.Context) error {
 	extension := split[1] // something like 'jpeg'
 
 	// concatenate the cleaned up first bytes with the existing bytes still in the reader (thanks Mara)
-	readerToStore := io.MultiReader(bytes.NewBuffer(firstBytes), reader)
+	readerToStore := io.MultiReader(bytes.NewBuffer(firstBytes), rc)
```
NyaaaWhatsUpDoc, line 313: Commenting here because it won't let me on a lower line. But on line 330 you don't need to attempt to close `readerToStore` since `MultiReader` doesn't handle closers (also we're closing the readcloser in the defer above)

NyaaaWhatsUpDoc: Oh wait ignore me I see it's "post" as in after, not as in `GET` `POST`.

internal/media/processingmedia.go
```diff
@@ -263,24 +263,31 @@ func (p *ProcessingMedia) store(ctx context.Context) error {
...
 	defer func() {
-		if rc, ok := reader.(io.ReadCloser); ok {
-			if err := rc.Close(); err != nil {
-				log.Errorf("store: error closing readcloser: %s", err)
+		if err := rc.Close(); err != nil {
+			log.Errorf("store: error closing readcloser: %s", err)
+		}
+	}()
+
+	// execute the postData function no matter what happens
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: Oh wait ignore me I see it's "post" as in after, not as in `GET` `POST`.

NyaaaWhatsUpDoc:
A singular comment, though relating to code not a part of this PR technically (though related). Otherwise ready to go :)

internal/media/processingmedia.go
tsmethurst: Aha! Will change it :) Two ticks..

tsmethurst: Ah hold on, that close is done on the thing returned from Terminator, which is one end of a pipe reader/writer. And I think pipereaders do implement Close(), so we should still try to close it.  I'll tidy it up to make this a bit clearer cuz it's vague rn

tsmethurst: There :) https://github.com/superseriousbusiness/gotosocial/pull/941/commits/fcc22d1ca034b7183d000b982ea2b88eb25a16e8

PR #942, opened 2022-11-01
[frontend] change bundler to skulk

f0x52:
todo: also note that the recommended way to use the webdev setup is:
`GTS_PORT=8081 go run ./cmd/gotosocial testrig start`
`NODE_ENV=development node web/source`
and open the livereloading pages at localhost:8080


web/source/index.js
```diff
@@ -1,64 +1,69 @@
...
-
-	This program is distributed in the hope that it will be useful,
-	but WITHOUT ANY WARRANTY; without even the implied warranty of
-	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-	GNU Affero General Public License for more details.
-
-	You should have received a copy of the GNU Affero General Public License
-	along with this program.  If not, see <http://www.gnu.org/licenses/>.
-*/
-
```
tsmethurst, line 18: Oh! Why the license text deletion? does it end up duplicated or something?

tsmethurst (approved):

PR #955, opened 2022-11-04
Indicate that media and video sizes are also for download

EchedeyLR:
Explicitly tell that video and image size are for both upload and download.

### DIFF
--- a/docs/configuration/media.md
+++ b/docs/configuration/media.md
@@ -9,12 +9,12 @@
 
 # Config pertaining to user media uploads (videos, image, image descriptions).
 
-# Int. Maximum allowed image upload size in bytes.
+# Int. Maximum allowed image size in bytes to upload to this instance or download from other instances.
 # Examples: [2097152, 10485760]
 # Default: 10485760 -- aka 10MB
 media-image-max-size: 10485760
 
-# Int. Maximum allowed video upload size in bytes.
+# Int. Maximum allowed video size in bytes to upload to this instance or download from other instances.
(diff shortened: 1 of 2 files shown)


EchedeyLR:
I recently realized by a conversation with @foreverxml what it seems the mix between upload and download sizes for image and video in the settings as well as its clarification in documentation.

I guess, that similar to what it happens with emojis, this should be changed to local and remote settings to provide wider configuration.

This pull request is made with the guess that it is really like this.

tsmethurst:
IIRC it isn't the case that `media-image-max-size` and `media-video-max-size` also limit what your instance will download. The maximum fetchable body size is hardcoded right now to 40MB (https://github.com/superseriousbusiness/gotosocial/blob/main/internal/httpclient/client.go#L108-L111), but I don't believe we limit the size of downloaded attachments beyond that. Unless i'm missing something. Which is very possible because it's been a while since I poked at this part of the codebase :')

EchedeyLR:
Could we leave this pull request open for discussion or similar about it?

To be exact, I was confused about it when I first spoke to @foreverxml as they had a problem getting higher files than this if I remember correctly.

However, I opened this being unsure. Maybe things could get worked to create a variable that is specific for download?

tsmethurst:
> Maybe things could get worked to create a variable that is specific for download?

that's a good idea :)

tsmethurst:
Heya, I opened https://github.com/superseriousbusiness/gotosocial/issues/1029 so we can discuss this there rather than leaving the PR open :)

PR #956, opened 2022-11-05
[bugfix] use the landing page user directly

LawnGnome:
If set, the landing page user configuration value is used as a Gin context parameter, which seems incorrect, since a normal request isn't going to have a parameter named after an arbitrarily configured user. Instead, the user name should be used directly when building the redirect URL.


tsmethurst:
Thanks! :)

Issue #917, opened 2022-10-20
[feature] Use the `sub` claim for OIDC Accounts


theSuess:
### Is your feature request related to a problem ?

As currently OIDC accounts are tied to email address, this can lead to issues when changing the email of a user. Per the OIDC spec, the `email` claim is neither guaranteed to be stable or unique. The only reliable identifier is the `sub` claim.

### Describe the solution you'd like.

For externally authenticated user, store an additional field to have a stable reference to the correct user. This prevents account takeovers (accidental or with malicious intent)

### Describe alternatives you've considered.

There really is no other stable way to prevent hostile account takeovers in a self-service OIDC environment

### Additional context.

Relevant OIDC spec for reference: https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability

tsmethurst:
closed by https://github.com/superseriousbusiness/gotosocial/pull/961

Issue #763, opened 2022-08-22
[feature] Make fields for OIDC username/display name etc configurable


tsmethurst:
Right now we just use the `Name` field of the OIDC claim for username. We should make it possible for instance admins to configure which claim field(s) should map to which User fields.

This will require adding some new config options in the `internal/config` package (and updating cli tests), fleshing out the `Claims` struct in `internal/oidc` to add more standard OIDC fields, and also updating the `parseUserFromClaims` function in `internal/api/client/auth/callback.go` to map using the configured fields rather than hardcoded ones.

To avoid breaking any existing deployments, the config should still use Name as the default Username mapping.

To support all this, we should write some decent OIDC documentation!

Handy document explaining different claims + scopes etc: https://developer.okta.com/blog/2017/07/25/oidc-primer-part-1

theSuess:
I really like the gitea approach: On first registration prompt the user for a desired username/displayname but prefill it with claims from the token.

This also offers an easy fallback regarding #309

tsmethurst:
closed by https://github.com/superseriousbusiness/gotosocial/pull/961

Issue #309, opened 2021-11-16
[feature] OIDC: support identity providers without name in id_token


LittleFox94:
While trying to configure GTS to use my personal Gitlab as OIDC provider, it failed authenticating me since Gitlab does not set a `Name` claim in the `id_token`. For now I patched my local code to use the local part of the `EMail` claim instead, but this is only a hacky workaround.

Is proper support for such IDPs wanted? I guess some kind of IDP interface to handle such quirks would be a good way to tackle this, users then configuring which kind of IDP it is (`Generic`, `Gitlab`, ..) and the Gitlab implementation would make an API call to get the name.

tsmethurst:
> Is proper support for such IDPs wanted

Yes :) I got OIDC in as quickly as possible a while ago because I wanted to be able to use Dex as an OIDC provider, but after getting it working I didn't test it with other providers.

I think one solution here--rather than having separate interfaces per-provider--would be to allow more granular configuration under the `oidc` section, so that instance admins can specify where certain values should be selected from. I know I've seen this pattern elsewhere--I think in Synapse?

LittleFox94:
Problem for Gitlab is, it does not provide the name in the `id_token` at all - stating you have to call an API to get it, as written in this doc below the table:
https://docs.gitlab.com/ee/integration/openid_connect_provider.html#shared-information

I've seen something like a common base class and implementations for different provider quirks (yeah, for me they are quirks - maybe I'll go fix gitlab in this regard) and let the user choose one of that quite often - like in Gitlab's Omniauth

LittleFox94:
We could add a config `URL+jsonpath to retrieve name from` though, this isn't completely generic but might work in enough cases for now :tm:

LittleFox94:
Just looked at the OIDC spec to find if `name` is a required claim and found this info about claims "They can be requested to be returned either in the UserInfo Response, per Section 5.3.2, or in the ID Token, per Section 2.". Looking at `Section 5.3.2`, I learned the API route Gitlab wants applications to use is actually one of the valid ways in OIDC - the route where to retrieve that is given in the discovery document.

I think I can build something to retrieve the additional claims from that route when some are missing :)

tsmethurst:
closed by https://github.com/superseriousbusiness/gotosocial/pull/961

tsmethurst:
Closed by accident!

daenney:
[...]

PR #961, opened 2022-11-05
[feature] overhaul the oidc system

theSuess:
Since I'm very interested in this topic, I've built a draft to implement my feedback on #763, #917 and #309 

This is by no means final, and very much intended as a base for further discussions on these issues.

It works by intercepting the first OIDC callback of a user and returns a small page for customizing the username if no matching user is found. The text field is prefilled with the `preferred_username` claim.

image

It also performs validation and checks the DB for colliding usernames

The biggest open topic IMHO is finding a way to migrate existing OIDC setups to the new `ExternalID` field. I honestly have no idea what would be the best approach here


internal/api/client/auth/callback.go
```diff
@@ -110,56 +116,124 @@ func (m *Module) CallbackGETHandler(c *gin.Context) {
...
+		m.clearSession(s)
+		api.ErrorHandler(c, gtserror.NewErrorInternalError(err), m.processor.InstanceGet)
+		return
+	}
+	c.Redirect(http.StatusFound, OauthAuthorizePath)
+}
 
+func (m *Module) fetchUserForClaims(ctx context.Context, claims *oidc.Claims, ip net.IP, appID string) (*gtsmodel.User, gtserror.WithCode) {
+	if claims.Sub == "" {
+		return nil, gtserror.NewErrorBadRequest(errors.New("no sub claim found - is your provider OIDC compliant?"))
```
tsmethurst, line 223: because this is a helpful error to users, you should provide it as the helpText as well imo

so like:

```go
err := errors.New("no sub claim found - is your provider OIDC compliant?")
return nil, gtserror.NewErrorBadRequest(err, err.Error())
```

That way the end user will see it returned from the API

internal/oidc/claims.go
```diff
@@ -18,10 +18,18 @@
...
+	Sub               string   `json:"sub"`
+	Email             string   `json:"email"`
+	EmailVerified     bool     `json:"email_verified"`
+	Groups            []string `json:"groups"`
+	Name              string   `json:"name"`
+	PreferredUsername string   `json:"preferred_username"`
+}
+
+func init() {
+	gob.Register(&Claims{})
```
tsmethurst, line 34: Could you explain this? I'm not sure what it does :P

cmd/gotosocial/action/admin/account/account.go
```diff
@@ -78,7 +78,7 @@ var Create action.GTSAction = func(ctx context.Context) error {
 		return err
 	}
 
-	_, err = dbConn.NewSignup(ctx, username, "", false, email, password, nil, "", "", false, false)
+	_, err = dbConn.NewSignup(ctx, username, "", false, email, password, nil, "", "", false, "", false)
```
tsmethurst, line 81: Man this function is becoming a bit messy huh? Out of scope for this PR, but in future we should make that NewSignup function accept a struct with parameters instead :')

internal/cache/user.go
```diff
@@ -87,6 +87,11 @@ func (c *UserCache) GetByEmail(email string) (*gtsmodel.User, bool) {
 	return c.cache.GetBy("email", email)
 }
 
+// GetByExternalID attempts to fetch a user from the cache by its external id, you will receive a copy for thread-safety
+func (c *UserCache) GetByExternalID(id string) (*gtsmodel.User, bool) {
+	return c.cache.GetBy("external_id", id)
```
tsmethurst, line 92: Does this need to be registered with the cache as a lookup @NyaaaWhatsUpDoc ?

tsmethurst (commented):
This is looking pretty tasty IMO. Got a few comments about clarification and stuff. Overall, this looks like it's going in a great direction. I'm glad you bit the bullet and just jumped into it, it's always a nice surprise to get a really good PR like this :)

internal/oidc/claims.go
theSuess: Sure: the way I implemented the finalization page, the claim temporarily gets stored in the session. Since the session package uses gob to serialize the struct, we need to register it as serializable. I'm open to cleaner approaches - am not too happy with this as well

theSuess:
> I had an idea for a graceful migration path:
> 
>     * Introduce a config flag to try matching the email to existing users
> 
>     * If a user already exists for a login request but does _not_ have the externalID set, ask the user if they want to link the account
> 
>     * If yes, populate the externalID field
> 
>     * If not, ask the user to provide a new username
> 
> 
> What do you think about this?

I've now implemented this (without the step of asking for a new username for now).

With this, we have a migration path for servers already using OIDC

tsmethurst:
Heya, apologies for being slow to respond to this, it's been rather a hectic week :')

This looks good to me so far! At least, I don't see any glaring issues, but will have to try it out on gts.superseriousbusiness.org (which uses oidc) to see how it works :)

One thing you missed is adding the new config flag to `examples/config.yaml`, and also updating the documentation here https://github.com/superseriousbusiness/gotosocial/blob/main/docs/configuration/oidc.md.

Once those things are done, I think you can mark this as ready for review and we can have a proper look for it.

Thank you, this is great work!

theSuess:
No problem, I'm just happy to hack on something not work-related :grin: 

I've added the docs and example config. I always seem to forget at least one place when adding new config options :D

Also marked this PR as ready :crossed_fingers:

tsmethurst:
This looks good to me :) @NyaaaWhatsUpDoc do you also wanna review or nah?

tsmethurst:
hey @theSuess , can you merge main into this or rebase to resolve the merge conflicts? then i'll look at it afresh

theSuess:
Rebased and tested

web/template/finalize.tmpl
```diff
@@ -0,0 +1,31 @@
...
+              To ensure the best experience for you, we need you to provide some additional details.
+            </p>
+            {{if .error}}
+              <section class="error">
+                <span>❌</span> <pre>{{.error}}</pre>
+              </section>
+            {{end}}
+            <div class="callout">
+              <p class="callout-title">Important</p>
+              <p>Due to the way the ActivityPub standard works, you <storng>cannot</strong> change your username after it has been set.</p>
```
dwmunster, line 17: `<storng>` should probably be `<strong>`.

theSuess: Derp, thanks for catching this!

tsmethurst:
No worries pal :)

docs/configuration/oidc.md
```diff
@@ -76,48 +83,36 @@ When OIDC is enabled on GoToSocial, the default sign-in page redirects automatic
...
-For example, say that an OIDC provider returns the username `Marx Is Great` for a sign in, which doesn't fit the pattern because it contains upper-case letters and spaces.
-
-In this case, GtS will convert it into `marx_is_great` by applying the following rules:
-
-1. Trim any leading or trailing whitespace.
-2. Convert all letters to lowercase.
-3. Replace spaces with underscores.
-
-Unfortunately, at this point GoToSocial doesn't know how to handle returned usernames containing special characters such as `@` or `%`, so these will return an error.
+This then allows you to change the username on a provider level without loosing
```
NyaaaWhatsUpDoc, line 97: * `without losing`

NyaaaWhatsUpDoc:
I have started to review this, but I have ran out of spoons for the evening, and it is late. I shall get back to it tomorrow evening most likely :)

tsmethurst:
i suggest we merge this once 0.6.0 is out, so we have time to mess about with it :)

internal/api/client/auth/callback.go
```diff
@@ -110,115 +116,152 @@ func (m *Module) CallbackGETHandler(c *gin.Context) {
...
+	claims, ok := s.Get(sessionClaims).(*oidc.Claims)
+	if !ok {
+		err := fmt.Errorf("key %s was not found in session", sessionClaims)
+		api.ErrorHandler(c, gtserror.NewErrorBadRequest(err, oauth.HelpfulAdvice), m.processor.InstanceGet)
+		return
+	}
+	user, errWithCode := m.createUserFromOIDC(c.Request.Context(), claims, form, net.IP(c.ClientIP()), appID)
+	if errWithCode != nil {
+		m.clearSession(s)
+		api.ErrorHandler(c, errWithCode, m.processor.InstanceGet)
```
royalfork, line 208: Should this return here?  Otherwise, `user == nil` and the below call to `s.Set(sessionUserID, user.ID)` panics from a nil pointer dereference.

theSuess: good catch - it's fixed in the newest commit

internal/api/client/auth/callback.go
```diff
@@ -110,115 +116,153 @@ func (m *Module) CallbackGETHandler(c *gin.Context) {
...
 
-	user, errWithCode := m.parseUserFromClaims(c.Request.Context(), claims, net.IP(c.ClientIP()), app.ID)
+	user, errWithCode := m.fetchUserForClaims(c.Request.Context(), claims, net.IP(c.ClientIP()), app.ID)
 	if errWithCode != nil {
 		m.clearSession(s)
 		api.ErrorHandler(c, errWithCode, m.processor.InstanceGet)
 		return
 	}
-
+	if user == nil {
```
NyaaaWhatsUpDoc, line 125: Again this is quite a heavy unseparated function block. I know this is quite nitpicky I'm just a bit of a stickler for readability!

theSuess: Also added some comments here. We could also split this up into its own function, but not sure if that would increase readability by much

PR #977, opened 2022-11-06
Add quick start docker compose example

manfre:
Modified the Dockerfile to create the folders necessary to use the example with a volume mount or local volume bind.

Provide an optional entrypoint that will create the initial admin user based upon provided environment variables.
* GTS_ADMIN_USERNAME
* GTS_ADMIN_EMAIL
* GTS_ADMIN_PASSWORD

### DIFF
--- a/docs/installation_guide/docker.md
+++ b/docs/installation_guide/docker.md
@@ -78,7 +78,7 @@ image: superseriousbusiness/gotosocial:0.3.1
 
 #### Host
 
-Change the `GTS_HOST` environment variable to the domain you are running GoToSocial on.
+Change the `GTS_HOST` environment variable to the domain you are running GoToSocial on. If you want your ActivityPub identity to be different than the host name, set the `GTS_ACCOUNT_DOMAIN` environment variable. For example, your `GTS_HOST` is `gts.example.org`, but you want your user to be `myuser@example.org`, instead of `myuser@gts.example.org`, you would get `GTS_ACCOUNT_DOMAIN` as `example.org`.
 
 #### User (optional / probably not necessary)
 
@@ -167,3 +167,8 @@ GoToSocial should now be running on your machine! To verify this, open your brow
 ## (Optional) Reverse Proxy
 
 If you want to run other webservers on port 443, or want to add an additional layer of security you might want to add [NGINX](https://nginx.org), [Traefik](https://doc.traefik.io/traefik/), or [Apache httpd](https://httpd.apache.org/) into your docker-compose to use as a reverse proxy.
+
+
+## (Optional) Quick Initialize User
+
+If you are deploying a single user install where shell access and/or executing commands on the container is complicated or undesirable, it is possible to initialize an admin user with a different `ENTRYPOINT`. Use `wget` to download the latest quickstart [docker-compose.yml](https://raw.githubusercontent.com/superseriousbusiness/gotosocial/main/example/docker-compose-quickstart/docker-compose.yml) example and customize as needed.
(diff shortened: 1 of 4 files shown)


scripts/quick-start.sh
```diff
@@ -0,0 +1,21 @@
...
+        echo "Initializing Admin user ${GTS_ADMIN_USERNAME}"
+        /gotosocial/gotosocial admin account create --username ${GTS_ADMIN_USERNAME} --email ${GTS_ADMIN_EMAIL} --password '${GTS_ADMIN_PASSWORD}'
+
+        /gotosocial/gotosocial admin account promote --username ${GTS_ADMIN_USERNAME}
+    else
+        echo "Existing admin user ${GTS_ADMIN_USERNAME} found"
+    fi
+fi
+
+/gotosocial/gotosocial server start
```
tsmethurst, line 21: I think this should use `exec`, since we don't want the script to be the parent

scripts/quick-start.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+set -e
+
+if [ "$GTS_ADMIN_USERNAME" != "" ]
+then
```
tsmethurst, line 6: I would add some checks here to ensure that the other necessary environment variables are also set

example/docker-compose-quickstart/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3.8"
```
tsmethurst, line 1: IMO it overcomplicates things to have a separate docker compose for the quick start. I would rather put this in the existing docker-compose file and just comment out the quickstart script variables, with a note to uncomment them if that functionality is desired

tsmethurst (commented):
I'm not 100% sure about this PR. In some ways I think it makes things *more* complicated, but I think it could be streamlined. @NyaaaWhatsUpDoc what do you think?

docs/installation_guide/docker.md
```diff
@@ -78,7 +78,7 @@ image: superseriousbusiness/gotosocial:0.3.1
 
 #### Host
 
-Change the `GTS_HOST` environment variable to the domain you are running GoToSocial on.
+Change the `GTS_HOST` environment variable to the domain you are running GoToSocial on. If you want your ActivityPub identity to be different than the host name, set the `GTS_ACCOUNT_DOMAIN` environment variable. For example, your `GTS_HOST` is `gts.example.org`, but you want your user to be `myuser@example.org`, instead of `myuser@gts.example.org`, you would get `GTS_ACCOUNT_DOMAIN` as `example.org`.
```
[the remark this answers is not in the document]
manfre: A fair concern. Would it be helpful if I provided (independent of this PR) documentation related to using `account-domain`? Or would that potentially cause more support issues by making that functionality more accessible before it's potentially fully ready for use?

example/docker-compose-quickstart/docker-compose.yml
manfre: I opted for a separate compose file to help identify it as being not part of the current, standard flow of running from a shell with local files/folders. Combining and commenting to provide example of the quickstart without the extra mounted volumes works for me.

manfre:
> I'm not 100% sure about this PR. In some ways I think it makes things _more_ complicated, but I think it could be streamlined.

Thank you for taking the time to review my PR. I don't want to add complication or potentially increase the support burden on the project. I'm closing this PR to reduce cognitive load on y'all. Discussion about if this functionality is wanted and how to achieve it without adding complication can continue on [the related issue](https://github.com/superseriousbusiness/gotosocial/issues/978).

Issue #990, opened 2022-11-07
[bug] fix nofollow on sanitization to exclude adding nofollow on server local links


buro9:
### Describe the bug with a clear and concise description of what the bug is.

In the present sanitization policy `rel="nofollow"` is added to all links, but you probably don't want to add these to server local links such as to profiles on the same instance or the About page.

I am the author of the bluemonday sanitization package and am proposing a small improvement in your policy.

### What's your GoToSocial Version?

latest - I'm just reading the code

### What happened?

Adds `rel="nofollow"` everywhere

### What you expected to happen?

Add `rel="nofollow"` only on fully qualified (i.e. external) links, and not on local relative links

### How to reproduce it?

Just post a link and view source.

### Anything else we need to know?

I am the author and maintainer of the bluemonday package, this is just a quality suggestion for your sanitization policy... a PR with the change is inbound

PR #991, opened 2022-11-07
Fix rel="nofollow" so that it only applies to external links

buro9:
Resolves https://github.com/superseriousbusiness/gotosocial/issues/990


tsmethurst:
Thank you :) And many thanks for bluemonday, it's a great library!

PR #992, opened 2022-11-07
refer to the latest release version

kaja47:


tsmethurst:
Thanks! We will need a more durable solution in future for this, but at least this brings the existing docs up to date :P

PR #1016, opened 2022-11-10
[chore] close in-storage media readers beforeopening write

NyaaaWhatsUpDoc:


tsmethurst (commented):
LGTM. Is it okay to close a closer twice? I guess the second close would be a noop then?

PR #1018, opened 2022-11-11
fix logo display on the web index page.

LegendRegent:
Hi~ I am a student and just start learning to join the open-source community. Please tell me if I did anything wrong. Currently, I am pretty novice but I wish I could help :)

So I found that on the index page, the header template used "instance.Thumbnail" to show the logo. This variable does not contain the port number and thus may not be able to present properly. (http://localhost/assets/logo.png can't show the logo, but http://localhost:8080/assets/logo.png)

I got the port from config and added it to instance.Thumbnail to solve this.

I apologize for not commenting on an issue in advance. I will do better next time.

### DIFF
--- a/internal/typeutils/internaltofrontend.go
+++ b/internal/typeutils/internaltofrontend.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


LegendRegent:
I am so sorry. According to the log, there's indeed a port number. I will close this.

Issue #1023, opened 2022-11-11
[bug] Example systemctl config does not work on Debian


sargant:
### Describe the bug with a clear and concise description of what the bug is.

When installing the binary version using the supplied documentation, the service installation steps don't work without further configuration.

### What's your GoToSocial Version?

0.5.2

### GoToSocial Arch

amd64 Binary

### What happened?

Unable to start the systemctl service using the binary installation instructions:

```
root@gotosocial:/opt/gotosocial# systemctl start gotosocial
Job for gotosocial.service failed because the control process exited with error code.
See "systemctl status gotosocial.service" and "journalctl -xe" for details.
```

### What you expected to happen?

Be able to start the service successfully without further configuration

### How to reproduce it?

Follow documented binary service installation instructions on Debian 11

owovin:
Had same results, running `systemctl -u gotosocial.service -b" gave me a constant result of 
```
Nov 11 17:51:48 Sol systemd[1]: Starting GoToSocial Server...
Nov 11 17:51:48 Sol systemd[160355]: gotosocial.service: Failed to set up standard output: No such file or directory
Nov 11 17:51:48 Sol systemd[160355]: gotosocial.service: Failed at step STDOUT spawning /gotosocial/gotosocial: No such file or directory
```
if you can crossreference to the same error, that'll help the developers out 🤝
the service file looks just fine from second/third checking, so i dont know what it would be

sargant:
Alternative PR in #1030 that instead changes the default log path to involve no new subdirectory creation (defaults to a new file in `/var/log`), which should also avoid permissions errors.

PR #1024, opened 2022-11-11
Add documentation for log directory creation

sargant:
See #1023

### DIFF
--- a/docs/installation_guide/binary.md
+++ b/docs/installation_guide/binary.md
@@ -128,7 +128,13 @@ Copy it to `/etc/systemd/system/gotosocial.service`.
 sudo cp /gotosocial/example/gotosocial.service /etc/systemd/system/
 ```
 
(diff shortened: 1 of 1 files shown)


NyaaaWhatsUpDoc:
Looks good, though should created log directory have permissions `0700` ?

sargant:
True - maybe changing the log path to /var/log/gotosocial.log would be better? Then permissions aren't an issue and there's no additional manual user instructions. Don't want to be moving log file targets unexpectedly though.

tsmethurst:
Oh... I think it's likely that it was meant to be in `log` rather than `logs` originally, and this was an error... Feel free to change the path imo, for consistency

sargant:
OK - there's an alternative PR in !1030 that changes the default log path instead and leaves the documentation unchanged

tsmethurst:
Thank you! I merged the other PR so i'll close this one now :)

Issue #1032, opened 2022-11-12
[feature] Support different formats (markdown) for Bio


manfre:
### Is your feature request related to a problem ?

I'd like to use Markdown to format my user Bio.

### Describe the solution you'd like.

* Add a `NoteFormat` field to account that behaves in a similar fashion as `StatusFormat`.
* Add the format select drop down to `/settings/user/profile` with a default of "plain".
* Update `processNote` to toggle the `From*` behavior based upon `NoteFormat`

### Describe alternatives you've considered.

Considered putting the format select on `/settings/user/settings`, but that would be odd UX.

### Additional context.

![Screenshot 2022-11-12 093754](https://user-images.githubusercontent.com/30523/201479241-02f0191e-82e0-4f82-8f90-00e9f6784e38.png)

manfre:
Proof of concept implementation at https://github.com/manfre/gotosocial/tree/noteformat

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/pull/1037

PR #1037, opened 2022-11-12
Support markdown format for Account bio/note

manfre:
Implements https://github.com/superseriousbusiness/gotosocial/issues/1032

### DIFF
--- a/docs/user_guide/posts.md
+++ b/docs/user_guide/posts.md
@@ -106,7 +106,7 @@ When set to `false`, likes/faves of your post will not be accepted by your GoToS
 
 ## Input Types
 
-GoToSocial currently accepts two different types of input for posts. These are:
+GoToSocial currently accepts two different types of input for posts (and user bio). These are:
 
 * `plain`
 * `markdown`
--- a/docs/user_guide/user_panel.md
+++ b/docs/user_guide/user_panel.md
@@ -32,6 +32,8 @@ Your bio is a longer text that introduces your account and your self. Your bio i
 - Describe your boundaries and preferences when it comes to other people interacting with you
 - Link hashtags that you often use when you post
 
+The bio accepts either `plain` or `markdown` formatting. This is set by the default post format setting described in [Post Settings](#post-settings).
+
 After updating your display name and bio, click on the `Save profile info` button at the bottom of the Profile Info section to save your changes.
 
 ### Manually Approve Followers / Lock Your Account
--- a/internal/processing/account/update.go
+++ b/internal/processing/account/update.go
@@ -66,7 +66,7 @@ func (p *processor) Update(ctx context.Context, account *gtsmodel.Account, form 
 		account.NoteRaw = *form.Note
 
 		// Process note to generate a valid HTML representation
-		note, err := p.processNote(ctx, *form.Note, account.ID)
+		note, err := p.processNote(ctx, *form.Note, account)
 		if err != nil {
 			return nil, gtserror.NewErrorBadRequest(err)
 		}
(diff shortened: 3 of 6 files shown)


internal/validate/formvalidation.go
```diff
@@ -137,6 +137,19 @@ func Note(note string) error {
 	return nil
 }
 
+// NoteFormat checks that the desired note format setting is valid.
+func NoteFormat(noteFormat string) error {
+	if noteFormat == "" {
+		return nil
+		// return fmt.Errorf("empty string for note format not allowed")
```
manfre, line 144: I'm not sure if it should require a value given that it will fallback to doing `fromPlain` if the `NoteFormat` is any value other than "markdown"

tsmethurst (commented):
The code looks OK to me on a first read through, but I'm not entirely sure how necessary it is to have a setting + db column just for bio/note format parsing, since people update their bios far less often than they post notes. Might it be better to just use the flag we already have for posts, and add a note somewhere that when you set your post parsing to 'markdown' it also applies for bio/note updates?

Not married to the above idea or saying 'we have to do it this way', just to be clear :P Genuinely wondering which way round would be best.

NyaaaWhatsUpDoc:
> The code looks OK to me on a first read through, but I'm not entirely sure how necessary it is to have a setting + db column just for bio/note format parsing, since people update their bios far less often than they post notes. Might it be better to just use the flag we already have for posts, and add a note somewhere that when you set your post parsing to 'markdown' it also applies for bio/note updates?
> 
> Not married to the above idea or saying 'we have to do it this way', just to be clear :P Genuinely wondering which way round would be best.

Yes I am also in agreement with this, it's an extra db column for something relatively small. I'd really like to exhaust other options before coming to that.

manfre:
The main reasons for having separate format fields are:
* Small posts are shorter and typically less structured than a bio. Users may want "plain" for posts and "markdown" for Bio
* Overloading meaning of StatusFormat, even with documentation, will confuse users and result to support questions
* Future support for additional format types that may not make sense for both types of content. E.g. It might make sense to allow users to use "html" (still sanitized a little) in their Bio, but that wouldn't make sense for posts.

Can you elaborate on your concerns with adding this additional db column? I'd like to know for this issue and a more generic view for any future PRs.

tsmethurst:
> It might make sense to allow users to use "html" (still sanitized a little) in their Bio, but that wouldn't make sense for posts.

They can already do this a bit, since our markdown parser also supports html :)

> Overloading meaning of StatusFormat, even with documentation, will confuse users and result to support questions

Yeah that's indeed a concern. Perhaps we can just change the lingo to `InputFormat` instead, to denote that it's for both bios and notes (the only two pieces of text that users ever really 'input' apart from shorter things like display names etc).

> Can you elaborate on your concerns with adding this additional db column?

In my case, my concerns with adding db columns when it's not necessary are mainly to do with storage space and db query performance, though i admin in this case it's pretty minor since it's just one small text column.

manfre:
Now that I know the markdown parser supports html, I see a different approach that can be taken. Would it make sense to always parse the bio field as markdown? In the (likely rare) situations where markdown parsing is problematic, users can use html. Let me know if this approach is more acceptable and I can update the PR.

tsmethurst:
> Would it make sense to always parse the bio field as markdown?

Yeah perhaps... Then in the settings page after 'bio' we could just add in brackets `(markdown supported)` or something like that. That requires even less work :D

tsmethurst:
@mirabilos do you have any input on this? I know you've been having a time with markdown posting lately :P

mirabilos:
tobi dixit:

***@***.*** do you have any input on this? I know you've been having a
***@***.*** with markdown posting lately :P

Just because I want some formatting, though I’d be happy with HTML.

Though both the “(markdown supported)” note and a possible toggle for
user preferences have the problem of how to communicate them to the
client used to change the profile…

I don’t have any ideas for this right now. Busy with work again. ☹

bye,
//mirabilos
-- 
  "Using Lynx is like wearing a really good pair of shades: cuts out
   the glare and harmful UV (ultra-vanity), and you feel so-o-o COOL."
                                         -- Henry Nelson, March 1999

tsmethurst:
Alright, so I think the best way to go with this would be to:

1. Add a note next to the default post format selector in the user settings panel to indicate that this selection applies to bio writing as well.
2. Use that setting when parsing bio updates.

For users who don't select markdown but keep the plain default, this will not break any expectations for how to write posts and bios.

For users who select 'markdown' specifically, they then know what they're letting themselves in for.

What do you think?

manfre:
@tsmethurst Updated PR with suggested changes.

tsmethurst:
Nice one :) Could you add a test in here somewhere to verify that the bio/note markdown parsing works as expected? https://github.com/superseriousbusiness/gotosocial/blob/main/internal/api/client/account/accountupdate_test.go

manfre:
Added a test to `internal/processing/account/update_test.go` which seemed a more appropriate place.

tsmethurst:
Great, thank you very much!

Issue #1031, opened 2022-11-12
[feature] Allow access to public timeline without authorization


sargant:
### Is your feature request related to a problem ?

No

### Describe the solution you'd like.

I would like to be able to query `/api/v1/timelines/public` without authorization. Perhaps enabled behind a config flag like `instance-expose-public-timeline` with a default of `false`?

### Describe alternatives you've considered.

NONE

### Additional context.

Suggested PR in #1039

tsmethurst:
closed by https://github.com/superseriousbusiness/gotosocial/pull/1039

PR #1039, opened 2022-11-12
[feature] add instance-expose-public-timeline flag

sargant:
Adds a config flag that allows unauthenticated access to `/api/v1/timelines/public`. Defaults to false to replicate existing behaviour.

Implements #1031 

(Also first ever time writing Go! Sorry for being part of the Eternal September...)


tsmethurst:
> Sorry for being part of the Eternal September

I keep seeing that phrase but I have no idea what it means :') The pull request is much appreciated! Will review it shortly :)

sargant:
> > Sorry for being part of the Eternal September
> 
> I keep seeing that phrase but I have no idea what it means :')

Back in the very early days of the internet, people only got access to Usenet discussions when they started university in September, so there was a sudden downtick in etiquette, then people got better behaved. Then AOL gave all its users access to Usenet, and there were people new to the system all the time, and the need for constant moderation began ;)

https://en.wikipedia.org/wiki/Eternal_September#History

The same thing is happening with ActivityPub/Mastodon now with the exodus from the "other place", and I'm part of that exodus - sorry!

internal/api/client/timeline/public.go
```diff
@@ -110,10 +111,14 @@ import (
 //		'400':
 //			description: bad request
 func (m *Module) PublicTimelineGETHandler(c *gin.Context) {
-	authed, err := oauth.Authed(c, true, true, true, true)
-	if err != nil {
-		api.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGet)
-		return
+	authed := &oauth.Auth{}
+	if !config.GetInstanceExposePublicTimeline() {
```
tsmethurst, line 115: To handle cases where a token is not required but it *is* supplied, I would rewrite this as:

```go
	var authed *oauth.Auth
	var err error
	
	if config.GetInstanceExposePublicTimeline() {
		authed, err = oauth.Authed(c, false, false, false, false)
	} else {
		authed, err = oauth.Authed(c, true, true, true, true)
	}

	if err != nil {
		api.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGet)
		return
	}
```

tsmethurst (commented):
This looks good :) Just the change I recommended in the comment. @NyaaaWhatsUpDoc do you have any thoughts on this PR overall?

internal/api/client/timeline/public.go
sargant: Done - thanks for the feedback

internal/api/client/timeline/public.go
```diff
@@ -110,7 +111,15 @@ import (
 //		'400':
 //			description: bad request
 func (m *Module) PublicTimelineGETHandler(c *gin.Context) {
-	authed, err := oauth.Authed(c, true, true, true, true)
+	var authed *oauth.Auth
+	var err error
+	
+	if config.GetInstanceExposePublicTimeline() {
+		authed, err = oauth.Authed(c, false, false, false, false)
```
NyaaaWhatsUpDoc, line 118: I think it's worth adding a comment here explaining exactly what's going on, especially given the nature of it.

sargant: Done

tsmethurst:
Thank you! :)

PR #1040, opened 2022-11-12
[chore] update database caching library

NyaaaWhatsUpDoc:
Updates to use `result.Cache{}` as the database caching library, which reduces need for boiler plate code, and allows for caching of negative results.

The caches themselves will be a bit less performant than previous iterations due to the use of reflection, and having to stringify any generic key type, and internal cache ordering / size limits (a requested feature). But more than made up for by the caching of negative database results, so no need to recheck values of known non-existent keys. And again, configurable size limits!

Configuration of the cache size limits will come at a later date.


NyaaaWhatsUpDoc:
This has been running on `k.iim.gay` now for over 24hrs without issues :)

internal/db/bundb/tombstone.go
```diff
@@ -43,7 +43,7 @@ func (t *tombstoneDB) init() {
 		t2 := new(gtsmodel.Tombstone)
 		*t2 = *t1
 		return t2
-	}, 1000)
+	}, 100)
```
tsmethurst, line 46: Any special reason to have this be 100 when the other ones are normally 1000?

internal/db/bundb/user.go
```diff
@@ -22,119 +22,129 @@ import (
...
-		Model(user).
-		Where("? = ?", bun.Ident("user.id"), user.ID).
-		Column(columns...).
-		Exec(ctx); err != nil {
-		return nil, u.conn.ProcessError(err)
-	}
-
-	u.cache.Invalidate(user.ID)
-	return user, nil
+	return u.cache.Store(user, func() error {
```
tsmethurst, line 140: Is there no way to keep the columns here (and in other places)? It's more efficient if we can only update the columns the caller specifies.

tsmethurst (commented):
Largely looks very nice :) Just a couple queries

Issue #1010, opened 2022-11-09
[bug] Unable to login from Mastodon iOS app


xplosionmind:
### Describe the bug with a clear and concise description of what the bug is.

After successfully inserting the credentials in the Mastodon iOS app, the login window disappears and there is no progress.

### What's your GoToSocial Version?

YunoHost 0.5.2~ynh1

### What happened?

![error](https://user-images.githubusercontent.com/42862428/200890420-26d12ab9-1309-46b8-a445-b9ed280f21e3.gif)

### What you expected to happen?

The login should successfully take place

tsmethurst:
Thanks :) We still need someone with an iphone to investigate this, since none of the core devs use iphone :crying_cat_face:

june-coloc:
Hello, not a dev but I have an iPhone and a GTS account!

xplosionmind:
No hurries, I simply switched to Metatext for the moment

sargant:
Had issues myself with GTS 0.5.2 and the Mastodon iOS app. Weirdly, it fails _90%_ of the time - if you try enough times, it does eventually work.

I'm reticent to log out and capture logs now it's working - maybe I'll try later today...

PR #1042, opened 2022-11-13
Fix login on Mastodon iOS app for users with no statuses

dbkr:
Mastodon for iOS can't cope with an empty string for a date and expect a JSON `null` instead.

Fixes https://github.com/superseriousbusiness/gotosocial/issues/1010


tsmethurst:
Neat, thanks for the fix :)

Issue #815, opened 2022-09-10
[frontend] Display admin role on public profile


f0x52:
Especially wrt moderation it's useful to see if a remote user is that instance's admin. There's no client API to show a user is admin so existing fediverse implementations leave it to the web UI to show that.

Issue #814, opened 2022-09-10
[feature] admin status in Account response


f0x52:
for clients to tell if the user that logged in is admin or not, and can perform admin actions. Mastodon API does not seem to have an equivalent anywhere, but uses the term `role` for this I think?

Could be returned in `api/v1/verify_credentials` and `/api/v1/accounts/{id}`
```json
{
  "acct": "some_user@example.org",
  "role": "admin",
  ...
}
```

f0x52:
It would also be useful in `/api/v1/accounts/{id}` responses, so clients can show roles too (open mastodon issue from 2018 https://github.com/mastodon/mastodon/issues/7108#issuecomment-395953636)

tsmethurst:
I had a look around!

Mastodon user roles are defined here: https://github.com/mastodon/mastodon/blob/main/app/models/user_role.rb

And they're serialized to the api via the admin view of accounts: https://github.com/mastodon/mastodon/blob/main/app/serializers/rest/admin/account_serializer.rb#L23-L25 and verify_credentials: https://github.com/mastodon/mastodon/blob/main/app/serializers/rest/credential_account_serializer.rb#L21-L23

It looks like Mastodon returns an object there representing the user role and the various permissions and flags that user role allows, but I think we can keep this simple and just return `"admin"` if the user is an admin, and either `null` or `"user"` or something if they're not.

And yeah indeed we should add that to the /api/v1/accounts/{id} representation as well, agree with you there.

tsmethurst:
Aha! It's serialized like this: https://github.com/mastodon/mastodon/blob/main/app/serializers/rest/role_serializer.rb

> it has `id` (arbitrary string, but numeric in practice), `name`, `permissions` (it's a large integer in string representation, it's actually a bitfield of individual permissions), `color` (hex code), `highlighted` a boolean which does… something (i think it's to display the thing at all on public pages)

Thanks @ClearlyClaire !

PR #1045, opened 2022-11-14
[feature] Serialize local account role via API, and show it via web view

tsmethurst:
Local account roles will be one of `user`, `moderator`, or `admin`. For remote accounts, `role` will not be serialized in the json.

The role of an account will now be shown on the web view of that account's profile (see screenshots below).

I don't *think* this will trip up clients that use the Mastodon API, but we'll have to see. We might need to make this a config option for optimum masto API compatibility, but I hope not.

Screenshot from 2022-11-14 13-38-06
Screenshot from 2022-11-14 13-37-56
Screenshot from 2022-11-14 13-37-30

Closes https://github.com/superseriousbusiness/gotosocial/issues/814
Closes https://github.com/superseriousbusiness/gotosocial/issues/815


NyaaaWhatsUpDoc:
Backend changes look good :), my only critique is a UI thing, in that I think we should only display non-default roles. i.e. most users' profile pages appear as they do currently (pre-PR), but only moderators and admins get a role tag on their profile.

tsmethurst:
> we should only display non-default roles

good idea... will change it

PR #1049, opened 2022-11-14
[bugfix] Fix unicode-unaware word boundary check in hashtags

illfygli:
Go `\b` does not care for Unicode, and without lookahead in the regex engine, the workarounds got very ugly. So I replaced the regex with a parser that uses Go's Unicode functions directly.

The parser runs in O(n) time and performance should be comparable.

Might even be faster since I think this is less work, but I don't know how to benchmark. :)

I'm sure it's all sorts of unidiomatic but maybe it can be salvaged!

Hashtags that wouldn't work before were those ending in non-ASCII characters, like `#blå`, since `\b` would consider `l` followed by `å` to be a word boundary.

### DIFF
--- a/internal/util/statustools.go
+++ b/internal/util/statustools.go
@@ -25,6 +25,10 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/regexes"
 )
 
+const (
+	maximumHashtagLength = 30
+)
+
 // DeriveMentionNamesFromText takes a plaintext (ie., not html-formatted) text,
 // and applies a regex to it to return a deduplicated list of account names
 // mentioned in that text, in the format "@user@example.org" or "@username" for
@@ -96,9 +100,11 @@ func FindHashtagSpansInText(text string) []Span {
 	return tags
 }
 
-func appendTag(tags *[]Span, text string, start int, endEx int) {
-	if 1 < endEx-start {
-		*tags = append(*tags, Span{First: start, Second: endEx})
+func appendTag(tags *[]Span, text string, start int, end int) {
+	l := end - start - 1
+	// This check could be moved out into the parsing loop if necessary!
+	if 0 < l && l <= maximumHashtagLength {
+		*tags = append(*tags, Span{First: start, Second: end})
 	}
 }
 
@@ -117,9 +123,12 @@ func isPermittedInHashtag(r rune) bool {
 	return unicode.IsLetter(r) || unicode.IsNumber(r)
 }
 
+// Decides where to break before or after a hashtag.
 func isHashtagBoundary(r rune) bool {
-	return r == '#' ||
-		unicode.IsSpace(r) ||
-		unicode.IsControl(r) ||
-		('&' != r && '/' != r && unicode.Is(unicode.Categories["Po"], r))
+	return r == '#' || // `###lol` should work
+		unicode.IsSpace(r) || // All kinds of Unicode whitespace.
+		unicode.IsControl(r) || // All kinds of control characters, like tab.
+		// Most kinds of punctuation except "Pc" ("Punctuation, connecting", like `_`).
+		// But `someurl/#fragment` should not match, neither should HTML entities like `&#35;`.
+		('/' != r && '&' != r && !unicode.Is(unicode.Categories["Pc"], r) && unicode.IsPunct(r))
 }
--- a/internal/util/statustools.go
+++ b/internal/util/statustools.go
@@ -19,7 +19,8 @@
 package util
 
 import (
-	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/superseriousbusiness/gotosocial/internal/regexes"
 )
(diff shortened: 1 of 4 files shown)


NyaaaWhatsUpDoc:
Unidiomatic? This is miles better than so much Go code I see, it's great!

internal/util/statustools.go
```diff
@@ -36,16 +37,69 @@ func DeriveMentionNamesFromText(text string) []string {
...
+// and returns a list of pairs of indices into the original string, where
+// hashtags are located.
+func FindHashtagSpansInText(text string) []Span {
+	tags := []Span{}
+	start := 0
+	// Keep one rune of lookbehind.
+	prev := ' '
+	inTag := false
+
+	for i, r := range text {
```
NyaaaWhatsUpDoc, line 76: Compared to previously, we're now missing a maximum hashtag length check, which I think would be worth adding back as a part of this function. What do you think?

tsmethurst:
Yeah what kim said. I'll read through this properly tomorrow but on a first read through it looks really good.

NyaaaWhatsUpDoc:
Other than the one comment, I'm otherwise happy with this :), once it's sorted it'll be good to go imo

internal/util/statustools.go
```diff
@@ -96,9 +100,10 @@ func FindHashtagSpansInText(text string) []Span {
 	return tags
 }
 
-func appendTag(tags *[]Span, text string, start int, endEx int) {
-	if 1 < endEx-start {
-		*tags = append(*tags, Span{First: start, Second: endEx})
+func appendTag(tags *[]Span, text string, start int, end int) {
+	l := end - start
```
NyaaaWhatsUpDoc, line 104: Might be worth adding a note here that if necessary we can move the length check to the loop. I'm just basing this entirely on a hunch but my immediate thought was that this should be among the loop check itself to prevent exploiting the extra CPU time checking overly large hashtags, _but_ I don't know the specifics on how much CPU time these unicode checks in the loop will actually take so I'm think it's worth at least noting down the thought if it is ever an issue :p

illfygli: I was thinking that avoiding conditionals in the loop would be good, but ur right, maybe the other work outweighs that and it's already full of `if`s!

NyaaaWhatsUpDoc: Yeah that was exactly my thoughts too and ultimately wasn't sure. Whether you want to include it as a comment or update to add max length checks in the loop I'm happy either way :)

NyaaaWhatsUpDoc: Actually sorry to be a pain I'm realising now it might be easier to stick with the comment for now, I have just noticed a bug in the updated code. If we mark it as not `inTag` on reaching max length, then that means we could have a hashtag within a long hashtag recognized if that makes sense? e.g. `#ReallyLongHashtagThatWillNotGetRecognized#ButThisWill`

illfygli: Alright, went with a comment. :)

NyaaaWhatsUpDoc:
Alright then this looks good to me :). @tsmethurst you will probably want to catch up on the above conversations as well for some further context on things when you get to reviewing this

internal/util/statustools.go
```diff
@@ -58,3 +118,14 @@ func DeriveEmojisFromText(text string) []string {
...
+
+func isPermittedInHashtag(r rune) bool {
+	return unicode.IsLetter(r) || unicode.IsNumber(r)
+}
+
+func isHashtagBoundary(r rune) bool {
+	return r == '#' ||
+		unicode.IsSpace(r) ||
+		unicode.IsControl(r) ||
+		('&' != r && '/' != r && unicode.Is(unicode.Categories["Po"], r))
```
tsmethurst, line 130: i think it's worth putting a comment here to explain this

internal/util/statustools.go
```diff
@@ -36,16 +41,71 @@ func DeriveMentionNamesFromText(text string) []string {
...
+// hashtags are located.
+func FindHashtagSpansInText(text string) []Span {
+	tags := []Span{}
+	start := 0
+	// Keep one rune of lookbehind.
+	prev := ' '
+	inTag := false
+
+	for i, r := range text {
+		if r == '#' && isHashtagBoundary(prev) {
```
tsmethurst, line 81: worth replacing these with a switch statement or nah? just a style thing

tsmethurst (commented):
Sick, I love it :heart: 

Just a couple little comments :)

internal/util/statustools.go
illfygli: Cool I didn't know about `switch /* nothing here */`!
I tried it but then I cuoldn't do assignment in a case, so I left it like that instead of duplicating that bit. :)

tsmethurst:
Looks great to me now, good work! If you're happy with it i'll merge it :)

illfygli:
Yes :D

tsmethurst:
Boom! thank you!!

PR #1050, opened 2022-11-15
[frogend] fix profile spacing on very small screens

f0x52:


web/source/css/profile.css
```diff
@@ -171,6 +171,7 @@ main {
 		}
 
 		.basic {
+			margin-top: calc(-22vw + 1rem);
```
tsmethurst, line 174: did we have some troubles with this calc thing on chrome or that other browser? :thinking: I can't remember which it was that was causing problems

PR #1053, opened 2022-11-15
[docs] Both HTTP proxies and NAT can cause rate limiting issues

technomancy:
In the case of the clients being treated as the same IP when there is no HTTP proxy, I believe the only fix is to disable rate limiting altogether, since there seems to be no way to get the actual IP.

Also explain specifically that setting the rate limit to zero turns off rate limiting rather than actually setting the limit to zero, since this is not intuitive.


tsmethurst:
Ahh thank you, that's much better explained :)

PR #1054, opened 2022-11-15
docs: Add AUR Reference

jelly:
The AUR is the Arch unofficial repository, where users can upload PKGBUILD's for Arch Linux users to build a package.


tsmethurst:
Thanks!

PR #1059, opened 2022-11-16
[chore] reversion: use specific columns for updating user again

tsmethurst:
Prevents trying to set columns to null when doing updates to users via the cli commands


NyaaaWhatsUpDoc:
*uno reverse card* !

PR #1071, opened 2022-11-18
[chore] Little Dockerfile updates

tsmethurst:
Explicitly create user, storage dir, and indicate that storage should be mounted from a volume

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,6 @@
 # syntax=docker/dockerfile:1.3
+# Dockerfile reference: https://docs.docker.com/engine/reference/builder/
+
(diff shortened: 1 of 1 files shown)


metalmatze:
Is it going to work without a volume though? Just if someone wants to give it a try for a bit without a volume?

tsmethurst:
> Is it going to work without a volume though? Just if someone wants to give it a try for a bit without a volume?

yes :) the directory will still be created, it just won't contain anything

hnrd:
> Is it going to work without a volume though? Just if someone wants to give it a try for a bit without a volume?

docker will create an anonymous volume which will stick around after you delete the container. so if you do that often you'll have to clean up old volumes sooner or later

PR #1073, opened 2022-11-18
[performance] replace status query joins with separate calls in order to rely on caches

NyaaaWhatsUpDoc:


NyaaaWhatsUpDoc:
As pointed out by @tsmethurst , this may mean we can get read of the `status_to_emojis` table.

PR #1078, opened 2022-11-18
[chore] move caches to a separate State{} structure

NyaaaWhatsUpDoc:
This moves toward consolidating much of our means of dependency injection. It will also make it easier sharing resources across the codebase, and in the particular case of the caches, invalidating the necessary caches on particular changes.


internal/state/cache.go
```diff
@@ -0,0 +1,243 @@
...
+
+type Caches struct {
+	// GTS provides access to the collection of gtsmodel object caches.
+	GTS GTSCaches
+
+	// AP provides access to the collection of ActivityPub object caches.
+	AP APCaches
+
+	// prevent pass-by-value.
+	_ nocopy
```
tsmethurst, line 19: never seen this before... what's it do?

tsmethurst (commented):
looks like it's going in a good direction!

is it worth making state.State into an interface instead of just using the struct, so that in future we can use different kinds of state objects? (just thinking years in the future here, when we start messing about with remote state etc)

internal/state/cache.go
NyaaaWhatsUpDoc: nocopy when embedded will signal linter to show errors on pass-by-value of the parent struct. it's the same error that appears if you try pass-by-value a `sync.Mutex{}` :)

NyaaaWhatsUpDoc:
@tsmethurst I started to make the `State{}` itself an interface, but it complicates being able to be granular about initiating individual parts without having tonnes of interface methods. Instead what I went for was making each individual attached piece being an interface. The DB obviously already is, so I updated the caches to now be interfaces which could allow swapping them out for a multi-node compatible deployment in the future.

tsmethurst:
lmk when you want me to review this again kimbe :)

internal/db/bundb/status.go
```diff
@@ -26,36 +26,16 @@ import (
...
-	}, 1000)
-
-	// Set cache TTL and start sweep routine
-	s.cache.SetTTL(time.Minute*5, false)
-	s.cache.Start(time.Second * 10)
+	conn  *DBConn
+	state *state.State
 }
 
 func (s *statusDB) newStatusQ(status interface{}) *bun.SelectQuery {
```
tsmethurst, line 41: just a reminder to myself that we need to take the relations out of this at some point

internal/state/state.go
```diff
@@ -0,0 +1,31 @@
+package state
```
tsmethurst, line 1: license text

internal/state/state.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/db"
+)
+
+// State provides a means of dependency injection and sharing of resources
+// across different subpackages of the GoToSocial codebase. DO NOT assume
+// that any particular field will be initialized if you are accessing this
+// during initialization. A pointer to a State{} is often passed during
+// subpackage initialization, while the returned subpackage type will later
+// then be set and stored within the State{} itself.
+type State struct {
```
tsmethurst, line 14: :+1: nice and clean

tsmethurst (approved):
I had a few comments but nothing significant. This looks really great, and is good to go afaict :) Amazing work kimbe

internal/cache/util.go
```diff
@@ -0,0 +1,18 @@
...
+// nocopy when embedded will signal linter to
+// error on pass-by-value of parent struct.
+type nocopy struct{}
+
+func (*nocopy) Lock() {}
+
+func (*nocopy) Unlock() {}
+
+// tryUntil will attempt to call 'do' for 'count' attempts, before panicking with 'msg'.
+func tryUntil(msg string, count int, do func() bool) {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: It isn't, but this is just to make sure :)

A cache failing to start / stop silently would lead to some very difficult to diagnose issues i'm sure :p

cmd/gotosocial/action/admin/account/account.go
```diff
@@ -27,17 +27,24 @@ import (
...
+	var state state.State
+	state.Caches.Init()
+
+	dbConn, err := bundb.NewBunDBService(ctx, &state)
 	if err != nil {
 		return fmt.Errorf("error creating dbservice: %s", err)
 	}
 
+	// Set the state DB connection
+	state.DB = dbConn
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: The only issue is then, it feels a bit odd to say have a constructor like this that takes the `db` parameter, but then have later fields manually set like this. But then on the opposite hand, if we make state take as many parameters as it can on initialization, then the `state.New()` function basically becomes the constructor for the entire codebase instead of the server actions like we have. Basically very hard to maintain consistency, and doing it this way it means the state pointer can be passed around and having different fields initialized at different points :)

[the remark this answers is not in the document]
tsmethurst: OK, ack, that makes sense :)

Issue #929, opened 2022-10-28
[bug] First lookup of a remote account doesn't work, second one does


tsmethurst:
I've seen this in pinafore a few times. When looking up an account your instance doesn't know about yet, the first search returns 0 results, but the second search (literally just clicking the button again immediately) returns the expected account. Argh?

xmgz:
yes, it also happens when searching for single post (using full URL).

tsmethurst:
Seems to have been resolved by https://github.com/superseriousbusiness/gotosocial/pull/1082 :)

PR #1082, opened 2022-11-19
[chore] Tidy up some of the search logic

tsmethurst:
GetRemoteAccount and GetRemoteStatus are now renamed in order to signal that you *can* pass a request for a local status/account into them, they'll just return the status/account early + not do any "remote" lookups. (I think in future instead of having these functions attached to a federator, we can have them attached to some kind of account/status 'store' which can explicitly return either local or remote things, as appropriate.)

Now that faux-remote lookups are properly guarded against in those functions, the processor search logic has been tidied up to just call those functions directly, rather than doing db calls first (and then pointlessly repeating them again in some cases).

Rather than allowing all errors coming from those functions to go ignored, this PR also introduces a new `dereferencing.Error` type and different versions thereof (see `internal/federation/dereferencing/error.go`). These types can be checked on by callers so that they only log errors where appropriate. This is now implemented in the processor: errors aren't logged by searching when something is just not retrievable, but if there's a database or significant transport error, they will be logged and result in an internal server error to denote that something is definitely not right.

Right now, this better error checking + returning is just implemented for the GetStatus and GetAccount functions, because I ran out of spoons. In future, we should also implement the error checking for dereferencing media and instances and whatnot.

This should help us debug https://github.com/superseriousbusiness/gotosocial/issues/929 but it remains to be seen if it closes it.


NyaaaWhatsUpDoc:
@tsmethurst let me know once this is passing tests / ready and i'll take a look :)

tsmethurst:
note -- DO NOT MERGE THIS until https://github.com/superseriousbusiness/gotosocial/pull/1119 is in; we need to merge main into this branch once that's the case, so i can do further tweaks

tsmethurst:
@NyaaaWhatsUpDoc this is ready to go :)

internal/federation/dereferencing/account.go
```diff
@@ -135,11 +135,33 @@ func (d *deref) GetRemoteAccount(ctx context.Context, params GetRemoteAccountPar
...
+		a, dbErr := d.db.GetAccountByURI(ctx, uriString)
+		switch {
+		case dbErr == nil:
 			foundAccount = a
-		} else if dbErr != db.ErrNoEntries {
+		case errors.Is(dbErr, db.ErrNoEntries):
+			// no problem
+		default:
+			// a real error
 			err = fmt.Errorf("GetRemoteAccount: database error looking for account with uri %s: %s", uri, err)
```
NyaaaWhatsUpDoc, line 149: Should we not return early on a non-"not found" error?

internal/api/client/search/searchget_test.go
```diff
@@ -0,0 +1,344 @@
...
+	"github.com/stretchr/testify/suite"
+	"github.com/superseriousbusiness/gotosocial/internal/api/client/search"
+	apimodel "github.com/superseriousbusiness/gotosocial/internal/api/model"
+)
+
+type SearchGetTestSuite struct {
+	SearchStandardTestSuite
+}
+
+func (suite *SearchGetTestSuite) TestSearchRemoteAccountByURI() {
```
NyaaaWhatsUpDoc, line 38: There's *a lot* of repeated logic in these tests. Do you think something like a `func testSearch(query string, resolve bool) (*apimodel.SearchResult, error)` would help sismplify a lot of this?

internal/processing/search.go
```diff
@@ -78,12 +77,13 @@ func (p *processor) SearchGet(ctx context.Context, authed *oauth.Auth, search *a
 
 	if username, domain, err := util.ExtractNamestringParts(maybeNamestring); err == nil {
 		l.Debugf("search term %s is a mention, looking it up...", maybeNamestring)
```
NyaaaWhatsUpDoc, line 79: Is it worth changing these debug messages to instead be trace? Very nitpicky I know but it definitely feels like more a trace-level thing. But this is also very subjective :p

internal/processing/search.go
```diff
@@ -78,12 +77,13 @@ func (p *processor) SearchGet(ctx context.Context, authed *oauth.Auth, search *a
...
-		if foundAccount, err := p.searchAccountByMention(ctx, authed, username, domain, search.Resolve); err == nil && foundAccount != nil {
+		foundAccount, err := p.searchAccountByMention(ctx, authed, username, domain, search.Resolve)
+		if err != nil {
+			l.Debugf("error looking up account %s: %s", maybeNamestring, err)
```
NyaaaWhatsUpDoc, line 82: Should we be checking the error here and returning in some cases? e.g. when not a `errors.Is(err, db.ErrNoEntries)`

internal/processing/search.go
```diff
@@ -92,24 +92,46 @@ func (p *processor) SearchGet(ctx context.Context, authed *oauth.Auth, search *a
...
+				// don't attempt to resolve (ie., dereference) local accounts/statuses
+				resolve := search.Resolve
+				if uri.Host == config.GetHost() || uri.Host == config.GetAccountDomain() {
+					resolve = false
+				}
+
+				// check if it's a status or an account
+				foundStatus, err := p.searchStatusByURI(ctx, authed, uri, resolve)
+				if err != nil {
+					log.Debugf("error searching status by uri %s: %s", uri, err)
```
NyaaaWhatsUpDoc, line 108: Again for both this error and the one below should we be checking the error for possible early return?

internal/federation/dereferencing/account.go
tsmethurst: Yes! Thanks, well spotted. Fixed https://github.com/superseriousbusiness/gotosocial/pull/1082/commits/98c41cef3f6e717d9a79aa15a36962597f99fc75

internal/federation/dereferencing/account.go
```diff
@@ -135,11 +135,33 @@ func (d *deref) GetRemoteAccount(ctx context.Context, params GetRemoteAccountPar
...
+		if foundAccount == nil {
+			a, dbErr = d.db.GetAccountByURL(ctx, uriString)
+			switch {
+			case dbErr == nil:
+				foundAccount = a
+			case errors.Is(dbErr, db.ErrNoEntries):
+				// no problem
+			default:
+				// a real error
+				err = fmt.Errorf("GetRemoteAccount: database error looking for account with url %s: %s", uri, err)
```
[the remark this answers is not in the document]
tsmethurst: We don't need to break here since it's at the end of the parent switch clause, but i reworked it here for clarity because it was a bit obtuse: https://github.com/superseriousbusiness/gotosocial/pull/1082/commits/98c41cef3f6e717d9a79aa15a36962597f99fc75

internal/api/client/search/searchget_test.go
tsmethurst: yuss, good idea, did it here :) https://github.com/superseriousbusiness/gotosocial/pull/1082/commits/a538870c3239e1462eac2356f10f74e590b1db73

internal/processing/search.go
tsmethurst: Nah you're right, think i had these as debug just while I was testing, will change them to trace though

internal/processing/search.go
tsmethurst: fixed!

internal/processing/search.go
tsmethurst: fixed!

internal/processing/search.go
tsmethurst: fixed!

internal/federation/dereferencing/error.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/superseriousbusiness/gotosocial/internal/transport"
+)
+
+// Error aliases common error types that occur when doing dereferencing.
+type Error error
```
NyaaaWhatsUpDoc, line 29: You shouldn't need this error alias, you already have separate types below that you check for. 

(general note: Go idioms generally just prefer `error` as the return type instead of a custom one, then you check specifics based on the error interface type)

internal/federation/dereferencing/status.go
```diff
@@ -63,15 +63,15 @@ func (d *deref) EnrichRemoteStatus(ctx context.Context, username string, status
...
 	// try to get by URI first
-	status, err := d.db.GetStatusByURI(ctx, uriString)
-	if err != nil {
-		if !errors.Is(err, db.ErrNoEntries) {
+	status, dbErr := d.db.GetStatusByURI(ctx, uriString)
+	if dbErr != nil {
+		if !errors.Is(dbErr, db.ErrNoEntries) {
 			// real error
-			return nil, nil, fmt.Errorf("GetRemoteStatus: error during GetStatusByURI for %s: %s", uriString, err)
+			return nil, nil, newErrDB(fmt.Errorf("GetRemoteStatus: error during GetStatusByURI for %s: %w", uriString, dbErr))
```
NyaaaWhatsUpDoc, line 74: do we need to do the extra wrap before `newErrDB()` here? in doing so we've wrapped that `dbErr` twice in returning it :p

(same goes for similar situations with the new functions below)

tsmethurst: Ah probably not!

NyaaaWhatsUpDoc: though thinking about it, having the extra error information _is_ useful. we definitely have some inconsistent erroring across the codebase it would probably be worth having a ticket to tidy up codebase errors at some point and we can maybe address this then :p

PR #1086, opened 2022-11-20
[docs] Add first draft of FAQ

technomancy:
As discussed on the Matrix channel, I've added an FAQ for potential users so that we can try to set expectations to a little more realistic level. I feel that this should help reduce the amount of questions in the help channel around things which simply don't exist yet. When I was first setting up my server, I had some difficulty understanding which things I should expect to work, so I hope this helps.

If you have suggestions for further topics to cover, I can do my best to add them.

I know that debugging federation issues can be confusing; I had a friend who ran into this when he was setting up his GtS behind nginx, but I don't know the details, so I left it as a TODO.

Also I clarified some language about algorithms in the readme/index but if you feel that is too pedantic I can remove that change.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,6 +1,6 @@
 # Frequently Asked Questions
 
-- **Where's the user interface?** GoToSocial is just a bare server for the most part and is designed to be used thru external applications. [Pinafore](https://pinafore.social) and [Tusky](https://tusky.app/) are the best-supported, but anything that supports the Mastodon API should work, other than the features GoToSocial doesn't yet have.
+- **Where's the user interface?** GoToSocial is just a bare server for the most part and is designed to be used thru external applications. [Pinafore](https://pinafore.social) and [Tusky](https://tusky.app/) are the best-supported, but anything that supports the Mastodon API should work, other than the features GoToSocial doesn't yet have. Permalinks and profile pages are served directly thru GoToSocial as well as the admin panel, but most interaction goes thru the apps.
 
 - **What happened to the gifs?** While GoToSocial supports gifs, it doesn't support videos. This wouldn't be a big problem, except that Mastodon doesn't support gifs; it converts them into videos when they get uploaded. So if someone posts a gif from a Mastodon server, it won't be visible. At the time of this writing, the video will be dropped altogether, but [in the future there should be at least a placeholder link](https://github.com/superseriousbusiness/gotosocial/issues/765).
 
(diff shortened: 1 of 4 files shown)


docs/faq.md
```diff
@@ -0,0 +1,27 @@
...
+
+- **Why aren't my posts showing up on other servers?** First check the visibility as noted above. TODO: explain how to debug common federation issues
+
+- **Why am I getting frequent error responses?** GoToSocial is configured to use per-IP [rate limiting](https://docs.gotosocial.org/en/latest/api/ratelimiting/) by default, but in certain situations it can't accurately identify the remote IP and will treat all connections as coming from the same place. In those cases, the rate limiting needs to be disabled or reconfigured.
+
+- **How can I sign up for a server?** Right now the only way to create an account is by the server's admin to run a command directly on the server. A web-based signup flow is in the roadmap but not implemented yet.
+
+- **Why's it still in alpha?** Take a look at the [list of open bugs](https://github.com/superseriousbusiness/gotosocial/issues?q=is%3Aissue+is%3Aopen+label%3Abug) and the [roadmap](https://github.com/superseriousbusiness/gotosocial/blob/main/ROADMAP.md) for a more detailed rundown, but the main missing features at the time of this writing are:
```
ppetermann, line 15: backfilling seems a big deal thats missing too

docs/faq.md
```diff
@@ -0,0 +1,27 @@
+# Frequently Asked Questions
+
+- **Where's the user interface?** GoToSocial is just a bare server for the most part and is designed to be used thru external applications. [Pinafore](https://pinafore.social) and [Tusky](https://tusky.app/) are the best-supported, but anything that supports the Mastodon API should work, other than the features GoToSocial doesn't yet have.
```
tsmethurst, line 3: Would add a note here that while we don't have a user interface for making posts, we do have the web pages and the /settings panel.

docs/faq.md
tsmethurst: Good one. Maybe we can have one that's like:

**My instance is deployed and I'm logged in to a client but my timelines are empty, what's up there?** To see posts, you have to start following people! Once you've followed a few people and they've posted or boosted things, you'll start seeing them in your timelines. Right now GoToSocial doesn't have a way of 'backfilling' posts -- that is, fetching previous posts from other instances -- so you'll only see new posts of people you follow. If you want to interact with an older post of theirs, you can copy the link to the post from their web profile, and paste it in to your client's search bar.

mkdocs.yml
```diff
@@ -13,6 +13,7 @@ extra_css:
 
 nav:
   - "Home": "index.md"
+  - "FAQ": "faq.md"
```
tsmethurst, line 16: Can you also add a link to the FAQ document in the main README.md? :)

tsmethurst (commented):
Looks good so far, just some little things. Thanks a lot for making this document! As devs we kinda forget what people know/don't know about the software, so we're not the best at writing docs like this :P

technomancy:
> As devs we kinda forget what people know/don't know about the software, so we're not the best at writing docs like this :P

I maintain a programming language, and I know this exact feeling! That's why I took notes while I was setting up my instance, because I know once you get used to the codebase and what it's like to use, you lose the perspective that you need to understand the newbie experience.

I'll go thru and make the suggested changes. =)

technomancy:
OK, pushed the suggested changes. I linked to the FAQ from the readme under both "History and Status" as well as "Known Issues"; if you think it would be better somewhere else feel free to switch it around. Thanks.

tsmethurst:
Brilliant, thank you very much for this! :)

PR #1091, opened 2022-11-20
[bugfix] fix possible infinite loop on federated AP profile delete

NyaaaWhatsUpDoc:
Refactor federator account statuses delete to better catch errors, ensure next maxID is always set.


tsmethurst:
Looks good, thanks kimrade!

PR #1101, opened 2022-11-20
[chore] Enable more linters

Gusted:
- While reading the code, I noticed some unusual style of Go(read unnecessary labels), so I took the time to add some linters to improve the code quality of GtS.
- Enable gofumpt, this is a non-breaking change as currently it: removes double argument types(e.g. `func Foo(a int, b int)` -> `func Foo(a, b int)` and unnecessary empty lines.
- Enable gocritic, this is more breaking change, due to that the following changes have been made.
  - Unnecessary labels are removed.
  - Tries to simplify logic by removing unnecessary nesting.
  - Avoid shadowing builtin commands (in this case `delete`).
  - Add named results for some functions.
  - Use `http.NoBody` to signal there's no body instead of using `nil`.
  - Avoid a duplicated import.
  - Use pointers for huge structs.
  - Removed a `defer` in a loop(`defer`s aren't run after each run of the loop).
  - Avoid embedding `sync.Mutex`.
  - Simplify time comparison logic (`!foo.After` -> `foo.Before`).


NyaaaWhatsUpDoc:
I am going to prefix most of what I have to say here, with the main point that is: there isn't a single correct way to write Go. And the reason that a lot of the linters have been disabled is because they often get too opinionated.

Gusted:
I'm frustrated that my intentions came over like that. I would like to offer my apologies for that as that wasn't my intention.

This was an opportunistic PR to see what fits in GtS and what not, but I see that you would have preferred to have this discussion on Matrix first.

I do agree that code style is quite personal, which this PR certainly contained(see gofumpt) but there were also legitimately improved cases(e.g. time condition checks, reducing nesting, avoid duplicated import etc.) which I personally wouldn't classify as personal taste for code-style.

NyaaaWhatsUpDoc:
_Some_ of the changes were useful, yes. But the vast majority were coding style, and some cases were ignored for good reason.. But coming in with these sweeping changes without speaking to anybody, not just for this particular change but at all I am assuming, with this being your first PR... It is quite frustrating to see.

If you had spoken to any of the other core developers first, or if you were working on bugfixes and providing some of these changes alongside code where you noticed them, those would be very different situations.

tsmethurst:
Generally speaking I think linters often can get in the way and/or lead to wasted time discussing style. We arrived at the set of linters we have now quite purposefully, and it's a style that works pretty well for us 😊

I do think another PR that fixes some of the semantic issues you spotted would be very welcome though! If our linters can be tuned to catch only cases like that, rather than enforcing style just for the sake of it, that would also be welcome imo.

NyaaaWhatsUpDoc:
It might be worth having a ticket open to discuss linters in use. What to keep enabled, what to add, what they bring etc.

PR #1115, opened 2022-11-21
Enable / document non-buildx cross compilation for docker image

ForestJohnson:
I wanted to document the process I used to build GTS for my ARM-based Single-board-computer  

Yes,  I could have updated docker and installed/configured whatever buildx requires instead, but I chose to do this because we don't technically need buildx as go's cross compilation is 10/10 and needs no `qemu`  :smiling_imp: 

My pet peeve:  When there is an option that does not list all of its possible values. So I wanted to document those:
  -  `GOOS`/`GOARCH`  for `./build.sh`
  - and the  `--platform` argument in the `Dockerfile`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -330,7 +330,7 @@ If all goes according to plan, you should now have a number of multiple-architec
 
 ### Manually
 
-If you prefer a simple approach to building a Docker container, with fewer dependencies, you can also just build in the following way:
+If you prefer a simple approach to building a Docker container, with fewer dependencies (go-swagger, Node, Yarn), you can also just build in the following way:
 
 ```bash
 ./scripts/build.sh && docker buildx build -t superseriousbusiness/gotosocial:latest .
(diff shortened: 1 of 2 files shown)


tsmethurst:
I think this would be a perfect candidate for putting up on a blog somewhere, so we can link it in the docs as a kind of 'if you don't wanna install buildx to cross-compile, you can **do it this way**'. What do you think?

Dockerfile
```diff
@@ -1,6 +1,9 @@
 # syntax=docker/dockerfile:1.3
 # Dockerfile reference: https://docs.docker.com/engine/reference/builder/
 
+ARG BUILDPLATFORM
```
tsmethurst, line 4: Worth linking to here I think: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

So I guess what you're doing here is essentially providing these automatic global platform args (in a way) even when buildx is not being used, am i understand that right?

CONTRIBUTING.md
```diff
@@ -330,15 +330,24 @@ If all goes according to plan, you should now have a number of multiple-architec
...
 
-You don't need to install go-swagger, Node, or Yarn to build Docker images this way.
+
+If you want to build a docker image for a different CPU architechture without setting up buildx (for example for ARMv7 aka 32-bit ARM), you can use:
+
+```bash
+GOOS=linux GOARCH=arm ./scripts/build.sh && sudo docker build --build-arg BUILDPLATFORM=linux/amd64 --build-arg TARGETPLATFORM=linux/arm/v7 -t superseriousbusiness/gotosocial:latest .
+```
+
+See also: [exhaustive list of GOOS and GOARCH values](https://gist.github.com/lizkes/975ab2d1b5f9d5fdee5d3fa665bcfde6)
```
tsmethurst, line 348: cool!

CONTRIBUTING.md
```diff
@@ -330,15 +330,24 @@ If all goes according to plan, you should now have a number of multiple-architec
...
 ```
 
 The above command first builds the `gotosocial` binary, then invokes Docker buildx to build the container image.
 
-You don't need to install go-swagger, Node, or Yarn to build Docker images this way.
+
+If you want to build a docker image for a different CPU architechture without setting up buildx (for example for ARMv7 aka 32-bit ARM), you can use:
+
+```bash
+GOOS=linux GOARCH=arm ./scripts/build.sh && sudo docker build --build-arg BUILDPLATFORM=linux/amd64 --build-arg TARGETPLATFORM=linux/arm/v7 -t superseriousbusiness/gotosocial:latest .
```
tsmethurst, line 345: i'd drop the 'sudo' from this since we don't use it elsewhere in these example commands

tsmethurst (commented):
Cool :) Thanks for doing this. When I first read the PR description (which I thought was an issue, at first, sorry for the confusion) I thought this was going to be very complicated. Glad to see you found such a nice simple solution :) Happy to merge this once you've looked at the comments

tsmethurst:
good shit, thank you! :)

PR #1119, opened 2022-11-22
[feature/performance] Fail fast when doing remote transport calls inside incoming request contexts

tsmethurst:
Add a `Fastfail` context that can be passed in to external calls made by our httpclient transport.

The idea here is that sometimes we do outbound http requests inside an incoming request context (say when a user is doing a search for an account our instance doesn't know yet). In these cases, the caller should wrap the request passed to the dereferencing functions in a Fastfail context, which indicates to the transport logic that it should fail after the first attempt.

This PR prevents situations where the search fails but the transport keeps retrying, which makes the search request feel very slow and buggy. Now, fails after the first attempt in such situations will return immediately, making things feel a lot snappier for the client (even when the request fails -- at least they can try again immediately if they want).

### DIFF
--- a/internal/federation/authenticate.go
+++ b/internal/federation/authenticate.go
@@ -216,7 +216,7 @@ func (f *federator) AuthenticateFederatedRequest(ctx context.Context, requestedU
 		}
 
 		log.Tracef("proceeding with dereference for uncached public key %s", requestingPublicKeyID)
-		trans, err := f.transportController.NewTransportForUsername(ctx, requestedUsername)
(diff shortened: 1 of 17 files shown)


NyaaaWhatsUpDoc:
Have made some small tweaks to make the context key checking in-line with intended use, and updated to wrap fast-fail errors instead of adding another line of log entry :). Aside from these very small tweaks all was looking good! So once tests are passing will merge this

tsmethurst:
Sweet!

Issue #1129, opened 2022-11-23
[chore/frontend] Update user setting panel profile preview css


tsmethurst:
We broke the profile CSS a little bit, and now the user panel looks like this:

![Screenshot from 2022-11-23 15-05-45](https://user-images.githubusercontent.com/31960611/203566922-d22a3656-0e3d-4d72-9944-4f7079c95281.png)

So we kinda goofed it up. We need to update `web/source/settings/style.css` to bring it the profile preview in line with the changes we made earlier in `web/source/css/profile.css`

PR #1130, opened 2022-11-23
[frogend] fix profile preview css, use role info

f0x52:
fixes https://github.com/superseriousbusiness/gotosocial/issues/1129


tsmethurst:
nice one!!

PR #1133, opened 2022-11-24
Small copy improvements to binary installation doc

cimm:
Some small changes to the installation document that confused me a little when installing GoToSocial.

1. During my first setup, I was a little confused as to why I needed to promote my user to an admin user while it looked like I created an admin user a few steps earlier. Reading the document again I noticed it says "similar command" so the documentation is correct, but I misread it. I slightly changed the sentence in the hope to clarify that part a bit.
2. At the start of the installation a /gotosocial root directory is created without sudo, later when adding the systemd service it does mention the use of sudo. Wouldn't it be better to always/never mention sudo? I removed the sudo use in the second part of the document.
3. Used correct casing for GoToSocial.

I hope this clarifies the installation a little bit. Feel free to reject the pull request if you disagree.

### DIFF
--- a/docs/installation_guide/binary.md
+++ b/docs/installation_guide/binary.md
@@ -76,7 +76,7 @@ Note that for this example we're assuming that we're allowed to run on port 443 
 
 You can use the GoToSocial binary to also create, confirm, and promote your user account.
 
-Run the following command to create a new account:
+Run the following command to create a new admin account:
 
 ```bash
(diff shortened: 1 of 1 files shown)


cimm:
Oops, I got this all wrong. The first user created with `./config.yaml admin account create` does _not_ create an admin user but a standard user so you need to promote the user. The "admin" part in the "admin account create" got me confused.

PR #1140, opened 2022-11-24
[bugfix] Change emailVerified to true for admin account create

technomancy:
The NewSignup method was already being called with requireApproval=false, but it had emailVerified=false as well, which meant that it was necessary to use the `admin account confirm` command to verify the email before the newly-created user could log in.

I think that was probably an oversight; effectively it did require approval anyway. Changing emailVerified to true allows you to just create the account and log in immediately, streamlining the setup and reducing the opportunity for manual error to sneak in.

Also updated the docs to remove the mention of needing to confirm new accounts. However, I've left the confirmation command alone because I think once we have web signups, it will be needed in that context.

It's possible I've overlooked something about the `admin account confirm` and it does in fact still make sense even for accounts created thru the CLI; if so please feel free to close this!

### DIFF
--- a/cmd/gotosocial/action/admin/account/account.go
+++ b/cmd/gotosocial/action/admin/account/account.go
@@ -78,7 +78,7 @@ var Create action.GTSAction = func(ctx context.Context) error {
 		return err
 	}
 
-	_, err = dbConn.NewSignup(ctx, username, "", false, email, password, nil, "", "", false, false)
+	_, err = dbConn.NewSignup(ctx, username, "", false, email, password, nil, "", "", true, false)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


tsmethurst:
Looks good, thank you :)

technomancy:
I just realized I may have been a little hasty here; this is going to change the docs at https://docs.gotosocial.org/en/latest/ which means that people installing 0.5.2 will run into a situation where they will need to confirm but the docs don't explain it.

So unless we're like ... right on the cusp of releasing 0.6.0, I should probably add a note about the fact that confirm is necessary on some versions and not others.

Sorry I missed this earlier!

PR #1142, opened 2022-11-25
[packaging] embed assets in binary + use as fallback when they're missing on the filesystem

noerw:
Embed both assets and templates in the executable to simplify packaging and installation/updates even further.
It still tries to read files from disk (for ease of customization), but falls back to embedded files.

###  TODO
The functionality as described is working. For a meaningful change work is still to be done:
- [x] implement hybrid FS
- [x] use hybrid FS for remaining asset types (default avatars)
- [x] change builds to not bundle assets
- [x] add cli to extract assets for customization (analogous to [`gitea embedded`](https://docs.gitea.io/en-us/cmd-embedded/))
- [x] adapt documentation
- [ ] add e2e tests with different configurations (eg check landing page response on `GTS_WEB_TEMPLATE_BASE_DIR=/dev/null ./gotosocial testrig start`)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,15 @@
-# syntax=docker/dockerfile:1.3
+# syntax=docker/dockerfile:1.4
 # Dockerfile reference: https://docs.docker.com/engine/reference/builder/
 
 # stage 1: generate up-to-date swagger.yaml to put in the final container
 FROM --platform=${BUILDPLATFORM} quay.io/goswagger/swagger:v0.30.0 AS swagger
 
-COPY go.mod /go/src/github.com/superseriousbusiness/gotosocial/go.mod
-COPY go.sum /go/src/github.com/superseriousbusiness/gotosocial/go.sum
-COPY cmd /go/src/github.com/superseriousbusiness/gotosocial/cmd
-COPY internal /go/src/github.com/superseriousbusiness/gotosocial/internal
 WORKDIR /go/src/github.com/superseriousbusiness/gotosocial
-RUN swagger generate spec -o /go/src/github.com/superseriousbusiness/gotosocial/swagger.yaml --scan-models
+COPY go.mod go.sum ./
+COPY cmd ./
+COPY internal ./
+RUN swagger generate spec -o ./swagger.yaml --scan-models --exclude-deps
+
 
 # stage 2: generate the web/assets/dist bundles
 FROM --platform=${BUILDPLATFORM} node:16.15.1-alpine3.15 AS bundler
@@ -19,7 +19,25 @@ RUN yarn install --cwd web/source && \
     BUDO_BUILD=1 node web/source  && \
     rm -r web/source
 
-# stage 3: build the executor container
+
+# stage 3: build the application
+FROM --platform=${BUILDPLATFORM} golang:1.19-alpine3.16 AS gobuild
+# pre-copy/cache go.mod for pre-downloading dependencies and only redownloading them in subsequent builds if they change
+WORKDIR /go/src/github.com/superseriousbusiness/gotosocial
+COPY go.mod go.sum vendor ./
+RUN go mod download && go mod verify
+
+
+COPY . .
+# we need the built assets, so they can be bundled into the build.
+COPY --from=bundler web ./
+COPY --from=swagger /go/src/github.com/superseriousbusiness/gotosocial/swagger.yaml ./web/assets/
+
+ARG VERSION=SNAPSHOT
+RUN CGO_ENABLED=0 go build -v -tags=netgo,osusergo,static_build -ldflags='-s -w -extldflags -static -X main.Version=${VERSION}' -o gotosocial github.com/superseriousbusiness/gotosocial/cmd/gotosocial
+
+
+# stage 4: use the executor image which will copy in the remaining assets
 FROM --platform=${TARGETPLATFORM} alpine:3.15.4 as executor
 
 # switch to non-root user:group for GtS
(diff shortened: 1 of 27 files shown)


internal/email/noopsender.go
```diff
@@ -20,20 +20,18 @@ package email
 
 import (
 	"bytes"
-	"text/template"
+	"html/template"
```
noerw, line 23: For email templates I switched from text/template to html/template. If that's bad lmk, but it seems right to me?

internal/web/assets.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/superseriousbusiness/gotosocial/internal/config"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
+	"github.com/superseriousbusiness/gotosocial/web"
 )
 
 type fileSystem struct {
```
noerw, line 34: The previous solution had custom handling about directory listings with an index.html (see below).
My changes don't keep this behaviour, as I don't get exactly what this is about. Could someone clarify?

internal/router/templates/functions.go
```diff
@@ -0,0 +1,149 @@
...
+func emojify(emojis []model.Emoji, inputText template.HTML) template.HTML {
+	out := text.Emojify(emojis, string(inputText))
+
+	/* #nosec G203 */
+	// (this is escaped above)
+	return template.HTML(out)
+}
+
+// EngineTemplateFuncs is map of names usable in templates to call go functions
+var EngineTemplateFuncs = template.FuncMap{
```
noerw, line 139: I refactored this into its own module, as templates are used both within the gin engine as well as in `internal/email`. Apart from changing this to be a variable declaration, this file is unchanged

tsmethurst:
Hmm, do we really need this? What are the advantages of it, since we always package the binary in a tar along with the assets?

noerw:
My main motivation was to simplify the update process, so I don't have to move assets in place separately, being careful not to overwrite some local changes in the process.
One of GtS features to me is having a comparably very easy setup, and I see value in optimizing that property further.
The packaging process also gets simplified + more portable, idk if that's relevant to you ;)

EchedeyLR:
I like this. Having to upgrade the assets directory recursively over local changes make it prone to fail (I have some custom changes for a friend) and this makes it easier for me.

tsmethurst:
> so I don't have to move assets in place separately, being careful not to overwrite some local changes in the process

mm fair enough yeah, that use case makes sense...

Dockerfile
```diff
@@ -39,9 +39,5 @@ WORKDIR "/gotosocial"
 # copy the dist binary created by goreleaser or build.sh
 COPY --chown=1000:1000 gotosocial /gotosocial/gotosocial
 
-# copy over the web directories with templates, assets etc
-COPY --chown=1000:1000 --from=bundler web /gotosocial/web
-COPY --chown=1000:1000 --from=swagger /go/src/github.com/superseriousbusiness/gotosocial/swagger.yaml web/assets/swagger.yaml
```
noerw, line 44: Damn, non-goreleaser docker builds break with this, because the executable embedding the assets is built *before* the assets are made.

The current dockerfile is a weird mix of copying the precompiled binary, and building assets during image building selfcontained anyway, so to fix this I'd suggest to separate the gorelease / standalone build contexts:
Have a separate dockerfile used by goreleaser that only defines the executor stage. Then have a default dockerfile that is truly selfcontained, i.e. including the `go build` part.

NyaaaWhatsUpDoc:
My only hesitation to this is that we already have a very large binary size. What's the size like _after_ embedding these assets?

noerw:
@NyaaaWhatsUpDoc that's right, go doesn't produce smol binaries D:
it's zero-sum, the difference is 1:1 the asset size when stored as files:
```
$ ls -lh gts*
-rwxr-xr-x 1 foo foo 54M 25. Nov 18:04 gts
-rwxr-xr-x 1 foo foo 57M 25. Nov 17:06 gts_embedded
-rw-r--r-- 1 foo foo 15M 25. Nov 18:04 gts.tar.gz
-rw-r--r-- 1 foo foo 17M 25. Nov 18:03 gts_embedd.tar.gz

$ du -hs web/{assets,template}
3,4M    web/assets
76K     web/template
```
as it's not a frontend focused project, I assume web/assets won't grow that much, but idk

tsmethurst:
> go doesn't produce smol binaries

well, normally it does, the issue in GtS's case is the large amount of generated code from go-fed (i believe) -- which is something we'll look to fix in *waves hands vaguely* the future

tsmethurst:
I'm gonna convert this to a draft btw, since it's still in progress :)

Dockerfile
EchedeyLR: Changing stages fix this instead of separated Dockerfiles?

noerw: I'm not sure I understand what you mean. Do you mean to have a selfcontained build defined in Dockerfile and to just accept doing the build twice for docker and binary releases?

For now  implemented it with 2 docker files in 5278c379494803eb669085cee65a702ac917c5e6

noerw:
Ok, last force push for now ;) From my side this is not WIP anymore, until I get guidance about the above and below questions.

- Do you think  the `no_embedded_assets` build flag I added is actually needed?

- The tests are failing now because responses now include avatar urls, where they didn't before. I assume this is because of [this change](https://github.com/superseriousbusiness/gotosocial/pull/1142/commits/1b8299573c564324ff5f134c8b5dc1753e0e78f4#diff-3f7e21e3475361a5d11e9b51d0f3a536f815f9cc908a7c72c739611a053f5102L47-L48)
The new logic now works within the test environment but didn't before I think? I'm not entirely sure if that's what is going on.

- I'm not sure how to go about adding e2e tests to make sure the frontend still works without an existing `web-assets-base-dir`, as there is no e2e suite so far, and I lack experience setting something like this up. Would unit tests that make sure access to some list of templates is working be a suitable replacement?

tsmethurst:
Heya! Sorry for keeping you waiting so long, busy with lots of release, admin, project management stuff yadda yadda. It's been wild recently.

So! I talked the embedding idea over with the other main devs @NyaaaWhatsUpDoc and @f0x52 a bunch. We agreed that the situation where users have to diff assets to do upgrades if they have custom changes is definitely not ideal, and that's really the main thing that this PR aims to fix right? (Assuming I understand your comment here correctly -- https://github.com/superseriousbusiness/gotosocial/pull/1142#issuecomment-1327311497.)

The problem with the embedding solution as it stands is that it seems to introduce quite a bit of complexity in terms of packaging, and it also makes the code and location of the assets more difficult to reason about (both for coders and for most users i imagine). Unfortunately, it also doesn't seem to solve the diffing issue with custom assets, since there's still a script included to do it. While embedding is very cool, and having a single-binary with no other files is *also* very cool to think about, we already in essence have something like that with the tar and the docker container.

I think it might be a good idea to close this PR *for now* but we ought to make an issue for updating custom installs, so that we can try to find a process that is simple and works well. This PR and the discussion really highlighted the need for that. One possible solution that came up in our discussion was having two separate assets folders: one essentially read-only and one 'custom' folder that's loaded from first if it's present. Functionally similar to embedding, but without the extra packaging complication... But we can discuss that on an issue.

So yes! The work you put into this is really appreciated, but I think we need to go a bit slower and think through other possible solutions before merging this. 

What do you think?

PR #1145, opened 2022-11-25
Add support for a Request ID to correlate things

daenney:
I was looking to add some basic request tracing into GTS. Though the logs are useful, it's sometimes hard to understand which request caused what additional log lines to be emitted, queries that were run etc.

I've managed to figure out how to plumb a request ID through the API endpoints, so when an error is returned you now get the request ID and the logs show things like:

```
timestamp="25/11/2022 14:24:42.781" func=router.loggingMiddleware.func1 level=INFO latency="1.038625ms" clientIP="127.0.0.1" userAgent="Mozilla/5.0 (X11; Linux x86_64; rv:107.0) Gecko/20100101 Firefox/107.0" method="GET" statusCode=404 path="/assets/dist/base.css" requestid="1952bc1d-2629-4b91-a097-dac1ef364635" msg="Not Found: wrote 908B"
```

What I would like to do now is to ensure that request ID is passed through in the rest of the code base. This would ensure log lines that are emitted for DB queries (for example I'd like to have the `requestid` in the TRACE lines), or any outgoing requests, have the same `requestid` attached to them if they were triggered by an incoming HTTP request. If that were to happen and you ingest your logs somewhere centrally you'd have a much easier time understanding what GTS is doing and hopefully improve the debugging experience.

Looking at the code base, it seems most other functions call `log.WithField(s)` directly and set their fields. I can't really figure out an entry point that would allow me to universally set the `requestid` from the HTTP request and pass it through. I'm not sure if that's currently possible, or if there's some other way to do this?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/disintegration/imaging v1.6.2
 	github.com/gin-contrib/cors v1.4.0
 	github.com/gin-contrib/gzip v0.0.6
+	github.com/gin-contrib/requestid v0.0.6
 	github.com/gin-contrib/sessions v0.0.5
 	github.com/gin-gonic/gin v1.8.1
 	github.com/go-fed/httpsig v1.1.0
--- a/go.sum
+++ b/go.sum
@@ -200,6 +200,8 @@ github.com/gin-contrib/cors v1.4.0 h1:oJ6gwtUl3lqV0WEIwM/LxPF1QZ5qe2lGWdY2+bz7y0
 github.com/gin-contrib/cors v1.4.0/go.mod h1:bs9pNM0x/UsmHPBWT2xZz9ROh8xYjYkiURUfmBoMlcs=
 github.com/gin-contrib/gzip v0.0.6 h1:NjcunTcGAj5CO1gn4N8jHOSIeRFHIbn51z6K+xaN4d4=
 github.com/gin-contrib/gzip v0.0.6/go.mod h1:QOJlmV2xmayAjkNS2Y8NQsMneuRShOU/kjovCXNuzzk=
+github.com/gin-contrib/requestid v0.0.6 h1:mGcxTnHQ45F6QU5HQRgQUDsAfHprD3P7g2uZ4cSZo9o=
+github.com/gin-contrib/requestid v0.0.6/go.mod h1:9i4vKATX/CdggbkY252dPVasgVucy/ggBeELXuQztm4=
 github.com/gin-contrib/sessions v0.0.5 h1:CATtfHmLMQrMNpJRgzjWXD7worTh7g7ritsQfmF+0jE=
 github.com/gin-contrib/sessions v0.0.5/go.mod h1:vYAuaUPqie3WUSsft6HUlCjlwwoJQs97miaG2+7neKY=
 github.com/gin-contrib/sse v0.1.0 h1:Y/yl/+YNO8GZSjAhjMsSuLt29uWRFHdHYUb5lYOV9qE=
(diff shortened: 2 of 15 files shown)


tsmethurst:
> I can't really figure out an entry point that would allow me to universally set the requestid from the HTTP request and pass it through

i'll have a proper look + get back to you about this after we've released 0.6.0, since that'll be a good moment for looking at changes like this

but my initial inkling would be to set it on the `ctx` that's passed almost everywhere

daenney:
> [edit] Also, thanks for opening this :) I was thinking something similar a couple months ago about tagging requests with IDs so we can correlate them, cool to see someone else was thinking the same thing too

My pleasure! It's handy to have this thing plumbed through. I figure I'd start with requestid like this. Once we have that figured out I might take a look at getting OpenTelemetry Tracing support wired up to get really fancy.

NyaaaWhatsUpDoc:
> > [edit] Also, thanks for opening this :) I was thinking something similar a couple months ago about tagging requests with IDs so we can correlate them, cool to see someone else was thinking the same thing too
> 
> My pleasure! It's handy to have this thing plumbed through. I figure I'd start with requestid like this. Once we have that figured out I might take a look at getting OpenTelemetry Tracing support wired up to get really fancy.

I think it'd be worth laying all our tracing options on the table and seeing what best fits our bill. My biggest concern is overhead since we want this server to (ideally) be able to run on a Pi Zero

NyaaaWhatsUpDoc:
My plan for the request ID being passed around the codebase was to have log entries accept a context, and then allow setting a log hook which we can set to pull out request ID values whenever they're set

daenney:
The reason I was looking at something like a request ID is because plumbing it through effectively gives you a very basic form of tracing. You still have to correlate things from the logs together yourself, but having the ID helps tremendously.

Regardless of potentially doing something more, I think this is useful one way or another and having it available in the logs is worth it.

If you do want to take it one step further though, OpenTelemetry tracing support is magic once you try it. Integrating it is pretty simple since it hops onto the `ctx` that can be passed around through everything.

```go
newCtx, span := otel.Tracer(name).Start(ctx, "Name it something")
defer span.End()
```

(`ctx` being the normal `context.Context`)

In every subsequent function, you'd effectively do the same, adding a span. Since it's added to a context with an existing span, you end up building a tree of spans. You can set additional attributes on span's, like your request ID too or anything else that may be interesting to understand what's going on.

There's a more complete example at https://opentelemetry.io/docs/instrumentation/go/ if you're curious.

The dependency tree is really small, so from that point of view I doubt it would impact much. It's also designed to have as little performance impact as possible, since the idea is you should be able to use and run it anywhere. I would also suggest that if this is integrated it should be behind a flag that switches to the NoopTracer if not enabled.

Bun also has OTEL support wired through it, so you'd get insight all the way into the ORM too: https://github.com/uptrace/bun/tree/master/extra/bunotel. Theoretically at least this could be added to activity too.

You can dump traces to a file, or to an open source solution like Jaeger if you want to host yourself or a commercial provider like Honeycomb, Datadog, Grafana Cloud etc.

tsmethurst:
So how would this look if we did it with open telemetry (which seems like a good idea btw): i guess we would do something like:

```go
ctx, span := otel.Tracer(name).Start(c.Request.Context(), "GET /api/v1/statuses/01GKKTZSTG59PNBM4G3RRSPM4W")
defer span.End()
```

So, starting the context when the http request starts. Is that about right?

tsmethurst:
Oh, look at that: https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gin-gonic/gin

tsmethurst:
@daenney is it OK with you if I close this PR for now, and instead we can move discussion to this opentelemetry issue i created? https://github.com/superseriousbusiness/gotosocial/issues/1230

daenney:
@tsmethurst Yup yup!

Issue #860, opened 2022-09-28
[feature] Clean up orphaned media from storage directory


tsmethurst:
Right now I'm in a situation on goblin technology (which is admittedly completely my own fault), where I did a database export/import without deleting my storage folder in between. Now, I've got a storage folder that's full of a combination of attachments with database entries (no problem, they can be cleaned), and attachments with no database entries. This latter is a problem because the cleaning logic has no way of removing them. It might not be a bad idea to implement some kind of storage cleaning thingy that can remove orphaned media from inside the storage folders.

NyaaaWhatsUpDoc:
KVStore has an iterator logic that should allow you to step through all the "keys" in the media folder, and for each you can check for a DB entry and build a list of keys that need to be removed.

tsmethurst:
hell yeah, nice!

in the meantime I did this very hacky command to remove files older than 5 days, but I don't recommend anyone else reading this to do this:

```
for f in $(find storage/* -mtime +5 | grep -v certs | grep -v store\.lock | egrep 'storage/.*/.*/.*/.*\..*'); do rm $f; done
```

NyaaaWhatsUpDoc:
This might be easier as well if we pull in v2 of `go-store` which now implements S3 support, so we can use the `Iterate()` function on either storage backend

gw1urf:
Just a quick drive-by comment. My "go" skills aren't good enough to tell by perusing the code (I did have a look), but does removing the last file in a directory cause the directory to be removed (and so on up the tree)? With directories taking 4Kb on ext4 (for example), a few tens of thousands of empty directories starts to add up.

NyaaaWhatsUpDoc:
We did previously support cleaning up directories in this manner (as the underlying `storage.DiskStorage{}` has a `.Clean()` functions which does this), but it is temporarily disabled as other users were having issues putting their storage directory on the root of their drive (and so perms issues trying to wipe`lost+found` lol).

PR #1146, opened 2022-11-25
[feature] Add `admin media prune orphaned` CLI command

tsmethurst:
This PR adds a function to the media manager for pruning orphaned media, where 'orphaned' means media in storage that doesn't have a corresponding database entry.

It also adds a CLI command for invoking that function, which can be run in both dry-run mode (default) or for-real mode.

Results from goblin.technology:

```text
before
---
du -sh storage
9.6G	storage
---

timestamp="25/11/2022 14:05:54.831" func=prune.glob..func1 level=INFO msg=" 5155 stored items were orphaned and pruned"

after
---
du -sh storage
9.2G	storage
---
```

closes https://github.com/superseriousbusiness/gotosocial/issues/860


internal/regexes/regexes.go
```diff
@@ -140,6 +140,13 @@ var (
 	// BlockPath parses a path that validates and captures the username part and the ulid part
 	// from eg /users/example_username/blocks/01F7XT5JZW1WMVSW1KADS8PVDH
 	BlockPath = regexp.MustCompile(blockPath)
+
+	filePath = fmt.Sprintf(`^(%s)/([a-z]+)/([a-z]+)/(%s)\.([a-z]+)$`, ulid, ulid)
```
NyaaaWhatsUpDoc, line 144: i don't think you need the capture groups in here (the regular parentheses)

internal/media/pruneorphaned.go
```diff
@@ -0,0 +1,127 @@
...
+	// make sure we have some keys, and also advance
+	// the iterator to the first non-empty key
+	if !iterator.Next() {
+		return 0, nil
+	}
+
+	// for each key in the iterator, check if entry is orphaned
+	log.Info("got %d orphaned pruning candidates, checking for orphaned status, please wait...")
+	var checkedKeys int
+	orphanedKeys := make([]string, 0, matchCount)
```
NyaaaWhatsUpDoc, line 62: ooh i like this use of the match function to build up an approximate slice allocation count :D

tsmethurst: thought you might ;)

PR #1149, opened 2022-11-25
[docs] Explain that before 0.6.0, account confirm was necessary.

technomancy:
Since the documentation site only shows the latest version of the docs, we need the docs to explain how to use the latest stable release, not just the latest git version.

Without this I think #1140 will cause some confusion.


tsmethurst:
Ah, good thinking :)

PR #1150, opened 2022-11-25
[bugfix]: Fix IPv6 validation

daenney:
The current code considers ff00::/8 valid, but contrary to the comment that's not the global unicast range. ff-prefixes in IPv6 denote multicast.

This adapts the code to take the same approach as IPv4, explicitly blacklisting reserved internal/private ranges.

### DIFF
--- a/internal/netutil/validate.go
+++ b/internal/netutil/validate.go
@@ -33,7 +33,7 @@ var (
 		netip.MustParsePrefix("fec0::/10"),     // Site-local, deprecated
 	}
 
-	// IPvReserved contains IPv4 reserved IP prefixes.
+	// IPv4Reserved contains IPv4 reserved IP prefixes.
 	IPv4Reserved = [...]netip.Prefix{
 		netip.MustParsePrefix("0.0.0.0/8"),       // Current network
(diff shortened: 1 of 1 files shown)


daenney:
An alternative would be to whitelist the allocated ranges from https://www.iana.org/assignments/ipv6-unicast-address-assignments/ipv6-unicast-address-assignments.xhtml instead.

daenney:
If I'm reading the code right and seeing how this is always used through Sanitize on outgoing requests, then I think most outgoing communication with IPv6-enabled instances might be failing.

tsmethurst:
I will defer to @NyaaaWhatsUpDoc on this one since she authored the sanitize code :)

daenney:
I double-checked this in the Go Playground with an IPv6 address of one of my boxes: https://go.dev/play/p/X-VMfnKuQkT.

That results in `false`, whereas it's very much a valid IPv6 unicast address.

NyaaaWhatsUpDoc:
Yes that is very much an oops on my behalf :')

daenney:
Thanks for checking and merging this so fast!

PR #1152, opened 2022-11-26
Additional IP range validations

daenney:
Commits are individually reviewable.

Since I was looking at this code yesterday I decided to take another pass through it:
* Added a check on the network to ensure it's always TCP
* Added IPv4-mapped range to v6 blocklist and add test for it. The behaviour is different between net and netip so it's useful to test and ensure we don't break this at some point
* I've aligned the reserved blocks with the respective IANA Special Purpose registries for v4 and v6
* Added some tests to ensure things are blocked especially for all the magical `/32`'s on v4

### DIFF
--- a/internal/httpclient/sanitizer.go
+++ b/internal/httpclient/sanitizer.go
@@ -38,6 +38,10 @@ func (s *sanitizer) Sanitize(ntwrk, addr string, _ syscall.RawConn) error {
 		return err
 	}
 
+	if !(ntwrk == "tcp4" || ntwrk == "tcp6") {
+		return ErrInvalidNetwork
+	}
+
 	// Seperate the IP
 	ip := ipport.Addr()
 
--- a/internal/httpclient/client.go
+++ b/internal/httpclient/client.go
@@ -36,6 +36,9 @@ import (
 // ErrInvalidRequest is returned if a given HTTP request is invalid and cannot be performed.
 var ErrInvalidRequest = errors.New("invalid http request")
 
+// ErrInvalidNetwork is returned if the request would not be performed over TCP
+var ErrInvalidNetwork = errors.New("invalid network type")
+
 // ErrReservedAddr is returned if a dialed address resolves to an IP within a blocked or reserved net.
 var ErrReservedAddr = errors.New("dial within blocked / reserved IP range")
 
--- a/internal/netutil/validate.go
+++ b/internal/netutil/validate.go
@@ -24,13 +24,30 @@ import (
 
 var (
 	// IPv6Reserved contains IPv6 reserved IP prefixes.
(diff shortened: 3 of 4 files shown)


daenney:
This test feels a little redundant now: https://github.com/superseriousbusiness/gotosocial/blob/e6cd81babc712979f72987466ce98154d43dd399/internal/httpclient/client_test.go#L141

But it's also a little underspecified given what it tests for: https://github.com/superseriousbusiness/gotosocial/blob/e6cd81babc712979f72987466ce98154d43dd399/internal/httpclient/client_test.go#L33-L43

I'm not sure what to do about that one, because retaining at least one test to ensure the Sanitizer is properly hooked up is valuable.

internal/httpclient/sanitizer.go
```diff
@@ -38,6 +38,10 @@ func (s *sanitizer) Sanitize(ntwrk, addr string, _ syscall.RawConn) error {
 		return err
 	}
 
+	if !(ntwrk == "tcp4" || ntwrk == "tcp6") {
+		return ErrInvalidNetwork
+	}
+
 	// Seperate the IP
 	ip := ipport.Addr()
```
daenney, line 46: One thing we could do here is also check if the port is `80` or `443`. That should be gated behind the development flag I'm guessing, but for production I don't think it could ever be anything other than those 2 ports?

NyaaaWhatsUpDoc: In case anyone is hosting with an odd setup I'd probably leave the port check out. The resolved IP checks are pretty exhaustive :)

NyaaaWhatsUpDoc:
> This test feels a little redundant now:
> 
> https://github.com/superseriousbusiness/gotosocial/blob/e6cd81babc712979f72987466ce98154d43dd399/internal/httpclient/client_test.go#L141
> 
> But it's also a little underspecified given what it tests for:
> 
> https://github.com/superseriousbusiness/gotosocial/blob/e6cd81babc712979f72987466ce98154d43dd399/internal/httpclient/client_test.go#L33-L43
> 
> I'm not sure what to do about that one, because retaining at least one test to ensure the Sanitizer is properly hooked up is valuable.

Leave it for now. As you say it checks the sanitizer is at least hooked up correctly :)

NyaaaWhatsUpDoc:
If you're happy with everything here @daenney I'll merge this. Looks good to me, and thank you for the contribution!

daenney:
Good to go from my side!

(also wow you're fast, I was not expecting a review at this time on a weekend)

NyaaaWhatsUpDoc:
> Good to go from my side!
> 
> (also wow you're fast, I was not expecting a review at this time on a weekend)

I work on this project outside my main job so more often than not you'll probably find me working at odd hours :'). Also your code is well written and easily reviewable!

PR #1153, opened 2022-11-26
Tweak the User-Agent a bit

daenney:
I've tried to tweak the User-Agent a bit to better match my understanding/interpretation of RFC 7231. But in the real world nobody really seems to agree on how these things should be formatted and what they contain so idk if this change is worth it.

This also flips it to use the account domain instead of the host domain, so that when you're looking at the header you can better identify the source of the request.

### DIFF
--- a/internal/transport/controller.go
+++ b/internal/transport/controller.go
@@ -60,6 +60,7 @@ type controller struct {
 func NewController(db db.DB, federatingDB federatingdb.DB, clock pub.Clock, client pub.HttpClient) Controller {
 	applicationName := config.GetApplicationName()
 	host := config.GetAccountDomain()
+	proto := config.GetProtocol()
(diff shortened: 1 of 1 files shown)


NyaaaWhatsUpDoc:
Is it worth having the github url in it? i just worry it makes the UA a bit too long.

daenney:
> Is it worth having the github url in it? i just worry it makes the UA a bit too long.

Ya, I've been pondering about that too. It's getting a little extreme. Lets drop that.

daenney:
There, that should be better.

daenney:
@NyaaaWhatsUpDoc  I just realised I'm an idiot. We do want to use Host and not the AccountDomain, since Host would be the actual address the server runs on.

I have no idea how I read this the wrong way around.

PR #1154, opened 2022-11-26
Set User-Agent header in transport

daenney:
Currently requests set their own User-Agent. Set it in the transport instead, unless it's been explicitly set on the request. This retains the flexibility of tweaking the UA if necessary, without having to remember to set it.

### DIFF
--- a/internal/transport/transport.go
+++ b/internal/transport/transport.go
@@ -134,6 +134,8 @@ func (t *transport) do(r *http.Request, signer func(*http.Request) error, retryO
 		{"url", r.URL.String()},
 	}...)
 
+	r.Header.Set("User-Agent", t.controller.userAgent)
+
 	for i := 0; i < maxRetries; i++ {
 		// Reset signing header fields
 		now := t.controller.clock.Now().UTC()
--- a/internal/transport/deliver.go
+++ b/internal/transport/deliver.go
@@ -82,7 +82,6 @@ func (t *transport) Deliver(ctx context.Context, b []byte, to *url.URL) error {
 
 	req.Header.Add("Content-Type", string(api.AppActivityLDJSON))
 	req.Header.Add("Accept-Charset", "utf-8")
-	req.Header.Add("User-Agent", t.controller.userAgent)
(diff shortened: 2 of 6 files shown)


daenney:
One thing I noticed is that the code used to do `req.Headers.Add()` for the UA instead of `.Set()`. This could theoretically lead to the UA header being multi-valued, but I don't think that header is allowed/specified as list-based or multi-valued. It's why I opted for checking if the header is unset in `req` and going with `.Set()` to make that more explicit. Given we're always the ones who build the `req` this doesn't make a difference in practice.

internal/transport/transport.go
```diff
@@ -134,6 +134,10 @@ func (t *transport) do(r *http.Request, signer func(*http.Request) error, retryO
 		{"url", r.URL.String()},
 	}...)
 
+	if r.Header.Get("User-Agent") == "" {
+		r.Header.Set("User-Agent", t.controller.userAgent)
```
NyaaaWhatsUpDoc, line 138: to be honest i would just call .Set() here, no need to do the check for existing. We're not going to have different UA strings coming from the same server :p

daenney: I was thinking that too, but then I discovered that both Mastodon and Honk tweak their UA to include "Bot" when fetching images from Twitter:

* Honk: https://humungus.tedunangst.com/r/honk/v/tip/f/hoot.go L163
* https://github.com/mastodon/mastodon/blob/6cf57c676550068a59149ca82d63fcb5b5431158/app/services/fetch_link_card_service.rb#L48

Now I'm not sure if the dereference media path would ever get triggered for that, but I left it here because of that. I'm not familiar enough with how requests flow through gts. If this wouldn't ever end up in that path, then we can absolutely drop that.

NyaaaWhatsUpDoc: At the moment it wouldn't do. If we add that in the future I think I'd prefer a different method of managing it imo, e.g. initializing the per user transport with the correct UA :)

daenney: Fixed!

PR #1155, opened 2022-11-26
[feature] media: add webp support

ftrvxmtrx:
Along with https://github.com/superseriousbusiness/exif-terminator/pull/4 this made me able to find and follow a person on `mas.to` that had their avatar image in webp format. Otherwise it results in error related to not being able to process webp.
I can update this PR properly once the `exif-terminator` one is merged.


tsmethurst:
Thanks for opening this! I've commented on your PR over on the exif-terminator repo :)

> Otherwise it results in error related to not being able to process webp.

Darn! Unrelated to this PR then, but we should also update some of the account dereferencing logic to be able to handle unknown file types on avatars / headers (perhaps just skipping the image if we can't process it). This might also be related to an issue we've been having with Honk federation...

ftrvxmtrx:
@tsmethurst I guess it boils down to the question if it's fine to store (potentially) sensitive data since EXIF won't be stripped out of unsupported media?

tsmethurst:
Indeed! I'd rather not store the data at all then, if we can't be sure it doesn't contain sensitive information. But I think by just skipping it completely if it's an unknown type, we can avoid this :)

tsmethurst (approved):
Great, thank you! :) Will merge this after cutting 0.6.0

Issue #1089, opened 2022-11-20
[feature] Video Support


theSuess:
Since I'd really like to post some videos, I'm working on implementing this right now. I've created this issue as grounds for discussion on the implementation and feature set.

IMHO integrating with ffmpeg is required in some way as there are no standalone go libraries to work with various video formats.

Open Topics:

- [ ] How to interact with video data (executing `ffmpeg`/`ffprobe` vs binding to the library)
- [ ] Reencode them for optimizations?
- [ ] Strip metadata?

theSuess:
As a reference: I've built a small proof-of-concept by executing ffmpeg in a subprocess which can be found here: https://github.com/superseriousbusiness/gotosocial/compare/main...theSuess:gotosocial:video-support

NyaaaWhatsUpDoc:
Regarding implementation details, we're very hesitant to require CGO and hooking into ffmpeg bindings. It just makes building much more complex, and CGO calls come with a _somewhat_ performance penalty compared to regular Go. Personally I was hoping to look into transpiling necessary ffmpeg libraries in a similar manner to `modernc.org/sqlite`, but obviously if you beat me to a solution then that's wonderful :D.

theSuess:
I think transpiring ffmpeg is a huge endeavor - even if we only do the calls required (extracting Thumbnail + fetching size information), mainly because of the many codecs/options hidden inside a "simple" video file. But it's just a hunch based on my previous experience with audio/visual stuff so maybe it's easier than I imagine. Would love to be proven wrong

hikari-no-yume:
I'd be interested in this feature. Personally, most of the video I upload is clips that I've already cut down to size, so if GoToSocial would have a configuration where it doesn't attempt to do any re-encoding itself and just takes the file as it is, that'd work great for me. There are actually social platforms out there that work like this: so far as I can tell, if you post a video file in a Discord message, it doesn't try to re-encode it.

theSuess:
We'd still need a video parsing library to extract a thumbnail and get the video dimensions

hikari-no-yume:
Ah, does the ActivityPub server _have_ to provide a thumbnail?

hikari-no-yume:
For what it's worth, executing the ffmpeg CLI tool seems like a reasonable way to do things to me. It's easy to get hold of on any platform (in my experience) and I'm sure there's hundreds of tools out there that are wrappers around it.

tsmethurst:
[...]

PR #1165, opened 2022-11-28
Video support

theSuess:
Implements #1089 

Pretty much what it says in the title :grin: 

This add support for the `video/mp4` mime type with either ffmpeg based thumbnail generation or a go-native placeholder solution.

This is still a draft and further cleanup/optimizations might be required


web/source/css/status.css
```diff
@@ -256,6 +259,23 @@ main {
...
+				z-index: 4;
+			}
+			display: none;
+			position: absolute;
+			height: 100%;
+			width: 100%;
+			justify-content: center;
+			align-items: center;
+			font-size: 7em;
+			text-shadow: 0 0 10px black;
```
f0x52, line 275: what does this look like? we don't really use text-shadows anywhere else

f0x52 (commented):
web stuff looks alright

web/source/css/status.css
theSuess: image

It's just a black border around the play icon. Regular css border does not work as it would just draw a rectangle around the element

internal/media/video.go
```diff
@@ -0,0 +1,219 @@
...
+			return
+		}
+		cmdErrc <- fmt.Errorf("metadata probe subprocess failed:\n%s", out)
+	}()
+	select {
+	case err := <-cmdErrc:
+		if err != nil {
+			return nil, err
+		}
+	case <-time.After(time.Second):
```
tsmethurst, line 85: could this not take longer than 1s for large videos or slow machines? perhaps it's worth increasing this

edit: wait, i guess this just reads info from the header rather than looking through the whole file right? :P

internal/media/video.go
```diff
@@ -0,0 +1,219 @@
...
+func decodeVideo(r io.Reader, contentType string) (*mediaMeta, error) {
+	if config.GetMediaVideoFFMPEGEnabled() {
+		return decodeVideoFFMPEG(r, contentType)
+	}
+	return decodeVideoNative(r, contentType)
+}
+
+func decodeVideoFFMPEG(r io.Reader, contentType string) (*mediaMeta, error) {
+	prog := "ffprobe"
+	args := []string{
```
tsmethurst, line 48: i think it's worth commenting these args / linking to docs for ffprobe for them if possible, to make this call easier to understand

internal/media/video.go
```diff
@@ -0,0 +1,219 @@
...
+	s := streamInfo.Streams[0]
+	return &mediaMeta{
+		width:  s.Width,
+		height: s.Height,
+		size:   s.Height * s.Width,
+		aspect: float64(s.Width) / float64(s.Height),
+	}, nil
+}
+func decodeVideoNative(r io.Reader, contentType string) (*mediaMeta, error) {
+	data, err := ioutil.ReadAll(r)
```
tsmethurst, line 109: reading the whole thing into memory seems expensive, since videos can be quite large... is there a way to avoid doing this?

internal/media/video.go
```diff
@@ -0,0 +1,219 @@
...
+
+func extractFromVideo(r io.Reader) (image.Image, error) {
+	if config.GetMediaVideoFFMPEGEnabled() {
+		return extractFromVideoFFMPEG(r)
+	}
+	return extractFromVideoNative(r)
+}
+func extractFromVideoFFMPEG(r io.Reader) (image.Image, error) {
+	prog := "ffmpeg"
+	args := []string{
```
tsmethurst, line 163: same comment as above re: args

internal/media/video.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+	case <-time.After(time.Second * 5):
+		return nil, fmt.Errorf("timeout during thumbnailing process")
+	}
+	return imaging.Decode(out, imaging.AutoOrientation(true))
+}
+func extractFromVideoNative(r io.Reader) (image.Image, error) {
+	// As of writing this, no go native way exists to extract a thumbnail from a
+	// video so we fall back by just returning a static file
+	file, err := os.Open(path.Join(config.GetWebAssetBaseDir(), "logo.png"))
```
tsmethurst, line 210: there's probably a better way of doing this than decoding the same logo every time :P

tsmethurst (commented):
Looking good :) Couple comments about the implementation stuff.

Re: calling ffmpeg/ffprobe as a command, I know @NyaaaWhatsUpDoc still had some doubts about this. Maybe we could talk about that here?

internal/media/video.go
theSuess: I think we should make the timeouts configurable as the performance will vary across hardware specifications

internal/media/video.go
theSuess: Nope, sadly not possible since the mp4 parser needs to be able to seek the stream at will. To avoid excessive memory usage one could write the stream into a file first and then read from disk

internal/media/video.go
theSuess: it should be possible to load it during package initialization, but not sure how the config module handles this. Otherwise I'll have to wrap the function onto a struct and initialize it somewhere else

tsmethurst:
I'm gonna close this now since we pulled in video support (based on the excellent work done in this PR) in #1274. Thank you Dominik!

PR #1166, opened 2022-11-28
[docs] encourage using loopback bind address (fixes #1099)

### LINKED ISSUE #1099: [docs] Setup/config guide doesn't warn against binding to 0.0.0.0/[::] in reverse-proxy use-case
Hi. I just set up a GoToSocial server in a reverse-proxy configuration. I noticed there isn't a mention in the config example of the difference between binding to `0.0.0.0`/`[::]` and `127.0.0.1`/`[::1]`/`localhost`:

https://github.com/superseriousbusiness/gotosocial/blob/4a9538593c6110b5972d802aebf1c4990edafe38/example/config.yaml#L77-L84

Likewise I didn't see it mentioned in the [reverse proxy guide for nginx](https://docs.gotosocial.org/en/latest/installation_guide/nginx/).

So far as I know, if you use `0.0.0.0` (IPv4) or `[::]` (IPv6), your server will usually become directly accessible from the public internet. That's the ideal in a normal configuration, but if you're using a reverse proxy running on the same host, then this is a liability, because now someone can access the server without going via the proxy.

I am thinking that it would be good to suggest using `localhost`, `127.0.0.1` (IPv4) or `[::1]` (IPv6) instead, which makes the server only accessible via loopback, in order to avoid this issue.

I'm happy to make a pull request myself to address this, I just wanted to check first whether this was intended somehow or if there's some reason people might object.

hikari-no-yume:


docs/installation_guide/apache-httpd.md
```diff
@@ -44,6 +44,8 @@ sudoedit /gotosocial/config.yaml
 
 Then set `letsencrypt-enabled: false`.
 
+If the reverse proxy will be running on the same machine, set the `bind-address` to `"localhost"` so that the GoToSocial server is only accessible via loopback. Otherwise it may be possible to bypass your proxy by connecting to GoToSocial directly, which might be undesirable.
```
tsmethurst, line 47: i'd change 'same machine' to 'same machine, inside the same network' -- people running nginx alongside gts in a docker-compose, for instance, would not be able to get gts working properly if they set their listen address to localhost

hikari-no-yume: Ah, hmm, I'm not that familiar with how Docker works, so that could be right, but I think “inside the same network” might be a detail of Docker that the user wouldn't be familiar with? I'm not sure what to write instead though.

tsmethurst: Ah actually it's my bad, I just misunderstood something :')

tsmethurst:
Thanks!

Issue #1112, opened 2022-11-21
[feature] Bookmarks


matthewp:
### Is your feature request related to a problem ?

Mastodon supports bookmarks in the form of:

```
/api/v1/statuses/[id]/bookmark
```

### Describe the solution you'd like.

Implement the API

### Describe alternatives you've considered.

NONE

### Additional context.

Currently hitting `/api/v1/statuses/[id]/bookmark` gives a 404.

I'd possibly be willing to take on the work for this, if given a little guidance. Thank you!

tsmethurst:
Yeah I think I just made a typo in the roadmap tbh :')

matthewp:
Typo by marking it out as to not be implemented, you mean?

So you'd accept a contribution for this?

tsmethurst:
Yes exactly :D And for sure!

matthewp:
Little update here, I have bookmarking and unbookmarking locally. I still need to:

* Write tests for the POST handlers
* Implement GET

But good progress, very happy with how easy GoToSocial is to contribute to.

dmacvicar:
>But good progress, very happy with how easy GoToSocial is to contribute to.

Hey @matthewp , may be you can share a bit about your local dev setup. eg how you test locally with a client?

matthewp:
I haven't been testing with a frontend. Just running the test suite. I would think you could run [Pinafore](https://github.com/nolanlawson/pinafore) and point it at a locally running instance if you wanted full end-to-end testing.

matthewp:
PR https://github.com/superseriousbusiness/gotosocial/pull/1168

PR #1168, opened 2022-11-29
[feature] bookmarks

matthewp:
Fixes https://github.com/superseriousbusiness/gotosocial/issues/1112

This is my first contribution to GoToSocial, let me know if I did anything wrong. And thanks!


matthewp:
I'm not sure why the tests are failing. I think perhaps I'm not cleaning up some resources or something... if someone has some hints please let me know.

tsmethurst:
> I'm not sure why the tests are failing

Was just a flakey test, I ran it again and it passes now :)

Thanks for the PR, will review soon(tm)

internal/processing/account/getbookmarks.go
```diff
@@ -0,0 +1,83 @@
...
+	if requestingAccount == nil {
+		return nil, gtserror.NewErrorForbidden(fmt.Errorf("cannot retrieve bookmarks without a requesting account"))
+	}
+
+	statuses, err := p.db.GetBookmarks(ctx, requestingAccount.ID, limit, maxID, minID)
+	if err != nil {
+		return nil, gtserror.NewErrorInternalError(err)
+	}
+
+	var filtered []*gtsmodel.Status
```
tsmethurst, line 41: replace with `filtered := make([]*gtsmodel.Status, 0, len(statuses)`

internal/processing/status/bookmark.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+
+	apimodel "github.com/superseriousbusiness/gotosocial/internal/api/model"
+	"github.com/superseriousbusiness/gotosocial/internal/db"
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/id"
+)
+
+func (p *processor) Bookmark(ctx context.Context, requestingAccount *gtsmodel.Account, application *gtsmodel.Application, targetStatusID string) (*apimodel.Status, gtserror.WithCode) {
```
tsmethurst, line 32: i don't think the application is used for anything here is it?

internal/processing/status/unbookmark.go
```diff
@@ -0,0 +1,63 @@
...
+func (p *processor) UnBookmark(ctx context.Context, requestingAccount *gtsmodel.Account, application *gtsmodel.Application, targetStatusID string) (*apimodel.Status, gtserror.WithCode) {
+	targetStatus, err := p.db.GetStatusByID(ctx, targetStatusID)
+	if err != nil {
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("error fetching status %s: %s", targetStatusID, err))
+	}
+	if targetStatus.Account == nil {
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("no status owner for status %s", targetStatusID))
+	}
+
+	// TODO the real logic
```
tsmethurst, line 40: ?

tsmethurst (changes requested):
Thanks for this. This is coming along but not quite there yet, see comments.

You also need to update the swagger documentation for this, since you're adding endpoints; see the contributing.md for how to do this.

matthewp:
Figured out the swagger doc and updated that. This is ready for another review.

tsmethurst:
> This is ready for another review

Are you sure? It looks like it's not compiling or passing tests yet :P

matthewp:
@tsmethurst Ah my bad, I thought the failing CI was a flakey test like last time. I'll take a look, I probably missed updating a function name or something.

tsmethurst:
No problem. It's a good idea to just run tests locally yourself every time before you push, just to make sure.

matthewp:
@tsmethurst Tests are passing now.

tsmethurst:
Great stuff, thank you! I will review again in the morning :)

docs/api/swagger.yaml
```diff
@@ -3326,6 +3326,34 @@ paths:
...
+            responses:
+                "200":
+                    description: Array of bookmarks
+                    headers:
+                        Link:
+                            description: Links to the next and previous queries.
+                            type: string
+                    schema:
+                        items:
+                            $ref: '#/definitions/bookmark'
```
tsmethurst, line 3344: ah right, i misunderstood something before; so the bookmarks endpoints returns an array of statuses right?

so my previous comment was wrong, and this ref needs to be changed to `'#/definitions/status'`, i believe

internal/api/client/bookmarks/bookmarks_test.go
```diff
@@ -0,0 +1,149 @@
...
+	// check response
+	suite.EqualValues(http.StatusOK, recorder.Code)
+
+	result := recorder.Result()
+	defer result.Body.Close()
+	b, err := ioutil.ReadAll(result.Body)
+	suite.NoError(err)
+
+	var statuses []model.Status
+	//statusReply := &model.Status{}
```
tsmethurst, line 140: delete unused line

internal/db/bundb/account.go
```diff
@@ -463,6 +463,38 @@ func (a *accountDB) GetAccountWebStatuses(ctx context.Context, accountID string,
...
+	statusIDs := []string{}
+
+	q := a.conn.
+		NewSelect().
+		TableExpr("? AS ?", bun.Ident("status_bookmarks"), bun.Ident("status_bookmark")).
+		Column("status_bookmark.status_id").
+		Order("status_bookmark.id DESC")
+
+	if accountID != "" {
+		q = q.Where("? = ?", bun.Ident("status_bookmark.account_id"), accountID)
```
tsmethurst, line 476: accountID should always be set, right? since bookmarks are made per account, and there's no reason to get a list of all bookmarks; perhaps it's worth erroring instead of accountID == ""

internal/processing/account/getbookmarks.go
```diff
@@ -0,0 +1,83 @@
...
+			prevMinIDValue = item.GetID()
+		}
+
+		items = append(items, item)
+	}
+
+	return util.PackagePageableResponse(util.PageableResponseParams{
+		Items:            items,
+		Path:             "/api/v1/bookmarks",
+		NextMaxIDValue:   nextMaxIDValue,
```
tsmethurst, line 78: The way this code currently is, the next and previous id values are derived from the first and last statuses in the slice, but this logic is actually incorrect: it should be derived from the first and last *bookmark IDs*. I believe you can use the maxID and minID parameters passed into this function to do this instead.

internal/processing/status/bookmark.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (p *processor) Bookmark(ctx context.Context, requestingAccount *gtsmodel.Account, targetStatusID string) (*apimodel.Status, gtserror.WithCode) {
+	targetStatus, err := p.db.GetStatusByID(ctx, targetStatusID)
+	if err != nil {
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("error fetching status %s: %s", targetStatusID, err))
+	}
+	if targetStatus.Account == nil {
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("no status owner for status %s", targetStatusID))
+	}
+
```
tsmethurst, line 40: we should check here to make sure the bookmarked status is actually visible to the requester, otherwise someone could eg., submit a bookmark to a status they're not allowed to view, and then when they request their bookmarks, they'd be able to see the status

internal/processing/status/unbookmark.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (p *processor) Unbookmark(ctx context.Context, requestingAccount *gtsmodel.Account, targetStatusID string) (*apimodel.Status, gtserror.WithCode) {
+	targetStatus, err := p.db.GetStatusByID(ctx, targetStatusID)
+	if err != nil {
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("error fetching status %s: %s", targetStatusID, err))
+	}
+	if targetStatus.Account == nil {
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("no status owner for status %s", targetStatusID))
+	}
+
```
tsmethurst, line 39: same comment as above: check if they're allowed to view the status before processing

tsmethurst (changes requested):
Looking a lot better :) Few things left to check/change though! Thank you for your work so far!

internal/processing/account/getbookmarks.go
matthewp: Think I got this working.

matthewp:
@tsmethurst ready for another review when you get a chance. Changes are all in dbf42c105f2ca9bb52be0860fb731e1c5bf27bf0

tsmethurst:
Looks good now, so I will merge it! Thank you for your hard work, and for making all the changes :) Very much appreciated, looking forward to trying this out

Issue #962, opened 2022-11-05
[bug] Internal oauth error after authenticating brid.gy


doenietzomoeilijk:
### Describe the bug with a clear and concise description of what the bug is.

After going through oauth authentication for brid.gy (I get redirected to login, I login and authenticate, get redirected back to brid.gy) I receive a 400 error.

The URL on brid.gy seems to be their callback endpoint. It contains a bit of state so I won't reproduce it here.

The logs on my end show this:

```
gotosocial    | timestamp="05/11/2022 18:56:43.820" func=oauth.New.func1 level=ERROR msg="internal oauth error: parse \"https://brid.gy/delete/finish\\nhttps://brid.gy/publish/mastodon/finish\\nhttps://brid.gy/micropub-token/mastodon/finish\\nhttps://brid.gy/mastodon/delete/finish\\nhttps://brid.gy/mastodon/callback\": net/url: invalid control character in URL"
```

The newlines might be the invalid control characters mentioned...

When looking at the brid.gy code, the endpoint controller seems to be [here](https://github.com/snarfed/bridgy/blob/f8d05833f9e4e2cf31efaf2464363c5d151857e1/mastodon.py#L193) (I think - I don't do Python/flask development, but it looks promising).

I have no idea why that new-line separated string ends up in my logs, it doesn't seem like that brid.gy controller is supposed to output that stuff.

### What's your GoToSocial Version?

0.5.1-SNAPSHOT git-15be356

### GoToSocial Arch

amd64 docker

### Browser version

Firefox 104.0.2

### What happened?

Tried to authenticate the Brid.gy Mastodon bits [here](https://brid.gy/mastodon/start?x=79&y=26&feature=listen&instance=&scope=read%3Aaccounts+read%3Ablocks+read%3Anotifications+read%3Asearch+read%3Astatuses). 

### What you expected to happen?

End up authenticated at brid.gy.

### How to reproduce it?

Follow [this link](https://brid.gy/mastodon/start?x=79&y=26&feature=listen&instance=&scope=read%3Aaccounts+read%3Ablocks+read%3Anotifications+read%3Asearch+read%3Astatuses), enter a GTS instance URL, authenticate / authorize at the GTS end.

### Anything else we need to know?

[...]

PR #1169, opened 2022-11-29
Support newline separated redirect_uris

matthewp:
Fixes https://github.com/superseriousbusiness/gotosocial/issues/962

### DIFF
--- a/internal/processing/app.go
+++ b/internal/processing/app.go
@@ -20,6 +20,7 @@ package processing
 
 import (
 	"context"
+	"strings"
 
 	"github.com/google/uuid"
 	apimodel "github.com/superseriousbusiness/gotosocial/internal/api/model"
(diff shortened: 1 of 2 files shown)


internal/processing/app.go
```diff
@@ -50,12 +51,21 @@ func (p *processor) AppCreate(ctx context.Context, authed *oauth.Auth, form *api
 		return nil, gtserror.NewErrorInternalError(err)
 	}
 
+	// Find the first real URI if there is one
+	var redirectURI string
```
tsmethurst, line 55: I'm curious about this: doesn't this mean that we basically throw away any other redirect URIs after the first non-oob one that we find?

tsmethurst:
I'm not 100% sure about this one. It's nice to get client stuff working, but this seems like a workaround for a weird Mastodon API thingy. Does it conform to oauth2 spec? In general we're pretty flexible about Mastodon API quirks, but I don't want to do anything that breaks oauth2 spec conformance, so I would like to validate this first. Would you mind looking into it?

matthewp:
@tsmethurst Yeah sure, but when you search for `redirect_uris` you only find references to Google APIs. I'm not sure it's an oauth thing.

Here's what makes sense to me though. When you register an app you can send it multiple redirect uris (maybe you have one for local testing, one for staging, one for prod).

Then when you go to do an authorization you send `redirect_uri` and the server makes sure it matches one of the ones registered for the app.

So perhaps what GoToSocial is doing wrong is it's using the value in the database to try to redirect to, rather than using the one that comes in the request. I'll look and see if that's something we can change.

matthewp:
Ok, I have a setup locally where I can test against the client that is failing. This will help track down the problem better, will submit a new PR when I have a better idea of the problem.

matthewp:
Ok, so I understand what is going on here now. Reopening the PR for discussion although it's not correct in its current form.

GTS saves the RedirectURIs as the Domain on the client here: https://github.com/superseriousbusiness/gotosocial/blob/a1dda2267260472aa777ef42d9d2f5c36df648f3/internal/processing/app.go#L73

When creating the auth code, the oauth library checks to see if the redirect uri is part of the client's domain. Since our client's domain is not a valid URI, it throws.

So *I think* the proper solution would be to create multiple clients, one for each redirect URI. But that would probably also mean making multiple Applications since those contain the client id / secret as well. And that doesn't feel right.

So that's the issue, would love to hear your thoughts @tsmethurst

matthewp:
The client I was using is no longer sending redirect_uris. So I'm going to close this issue. I can move my last comment over to #962 so we can continue the discussion there.

PR #1178, opened 2022-11-29
[feature] domain block wildcarding

NyaaaWhatsUpDoc:
For domain blocks, it would now also block all subdomains under that given domain.

e.g. if you block `example.com` then the following would also be blocked
- `gts.example.com`
- `pleroma.example.com`
- `gts.pleroma.example.com`

### DIFF
--- a/internal/cache/domain/domain.go
+++ b/internal/cache/domain/domain.go
@@ -26,12 +26,27 @@ import (
 	"github.com/miekg/dns"
 )
 
+// BlockCache provides a means of caching domain blocks in memory to reduce load
+// on an underlying storage mechanism, e.g. a database.
+//
+// It consists of a TTL primary cache that stores calculated domain string to block results,
+// that on cache miss is filled by calculating block status by iterating over a list of all of
+// the domain blocks stored in memory. This reduces CPU usage required by not need needing to
+// iterate through a possible 100-1000s long block list, while saving memory by having a primary
+// cache of limited size that evicts stale entries. The raw list of all domain blocks should in
+// most cases be negligible when it comes to memory usage.
+//
+// The in-memory block list is kept up-to-date by means of a passed loader function during every
+// call to .IsBlocked(). In the case of a nil internal block list, the loader function is called to
+// hydrate the cache with the latest list of domain blocks. The .Clear() function can be used to invalidate
+// the cache, e.g. when a domain block is added / deleted from the database. It will drop the current
+// list of domain blocks and clear all entries from the primary cache.
 type BlockCache struct {
-	pcache *ttl.Cache[string, bool]
-	blocks []block
-	loaded bool
+	pcache *ttl.Cache[string, bool] // primary cache of domains -> block results
+	blocks []block                  // raw list of all domain blocks, nil => not loaded.
 }
 
+// New returns a new initialized BlockCache instance with given primary cache capacity and TTL.
 func New(pcap int, pttl time.Duration) *BlockCache {
 	c := new(BlockCache)
 	c.pcache = new(ttl.Cache[string, bool])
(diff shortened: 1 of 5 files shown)


NyaaaWhatsUpDoc:
Marking as draft as there's still some discussions to be had regarding
- separation of logic, putting this in the database feels like putting too much into the database itself
- how many max subdomain parts is a reasonable amount?

NyaaaWhatsUpDoc:
Thinking about this now... To help separate the logic, once the state PR #1078 has been merged, we could move this logic out of the database quite easily since the main part that holds it here right now is the access to `cache.Has(...)`

tsmethurst:
This looks good so far, very neat implementation :heart_eyes: My only concern is that it adds a lot of database calls (up to three extras, right?) during authorization, which is a very hot path called on every incoming request. I know the cacheing will help a lot but this still makes me a little nervous. What do you think?

igalic:
what about reversing a domain and comparing it?

```
reverse('gts.example.com').startsWith(reverse('example.com') +  '.')
```

shouldn't matter how many domain parts there are then

maybe there's even an `endsWith()` function, so you don't need to reverse it

NyaaaWhatsUpDoc:
> what about reversing a domain and comparing it?
> 
> ```
> reverse('gts.example.com').startsWith(reverse('example.com') +  '.')
> ```
> 
> shouldn't matter how many domain parts there are then
> 
> maybe there's even an `endsWith()` function, so you don't need to reverse it

This is helpful if we have the domain block to compare an incoming domain against. But how do we structure this as part of a database call and a hashmap (cache) lookup?

NyaaaWhatsUpDoc:
> This looks good so far, very neat implementation heart_eyes My only concern is that it adds a lot of database calls (up to three extras, right?) during authorization, which is a very hot path called on every incoming request. I know the cacheing will help a lot but this still makes me a little nervous. What do you think?

I agree, it's definitely a work in progress. If you have any brainwaves on it let me know! I'm going to keep pondering on it myself :thinking:

NyaaaWhatsUpDoc:
> This looks good so far, very neat implementation heart_eyes My only concern is that it adds a lot of database calls (up to three extras, right?) during authorization, which is a very hot path called on every incoming request. I know the cacheing will help a lot but this still makes me a little nervous. What do you think?

I agree, it's definitely a work in progress. If you have any brainwaves on it let me know! I'm going to keep pondering on it myself :thinking:

internal/db/bundb/domain.go
```diff
@@ -82,17 +95,62 @@ func (d *domainDB) CreateDomainBlock(ctx context.Context, block *gtsmodel.Domain
...
+		// Check if this lookup result already cached
+		cached := d.cache.Has("Domain", lookup)
+
+		// Attempt to fetch domain block for lookup
+		block, err := d.getDomainBlock(ctx, lookup)
+		if err != nil {
+			// We return early if this is one of:
+			// - NOT db.ErrNoEntries, i.e. unrecoverable error
+			// - cached db.ErrNoEntries, i.e. have tried this before
+			if !errors.Is(err, db.ErrNoEntries) || cached {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: I may have brain-farted this logic. Cached should indeed be checked first :sweat_smile:

NyaaaWhatsUpDoc:
Just had a thought -- the caching side of things could be sped up with a custom lookup system using a radix-trie to do glob-base matching, similar to what I have here: https://codeberg.org/gruf/go-glob/src/branch/main/multi.go.

Still need to figure out the database though.

NyaaaWhatsUpDoc:
I actually came up with quite a nice solution for this, not too complicated but should handle things well :grinning: 

Only thing left is I need to write some tests for the new `domain.BlockCache{}` type!

internal/db/bundb/domain.go
```diff
@@ -50,46 +50,51 @@ func normalizeDomain(domain string) (out string, err error) {
...
 		return nil, err
 	}
 
-	return d.state.Caches.GTS.DomainBlock().Load("Domain", func() (*gtsmodel.DomainBlock, error) {
-		// Check for easy case, domain referencing *us*
-		if domain == "" || domain == config.GetAccountDomain() {
-			return nil, db.ErrNoEntries
-		}
+	// Check for easy case, domain referencing *us*
+	if domain == "" || domain == config.GetAccountDomain() {
```
tsmethurst, line 82: Should `domain == config.GetHost()` also be included here? Or is that case handled elsewhere?

tsmethurst (commented):
This is looking really nice :)

internal/cache/domain/domain.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// block represents a domain block, and stores the
+// deconstructed labels of singular domain block.
+// e.g. []string{"gts", "superseriousbusiness", "org"}.
+type block struct {
+	labels []string
+}
+
+func (b block) Blocks(labels []string) bool {
```
tsmethurst, line 141: It's not so easy to understand what's happening here. I think a bit more commenting and/or some examples here would help a lot. Maybe a step-by-step or something? Not necessarily inside the function, but perhaps as a comment on the function?

tsmethurst (commented):
Looks great :) Just a comment on the comprehensibility of one of the functions (arguably the most important one) :P

PR #1183, opened 2022-11-29
Add AppArmor profile for Debian and Ubuntu installations

kernelmethod:
This PR adds an AppArmor profile for GoToSocial to `examples/apparmor/gotosocial`. The security guarantees provided by the profile should be roughly comparable to those provided by SELinux, but will work on Debian-family Linux distros (as well as some others, like OpenSUSE). The profile I've provided has been primarily tested and verified to work with the Docker-based deployment for GTS, but should also work for people who use the binary installation method.

I've also updated the security documentation (`Federation > Security`) to include an "application sandboxing" section, which currently has some AppArmor setup instructions that I've included as well as instructions for SELinux (which were previously provided in `Installation Guide > Binary Installation From Release`.

### DIFF
--- a/docs/installation_guide/advanced.md
+++ b/docs/installation_guide/advanced.md
@@ -164,3 +164,95 @@ Yes! GoToSocial supports canonical environment variables for doing this: `HTTP_P
 The http client that GoToSocial uses will be initialized with the appropriate proxy.
 
 The environment values may be either a complete URL or a `host[:port]`, in which case the "http" scheme is assumed. The schemes "http", "https", and "socks5" are supported.
+
+## Application sandboxing
+
+Although GoToSocial does not currently have any known vulnerabilities, it's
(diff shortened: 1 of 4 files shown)


kernelmethod:
Let me know if there's a better place to put the new documentation. My understanding is that the SELinux documentation was put in the `Binary Installation From Release` instructions since it wasn't intended to be compatible with Docker installations (whereas the AppArmor profile should be compatible with both methods that are currently specified).

tsmethurst:
Thanks for the PR :)

I think instead of putting it in federation/security, it should go in the Advanced section of the install instructions. What do you think?

kernelmethod:
Sure! That sounds reasonable to me. I've just updated my branch so that the new documentation is in `Installation Guide > Advanced`.

tsmethurst:
Great, thank you! This is very well written and clear, i look forward to trying it myself :D

Issue #1026, opened 2022-11-12
[feature] Allow S3 bucket name to be blank to better support CDN with S3 buckets


joelghill:
### Is your feature request related to a problem ?

I have configured an S3 storage bucket for server media. I also configured a CDN for said bucket at a subdomain.

When configuring the server, there do not appear to be any settings to allow for the CDN to function correctly. I can set the S3 host to the CDN url, but GTS inserts the bucket name in the media urls and this is inappropriate for the CDN.

### Describe the solution you'd like.

Allow the S3 bucket name to be blank and document this use case in the config file. This should produce correct URLs

### Describe alternatives you've considered.

None

### Additional context.

I can provide my current server config, hostname, S3 bucket, CDN url, etc, if any of that will help.

joelghill:
Did a little bit of research here and it looks like allowing a blank bucket name is probably not going to work, reading and writing to S3 needs a bucket name.

I think the solution here might be to add a new setting for a CDN endpoint. The S3 config would all be the same, however if a CDN is configured creating URLs to read media the CDN endpoint should be used instead of the S3 URL + bucket name.

daenney:
I'm going to close this in favour of #2574 where we've got the generic solution and different use cases documented. If you're still interested in this feature, feel free to subscribe to that issue.

Issue #864, opened 2022-09-28
[feature] S3 browser caching


LittleFox94:
Current S3 handling prevents any caching of media attachments, because we have a new URL every time we request a piece of media (as it includes a signature for auth/authz). This is especially problematic with browsers.

There is at least one way to work around this problem: instead of generating a URL that is valid "for $x minutes starting now", we could generate one that is valid until a semi-fixed timestamp. That semi-fixed timestamp could be something like "next hour from now, minute 42 second 23", giving us an URL that is valid for one hour and does not change, regardless of the time when you get it. See here for some more: https://stackoverflow.com/a/56543573/2537210

## Proposal

When media is requested by a local, authenticated user, we hand out signed URLs for a semi-fixed timestamp of "$tomorrow $fixed-time". This URL is stored in database and every local authenticated user requesting that media gets redirected to this URL.

With this we would have a up-to-24h cache, much better than no cache at all.

tsmethurst:
That seems sensible!

NyaaaWhatsUpDoc:
I started brainstorming this and i don't think we even need it to be handled by the database. just a ttl cache of presigned fetch urls would do the trick

LittleFox94:
Where would that cache be stored?

NyaaaWhatsUpDoc:
In memory. We don't support running as a cluster so it's not an issue. Worst case it gets restarted, lose the cache and regenerate them again when necessary.

Or is there something I'm not seeing here?

LittleFox94:
> We don't support running as a cluster so it's not an issue.

Oh, we don't? What's preventing that if I have the database and media storage external (postgresql and S3)?

NyaaaWhatsUpDoc:
A few things: mainly caching, and the way the go-fed/activity library is designed and how we have then implemented ourselves around it. We'd have to rearchitect a lot of things to support it

tsmethurst:
closed by https://github.com/superseriousbusiness/gotosocial/pull/1194

PR #1194, opened 2022-12-01
[performance]: make s3 urls cacheable

theSuess:
Implements #864 and should speed up s3 based installations by a lot.

With more static URL, we can then also implement #1026 for even better performance when used in conjunction with CDNs

I hope that I used the cache library correctly - oriented myself on the `eTagCache`

(Also tweaked the testrig to use the actual storage implementation for building the S3 config)

### DIFF
--- a/internal/storage/storage.go
+++ b/internal/storage/storage.go
@@ -26,11 +26,17 @@ import (
 	"path"
 	"time"
 
+	"codeberg.org/gruf/go-cache/v3"
 	"codeberg.org/gruf/go-store/v2/kv"
 	"codeberg.org/gruf/go-store/v2/storage"
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/credentials"
 	"github.com/superseriousbusiness/gotosocial/internal/config"
+)
+
+const (
+	urlCacheTTL             = time.Hour * 24
+	urlCacheExpiryFrequency = time.Minute * 5
 )
 
 // ErrAlreadyExists is a ptr to underlying storage.ErrAlreadyExists,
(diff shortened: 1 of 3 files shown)


internal/storage/storage.go
```diff
@@ -26,13 +26,19 @@ import (
...
+	"codeberg.org/gruf/go-cache/v3"
 	"codeberg.org/gruf/go-store/v2/kv"
 	"codeberg.org/gruf/go-store/v2/storage"
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/credentials"
 	"github.com/superseriousbusiness/gotosocial/internal/config"
 )
 
+const (
+	urlCacheTTL             = time.Hour * 24
```
theSuess, line 38: Maybe this should be configurable, but I'm not too sure if it is actually that useful

NyaaaWhatsUpDoc: We have many configuration options we mean to sort-out in time. We can leave it until then :')

internal/storage/storage.go
```diff
@@ -57,73 +64,87 @@ func (d *Driver) URL(ctx context.Context, key string) *url.URL {
 	}
 
 	// If URL request fails, fallback is to fetch the file. So ignore the error here
-	url, _ := s3.Client().PresignedGetObject(ctx, d.Bucket, key, time.Hour, url.Values{
+	if u, ok := d.PresignedCache.Get(key); ok {
```
NyaaaWhatsUpDoc, line 67: I think this was a little neater here if we separated these parts out, and now check the error so we can return early, and on success case return and add to cache.

e.g.
```
if u, ok := d.PresignedCache.Get(key); ok {
    return u
}

u, err := s3.Client().PresignedGetObject(...)
if err != nil {
    return nil
}

d.PresignedCache.Set(key, u)
return u
```

theSuess: good point, I've also moved the comment to the correct location now

tsmethurst:
this looks OK to me if you're both happy with it :)

poxelman (commented):
Leave a review

internal/storage/storage.go
poxelman: Leave areaply

PR #1196, opened 2022-12-01
[frogend] Implement editing domain blocks by delete + recreate

f0x52:
Currently there's no API endpoint to PATCH a domain block (to update reasoning/obfuscation).

There was already frontend code that just sent a POST again, doing nothing. Changed the code to DELETE the old id, and then just re-create the block for that domain with the new settings.

Not sure if deleting + recreating a block has disproportionate load on the backend? Or if it's better to just remove this part of the interface until we have an actual PATCH endpoint for it to use

### DIFF
--- a/web/source/settings/lib/api/admin.js
+++ b/web/source/settings/lib/api/admin.js
@@ -59,8 +59,10 @@ module.exports = function ({ apiCall, getChanges }) {
 
 		updateDomainBlock: function updateDomainBlock(domain) {
 			return function (dispatch, getState) {
+				const state = getState().admin.newInstanceBlocks[domain];
 				return Promise.try(() => {
-					const state = getState().admin.newInstanceBlocks[domain];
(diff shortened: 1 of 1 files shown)


tsmethurst:
Mmm it will cause some db spam as it iterates through accounts to select posts from them. I think it might be better for now to indeed just remove this part of the interface instead.

f0x52:
would that be actual strain though given they were already removed by the original block?

tsmethurst:
hmm I'm not sure, I'll have to check; I don't think we *remove* remove accounts when a domain is blocked, we just mark them as suspended, so they're still there to iterate through, but i'll double check

tsmethurst:
Yeah, so it'll still cause accounts to be iterated through: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/processing/admin/createdomainblock.go

So if you blocked a domain with like 100 accounts, that's 100 accounts getting sent to the workers to be processed again, which is not ideal.

Issue #81, opened 2021-07-07
[infrastructure] Move away from Github to a better solution (Gitea?)


tsmethurst:
Since Github is full-on stealing licensed code these days, among other problems, GoToSocial shouldn't stay here in the long run.

One good looking alternative is https://gitea.io/en-us/ which we could self-host somewhere.

We'd have to decide *where* to host it exactly -- on some cloud infrastructure? On a raspberry pi somewhere? (i have a spare in my drawer).

We'd also have to decide under which domain name to host it. `git.superseriousbusiness.org` is the most likely candidate, but we also could do `git.gotosocial.org` if it's just going to have GtS on it.

Seirdy:
I'd like to suggest using multiple remotes. I've [written](https://seirdy.one/2020/11/18/git-workflow-1.html) about this approach in a bit of depth, but in summary: it's quite trivial to push to or pull from multiple remotes in parallel.

I think that Gitea is approachable for people who've taught themselves how to use GitHub, but it requires creating an account. Sourcehut doesn't require creating an account and has the smallest amount of vendor lock-in among any forge I've tried, but it doesn't use the same workflow as GitHub. I think that mirroring across both platforms would allow the greatest amount of contribution without bending a knee to the network effect.

aschrijver:
Codeberg is a gitea instance that may be a good alternative to save on the burden of self-hosting. It is especially for FOSS and growing fast (network effects). They'll soon have Codeberg Pages with custom domain name support and CI based on Woodpecker (OSS drone.io clone).

daenney:
Done in #3769.

PR #1204, opened 2022-12-03
[feat] Add CodeQL based security scanning

daenney:
CodeQL can be used to scan for common vulnerabilities in code and supports Go. This enables weekly scanning because currently the analysis takes up to 8min (due to having to extract all the types from the activity package).

It also includes a fix for one of the two issues CodeQL will raise, the integer conversion. Right now we parse the integer from a string using 64 for a bit size and then convert into `int`. However, int in Go is platform dependent and is only guaranteed to be at least 32 bits. Since it seems likely that GTS can run on 32-bit platforms like single board computers, the current conversion might result in the number being truncated (https://go.dev/ref/spec#Conversions) which may result in issues. Given what `limit` and `offset` are used for, 32-bit seems plenty so this PR switches the bit-size to 32.

There are two alternative ways to fix it:
* We retype `limit` and `offset` to be `int64` everywhere so that it works with values that big on 32-bit too
* Guard the `int(i)` conversion with `if i >= math.MinInt && <= math.MaxInt` but it means that GTS may behave differently on 32-bit vs 64-bit which seems less ideal

It also switches the port conversion in the testrig to use a bitsize of 16, since port numbers don't go above that.

When currently run it will also raise an issue for https://github.com/superseriousbusiness/gotosocial/blob/bc917a4085f8f6195e70d55b751b08cc5a93059c/internal/db/bundb/bundb.go#L333
There's no way to tell it it's a false positive but the warning can be dismissed in the security tab and shouldn't recur.


daenney:
I think in this case we may also want change the OpenAPI annotations to make it clear that `limit` and `offset` should always be in32?

igalic:
how soon are we going to run into limitations of we use `int32` instead of `int64` in certain parts?

daenney:
I suspect not ever. MaxInt32 is big, 2,147,483,647. That's a lot of custom emoji for a single instance and if someone wanted to query for all you can still set the limit to 0.

I'm a little unsure about offset, but normally you'd do something like set the limit to $num and then on the first result you'd get back how many pages of results there are. Then you'd follow up with subsequent requests to get the next page and so on through the offset. At least in theory that would mean MaxInt32 * MaxInt32 amount of results.

I'm a little puzzled as to how some endpoints only have a limit and not an offset. Without the offset it seems like you wouldn't be able to page through results, which at least in part feels like it defeats the purpose of limit since clients would be inclined to set it to 0 in that case?

tsmethurst:
Can only speak for myself here, but I don't really find things like CodeQL very helpful, for a couple reasons:

1. It ties us further in to the github ecosystem, when we already have plans to migrate away from it at some point: #81 
2. As you mention, it flags false positives, which at best cause notification fatigue, and at worse lead us to waste time investigating them.
3. The issues they do catch tend to be minor/edge cases, in my experience: the int stuff you mention above is a case in point.
4. It's too easy to start falsely relying on these tools as a stamp of code quality, when they're generally arbitrary measurements.

We rejected a similar PR a while ago (https://github.com/superseriousbusiness/gotosocial/pull/1101) though this was a linter PR rather than code analysis PR.

So I'm not really convinced that we need/want something like this, but I'm happy to have my eyes opened if it turns out something like codeql is actually really indispensable :P

daenney:
I'm honestly not sure. I think the int thing is worth fixing because "Go will truncate the value to something" doesn't seem super great. Other than that, no particular feelings about it.

daenney:
Though give the size of the code base, 1 false positive doesn't seem like a particularly bad track record :smile:.

tsmethurst:
> I think the int thing is worth fixing because "Go will truncate the value to something" doesn't seem super great

Yeah i guess! I mean, I'd also kinda argue that this is a false positive, because given what those ints are used for (defining limits and offsets during API queries), there's almost no way that it would become relevant whether they're an int32 or an int64. A PR that just changes these values without enabling codeql would be welcome though imo.

PR #1206, opened 2022-12-04
[docs] Remove filesystem logging directives from example systemd unit config

f0x52:
I was a bit surprised to see these two lines in our example Systemd unit config.

As far as I know this is a very uncommon default, given Systemd logs to journald by default. Users can then just use journalctl (as expected).


daenney:
It's definitely a little odd and not what you'd typically see. It's also a bit unhelpful in its current state because it doesn't prompt you to set up log rotation so these files will grow endlessly.

PR #1207, opened 2022-12-04
[chore] Fix a few possible cases of int truncation

daenney:
Closes: #1204 

This fixes a couple of cases where due to int being platform dependent a value could get truncated if running on 32bits.


tsmethurst:
Thanks @daenney :)

PR #1221, opened 2022-12-07
[feature] Start implementing refetch of media items missing from storage

tsmethurst:
This PR adds an api endpoint `/api/v1/admin/media_refetch` which accepts `POST` requests from instance admins.

On successful `POST`, this triggers a process in the media manager which iterates through media entries in the database, and for each item checks whether an entry for that item exists in storage. If it doesn't, it will be refetched by making a call to a remote instance, and stored if possible.

This is intended for scenarios where you've accidentally deleted items from your storage directory, and you want to recover them without having to take painstaking repetitive manual actions.

The `POST` request accepts a query parameter, `domain`, which can be used to specify the domain that the instance should check media from. This defaults to all domains.

The PR is tested + working on goblin.technology where I, guess what, accidentally deleted some emojis from storage :')

**currently this only implements refetching of emojis!**


internal/media/refetch.go
```diff
@@ -0,0 +1,150 @@
...
+		totalRefetched     int
+		maxShortcodeDomain string
+		refetchIDs         []string
+	)
+
+	if domain == "" {
+		domain = db.EmojiAllDomains
+	}
+
+	log.Info("checking emojis for refetch candidates...")
```
NyaaaWhatsUpDoc, line 48: I think the info level logs in this function should be updated to be a mixture of debug / trace level logs. particularly the within-loop (non warn / error) entries, those should definitely be trace level. it's quite deep in the application logic to be info level imo

internal/processing/admin/mediarefetch.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (p *processor) MediaRefetch(ctx context.Context, requestingAccount *gtsmodel.Account, domain string) gtserror.WithCode {
+	transport, err := p.transportController.NewTransportForUsername(ctx, requestingAccount.Username)
+	if err != nil {
+		err = fmt.Errorf("error getting transport for user %s during media refetch request: %w", requestingAccount.Username, err)
+		return gtserror.NewErrorInternalError(err)
+	}
+
+	go func() {
+		refetched, err := p.mediaManager.RefetchEmojis(context.Background(), domain, transport.DereferenceMedia)
```
NyaaaWhatsUpDoc, line 38: it might be worth adding an info log here stating `MediaRefetch: starting emoji refetch`, instead of  the info level entries in the other function. When we add further fetetched media types here we could also add further entries to state when each part of the process is starting :)

internal/media/refetch.go
```diff
@@ -0,0 +1,150 @@
...
+		for _, emoji := range emojis {
+			totalChecked++
+			if totalChecked%50 == 0 {
+				log.Infof("checked %d emojis so far...", totalChecked)
+			}
+
+			refetch, err := m.emojiRequiresRefetch(ctx, emoji)
+			if err != nil {
+				return 0, fmt.Errorf("error checking refetch requirement for emoji %s: %w", util.ShortcodeDomain(emoji), err)
+			}
```
NyaaaWhatsUpDoc, line 70: It might look a little neater here to do something like:

```go
} else if !refetch {
    continue
}

refetchIDs = append(refetchIDs, emoji.ID)
```

this is something i generally try to conform to: https://dave.cheney.net/practical-go/presentations/gophercon-israel.html#_return_early_rather_than_nesting_deeply I know it's not 100% like-for-like situation here, but the general takeaway i apply (where possible) is preferring to return/continue/break early on the "negative" code path and preferring not to nest on the positive codepath if i can help it

sorry if this comes across as really pretentiouos my coworker keeps sending me these articles and they've just kind of clicked with my obsessive personality and getting anal over how my code looks :')

internal/media/refetch.go
```diff
@@ -0,0 +1,150 @@
...
+	for _, emojiID := range refetchIDs {
+		emoji, err := m.db.GetEmojiByID(ctx, emojiID)
+		if err != nil {
+			// this shouldn't happen so return if it does
+			return 0, fmt.Errorf("error getting emoji %s: %w", emojiID, err)
+		}
+		shortcodeDomain := util.ShortcodeDomain(emoji)
+
+		if emoji.ImageRemoteURL == "" {
+			log.Warn("emoji %s could not be refreshed because it has no ImageRemoteURL set", shortcodeDomain)
```
NyaaaWhatsUpDoc, line 98: Does this mean that it's a local emoji? If so do we need to be warn logging?

internal/media/refetch.go
```diff
@@ -0,0 +1,150 @@
...
+			continue
+		}
+
+		emojiImageIRI, err := url.Parse(emoji.ImageRemoteURL)
+		if err != nil {
+			log.Warn("emoji %s could not be refreshed because its ImageRemoteURL (%s) is not a valid uri", shortcodeDomain, emoji.ImageRemoteURL)
```
NyaaaWhatsUpDoc, line 104: For these cases do we want to consider removing them from the database, given that they're broken? Also that log entry should be a `Warnf` :p

internal/media/refetch.go
```diff
@@ -0,0 +1,150 @@
...
+			Disabled:             emoji.Disabled,
+			VisibleInPicker:      emoji.VisibleInPicker,
+		}, true)
+		if err != nil {
+			log.Warn("emoji %s could not be refreshed because of an error during processing: %w", shortcodeDomain, err)
+			continue
+		}
+
+		if _, err := processingEmoji.LoadEmoji(ctx); err != nil {
+			log.Warn("emoji %s could not be refreshed because of an error during loading: %w", shortcodeDomain, err)
```
NyaaaWhatsUpDoc, line 125: Should this be an error level log? Also would be `Errorf` :p

NyaaaWhatsUpDoc: Same goes for line 120 also

internal/media/refetch.go
tsmethurst: ack, will change

internal/media/refetch.go
```diff
@@ -0,0 +1,150 @@
...
+				// an actual error has occurred
+				log.Errorf("error fetching emojis from database: %s", err)
+			}
+			break
+		}
+
+		for _, emoji := range emojis {
+			totalChecked++
+			if totalChecked%50 == 0 {
+				log.Infof("checked %d emojis so far...", totalChecked)
```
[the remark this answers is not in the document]
tsmethurst: Ah, the idea was basically just to give some information that the process is working; if we don't log anything it can look like the instance just swallowed the command, so I'd like to still log *something* here

internal/media/refetch.go
tsmethurst: nah that totally makes sense, i will change it :)

internal/media/refetch.go
tsmethurst: mm i'll double check

internal/media/refetch.go
tsmethurst: > Also that log entry should be a Warnf

Thanks! Well spotted, will fix it.

> For these cases do we want to consider removing them from the database, given that they're broken

Maybe, but it seems a bit out of scope for this function... perhaps some 'clean' function in future, or 'validate' could do something like this? :thinking: what do you think?

tsmethurst:
Fixed 'er up a bit :)

internal/media/refetch.go
igalic: i love return early code

the alternative is: https://github.com/icing/blog/blob/main/gotos.md

internal/processing/admin/mediarefetch.go
tsmethurst: I forgot to do this and now i'm knees-deep in another branch... will fix it up tomorrow :)

internal/media/refetch.go
NyaaaWhatsUpDoc: Agreed, seems like a future thing

NyaaaWhatsUpDoc:
Once you've fixed that above unresolved comment you can go ahead and merge this, otherwise looks great :)

PR #1222, opened 2022-12-07
[frogend] Emoji copy "Steal this toot"

f0x52:
Allows pasting a remote toot url, which will then present the custom emoji in it for either copying or disabling on the local instance.
Shortcodes can be adapted, and will be validated against already existing local shortcodes.

Not actually tested yet, because the testrig can't fetch remote toots over federation, and already local emoji can't be copied 
image


f0x52:
tested and works on a 0.6.0 instance, I can fetch emoji from my mastodon instance :D
image

web/source/settings/admin/emoji/remote/index.js
```diff
@@ -0,0 +1,53 @@
...
+const query = require("../../../lib/query");
+const Loading = require("../../../components/loading");
+
+module.exports = function RemoteEmoji() {
+	const {
+		data: emoji = [],
+		isLoading,
+		error
+	} = query.useGetAllEmojiQuery({filter: "domain:local"});
+
```
tsmethurst, line 34: i think it's worth adding a comment here to explain that you're fetching local emoji codes so that you can compare them to imported emojis + make sure there's no conflicts; otherwise it's not immediately obvious why you're doing a `domain:local` query here in the remote emoji component

web/source/settings/admin/emoji/remote/index.js
```diff
@@ -0,0 +1,53 @@
...
+		return new Set(emoji.map((e) => e.shortcode));
+	}, [emoji]);
+
+	return (
+		<>
+			<h1>Custom Emoji (remote)</h1>
+			{error && 
+				<div className="error accent">{error}</div>
+			}
+			{isLoading
```
tsmethurst, line 45: love how easy rtk query makes checking loading :D

web/source/settings/admin/emoji/remote/parse-from-toot.js
```diff
@@ -0,0 +1,318 @@
...
+			/>
+		);
+	}, [isSuccess, data, instanceDomain, emojiCodes]);
+
+	function submitSearch(e) {
+		e.preventDefault();
+		searchStatus(url);
+	}
+
+	return (
```
tsmethurst, line 74: is it worth refactoring this into smaller components to avoid all the nesting?

fine if you'd rather not, just checking

web/source/settings/admin/emoji/remote/parse-from-toot.js
```diff
@@ -0,0 +1,318 @@
...
+						{shortcode}: <span style={{ fontWeight: "initial" }}>{msg}</span>
+					</div>
+				)));
+			} else {
+				setError(e);
+			}
+		});
+	}
+
+	return (
```
tsmethurst, line 239: same comment here re: nesting.. though i guess if we're not going reuse any of these subcomponents elsewhere maybe there's not much point

tsmethurst (approved):
Looks great :D Just a few comments about readability and that's it; feel free to merge once you've decided what to do with those!

web/source/settings/admin/emoji/remote/parse-from-toot.js
f0x52: I looked over at these again but don't think splitting it into separate components would improve readability too much, and we're not re-using them either

Issue #1205, opened 2022-12-03
[investigation/performance] Unexpectedly high RAM pressure when a post goes "viral"


trysdyn:
I'm filing this as much as a "story" or "what do y'all think?" as I am a bug. Sorry for the weird report.

Version 0.6.0-rc1. I don't believe anything in rc2 or rc3 would impact this but I plan to update soon just to be absolutely sure.

I have my GtS instance on system level monitoring and have been seeing distressingly large spikes in GtS's RAM usage when a post gets carried out beyond the usual reach. A couple of boosts that carry a post beyond normal reach is sufficient to cause a 200+ mb blip in GtS's usage for about a minute. In the case of a small instance hosted on a VPS, a post going slightly "viral" might be enough to bring the instance down.

Fun fact: I mentioned the gotosocial fedi account with this and when they replied, my GtS instance *did* effectively crash under the impending load after GtS gnawed about 500mb and went well into my swap.

Here's my monitoring of available RAM on the box. This is on a bog standard $5 VPS (1gb RAM with another for swap) using postgres as a data store, which I think is the base SKU for anything targeting "small and self-hosted".

![gtsmem](https://user-images.githubusercontent.com/1696997/205465545-f80e4aa6-82e2-4306-8792-e868c47c1932.png)

To be clear: it's GtS itself growing during these spikes, not postgres or anything else.

RAM usage is stable (with a small persistent increase) for days except for those large dips. The first occurred with just a few boosts on a post I made, the second occurred when the gotosocial fedi account replied to me.

Logs just show a *lot* of requests for the statuses in question and my webfinger/profile, with some very unhappy warnings about DB query time. The slow query log hits look like this:

```
[...]

PR #1226, opened 2022-12-07
Add caching to webfinger responses

mperham:
Per #1205, this adds a short period of caching to the webfinger response to minimize the work necessary to handle the duplicate requests in a viral thundering herd.

Any other endpoints need this? Is 5 minutes a reasonable value?

### DIFF
--- a/internal/api/s2s/webfinger/webfingerget.go
+++ b/internal/api/s2s/webfinger/webfingerget.go
@@ -98,5 +98,9 @@ func (m *Module) WebfingerGETRequest(c *gin.Context) {
 		return
 	}
 
+	// allow upstream HTTP proxies to cache this result for a short period.
+	// mastodon and others can cause a massive spike in webfinger requests
+	// for an author when their toot goes viral.
+	c.Header("Cache-Control", "public, max-age=300")
 	c.JSON(http.StatusOK, resp)
 }


mperham:
Many caches will silently give up if the request has query parameters, they only cache based on the /url/file/path. Unclear if Caddy does this.

mperham:
My understanding is that if @foo@bar.social authors a viral toot or reply, you'll get 1000s of calls to `https://bar.social/.well-known/webfinger?resource=acct:foo@bar.social`. If you have Caddy, nginx, varnish or similar http proxy in front of GtS, this will turn those 1000s of calls into 1 call to GtS.

Agreed, there are likely much heavier-weight endpoints to cache. This is just my first and simplest PR.

tsmethurst:
Ahhh okay, my apologies, I'm tired and misunderstood the workings of the cache. I thought it was intended for remote *servers* to cache the response, not the upstream reverse proxy. In that case, cool. But then we need to make sure that, like you warned, the upstream reverse proxy ought to differentiate based on the query, otherwise this could end up breaking things right?

(Also, the linter change should not be included in this pr, since it's not related to the cache thing we're discussing)

mperham:
My mistake, I forgot to make branches for the different issues.

mperham:
I have no idea how to remove the linter commit. Do I `git revert` that commit or use some other black magic?

igalic:
many caches, in the past, did not cache stuff of it had query strings

i hope that has improved…

mperham:
> we need to make sure that, like you warned, the upstream reverse proxy ought to differentiate based on the query

Exactly. This is the hard part, we'll need custom logic to tell the proxy that the resource parameter is the actual cache key here. Do you have a caching proxy you prefer?

I've worked with Varnish in the past. Varnish's VCL allows us to build that logic. We could run Varnish "in front of" the GtS binary in the Docker image. It adds a bit of production complexity but I think it's really a critical tool to handle those thundering herds built into the AP standard and Mastodon impl.

https://varnish-cache.org/docs/7.2/users-guide/

daenney:
I don't think the container image should contain multiple processes, like Varnish and GTS. In many cases, folks will already be using something like nginx or Apache in front of their GTS instance. For caching webfinger I think that's a better place to do it than in GTS, as that would also let operators more easily control whether they want to serve stale responses etc. in case GTS itself is unavailable.

Providing some documentation on how to configure the caching proxy behaviour correctly is helpful. https://www.sheshbabu.com/posts/nginx-caching-proxy/ provides a nice read on this for nginx for example.

PR #1228, opened 2022-12-07
Remove deprecated linters

mperham:
These linters print warnings when used in recent versions of golangci-lint.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,10 +20,8 @@ linters:
     - gocritic
     - gofmt
     - gosec
-    - ifshort
     - nilerr
     - revive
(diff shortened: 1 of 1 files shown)


tsmethurst:
Just to jot the warnings down for future reference:

```
WARN [runner] The linter 'ifshort' is deprecated (since v1.48.0) due to: The repository of the linter has been deprecated by the owner.  
WARN [linters_context] wastedassign is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649.
```

And a little reminder that we should reenable wastedassign when golangci-lint reenables it on their end too

PR #1239, opened 2022-12-09
[chore] make single pull request template

tsmethurst:
When you have multiple templates, they can only be selected by the API, so you can't just open a pull request via the web UI and have the template presented to you. This should fix that by having just one pull request template.


.github/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+> If this is a documentation change, please briefly describe what you've changed and why.
+
+This pull request implements xyz or fixes abc.
+
+closes #(issue)
+closes #(another issue)
+
+## Checklist
+
+Please put an x inside each checkbox to indicate that you've read and followed it: `[ ]` -> `[x]`
```
igalic, line 14: this should be an HTML comment

PR #1243, opened 2022-12-09
[feature] Ensure support to older browsers

EchedeyLR:
# Description

Current `.browserslistrc` approach doesn't cover UXP-based web browsers (currently tagging as Firefox 68) as showed in https://browserslist.dev/?q=PiAwLjI1JQ%3D%3D

Given that this is already implemented, I mean the idea of using `.browserslistrc`, I think this can be added also (as its existence is already preventive) to ensure support on the platform.

## Checklist


tsmethurst:
In the browserlist thing you linked, I see that 'since 2015' also includes browsers that are used by 0% of people according to the stats there, including very old versions of Edge and stuff. I'm not really keen on implicitly promising to support such browsers, for the sake of moving our supported users thingy from 93% (which it is currently) to 98% (which it becomes with this change). To me, the current settings seem fine: can you point to a concrete case or issue where the current settings prevent someone from using the software?

f0x52:
Currently this browserslist change leaves our bundles identical, except for one case where it actually removes some `-webkit` compatibility rules.

Going into the future though, I very much agree with
> I'm not really keen on implicitly promising to support such browsers

We might do fixes on a case-by-case basis as they arise, but we're not going to actively test for ancient browsers. There's also only so much autoprefixer and polyfills can achieve, as shown by earlier CSS Grid issues.

EchedeyLR:
> There's also only so much autoprefixer and polyfills can achieve, as shown by earlier CSS Grid issues.

I tried to add polyfills as suggested before and in the Matrix channel using a plugin and you or another person told me that current babel plugin performed that (which I ignored a bit).

With that browserlist and the plugin I integrated, the size of the file was increased by 2 or 3 times in size...

EchedeyLR:
I would like to reword some things to avoid misunderstanding, as I reread my own texts and I always get the same issues no matter how much effort I put in caring how I say things:

The idea with the fix was to avoid you having to test as I am also not a core contributor/developer of this project and, as result, not someone who you could trust to perform actual tests over this.

I already expected this because is a recurrent I have experienced and this is why I preferred a different approach of trying to do it myself, even if I have to learn about your configuration and coding style in the process.

Anyways, I am sorry I made you lose any bit of time.

Issue #1244, opened 2022-12-09
[chore] cleanup server CLI / environment / configuration file parsing


NyaaaWhatsUpDoc:
Current issues:
- we accept almost every single configuration variable as CLI flags, our usage printing is already rather absurd but it will get worse as we add more configurables
- all of our configuration variables live at the root json / toml / yaml "namespace"
- ephemeral configurables are stored in the main configuration struct, and available all over the codebase via `config.Get___()` despite only ever being used in `cmd/gotosocial/action/...`
- we have no way of outputting updated configuration files based on current settings

My proposed change:
- we drastically reduce the CLI flags that we support to only the essentials to get the server running: e.g. ports, lets-encrypt, config path, database, storage path
- move configuration variables into grouped structures e.g. `DatabaseConfiguration{}`, `StorageConfiguration{}` etc
- ephemeral configurables are passed to gts action functions by means of context variables
- move `debug config` action to be under `config dump-json`, and then add `dump-yaml` and `dump-toml` actions to allow printing current configuration to file. useful if you have an old configuration file that you want to easily update with the latest flags without manually doing it
- to make migration from old configuration file format easier, create a `cmd/config-migrate` utility that allows converting from old to new configuration file format. environment variable parsing and those relying on CLI flags will unfortunately run into breakages, but ultimately we _are alpha_ and I don't think many will be relying on them for the finer configurables (e.g. oidc)

NyaaaWhatsUpDoc:
@tsmethurst if you're good with this, I have the start of some changes, and the rest of them ready in my head :)

tsmethurst:
@NyaaaWhatsUpDoc should we still consider this issue open? Or should we close it? I don't think we're having too much trouble with our config stuff these days.

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/1244.

PR #1246, opened 2022-12-09
[feature] support configuring database caches

NyaaaWhatsUpDoc:
# Description

- adds `CacheConfiguration{}` options to main configuration struct, allowing setting via config file
- update config helpers generator to support nested structs
- add example cache config variables to config tests

## Checklist


NyaaaWhatsUpDoc:
Only thing still remaining is to update example configuration file, and possibly update other areas of documentation?

NyaaaWhatsUpDoc:
In time I want to add intelligent default for cache sizes.

NyaaaWhatsUpDoc:
This lays the groundwork for any changes (if we decide to make them) in #1244

tsmethurst:
Nice one kim! this looks really good :)

We gotta remember when we cut a release including these changes to include a migration note about the config format changing to allow nesting.

Also, do you foresee any issues with the nesting? I mean, could it break if one provides a config that doesn't nest?

NyaaaWhatsUpDoc:
> Nice one kim! this looks really good :)
> 
> We gotta remember when we cut a release including these changes to include a migration note about the config format changing to allow nesting.
> 
> Also, do you foresee any issues with the nesting? I mean, could it break if one provides a config that doesn't nest?

for this specific PR there shouldn't be any breaking changes. the only nested values are new ones so shouldn't be an issue. for later PRs though -- I added a note about this in the issue #1244 :)

internal/config/state.go
```diff
@@ -41,8 +41,8 @@ func NewState() *ConfigState {
 	viper := viper.New()
 
 	// Flag 'some-flag-name' becomes env var 'GTS_SOME_FLAG_NAME'
-	viper.SetEnvPrefix("gts")
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+	viper.SetEnvPrefix("gts")
```
tsmethurst, line 45: Is the order important here? If so, worth commenting it imo

tsmethurst (approved):
Great :) Just the one comment

internal/config/state.go
NyaaaWhatsUpDoc: oh no importance, was literally just a "this looks better this way round :')"

I have very neurotypical brain

PR #1247, opened 2022-12-10
[feature] sqlite optimizations

NyaaaWhatsUpDoc:
# Description

- moves database configurables to their own namespaces
- allows configuring certain SQLite pragmas, namely: journal_mode,synchronous,cache_size

PR not yet finished, but opening it to allow discussion during development.

## Checklist


NyaaaWhatsUpDoc:
Also starts to address some of the issues in #1244 .

example/config.yaml
```diff
@@ -103,66 +103,136 @@ trusted-proxies:
...
+    tls-ca-cert: ""
+
+  sqlite:
+    # String. SQLite journaling mode.
+    # See: https://www.sqlite.org/pragma.html#pragma_journal_mode
+    journal-mode: "WAL"
+
+    # String. SQLite synchronous mode.
+    # See: https://www.sqlite.org/pragma.html#pragma_synchronous
+    synchronous: "FULL"
```
tsmethurst, line 176: I think if we're using WAL mode, we might want to set NORMAL as the synchronous default, since it's apparently fairly safe

internal/config/defaults.go
```diff
@@ -35,14 +40,25 @@ var Defaults = Configuration{
...
+			Password:  "",
+			Database:  "gotosocial",
+			TLSMode:   "disable",
+			TLSCACert: "",
+		},
+
+		SQLite: SQLiteConfiguration{
+			JournalMode: "WAL",
+			Synchronous: "FULL",
+			CacheSize:   32 * bytesize.MiB,
```
tsmethurst, line 59: this is different from the value given in the config.yaml

internal/db/bundb/bundb.go
```diff
@@ -225,47 +209,68 @@ func NewBunDBService(ctx context.Context, state *state.State) (db.DB, error) {
...
 
 	// Drop anything fancy from DB address
-	dbAddress = strings.Split(dbAddress, "?")[0]
-	dbAddress = strings.TrimPrefix(dbAddress, "file:")
+	address = strings.Split(address, "?")[0]
+	address = strings.TrimPrefix(address, "file:")
 
 	// Append our own SQLite preferences
-	dbAddress = "file:" + dbAddress + "?cache=shared"
+	address = "file:" + address + "?cache=shared"
```
tsmethurst, line 222: should we remove the `?cache=shared` param now, since apparently it's deprecated and better to use WAL? https://www.sqlite.org/sharedcache.html#use_of_shared_cache_is_discouraged

tsmethurst (commented):
These look like sensible changes. Just a few comments!

ALSO: is it a reasonable idea to run the optimize pragma when shutting down the db connection (or indeed on a schedule once per day or so?): https://www.sqlite.org/pragma.html#pragma_optimize

PR #1251, opened 2022-12-11
[docs] Serve static assets with nginx

daenney:
# Description

This explains how to use nginx to serve static assets and offload GTS from that responsibility. It also shows how to have nginx add caching headers to indicate to clients how long they may cache an asset.

## Checklist


daenney:
> Also, is it worth putting these nginx things under some 'advanced' nginx subheading? The nginx document is becoming really big and I suspect that's gonna put people off as soon as they open it.

Ya, probably :sweat_smile:. Do you want it in the same doc still or a different page entirely?

> GtS uses etag + cache headers (if-none-match, etag) for asset cacheing to return 304s when an asset hasn't been modified: can nginx also do that?

Yes it can, the behaviour stays the same.

With the configuration I'm showing here (though I've tweaked mine to use 1d currently instead of max), nginx will respond on an asset like so:
```
HTTP/2 200 OK
server: nginx
date: Sun, 11 Dec 2022 12:27:22 GMT
content-type: image/jpeg
content-length: 336801
last-modified: Wed, 23 Nov 2022 17:04:10 GMT
etag: "637e528a-523a1"
expires: Mon, 12 Dec 2022 12:27:22 GMT
cache-control: max-age=86400
```

It computes the `etag` itself, and correctly sets `last-modified` based on the file timestamp, `cache-control` based on the fact that I have it set to `1d` and the `expires` header. With this configuration and refreshing my browser a few times I don't see Firefox repeating any additional requests.

In this case I think it may be worthwhile to add `add_header Cache-Control "public, no-transform";` to ensure that things along the way may cache it but not change anything about it.

Just to go full circle:
```
$ curl 'https://ap.dny.social/fileserver/01QM36GFRFX6BGXYEDV15PPDQP/attachment/original/0193V4VAXT9X4A9DFE5N54JV15.jpeg' -vvv
[..]
> GET /fileserver/01QM36GFRFX6BGXYEDV15PPDQP/attachment/original/0193V4VAXT9X4A9DFE5N54JV15.jpeg HTTP/2
> Host: ap.dny.social
> user-agent: curl/7.86.0
> accept: */*
> 
[..]
< HTTP/2 200 
< server: nginx
< date: Sun, 11 Dec 2022 12:35:56 GMT
< content-type: image/jpeg
< content-length: 336801
< last-modified: Wed, 23 Nov 2022 17:04:10 GMT
< etag: "637e528a-523a1"
< expires: Mon, 12 Dec 2022 12:35:56 GMT
< cache-control: max-age=86400
[..]
< accept-ranges: bytes
[..]
* Connection #0 to host ap.dny.social left intact
```
Now repeat with the `etag`:
```
$ curl 'https://ap.dny.social/fileserver/01QM36GFRFX6BGXYEDV15PPDQP/attachment/original/0193V4VAXT9X4A9DFE5N54JV15.jpeg' -H 'If-None-Match: "637e528a-523a1"' -vvv
[..]
> GET /fileserver/01QM36GFRFX6BGXYEDV15PPDQP/attachment/original/0193V4VAXT9X4A9DFE5N54JV15.jpeg HTTP/2
> Host: ap.dny.social
> user-agent: curl/7.86.0
> accept: */*
> if-none-match: "637e528a-523a1"
> 
[..]
< HTTP/2 304 
< server: nginx
< date: Sun, 11 Dec 2022 12:37:45 GMT
< last-modified: Wed, 23 Nov 2022 17:04:10 GMT
< etag: "637e528a-523a1"
< expires: Mon, 12 Dec 2022 12:37:45 GMT
* Connection #0 to host ap.dny.social left intact
```

tsmethurst:
Cool :) Thanks for the explanation.

Think my preference for the docs would be to put some kind of 'nginx' subheading under the 'advanced' document here: https://github.com/superseriousbusiness/gotosocial/blob/main/docs/installation_guide/advanced.md, then move the 'hardening' and other extra nginx things to there, and put a note at the bottom of the nginx doc linking to the nginx heading in the advanced doc, with something like 'advanced configuration/cacheing options for nginx are available here'.

Does that make sense? What do you think?

daenney:
Sounds good! Let me move that around.

daenney:
I added `add_header Cache-Control "public, immutable";` to `/fileserver`, but only `public` to `/assets` because they don't have unique names. Both Firefox (desktop and Android) and Safari (on iOS and macOS) understand the immutable and will forego revalidation requests entirely for resources that are still fresh. It cuts down the amount of requests a client has to do, especially on mobile, quite significantly.

tsmethurst:
Nice one! Thanks :)

igalic:
> I added `add_header Cache-Control "public, immutable";` to `/fileserver`, but only `public` to `/assets` because they don't have unique names.

maybe we should put our assets into a versioned directory, and mark them as immutable too?

tho we might need to provide some tooling for people customizing the look of their insurance then

igalic:
`web/source` is the *source* for `web/assets`.

daenney:
I don't think that's right? `web/assets` has all kinds of things like images that best I can tell don't come from `web/source`. `web/assets` doesn't get updated when `web/source` is updated either, so at least in the source tree they would appear disjoint.

PR #1266, opened 2022-12-14
Update Apache docs to use 127.0.0.1 instead of localhost

EchedeyLR:
# Description

Update Apache docs to use 127.0.0.1 instead of localhost as done with Nginx at #1264.

## Checklist

- [x] I we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [x] I we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [x] I we have performed a self-review of added code.
- [x] I we have written code that is legible and maintainable by others.
- [x] I we have commented the added code, particularly in hard-to-understand areas.
- [x] I we have made any necessary changes to documentation.

### DIFF
--- a/docs/installation_guide/apache-httpd.md
+++ b/docs/installation_guide/apache-httpd.md
@@ -87,8 +87,9 @@ MDCertificateAgreement accepted
 
   SSLEngine On
   ProxyPreserveHost On
-  ProxyPass / http://localhost:8080/
-  ProxyPassReverse / http://localhost:8080/
+  # set to 127.0.0.1 instead of localhost to work around https://stackoverflow.com/a/52550758
+  ProxyPass / http://127.0.0.1:8080/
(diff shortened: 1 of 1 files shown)


igalic (changes requested):
thanks! here's some feedback

igalic:
> thanks! here's some feedback

Well that didn't work

EchedeyLR:
Did I make something wrong? I cannot see the proposed changes.

igalic (approved):
after rereading your patch, and [the documentation](https://httpd.apache.org/docs/2.4/rewrite/), I've come to the conclusion that there's nothing to improve on here!

tsmethurst:
Thanks for this :)

PR #1281, opened 2022-12-18
[bugfix] Close reader gracefully when streaming recache of remote media to fileserver api caller

tsmethurst:
# Description

>

This PR implements graceful failure when streaming an image to an API client during a media recache, in cases where the media is no longer present on a remote instance, or some other failure occurs.

Here's the situation before:

1. a client request a piece of media that was cached previously, but needs to be recached now
2. the async Data function is passed back to the worker, with a tee reader pipe jammed into it which the API caller will slurp from as the media is recached into storage
3. the caller waits for data to be blasted into the pipe
4. the Data function fails
5. the pipe stays open, the caller keeps waiting for that data, which will never come 
6. request eventually times out (depending on reverse proxy / client settings), but the gin request hangs around inside gts forever blocked

With these changes now:

1. a client request a piece of media that was cached previously, but needs to be recached now
2. the async Data function is passed back to the worker, with a tee reader pipe jammed into it which the API caller will slurp from as the media is recached into storage
3. the caller waits for data to be blasted into the pipe
4. the Data function fails, **and the pipe reader is closed when this happens**
5. the caller in the API function checks if any data is coming through the pipe, or if there's an EOF
6. in case of error, a 404 is returned to the caller immediately and the request completes + gets cleaned up

## Checklist

- [x] I/we have made any necessary changes to documentation (N/A)

### DIFF
--- a/internal/api/client/fileserver/fileserver_test.go
+++ b/internal/api/client/fileserver/fileserver_test.go
@@ -1,0 +1,109 @@
+/*
+   GoToSocial
+   Copyright (C) 2021-2022 GoToSocial Authors admin@gotosocial.org
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


NyaaaWhatsUpDoc:
Checked over and approved this earlier. Once you're happy with the extra few changes I added @tsmethurst this is good to go :)

PR #1282, opened 2022-12-20
[feature] For video attachments, store + return fps, bitrate, duration

tsmethurst:
# Description

>

This PR reworks some of our mp4 parsing logic, to also derive video bit rate, duration, and framerate information. This is stored in the database and then serialized to clients via the API.

This brings GtS closer in line with the Mastodon API, which also serves these fields via the client API. This should hopefully resolve some issues where videos weren't playing in some clients.

Also adds tests for the above :)

## Checklist


internal/media/video.go
```diff
@@ -61,60 +62,81 @@ func decodeVideo(r io.Reader, contentType string) (*mediaMeta, error) {
...
-		return nil, nil
+		if d := float32(tr.Duration) / float32(tr.Timescale); d > duration {
+			duration = d
+			framerate = float32(len(tr.Samples)) / duration
+		}
+	}
+
+	var errs gtserror.MultiError
+	if width == 0 {
+		errs.Append(errors.New("video width could not be discovered"))
```
NyaaaWhatsUpDoc, line 107: since a MultiError is just a string slice you could simplify this to `errs = append(errs, "video width could not be determined"` (same also goes for below errors)

internal/media/video.go
```diff
@@ -61,60 +62,81 @@ func decodeVideo(r io.Reader, contentType string) (*mediaMeta, error) {
...
+	if framerate == 0 {
+		errs.Append(errors.New("video framerate could not be discovered"))
+	}
+
+	if bitrate == 0 {
+		errs.Append(errors.New("video bitrate could not be discovered"))
+	}
+
+	if errs != nil {
+		errs.Append(errors.New("this may not be a valid mp4"))
```
NyaaaWhatsUpDoc, line 127: if we already have the above errors added i don't think we need to add this part imo.

internal/typeutils/internaltofrontend.go
```diff
@@ -299,26 +301,36 @@ func (c *converter) AttachmentToAPIAttachment(ctx context.Context, a *gtsmodel.M
...
 	// nullable fields
-	if a.URL != "" {
-		i := a.URL
+	if i := a.URL; i != "" {
 		apiAttachment.URL = &i
 	}
 
-	if a.RemoteURL != "" {
+	if i := a.RemoteURL; i != "" {
 		i := a.RemoteURL
```
NyaaaWhatsUpDoc, line 309: you can get rid of this line now :p

internal/media/types.go
```diff
@@ -140,4 +140,9 @@ type mediaMeta struct {
 	aspect   float64
 	blurhash string
 	small    []byte
+
+	// video-specific properties
+	duration  float32
```
NyaaaWhatsUpDoc, line 145: why are we storing the aspect ratio as a float64 but framerate and duration as float32s? :thinking:

NyaaaWhatsUpDoc: if anything i'd say aspect ratio is float32 and duration is float64

internal/typeutils/internaltofrontend.go
```diff
@@ -299,26 +301,36 @@ func (c *converter) AttachmentToAPIAttachment(ctx context.Context, a *gtsmodel.M
...
 		apiAttachment.Description = &i
 	}
 
+	if i := a.FileMeta.Original.Duration; i != nil {
+		apiAttachment.Meta.Original.Duration = *i
+	}
+
+	if i := a.FileMeta.Original.Framerate; i != nil {
+		round := math.Round(float64(*i))
+		apiAttachment.Meta.Original.FrameRate = fmt.Sprintf("%.f/1", round)
```
NyaaaWhatsUpDoc, line 327: there's no need for this format string like this as you've already rounded it now.. and since you know the type going in you can just do `strconv.FormatInt(int64(round), 10)` :)

NyaaaWhatsUpDoc:
@tsmethurst have left some comments but after those are fixed up it looks good to me :)

internal/media/types.go
tsmethurst: > why are we storing the aspect ratio as a float64 but framerate and duration as float32s?

I... have no idea :D early gts code i think (when it was just me making wild decisions); fixed!

internal/typeutils/internaltofrontend.go
tsmethurst: Ooops :')

PR #1288, opened 2023-01-02
[performance] media processing improvements

### LINKED ISSUE #1279: [bug] Panic probably related to S3 presigned URL caching

Shortly after updating my GtS from v0.6.0-rc3 to v0.6.0 and then continuing to use it, GoToSocial server died from a panic. The panic output is too long to add here (GitHub refuses to save the issue then), so have a pastebin instead: https://pastebin.com/71J68ZYH

v0.6.0

### GoToSocial Arch

amd64 docker

### Browser version

_No response_

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

NyaaaWhatsUpDoc:
# Description
- pulls in latest `go-store` and `go-cache` versions
- wraps media encoding/decoding handling to be mostly contained within `gtsImage{}` and `gtsVideo{}` types for neatness
- decreases JPEG thumb quality to 70% (it's only a thumb, helps with filesize)
- uses memory pools for  JPEG and PNG encoding
- updates media processing to no longer use the atomic state variables (as these were mostly unused), all is now just done within mutex
- general tidy-up of media processing which improves: neatness, memory usage, error handling
- use larger write buffer size when streaming data into storage (now uses 16KiB, previously used 4KiB which is quite small)
- fix presigned S3 URL cache call (fixes #1279)
- changes test filenames / paths to fit with updated changes
- removes tee'd streaming of recache to both client and storage: in cases of slow client connections each tee stream could block a media worker pool function, which on instances with multiple users could very quickly exhaust available media workers
- run a disk storage clean operation (removes empty folders) on storage open

# Checklist


internal/storage/storage.go
```diff
@@ -64,8 +66,8 @@ func (d *Driver) URL(ctx context.Context, key string) *url.URL {
 	}
 
 	// access the cache member directly to avoid extending the TTL
-	if u, ok := d.PresignedCache.Cache.Get(key); ok {
-		return u.Value
+	if u, ok := d.PresignedCache.Get(key); ok {
```
theSuess, line 69: Have there been any changes I'm not seeing in the cache library? We had this at one point, but the issue is that accessing the cache will extend the TTL which results in returning expired presigned URLs #1208

internal/iotools/io.go
```diff
@@ -119,3 +119,41 @@ func (w *SilentWriter) Write(b []byte) (int, error) {
...
+func StreamReadFunc(read func(io.Reader) error) io.Writer {
+	// In-memory stream.
+	pr, pw := io.Pipe()
+
+	go func() {
+		var err error
+
+		defer func() {
+			// Always pass along error.
+			pr.CloseWithError(err)
```
tsmethurst, line 132: Wouldn't this always panic if the error is nil, or am I missing something?

tsmethurst:
Still reviewing but it looks good so far. Does it make sense to try to include https://github.com/superseriousbusiness/gotosocial/issues/1295 in this, or should that be a separate PR? Either is OK imo

internal/iotools/io.go
NyaaaWhatsUpDoc: nope the error gets converted by the underlying pipe if it's nil to either EOF or ErrClosedPipe :)

internal/storage/storage.go
NyaaaWhatsUpDoc: Ah understood, I hadn't remembered that previous PR. In that case I'll need to update the previous code to acquire a mutex lock (as the lack of lock was what was causing the panic).

NyaaaWhatsUpDoc:
> Still reviewing but it looks good so far. Does it make sense to try to include #1295 in this, or should that be a separate PR? Either is OK imo

it might as well be included in this PR tbh :)

tsmethurst (commented):
Looks fucking great in general. I have a few small comments, but this seems more or less ready to merge, to me. Thanks for your hard work on this!! <3

internal/processing/media/getfile.go
```diff
@@ -99,135 +98,59 @@ func (p *processor) getAttachmentContent(ctx context.Context, requestingAccount
...
-		// process it, then return the thumbnail data
-		data = func(innerCtx context.Context) (io.ReadCloser, int64, error) {
+		dataFn := func(innerCtx context.Context) (io.ReadCloser, int64, error) {
 			t, err := p.transportController.NewTransportForUsername(innerCtx, requestingUsername)
 			if err != nil {
 				return nil, 0, err
 			}
 			return t.DereferenceMedia(transport.WithFastfail(innerCtx), remoteMediaIRI)
 		}
-	} else {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: *pours my coffee on my foot*

for real though, it was some very nice code :heart:

actually brb gonna add a code comment referencing this

internal/media/types.go
```diff
@@ -24,13 +24,6 @@ import (
 	"time"
 )
 
-// maxFileHeaderBytes represents the maximum amount of bytes we want
-// to examine from the beginning of a file to determine its type.
-//
-// See: https://en.wikipedia.org/wiki/File_format#File_header
-// and https://github.com/h2non/filetype
-const maxFileHeaderBytes = 261
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: you're right i'll include this :)

internal/api/fileserver/servefile.go
```diff
@@ -118,13 +116,13 @@ func (m *Module) ServeFile(c *gin.Context) {
 	}
 
 	// try to slurp the first few bytes to make sure we have something
-	b := bytes.NewBuffer(make([]byte, 0, 64))
-	if _, err := io.CopyN(b, content.Content, 64); err != nil {
-		err = fmt.Errorf("ServeFile: error reading from content: %w", err)
-		apiutil.ErrorHandler(c, gtserror.NewErrorNotFound(err, err.Error()), m.processor.InstanceGet)
-		return
-	}
+	// b := bytes.NewBuffer(make([]byte, 0, 64))
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: i forgot i had commented this out (was doing it for testing). i added code back in (modified a little) if you could check it over pls :3

PR #1297, opened 2023-01-03
[feature] HTTP request throttling middleware

tsmethurst:
# Description

>

This pull request implements http request throttling middleware for GoToSocial ActivityPub endpoints (inboxes, user endpoints, etc), client API endpoints, and web endpoints. See the added docs for details of how this is configured + performed.

I've been running this on goblin.technology for a week or so now, and it's successfully prevented the instance from becoming unresponsive when interacting with accounts with many thousands of followers.

This relates to https://github.com/superseriousbusiness/gotosocial/issues/1205 -- it doesn't close the issue per se, but it should mitigate some of the problems described there.

## Checklist


internal/middleware/throttling.go
```diff
@@ -0,0 +1,155 @@
...
+
+	var (
+		limit           = cpuMultiplier * cpus
+		backlogLimit    = cpuMultiplier * limit
+		tokens          = make(chan token, limit)
+		backlogTokens   = make(chan token, limit+backlogLimit)
+		retryAfter      = "30" // seconds
+		backlogDuration = 30 * time.Second
+	)
+
```
tsmethurst, line 98: @NyaaaWhatsUpDoc should i log something here about the limits that are being used?

docs/api/throttling.md
```diff
@@ -0,0 +1,33 @@
...
+This leads to the following values for the default multiplier (8):
+
+```
+1 cpu = 08 in-process, 064 backlog
+2 cpu = 16 in-process, 128 backlog
+4 cpu = 32 in-process, 256 backlog
+8 cpu = 64 in-process, 512 backlog
+```
+
+New requests that overflow the in-process limit are held in the backlog queue, and processed as soon as a spot is freed up (ie., when a currently in-process request is finished). Requests that cannot be processed, and cannot fit in the backlog queue will be responded to with http code [429 - Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429), and the `retry-after` header will be set to `30` (seconds), to indicate that the caller should try again later.
```
Sentynel, line 21: Is this correct? 4xx errors are user errors, and specifically the definition of 429 "indicates _the user_ has sent too many requests in a given amount of time" (emphasis mine). But that's not necessarily what's happening here - it's more likely that _other_ users have sent too many requests. A 503 "indicates that the server is not ready to handle the request. Common causes are a server that is down for maintenance or that _is overloaded_."  `retry-after` is still accepted on a 503 response.

This might be spec pedantry, but on the other hand we don't want a situation where (say) somebody is developing a GoToSocial client or debugging a federation from another server and assumes that _they're_ failing to respect a rate limit when they get a 429 error.

Sentynel:
This looks very sensible to me - good work.

internal/middleware/throttling.go
NyaaaWhatsUpDoc: Given we're not doing this for rate limits yet i wouldn't worry about it :)

internal/middleware/throttling.go
```diff
@@ -0,0 +1,155 @@
...
+// write a JSON error into the response, set an appropriate Retry-After value, and set
+// the HTTP response code to 429: Too Many Requests.
+//
+// If the multiplier is <= 0, a noop middleware will be returned instead.
+//
+// Useful links:
+//
+//   - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After
+//   - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429
+func Throttle() gin.HandlerFunc {
```
NyaaaWhatsUpDoc, line 81: it's probably worth updating this to follow the same use as the rate limit middleware. so it accepts a limit argument, then initialize the `clThrottle` `s2sThrottle` and `fsThrottle` in `server.go` and pass them in as shared middleware in the same way as the rate limit one. what do you think?

internal/middleware/throttling.go
```diff
@@ -0,0 +1,155 @@
...
+		limit           = cpuMultiplier * cpus
+		backlogLimit    = cpuMultiplier * limit
+		tokens          = make(chan token, limit)
+		backlogTokens   = make(chan token, limit+backlogLimit)
+		retryAfter      = "30" // seconds
+		backlogDuration = 30 * time.Second
+	)
+
+	// prefill token buckets
+	for i := 0; i < limit+backlogLimit; i++ {
```
NyaaaWhatsUpDoc, line 100: might be a little simpler here to just have 2 for loops, one to fill the tokens channel and the second to fill the backlog tokens channel

docs/api/throttling.md
NyaaaWhatsUpDoc: This seems reasonable imo. While a rate limit _is_ a user error, throttling is an _us_ error.

tsmethurst: agreed! changed here: https://github.com/superseriousbusiness/gotosocial/pull/1297/commits/db8cb40119e148b7f890e98ba2f5a12a1fdb2ef9

internal/middleware/throttling.go
tsmethurst: Sounds good, I'll change it :)

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1297/commits/d39d7bf87f54e1502740e39de8bbfc55ceb738e0 done!

internal/middleware/throttling.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/1297/commits/79f66cc950371e272935ca3d3742177602a0bb01

Issue #1309, opened 2023-01-06
[feature] Allow GoToSocial to federate user reports (Flag) in and out


tsmethurst:
GoToSocial should be able to handle reports coming from other ActivityPub implementations via the Flag activity, and should also be able to send reports to other instances too. Admins of GtS instances should be able to view + act on received reports.

Task list for this:

- [x] Define + implement new database models https://github.com/superseriousbusiness/gotosocial/pull/1310
- [x] Implement federation of incoming/outgoing reports (Flag activity type)
  - [x] Incoming https://github.com/superseriousbusiness/gotosocial/pull/1382
  - [x] Outgoing https://github.com/superseriousbusiness/gotosocial/pull/1386
- [x] Client API endpoints to create reports https://github.com/superseriousbusiness/gotosocial/pull/1330
- [x] Client API endpoints for admins to interact with reports https://github.com/superseriousbusiness/gotosocial/pull/1378
- [x] Email notifications for admins
- [x] Document how GoToSocial federates reports in/out https://github.com/superseriousbusiness/gotosocial/pull/1393

PR #1310, opened 2023-01-06
[feature] Implement Report database model and utility functions

tsmethurst:
# Description

>

This pull request implements `gtsmodel.Report` to allow reports to be stored in the database.

This is modeled fairly closely on the report model used by Mastodon, since that seems to provide prettymuch all we need.

Reports are not actually used anywhere yet, this just adds them to the codebase (and adds a db table for them as a migration) to pave the way for upcoming report federation implementation.

Implements part of https://github.com/superseriousbusiness/gotosocial/issues/1309

## Checklist


internal/db/bundb/migrations/20230105171144_report_model.go
```diff
@@ -0,0 +1,75 @@
...
+				Index("report_uri_idx").
+				Column("uri").
+				Exec(ctx); err != nil {
+				return err
+			}
+
+			if _, err := tx.
+				NewCreateIndex().
+				Model(&gtsmodel.Report{}).
+				Index("report_account_id_idx").
```
NyaaaWhatsUpDoc, line 47: we're not using these lookup IDs are we? it's probably worth adding them as we need them just to make sure we don't end up with indices we don't need

tsmethurst: We're not in this PR, but we will be pretty soon in upcoming ones:

`report_uri_idx` -- will (probably) be used in the federating db to check if we already have the report in the db
`report_account_id_idx` -- will be used so that accounts can see all the reports they've submitted
`report_target_account_id_idx` -- will be used so that admins can see all reports that target a certain account

I'm happy to remove these for now and add them in another PR if you prefer though, all good :)

tsmethurst: Ah on second thought, I don't think report_uri_idx will be used (since we'll use a conflict to check if it's already in the db, rather than trying to select it first). So I will remove that one straight away.

NyaaaWhatsUpDoc: does that mean we can now also remove the "GetReportByURI" method and it's associated registered cache lookup??

tsmethurst: Right! Good point, forgot to clean those up too, will do that

internal/db/bundb/report.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+
+	return report, nil
+}
+
+func (r *reportDB) PutReport(ctx context.Context, report *gtsmodel.Report) db.Error {
+	return r.state.Caches.GTS.Report().Store(report, func() error {
+		_, err := r.conn.NewInsert().Model(report).Exec(ctx)
+		return err
```
NyaaaWhatsUpDoc, line 103: need to process the error here :)

tsmethurst: like this? https://github.com/superseriousbusiness/gotosocial/pull/1310/commits/51794a0aa9456d797d143cc2ce3844f3a4fcac06

PR #1315, opened 2023-01-06
[bugfix] return early in websocket upgrade handler

### LINKED ISSUE #1314: [bug] Open websockets shouldn't count towards connection throttling limit

Just gonna copy kim's comments from Matrix in here:

> i just took a look, and it basically does net/http hijack on the incoming UPGRADE request connection
> soooo, there might actually be an easy way of handling this
> we're currently holding open the main request handler with the streaming loop
> but if the net/http hijack logic allows us to return that handler and perform the stream loop in another goroutine, that would mean throttling won't be an issue with websockets :)
> in fact if that is the case, then that would be advantageous as it still allows us to throttle the streaming UPGRADE endpoint instead of completely ignoring it
> by the looks of things it should be fine to return from the upgrade handler: https://pkg.go.dev/net/http#Hijacker
> 
> basiccally all we should need to do is
> 
> ```go
> 1234567891011go func() {
>     // create new websocket context as the original net/http
>     // request context is only valid until the calling handler returns.
>     ctx, cncl := context.WithCancel(context.Background())
>     defer cncl()
> 
>     for {
>         // do some shit with wsConn, yada yada ...
>     }
> }()
> ```
> 
> and the logic from here onwards can go in that goroutine: https://github.com/superseriousbusiness/gotosocial/blob/0dbe6c514fb742b727694a6c8152fbd2a9caf035/internal/api/client/streaming/stream.go#L196

git

### GoToSocial Arch

_No response_

### Browser version

_No response_

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

NyaaaWhatsUpDoc:
Fixes #1314 

### DIFF
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -19,8 +19,8 @@
 package streaming
 
 import (
+	"errors"
 	"fmt"
-	"net/http"
 	"time"
 
 	"codeberg.org/gruf/go-kv"
@@ -30,16 +30,6 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/gorilla/websocket"
-)
-
-var (
-	wsUpgrader = websocket.Upgrader{
-		ReadBufferSize:  1024,
-		WriteBufferSize: 1024,
-		// we expect cors requests (via eg., pinafore.social) so be lenient
-		CheckOrigin: func(r *http.Request) bool { return true },
-	}
-	errNoToken = fmt.Errorf("no access token provided under query key %s or under header %s", AccessTokenQueryKey, AccessTokenHeader)
 )
 
 // StreamGETHandler swagger:operation GET /api/v1/streaming streamGet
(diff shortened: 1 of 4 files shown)


internal/api/client/streaming/stream.go
```diff
@@ -150,21 +141,20 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
...
-		accessToken = t
-	} else {
-		// no token
-		err := errNoToken
-		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGet)
-		return
+	var token string
+
+	// First we check for a query param provided access token
+	if token = c.Query(AccessTokenQueryKey); token == "" {
```
tsmethurst, line 147: i learn so much from reading your code kimbe, this little bit is just an example of that: it hadn't occurred to me to nest the checks like this, and also didn't think of using `const` to hint to the compiler like this

tsmethurst:
Looks good to me! Feel free to merge if you're happy with it :) Thanks for fixing my rather slapdash code :')

PR #1336, opened 2023-01-13
[feature] Inherit resource limits from cgroups

daenney:
# Description

When GTS is running in a container runtime which has configured CPU or memory limits or under an init system that uses cgroups to impose CPU and memory limits the values the Go runtime sees for GOMAXPROCS and GOMEMLIMIT are still based on the host resources, not the cgroup (if the cgroup has such limits set).

At least for the throttling middlewares which use GOMAXPROCS to configure their queue size, this can result in GTS running with values too big compared to the resources that will actuall be available to it.

This introduces 2 dependencies which can pick up resource contraints from the current cgroup and tune the Go runtime accordingly. This should result in the different queues being appropriately sized and in general more predictable performance. These dependencies are a no-op on non-Linux systems or if running in a cgroup that doesn't set a limit on CPU or memory.

The automatic tuning of GOMEMLIMIT can be disabled by either explicitly setting GOMEMLIMIT yourself or by setting AUTOMEMLIMIT=off. The automatic tuning of GOMAXPROCS can similarly be counteracted by setting GOMAXPROCS yourself.

## Checklist


cmd/gotosocial/action/server/server.go
```diff
@@ -54,10 +55,19 @@ import (
...
+	_ "github.com/KimMachineGun/automemlimit"
 )
 
 // Start creates and starts a gotosocial server
 var Start action.GTSAction = func(ctx context.Context) error {
+	maxprocsUndo, err := maxprocs.Set(maxprocs.Logger(log.Errorf))
+	if err != nil {
+		return fmt.Errorf("failed to set CPU limits from cgroup: %s", err)
+	}
+	defer maxprocsUndo()
```
NyaaaWhatsUpDoc, line 69: literally my only comment for this PR is I don't think we need to undo the memory limit here since on return we'll be exiting the Go runtime (well, about to). otherwise good to go!

daenney: Done!

Issue #1335, opened 2023-01-13
[feature] Support WAL mode for sqlite databases.


zaphar:
### Is your feature request related to a problem ?

Currently gotosocial opens sqlite databases in shared cache mode. This mode is considered deprecated and also conflicts with it's recommended replacement: WAL mode. See https://www.sqlite.org/sharedcache.html for details.

This prevents you from using a tool like litestream to keep a backup of your sqlite database since it requires WAL mode to work.

The line here: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/db/bundb/bundb.go#L243 forces shared_cache mode on. I would happily put together a PR to change it but I'm not exactly sure why it was set in the first place or if there are issues with using sqlite WAL mode instead.

I'm not super familiar with the pure go modernc port either so if that doesn't support this then this may be a non-starter for the moment.

### Describe the solution you'd like.

Ideally gotosocial would be able to open and use sqlite DBs in WAL mode and cache=shared would be a configuration option instead of required.

### Describe alternatives you've considered.

Gotosocial works just fine without this and you could cobble together backups using other methods. However litestream is probably the best solution out there for more robust sqlite backup solutions and makes the operational overhead of sqlite much easier to manage.

tsmethurst:
this will be closed by #1247 when we finish it :)

zaphar:
@tsmethurst Awesome. If I can help even just with testing let me know.

PR #1349, opened 2023-01-16
[feature] Tune sqlite pragmas

tsmethurst:
# Description

>

This pull request implements sqlite pragma tuning, based on https://github.com/superseriousbusiness/gotosocial/pull/1247 but using the current config structure (since the other method has trouble reading environment variables for nested config values).

When sqlite is used, then the following pragmas will be used by default:

```
# String. SQLite journaling mode.
# SQLite only -- unused otherwise.
# See: https://www.sqlite.org/pragma.html#pragma_journal_mode
db-sqlite-journal-mode: "WAL"

# String. SQLite synchronous mode.
# SQLite only -- unused otherwise.
# See: https://www.sqlite.org/pragma.html#pragma_synchronous
db-sqlite-synchronous: "NORMAL"

# Byte size. SQlite cache size.
# SQLite only -- unused otherwise.
# See: https://www.sqlite.org/pragma.html#pragma_cache_size
db-sqlite-cache-size: "64MiB"
```

The PR also disables `cache=shared` when not running using an in-memory database. It's kept for in-memory for testing purposes.

I'm testing the PR on goblin.technology currently. It seems *mostly* ok, but i am getting some locked errors every couple minutes:

```
Jan 16 13:47:40 goblin.technology gotosocial[505318]: timestamp="16/01/2023 13:47:40.327" func=concurrency.(*WorkerPool).Queue.func1 level=ERROR type=worker.Worker[*media.ProcessingMedia] error="error loading media redacted: database is locked (5) (SQLITE_BUSY)" msg="message processing error"
Jan 16 13:54:02 goblin.technology gotosocial[505318]: timestamp="16/01/2023 13:54:02.676" func=concurrency.(*WorkerPool).Queue.func1 level=ERROR type=worker.Worker[messages.FromFederator] error="unexpected error: GetRemoteStatus: error populating status fields: populateStatusFields: error populating status mentions: populateStatusMentions: error creating mention: database is locked (5) (SQLITE_BUSY)" msg="message processing error"
Jan 16 13:58:19 goblin.technology gotosocial[505318]: timestamp="16/01/2023 13:58:19.197" func=concurrency.(*WorkerPool).Queue.func1 level=ERROR type=worker.Worker[messages.FromFederator] error="unexpected error: GetRemoteStatus: error populating status fields: populateStatusFields: error populating status mentions: populateStatusMentions: error creating mention: database is locked (5) (SQLITE_BUSY)" msg="message processing error"
Jan 16 13:59:38 goblin.technology gotosocial[505318]: timestamp="16/01/2023 13:59:38.835" func=concurrency.(*WorkerPool).Queue.func1 level=ERROR type=worker.Worker[messages.FromFederator] error="unexpected error: GetRemoteStatus: couldn't get status author: database error during dereferencing: GetRemoteAccount: error updating remoteAccount: database is locked (5) (SQLITE_BUSY)" msg="message processing error"
```

For more info on the above errors, see https://www.sqlite.org/wal.html#sometimes_queries_return_sqlite_busy_in_wal_mode

I believe they're being caused by messages being delivered to more than one account at a time, and running multiple concurrent processes to update the same entry. We will probably need to add some kind of locking here to fix this before merging this PR.... @NyaaaWhatsUpDoc did you also run into this problem when testing, or did it not show on a single-user instance?

closes https://github.com/superseriousbusiness/gotosocial/issues/1335

## Checklist


internal/db/bundb/bundb.go
```diff
@@ -127,26 +113,34 @@ func doMigration(ctx context.Context, db *bun.DB) error {
...
+		return nil, fmt.Errorf("database type %s not supported for bundb", t)
 	}
 
 	// Add database query hook
 	conn.DB.AddQueryHook(queryHook{})
 
+	// execute sqlite pragmas *after* adding database hook;
+	// this allows the pragma queries to be logged
+	if t == "sqlite" {
+		if err := sqlitePragmas(ctx, conn); err != nil {
```
NyaaaWhatsUpDoc, line 139: if you make the sqlitePragmas() function accept the `sql.DB` type (or whatever it is), then you can do this part in the above switch ^. Just saves needing to add extra checks (also furthers our reliance on bun :p)

NyaaaWhatsUpDoc:
regarding the DB locked errors, my solution has been setting the `busy_timeout` parameter. It accepts values in milliseconds, so I'd have a configuration parameter of type `time.Duration` that you then  divide the value of by `time.Millisecond`.

internal/db/bundb/bundb.go
```diff
@@ -278,11 +271,48 @@ func sqliteConn(ctx context.Context) (*DBConn, error) {
...
+	}
+
+	if mode := config.GetDbSqliteSynchronous(); mode != "" {
+		// Set the user provided SQLite synchronous mode
+		pragmas = append(pragmas, []string{"synchronous", mode})
+	}
+
+	if size := config.GetDbSqliteCacheSize(); size > 0 {
+		// Set the user provided SQLite cache size (in kibibytes)
+		s := strconv.Itoa(int(size / bytesize.KiB))
```
NyaaaWhatsUpDoc, line 293: Since bytesize.Size is a uint64 alias, it probably makes more sense here to use `strconv.FormatUint()` just to prevent the (incredibly rare) chance of value truncation

internal/db/bundb/bundb.go
tsmethurst: Ah, I had something like that previously, but the issue is that then you don't get the queries logged nicely as part of the db hook, because the db hook is always set up after the db-type-specific stuff. Is there another way of doing it that i'm missing, which still lets you get the queries logged as part of our normal hook logging?

NyaaaWhatsUpDoc: ah yeah that is a fair enough :)

PR #1356, opened 2023-01-18
[feature] Don't expose `nodeinfo` by default

tsmethurst:
# Description

>

This PR makes the `.well-known/nodeinfo` and `/nodeinfo/*` endpoints turn-offable, and disabled by default.

Reasoning for this: some softwares like fediverse.observer scrape nodeinfo endpoints for, well, node info. This is the whole purpose of nodeinfo in the first place: https://nodeinfo.diaspora.software/ Unfortunately, this means that GtS instances end up getting indexed (along with masto instances etc) on services that admins might not be aware of. With this PR, enabling nodeinfo becomes explicitly an opt-in thing.

Unfortunately, this extra privacy feature does break some clients that use nodeinfo to determine what kind of server software an instance is running. In particular, elk.zone (which checks nodeinfo to see if a server is a GtS instance).

## Checklist


daenney:
Since we know this may break some things, should it log something on startup to draw admin's attention to that (which in turn you may want to be able to turn off through a setting)?

internal/web/robots.go
```diff
@@ -57,6 +55,16 @@ Disallow: /settings/`
...
 // depending on user preferences (see internal/web).
-func (m *Module) robotsGETHandler(c *gin.Context) {
-	c.String(http.StatusOK, robotsTxt)
+func (m *Module) robotsGETHandler() gin.HandlerFunc {
+	r := robotsTxt
+
+	if !config.GetInstanceExposeNodeinfo() {
+		r += `
+# metadata
+Disallow: /nodeinfo/`
```
NyaaaWhatsUpDoc, line 64: I think we might as well just always leave this in the robots text. Then it means people can't scan the gts robots texts to determine if things are enabled (or if they've say blocked access via nginx)

NyaaaWhatsUpDoc:
> Since we know this may break some things, should it log something on startup to draw admin's attention to that (which in turn you may want to be able to turn off through a setting)?

if we update the router wrapper type we have, we can have it do a `log.Info()` for registering each route and it should make situations like this evident :thinking: (also quite nice to see these things on startup, or maybe that's just me)

tsmethurst:
> > Since we know this may break some things, should it log something on startup to draw admin's attention to that (which in turn you may want to be able to turn off through a setting)?
> 
> if we update the router wrapper type we have, we can have it do a `log.Info()` for registering each route and it should make situations like this evident thinking (also quite nice to see these things on startup, or maybe that's just me)

I'm down with that, but this feels out of scope for this PR. How about I add a warning log for now, and in future we can do log.Info() for mounting endpoints?

internal/web/robots.go
tsmethurst: The only thing is that if it's always in, then good bots will never scan nodeinfo, whereas some instance admins might actually want that. You see what I mean?

tsmethurst:
I'm considering changing this PR so that nodeinfo is still mounted if `instance-expose-nodeinfo` is false, but just requires an authorization token. Similar to what we do already for `/api/v1/instance/peers`. That way, clients that use nodeinfo to guess capabilities can still access it, as long as they provide an authorization token in the header as with other API requests. What do you think @NyaaaWhatsUpDoc and @daenney ?

daenney:
By authorization token you mean the usual HTTP signature scheme employed with ActivityPub? I think that would work out in practice, since by the time a client would issue this request they'll probably have a logged in user already.

tsmethurst:
> HTTP signature scheme employed with ActivityPub

Nah, just a standard oauth bearer token that logged in users already use for API endpoints

daenney:
I might be missing something, but how does that work? That would be a token from a different instance, so that's validated with the other instance?

daenney:
I just realised I never actually answered the question. I like the idea of having nodeinfo exposed but requiring auth by default, which you can then disable if you want to.

That sounds like it would probably work out well enough for clients too, though based on some investigation in #1609 I've yet to see a client use it in practice. It seems everyone uses `/api/v1/instance` in actuality which makes sense since that exposes much more information.

With that in mind, I'm not sure closing off nodeinfo ends up improving things much from a privacy point of view. The same info that's available in nodeinfo is available in /api/v1/instance and then some and that endpoint is expected to be unauthenticated.

daenney:
I took a look at what Elk does with its trying to detect GtS through nodeinfo. The only piece of code I can find that's guarded behind a GtS check is https://github.com/elk-zone/elk/blob/3732a2cc16b4de6cd60f4836151d4a3292c627fa/composables/cache.ts#L72-L75. It might just be easiest for us to support the other query method for the account too? 

Then we would at least not break any known clients by making nodeinfo private/authed.

tsmethurst:
>  I'm not sure closing off nodeinfo ends up improving things much from a privacy point of view.

Indeed, I'm considering just closing this PR and we can come back to it at a later point perhaps when we've thought about it more...

tsmethurst:
Closing this for now, we can come back to it later if desired.

PR #1357, opened 2023-01-18
[chore] Settings refactor fix 2

f0x52:
Three small fixes building on `settings-refactor`:

- 'Emoji' RTK Query tag type got lost, re-added and renamed
- Replace 'localhost' with '127.0.0.1' in webdev setup proxy, because GTS doesn't listen on ::1
- Fix issue where login flow quickly flashed "expecting callback" error before redirecting to the instance login flow

### DIFF
--- a/web/source/index.js
+++ b/web/source/index.js
@@ -48,7 +48,7 @@ skulk({
 	},
 	servers: {
 		express: {
-			proxy: "http://localhost:8081",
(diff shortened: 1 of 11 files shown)


tsmethurst:
Feel free to merge if you're happy with it :) I'm out of spoons to review today but I trust ya

Issue #378, opened 2022-02-03
[feature] Opt-in public viewable instance suspension list


f0x52:
Similar to a Mastodon instance's /about/more, should be off by default I think (and enabled through the admin panel or cli)
Also would be nice to have a plaintext or json export so others can easily import

Issue #1360, opened 2023-01-19
[feature] API routes / database model for instance settings


f0x52:
Currently the closest thing we have to instance settings (also labelled as such in the admin settings) only relates to modifying the `/api/v1/instance` response, and the routes are named accordingly (GET and PATCH on that endpoint).

I think it would be more correct to label this "Instance information", and have a separate section + api for actual 'settings', currently prompted by needing a toggle for publishing the instance blocklist, but we'll undoubtedly have more settings like this in the future.

AFAIK there's not really prior art for naming this route, given Mastodon has no documented API for most admin things.

Having GET,PATCH `/api/v1/admin/settings` would work well enough I think, both only accessible for authorized admins, although some of the properties might also be exposed in the `/api/v1/instance` section `configuration` for use in clients etc.

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/1360.

PR #1362, opened 2023-01-19
[feature] Public list of suspended domains

f0x52:
# Description

Adds a opt-in config 'InstanceExposeSuspendedWeb' for serving a list of suspended domains on `/about/suspended`. 
fixes https://github.com/superseriousbusiness/gotosocial/issues/378

Ideally this will later be configurable in the settings interface, but that depends on https://github.com/superseriousbusiness/gotosocial/issues/1360

image

## Checklist

NLnet Milestone 10.a: Page for displaying block lists (if instance allows showing it)


internal/web/domain-blocklist.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	host := config.GetHost()
+	instance, err := m.processor.InstanceGet(c.Request.Context(), host)
+	if err != nil {
+		apiutil.ErrorHandler(c, gtserror.NewErrorInternalError(err), m.processor.InstanceGet)
+		return
+	}
+
+	domainBlocks, errWithCode := m.processor.AdminDomainBlocksGet(c.Request.Context(), authed, false)
```
tsmethurst, line 49: Rather than using AdminDomainBlocksGet, I think it's better to use InstancePeersGet with `includeSuspended` set to `true`. That function only includes public comments, and also does obfuscation of domains as desired.

Only downside is that it also requires `instance-expose-suspended` to be `true`.

web/source/settings/admin/emoji/local/overview.js
```diff
@@ -26,27 +26,35 @@ const NewEmojiForm = require("./new-emoji");
 const query = require("../../../lib/query");
```
tsmethurst, line 26: should the changes in this file be included in this pr? looks like it might have slipped in from something else you're working on :P

web/template/domain-blocklist.tmpl
```diff
@@ -0,0 +1,48 @@
...
+*/ -}}
+
+{{ template "header.tmpl" .}}
+<main>
+	<section>
+		<h1>Suspended Instances</h1>
+		<p>
+			The following list of domains have been blocked by the administrator(s) of this server.
+			All current and future accounts on that instance are blocked, and no data is federated to the remote server.
+			This extends to all subdomains as well, so an entry for 'example.com' includes 'social.example.com' as well.
```
tsmethurst, line 26: i'd remove the first 'as well' in this sentence.

tsmethurst (commented):
The new page and its styling look really nice. Just a couple comments, lemme know what you think :)

internal/web/domain-blocklist.go
f0x52: I would prefer having them as separate settings, because the api version is that much easier to scrape. Maybe abstract the current InstancePeersGet code for it into something that can be used in both then

web/source/settings/admin/emoji/local/overview.js
f0x52: oh yeah, I was going to PR those separately then rebase after, will fix

internal/web/domain-blocklist.go
tsmethurst: Alright, I can make a little fix for this if you want :) Should I do it on this branch, or a separate one?

turtlegarden:
this looks amazing! Until now I used a hacky solution and this looks much better

f0x52:
Rebased, good to go now!

internal/web/web.go
```diff
@@ -100,6 +100,10 @@ func (m *Module) Route(r router.Router, mi ...gin.HandlerFunc) {
 	r.AttachHandler(http.MethodGet, confirmEmailPath, m.confirmEmailGETHandler)
 	r.AttachHandler(http.MethodGet, robotsPath, m.robotsGETHandler)
 
+	if config.GetInstanceExposeSuspendedWeb() {
+		r.AttachHandler(http.MethodGet, domainBlockListPath, m.domainBlockListGETHandler)
```
tsmethurst, line 104: Since the logic inside the handler already deals with the unauthorized case, I think you can remove this check and always attach the handler.

internal/web/domain-blocklist.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (m *Module) domainBlockListGETHandler(c *gin.Context) {
+	authed, err := oauth.Authed(c, false, false, false, false)
+	if err != nil {
+		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGet)
+		return
+	}
+
+	if !config.GetInstanceExposeSuspendedWeb() && (authed.Account == nil || authed.User == nil) {
+		err := fmt.Errorf("peers open query requires an authenticated account/user")
```
tsmethurst, line 44: this error message needs to change

tsmethurst (commented):
almost there!

PR #1369, opened 2023-01-22
[chore] set max open / idle conns + conn max lifetime for both postgres and sqlite

tsmethurst:
# Description

>

This PR updates some of our database tuning. Previously we were only setting maxopenconns and maxidleconns for postgres, and only setting connmaxlifetime(0) for in-mem sqlite connections, to avoid the db being deleted on connection cleanup.

Now, this tuning has been combined and is done for both database types. This is partially because we should have already been setting maxopenconns and maxidleconns for sqlite but weren't, and also to see if some SQLITE_BUSY errors can be resolved by never letting sqlite connections die (see https://www.sqlite.org/wal.html#sometimes_queries_return_sqlite_busy_in_wal_mode)

## Checklist

- [x] I/we have added tests that cover new code. -- n/a

### DIFF
--- a/internal/db/bundb/bundb.go
+++ b/internal/db/bundb/bundb.go
@@ -130,6 +130,12 @@ func NewBunDBService(ctx context.Context, state *state.State) (db.DB, error) {
 		return nil, fmt.Errorf("database type %s not supported for bundb", t)
 	}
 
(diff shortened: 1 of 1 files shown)


tsmethurst:
This seems pretty decent now so i'm gonna merge and we can still play around with some of the values on main as necessary! :)

PR #1370, opened 2023-01-22
Advertise rich text formats, support content_type field

### LINKED ISSUE #1363: [feature] Status content_type param compatible with Glitch, Hometown, Pleroma

As a native client dev, I want to write a native client with rich text support that works with as many instance server implementations as possible, including GotoSocial. Unfortunately GotoSocial does rich text support a little differently from other Fedi servers (see #110 for where it was implemented). 

The Mastodon forks Glitch and Hometown, as well as Pleroma and its forks, accept an additional `content_type` parameter on the `POST /api/v1/statuses` endpoint, which is expected to be a MIME type. Glitch supports `text/plain`, `text/markdown`, and `text/html`, but does not provide a way to get a list of supported types. Pleroma and its forks may support more (for example, BBCode and Misskey-flavored Markdown), and declare which types they support in ad-hoc Pleroma-specific extensions to `/api/v1/instance` and `/.well-known/nodeinfo`.

The `POST /api/v1/statuses` endpoint should accept a `content_type` parameter instead of or in addition to `format`, which would be a full MIME type: currently `text/plain` or `text/markdown`, equivalent to `format` values `plain` and `markdown` respectively. If both are present, GtS should use `content_type` and ignore `format`.

There are workarounds:
- Clients can try to detect GotoSocial and switch which parameter they use to set the post's content type. This involves some additional client complexity.
- Both GtS and other Fedi servers also support setting a default content type per user, so that rich-text-unaware clients can send Markdown as plain text but have it interpreted as Markdown on the server. This is already the status quo, but doesn't allow selecting the post's content type on a per-post basis.

### Additional context.

I'd also like to propose an API extension for _detecting_ that a given Fedi server with a Mastodon-compatible API supports rich text authoring, and which formats it accepts. This is out of scope for this feature request so I'm mentioning it here only for background.

My ideal extension would be to add a `supported_mime_types` list to the `.configuration.statuses` object in `/api/v1/instance` and `/api/v2/instance`, alongside `max_characters`, analogous to the existing `.configuration.media_attachments.supported_mime_types` list. I plan to at least ask about getting this into Glitch, Hometown, and Akkoma as well as GtS.

Alternatively, Fedi servers could use the `.pleroma.metadata.post_formats` extension to the same API, which already exists for Pleroma and its forks, but kinda has the same aesthetic problem as vendor-prefixed namespaces in CSS.

VyrCossont:
# Description

This pull request adds two features:
- Accepts a `content_type` field when creating a status (this is the more standard version of `format`, as used by Glitch, Pleroma, the `toot` command-line client…)
- Reports supported status content types in `/api/v1/instance` as the list `.configuration.statuses.supported_mime_types` (as Glitch does now: glitch-soc/mastodon#2090)

Closes #1363.

## Checklist


tsmethurst:
This looks great :) If I'd done my homework and realized glitch parsed status format this way, I would have implemented it this way in the first place.

As far as I know, nobody really uses that `format` field, since most people--at least, anecdotally--just set their default content type via the settings panel and let their posts get parsed using that. So it might be worth just removing that `format` field entirely and always using `content_type` instead. @NyaaaWhatsUpDoc do you have any onions on this?

VyrCossont:
On the chance we want to do it that way, I've prepped 3b41ab8. If not, we can revert this patch to 64e3c0d. Either way is fine with me.

tsmethurst:
Thank you! Will let you know after we've cut 0.7.0 at some point soon (hopefully)

tsmethurst:
Hey @VyrCossont ! Since 0.7.0 (and indeed 0.7.1!) is out now, let's get this working. Do you want me to do the rebasing?

VyrCossont:
@tsmethurst sure, thank you!

tsmethurst:
Would you mind giving me permissions to update this branch? It's a bit easier than checking it out and making another pr: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

VyrCossont:
@tsmethurst Done. You should be able to edit this now.

tsmethurst:
Alright, think I got this up to date without breaking anything :D

Issue #1358, opened 2023-01-19
[feature] RSS Autodiscovery


kvibber:
### Is your feature request related to a problem ?

RSS readers (and some browsers, like Vivaldi) can detect when a feed is available based on a link rel=alternate in the html page's header. GTS currently does not include this.

### Describe the solution you'd like.

When a user has RSS enabled, add something like this to their profile page:

```
<link rel="alternate" type="application/rss+xml" href="/path/to/rss/feed" title="Username on GoToSocial" />
```

### Describe alternatives you've considered.

none

### Additional context.

I've taken a look at the template code and I'm willing to implement this myself and submit a PR if it sounds like something you'd want included.

tsmethurst:
This sounds very reasonable! Feel free to work on a PR for this if you like :)

kvibber:
Great! I'll give it a shot!

tsmethurst:
closed by https://github.com/superseriousbusiness/gotosocial/pull/1373

PR #1373, opened 2023-01-23
Add RSS autodiscovery on profiles that enable RSS

kvibber:
One-line change, implements this feature request:

https://github.com/superseriousbusiness/gotosocial/issues/1358

This adds a <link> element to profiles that have RSS feeds so that browsers and feed readers can automatically discover it.

Tested using the testrig and Vivaldi. As expected, the RSS icon appeared in the URL bar when viewing the profiles for the_mighty_zork and admin, but not 1happyturtle, and not on individual posts or the instance home page.


tsmethurst:
Great stuff, thanks for the change :)

PR #1386, opened 2023-01-25
[feature] Federate reports to remote instance as Flag (if desired)

tsmethurst:
# Description

>

This pull request implements outgoing federation of reports as the ActivityStreams `Flag` type.

Outgoing Flags look like this:

```json
{
  "actor": "http://example.org/users/example.org",
  "content": "dark souls sucks, please yeet this nerd",
  "id": "http://example.org/reports/01GP3AWY4CRDVRNZKW0TEAMB5R",
  "object": [
    "http://fossbros-anonymous.io/users/foss_satan",
    "http://fossbros-anonymous.io/users/foss_satan/statuses/01FVW7JHQFSFK166WWKR8CBA6M"
  ],
  "type": "Flag"
}
```

Internally, the `bto` field is used to address a Flag to the reported account, but this is stripped by the go-fed library before the activity is sent out. This is in line with how other AP implementations handle delivery of Flags.

As with other implementations, Flags are sent out using **the instance account** rather than the account of the user that created the report. This provides partially anonymity, at least for instances with more than one user. This is also in line with what other fedi implementations do.

Because the username of the instance account doesn't conform to the strict username used by 'normal' accounts, a new relaxed username regex has been introduced as well, which includes the '.' character. This regex isn't used for account sign ups (so users can't put a period in their username, still), but is used when parsing URIs, to allow the functions hooked into go-fed to recognize the instance account as a valid account, and use it when sending.

relates to https://github.com/superseriousbusiness/gotosocial/issues/1309

## Checklist


internal/federation/federatingdb/get.go
```diff
@@ -20,7 +20,7 @@ package federatingdb
```
tsmethurst, line 20: Note to reviewers, the code in here is just a little clean up while I was looking through + debugging., it doesn't make much functional difference.

internal/federation/federatingdb/util.go
```diff
@@ -229,60 +229,56 @@ func (f *federatingDB) ActorForInbox(ctx context.Context, inboxIRI *url.URL) (ac
 }
```
tsmethurst, line 229: Note to reviewers, the code in here is just a little clean up while I was looking through + debugging., it doesn't make much functional difference.

internal/regexes/regexes.go
```diff
@@ -77,40 +77,45 @@ var (
...
 	// Username can be used to validate usernames of new signups
 	Username = regexp.MustCompile(fmt.Sprintf(`^%s$`, usernameString))
 
-	userPathString = fmt.Sprintf(`^?/%s/(%s)$`, users, usernameString)
+	// usernameStringRelaxed is like usernameString, but also allows the '.' character,
+	// so it can also be used to match the instance account, which will have a username
+	// like 'example.org', and it has no upper length limit, so will work for long domains.
+	usernameStringRelaxed = `[a-z0-9_\.]{2,}`
+
+	userPathString = fmt.Sprintf(`^?/%s/(%s)$`, users, usernameStringRelaxed)
```
Sentynel, line 90: There's an inconsistency in the regexes here - can I just check this is intentional? This and a couple of others start with `^?/` - an optional start of string, followed by a required slash - which seems like an odd construction (an optional start of string is redundant). The others start with `^/?` - that is, start of string, and an optional slash.

tsmethurst: oh whoops, indeed that's an error, the question mark should be after the slash, well spotted!

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1386/commits/2fb87bad7ca51ab5b4e0dabddf9f59f29c91acbf

internal/processing/fromclientapi.go
```diff
@@ -922,3 +931,51 @@ func (p *processor) federateUnblock(ctx context.Context, block *gtsmodel.Block)
...
+	if err != nil {
+		return fmt.Errorf("federateReport: error parsing outboxURI %s: %w", report.TargetAccount.URI, err)
+	}
+	bTo := streams.NewActivityStreamsBtoProperty()
+	bTo.AppendIRI(reportTargetURI)
+	flag.SetActivityStreamsBto(bTo)
+
+	// deliver the flag using the outbox of the
+	// instance account to anonymize the report
+	instanceAccount, err := p.db.GetInstanceAccount(ctx, "")
```
Sentynel, line 969: We're fetching the instance account twice here - `ReportToASFlag` does it as well. Maybe get it further up this function and pass it or the `instanceAccount.URI` into `ReportToASFlag`? Not the end of the world by any means so may not be worth making the code less clean.

tsmethurst: Yeah, I considered that too. In this case it's not too big of a deal, because the instance account is cached in memory at this point (so it avoids a second db call). Good spot though

PR #1390, opened 2023-01-26
[frogend] (Mastodon) domain block CSV import

f0x52:
# Description
Adds CSV import and export to the bulk import/export interface, tested to be compatible with Mastodon too (glitch-soc).
Generally cleans up the code structure around it a bit too, both on the page itself and the exportType based branching in query.

Also added some information text about the page, and a table comparing the different export formats. Still not sure how well it fits there, might just have to move that to the docs, or a separate page?

## Checklist


tsmethurst:
This looks really good to me :) Just wondering about what we discussed briefly in matrix: domains that are wildcarded by starting with `*.`. Should we trim that off before it gets sent to gts, or should the server do it?

f0x52:
hmm I feel like the server should reject them if it doesn't handle it for you, because as a literal * it's never a valid domain. If we don't want to handle them in the backend I can update the client validation to match, and possibly add the trimming there

I'm actually not sure if Mastodon even handles those wildcards either..

tsmethurst:
I can add something to trim it in the server side, then it'll work via the API too and not just the settings panel: I needed a small thing to work on this afternoon anyway :)

web/source/settings/admin/federation/import-export.js
```diff
@@ -94,49 +95,72 @@ module.exports = function ImportExport() {
...
 				</h1>
-				<FormWithData
-					dataQuery={query.useInstanceBlocksQuery}
-					DataForm={ImportList}
-					list={parseResult.data}
-				/>
+				<div className="without-border">
+					<FormWithData
+						dataQuery={query.useInstanceBlocksQuery}
+						DataForm={ImportList}
```
tsmethurst, line 101: very small nitpick, feel free to ignore: is there a reason for the different casing of DataForm?

web/source/settings/lib/domain-block.js
```diff
@@ -0,0 +1,50 @@
...
+			Wildcard prefix *. can be stripped since it's equivalent to not having it,
+			but wildcard anywhere else in the domain is not handled by the backend so it's invalid.
+		*/
+		wildcard: false,
+		allowUnicode: true
+	});
+}
+
+/* 
+	Still can't think of a better function name for this,
```
tsmethurst, line 36: `canBlockHigher`? i can't really think of one either; think it's fine as-is though

tsmethurst (commented):
This looks good to me so far :)

web/source/settings/admin/federation/import-export.js
f0x52: because it's a React component, which are recommended as PascalCase for some reason

PR #1395, opened 2023-01-28
Add an example on how to setup redirect with Traefik

ketan-vijayvargiya:
# Description

There is a section in the documentation that explains how admin can setup a redirect from `example.org` to `fedi.example.org` using Nginx. I wanted to achieve the same through Traefik. It took me about an hour to get it working, so thought of sharing it with others too.

## Checklist


ketan-vijayvargiya:
On a separate note, I can raise another PR to add a section on "Reverse proxy with Traefik" if you folks think it'll help. (This will be in addition to Nginx, Apache HTTP Server and Caddy 2.

tsmethurst:
Thank you!

> I can raise another PR to add a section on "Reverse proxy with Traefik" if you folks think it'll help

That would be very welcome :)

PR #1396, opened 2023-01-28
docs: Simplify Apache httpd proxy documentation

igalic:
# Description

since 2.4.47 (released April 22nd 2021), Apache httpd can ProxyPass to websockets on the same URL, without mod_rewrite (and, without mod_proxy_wstunnel).

## Checklist

### DIFF
--- a/docs/installation_guide/apache-httpd.md
+++ b/docs/installation_guide/apache-httpd.md
@@ -96,6 +96,29 @@ MDCertificateAgreement accepted
 </VirtualHost>
 ```
 
+or, if you have [Apache httpd 2.4.47+](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#protoupgrade), you can get rid of both `mod_rewrite` and `mod_proxy_wstunnel` and simplify the whole config to:
+
+```apache
+MDomain example.com auto
+MDCertificateAgreement accepted
+
+<VirtualHost *:80 >
+  ServerName example.com
+</VirtualHost>
+
+<VirtualHost *:443>
+  ServerName example.com
+
+  SSLEngine On
+  ProxyPreserveHost On
+  # set to 127.0.0.1 instead of localhost to work around https://stackoverflow.com/a/52550758
+  ProxyPass / http://127.0.0.1:8080/ upgrade=websocket
+  ProxyPassReverse / http://127.0.0.1:8080/
+
+  RequestHeader set "X-Forwarded-Proto" expr=https
+</VirtualHost>
+```
+
 Again, replace occurrences of `example.com` in the above config file with the hostname of your GtS server. If your domain name is `gotosocial.example.com`, then `gotosocial.example.com` would be the correct value.
 
 You should also change `http://127.0.0.1:8080` to the correct address and port (if it's not on `127.0.0.1:8080`) of your GtS server. For example, if you're running GoToSocial on another machine with the local ip of `192.168.178.69` and on port `8080` then `http://192.168.178.69:8080/` would be the correct value.


docs/installation_guide/apache-httpd.md
```diff
@@ -96,6 +96,29 @@ MDCertificateAgreement accepted
 </VirtualHost>
 ```
 
+or, if you have [Apache httpd 2.4.47+](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#protoupgrade), you can get rid of both `mod_rewrite` and `mod_proxy_wstunnel` and simplify the whole config to:
```
tsmethurst, line 99: Looks good to me. Are there any extra modules that need to be installed, or is that it?

igalic: those are the ones that *don't* need to be loaded

given that we've forgone mentioning all *other* modules that need loading for any of this config to work, I've continued that tradition.

This is the set of modules i load to have a functional file server and ssl terminating proxy: https://codeberg.org/meena-co/puppet-environment/src/branch/production/site/profile/manifests/apache.pp#L15-L35

igalic: and application firewall, apparently…

EchedeyLR: I dont think extra modules need to be instailed.

When I performed the first simplification of the documentation, it was mentioned that everything was to be handled by `proxy_http` itself so, I don't think that something more is needed.

I am going to test this in some hours anyways, given that this was not specified in the official documentation for the 2.4 version.

igalic: https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#protoupgrade (this is one or two days old, see https://bz.apache.org/bugzilla/show_bug.cgi?id=58593#c13  )

vs

https://httpd.apache.org/docs/2.4/mod/mod_proxy_wstunnel.html (the Rewrite doc here)

EchedeyLR: Thanks, so is newly documented.

My first simplification involved using rewrite only and not wstunnel mod.

Thanks for this.

igalic: the mod_rewrite code relied on mod_proxy_wstunnel being loaded. maybe since 2.4.47 it didn't?
but now we can drop both modules

EchedeyLR: I am unsure, need to check official documentation again but I got the example there to use rewrite without wstunnel.
This is before the current change this year about avoiding it too.

EchedeyLR: I tested that works pretty well without extra modules.

Thank you again for this.

igalic: you can try by installing an 2.4.46 or older: if it doesn't work without mod_proxy_wstunnel, we know.

we could also just ask on the bugzilla: https://bz.apache.org/bugzilla/show_bug.cgi?id=58593#c21

my guess is: yupp
it'll just work

igalic: to quote https://bz.apache.org/bugzilla/show_bug.cgi?id=58593#c22

> I think it does work because the workaround sets the ws(s) scheme for mod_proxy, which mod_proxy_http 2.4.47+ treats as/like "upgrade=websocket" (for compatibility precisely).

so now we have our answer for why the mod_rewrite workaround works without mod_proxy_wstunnel from an official source!

EchedeyLR: Thanks.

tsmethurst:
Thanks for doing this! ^_^

PR #1406, opened 2023-02-01
Text formatting overhaul

### LINKED ISSUE #789: [bug] Detection of mentions and hashtags in markdown can be improved

When posting markdown:

```
_#hashtag_
```

Won't be converted to an italicized link (only italicized). However:

```
_#hashtag #hashtag #hashtag_
```

Will convert all three

v0.4.0

### GoToSocial Arch

amd64 binary

### What happened?

posting `_#hashtag_` doesn't convert the hashtag to a link

posting `_#hashtag_` converts the hashtag to a link

### How to reproduce it?

post `_#hashtag_` with markdown formatting turned on

_No response_

### LINKED ISSUE #1271: [bug] Mentions/hashtags in markdown show up in metadata even in code blocks

In markdown mode, mentions and hashtags that don't render as clickable tags (e.g. in code blocks), are included in the `tags` and `mentions` metadata of the post.

v0.6.0

### GoToSocial Arch

armv7 Binary

### Browser version

_No response_

### What happened?

Make a post like this:
~~~
How to mention a user
```
@user@floss.dude.io hey bud! nice #ObjectOrientedProgramming software you've been writing lately!
```
hope that helps
~~~
and the user tagged in the code block will be included in the `mentions` field of the status, as well as the hashtag being included in the `tags` field.

Neither the user nor the hashtag should be associated with the post metadata.

### How to reproduce it?

Make post as above.

This is because the hashtags and mentions are scanned from the post before it is rendered as markdown. They should probably be scanned afterward, or during the markdown parsing process somehow.

autumnull:
# Description

>

This pulled request implements all of the changes suggested [here](https://github.com/superseriousbusiness/gotosocial/issues/789#issuecomment-1355158802) as well as a few other so-called "improvements"

Summary of changes:
- Added minifying step to plaintext to HTML conversion
- Links in formatted text now don't have their scheme removed (this is kind of incidental, the Linkify extension in goldmark doesn't have an option to remove the scheme)
- Added `Formatter.FromEmojiOnly`, which makes use of the existing goldmark parser to find emojis in a text (e.g. a display name) without processing mentions or tags.
- Converted `Formatter.FromPlain` to use a modified goldmark parser.
- Moved all of the logic for "what mentions/emojis/tags are in this text" into the goldmark parser, specifically into the rendering step, so that the context of these is taken into account and only included if the thing is actually active in the text.
- ...as a result of this: moved a lot of tests around, and deleted some that weren't relevant anymore.
- Unified the set of characters which can delimit mentions and hashtags into one set (any whitespace or punctuation)
- [Normalize](https://www.unicode.org/reports/tr15/) unicode text in hashtags before validating, rendering or adding them to the database.

closes #1271
closes #789

## Checklist


internal/text/formatter.go
```diff
@@ -26,18 +26,19 @@ import (
...
 	// FromMarkdown parses an HTML text from a markdown-formatted text.
-	FromMarkdown(ctx context.Context, md string, mentions []*gtsmodel.Mention, tags []*gtsmodel.Tag, emojis []*gtsmodel.Emoji) string
-
-	// ReplaceTags takes a piece of text and a slice of tags, and returns the same text with the tags nicely formatted as hrefs.
-	ReplaceTags(ctx context.Context, in string, tags []*gtsmodel.Tag) string
-	// ReplaceMentions takes a piece of text and a slice of mentions, and returns the same text with the mentions nicely formatted as hrefs.
-	ReplaceMentions(ctx context.Context, in string, mentions []*gtsmodel.Mention) string
+	FromMarkdown(ctx context.Context, pmf gtsmodel.ParseMentionFunc, authorID string, statusID string, md string) *FormatResult
+	// FromEmojiOnly parses an HTML text from a plaintext, only parsing emojis and not mentions etc.
+	FromEmojiOnly(ctx context.Context, pmf gtsmodel.ParseMentionFunc, authorID string, statusID string, plain string) *FormatResult
```
tsmethurst, line 37: The naming of this function is pretty confusing. I'd consider changing it from `FromEmojiOnly` to `FromPlainEmojiOnly` to make clear that it's like FromPlain, but only parses emoji.

internal/text/markdown.go
```diff
@@ -60,19 +53,15 @@ func (f *formatter) FromMarkdown(ctx context.Context, markdownText string, menti
...
 	}
+	result.HTML = htmlContentBytes.String()
 
-	// clean anything dangerous out of the html
-	htmlContent = SanitizeHTML(htmlContent)
+	// clean anything dangerous out of the HTML
+	result.HTML = SanitizeHTML(result.HTML)
 
-	minified := minifyHTML(htmlContent)
+	// shrink ray
```
tsmethurst, line 63: love this for us

internal/text/markdown_test.go
```diff
@@ -77,108 +75,105 @@ const (
 	mdWithStrikethroughExpected     = "<p>I have <del>mdae</del> made an error</p>"
 	mdWithLink                      = "Check out this code, i heard it was written by a sloth https://github.com/superseriousbusiness/gotosocial"
 	mdWithLinkExpected              = "<p>Check out this code, i heard it was written by a sloth <a href=\"https://github.com/superseriousbusiness/gotosocial\" rel=\"nofollow noreferrer noopener\" target=\"_blank\">https://github.com/superseriousbusiness/gotosocial</a></p>"
+	mdObjectInCodeBlock             = "How to mention a user\n```\n@user@floss.dude.io hey bud! nice #ObjectOrientedProgramming software you've been writing lately! :rainbow:\n```\nhope that helps"
```
tsmethurst, line 78: would this mention get picked up anyway if it's not a mention of an account that's dereferenceable?

internal/text/replace.go
```diff
@@ -0,0 +1,111 @@
+package text
```
tsmethurst, line 1: license text

internal/text/goldmark_extension.go
```diff
@@ -154,25 +153,30 @@ func (p *hashtagParser) Parse(parent ast.Node, block text.Reader, pc parser.Cont
...
+				return nil
+			}
 			// End of hashtag
 			block.Advance(i)
 			return newHashtag(segment.WithStop(segment.Start + i))
 		}
 	}
-	// If we don't find invalid characters before the end of the line then it's good
-	block.Advance(len(s))
+	// If we don't find invalid characters before the end of the line then it's all hashtag, babey
```
tsmethurst, line 178: v. important code change :+1:

internal/util/statustools.go
```diff
@@ -26,12 +26,7 @@ func IsPermittedInHashtag(r rune) bool {
...
-// Decides where to break before or after a hashtag.
-func IsHashtagBoundary(r rune) bool {
-	return r == '#' || // `###lol` should work
-		unicode.IsSpace(r) || // All kinds of Unicode whitespace.
-		unicode.IsControl(r) || // All kinds of control characters, like tab.
-		// Most kinds of punctuation except "Pc" ("Punctuation, connecting", like `_`).
-		// But `someurl/#fragment` should not match, neither should HTML entities like `&#35;`.
-		('/' != r && '&' != r && !unicode.Is(unicode.Categories["Pc"], r) && unicode.IsPunct(r))
+// Decides where to break before or after a #hashtag or @mention
+func IsMentionBoundary(r rune) bool {
```
tsmethurst, line 36: if this now works for both mentions and domains, then I think the function name needs to change

i know it's clunky but perhaps IsMentionOrHashtagBoundary ?

internal/util/statustools.go
```diff
@@ -26,12 +26,7 @@ func IsPermittedInHashtag(r rune) bool {
 	return unicode.IsLetter(r) || unicode.IsNumber(r)
 }
 
-// Decides where to break before or after a hashtag.
-func IsHashtagBoundary(r rune) bool {
-	return r == '#' || // `###lol` should work
```
tsmethurst, line 128: I think it'd be nice to keep the comments where possible since they're p. useful for understanding stuff.

internal/util/statustools.go
```diff
@@ -26,12 +26,7 @@ func IsPermittedInHashtag(r rune) bool {
...
-		unicode.IsSpace(r) || // All kinds of Unicode whitespace.
-		unicode.IsControl(r) || // All kinds of control characters, like tab.
-		// Most kinds of punctuation except "Pc" ("Punctuation, connecting", like `_`).
-		// But `someurl/#fragment` should not match, neither should HTML entities like `&#35;`.
-		('/' != r && '&' != r && !unicode.Is(unicode.Categories["Pc"], r) && unicode.IsPunct(r))
```
tsmethurst, line 133: Has this logic been moved elsewhere? Or did you remove it? If the latter, how come?

internal/text/markdownextension.go
```diff
@@ -57,14 +58,13 @@ func (n *hashtag) Kind() ast.NodeKind {
 	return kindHashtag
 }
 
-// Dump is used by goldmark for debugging. It is implemented only minimally because
-// it is not used in our code.
+// Dump can be used for debugging.
 func (n *mention) Dump(source []byte, level int) {
-	ast.DumpHelper(n, source, level, nil, nil)
+	fmt.Printf("%sMention: %s\n", strings.Repeat("    ", level), string(n.Segment.Value(source)))
```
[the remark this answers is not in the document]
autumnull: the dump functions don't show up anywhere in our code at all -- they're not really useful except for if you want to debug this part of the code in which case you uncomment [this line](https://github.com/superseriousbusiness/gotosocial/blob/4ee4cd2da11bc0199ee99a40527e758d394dbc9c/vendor/github.com/yuin/goldmark/parser/parser.go#L884) in `vendor/` and it prints some useful stuff during parsing. it uses `fmt.Printf` because it ends up in a (relatively inelegant) recursive dump straight to stdout that looks something like this:
```
Document {
    Paragraph {
        RawText: "@the_mighty_zork"
        HasBlankPreviousLines: true
        Mention: @the_mighty_zork
    }
    FencedCodeBlock {
        RawText: "@the_mighty_zork
"
        HasBlankPreviousLines: true
    }
}
```
But they're compulsory in order to implement the InlineParser interface so i figured it would be best to just implement them as something useful, which is why this commit even exists.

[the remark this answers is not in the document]
tsmethurst: Ahh got it, ok :)

internal/util/statustools.go
autumnull: So, the basic answer is that this logic is not necessary any more. The specific answers are kind of subtle, i talked about them a bit in the original comment I made about all this stuff but here's a line-by-line:

- '#' no longer needs a special case because it's included in [`IsPunct`](https://www.compart.com/en/unicode/category/Po).
- `IsControl` is not needed because tab and the other normal whitespace characters are all included in [`IsSpace`](https://pkg.go.dev/unicode#IsSpace)
- The special casing of `/` and `&` for HTML entities and URLs is made obsolete by the context-aware parsing in goldmark. Goldmark's Linkify extension adds a parser for links, so an entire link will be parsed and stored and the other parsers can't interfere with that process. There's a test in there already for the link case but i will add one for the entity case.
- The special case for `Is("Pc")` (i.e. for underscore) is removed on purpose, in order to fix the issue where hashtags weren't parsed in italic sections like `_#hashtag_`.

tsmethurst: Cool, thank you, that makes sense! :)

autumnull:
> Another question, and again perhaps it's already explained in the code and i just missed it because brain tired: does a remote lookup still happen when you mention an account that hasn't been dereferenced yet? Eg., if I mention `@someone@somewhere.com` will my instance try to webfinger + GET the AP representation of that account from `somewhere.com`? It did already, just wanna check that it still does.

I kept all of the logic for ingesting mentions and hashtags, it's just been relocated. Specifically I'm pretty sure that this happens in the `parseMentionFunc` that used to be called in 2 different places ([1](https://github.com/superseriousbusiness/gotosocial/blob/4ee4cd2da11bc0199ee99a40527e758d394dbc9c/internal/processing/status/util.go#L221),[2](https://github.com/superseriousbusiness/gotosocial/blob/4ee4cd2da11bc0199ee99a40527e758d394dbc9c/internal/processing/account/update.go#L258)) but now is only called [here](https://github.com/superseriousbusiness/gotosocial/blob/082f7614df4fec51ddfad26751c2c0d5d0a70503/internal/text/replace.go#L22). Basically all of the logic you might worry about going missing has been moved into the 2 functions there in `internal/text/replace.go`.

internal/util/statustools.go
autumnull: Do u think it still needs comments given that most of the logic has been removed now?

internal/db/bundb/bundb.go
```diff
@@ -473,43 +473,39 @@ func sqlitePragmas(ctx context.Context, conn *DBConn) error {
...
-				tag.Listable = &listable
-			} else {
-				return nil, fmt.Errorf("error getting tag with name %s: %s", t, err)
+	tag := &gtsmodel.Tag{}
+	// we can use selectorinsert here to create the new tag if it doesn't exist already
+	// inserted will be true if this is a new tag we just created
+	if err := dbService.conn.NewSelect().Model(tag).Where("LOWER(?) = LOWER(?)", bun.Ident("name"), t).Scan(ctx); err != nil {
+		if err == sql.ErrNoRows {
+			// tag doesn't exist yet so populate it
+			newID, err := id.NewRandomULID()
```
NyaaaWhatsUpDoc, line 486: @tsmethurst do we want to be using `NewULID()` here or this `NewRandomULID()`? I know the latter uses a random timestamp.

internal/db/bundb/bundb.go
```diff
@@ -473,43 +473,39 @@ func sqlitePragmas(ctx context.Context, conn *DBConn) error {
...
-				tag.Useable = &useable
-				listable := true
-				tag.Listable = &listable
-			} else {
-				return nil, fmt.Errorf("error getting tag with name %s: %s", t, err)
+	tag := &gtsmodel.Tag{}
+	// we can use selectorinsert here to create the new tag if it doesn't exist already
+	// inserted will be true if this is a new tag we just created
+	if err := dbService.conn.NewSelect().Model(tag).Where("LOWER(?) = LOWER(?)", bun.Ident("name"), t).Scan(ctx); err != nil {
+		if err == sql.ErrNoRows {
```
NyaaaWhatsUpDoc, line 484: Would probably be a little neater here to instead check `if err != nil sql.ErrNoRows { return ... }` and return early. Go coding practices generally go for returning early on negative result before handling the other logic, helps in most cases with readability :)

NyaaaWhatsUpDoc: e.g.

```
tag := &gtsmodel.Tag{}

if err := dbService.conn.NewSelect().Model(
    ....
); err !=nil && err != sql.ErrNoRows {
    return nil, fmt.Errorf("error string ...")
}

if tag.ID == "" {
    // no tag was found by name, create new.
    ...
}
```

NyaaaWhatsUpDoc: It would also make sense to only call `time.Now()` the once. Each of these calls can rack-up over time.

internal/db/bundb/bundb.go
tsmethurst: I think `NewRandomULID` is OK here because we'll probably not be paging through hashtags chronologically (unless you can think of a case where we'll do that?).

internal/db/bundb/bundb.go
tsmethurst: mea culpa: these were my mistakes in the first place :P

internal/db/bundb/bundb.go
```diff
@@ -473,43 +473,39 @@ func sqlitePragmas(ctx context.Context, conn *DBConn) error {
...
+			// tag doesn't exist yet so populate it
+			newID, err := id.NewRandomULID()
+			if err != nil {
+				return nil, err
 			}
+			tag.ID = newID
+			tag.URL = fmt.Sprintf("%s://%s/tags/%s", protocol, host, t)
```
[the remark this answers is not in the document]
tsmethurst: another mea culpa, lord

internal/db/bundb/bundb.go
autumnull: yea i didn't change this part of the code very much on the grounds that the DB is not my specialty (i don't know how to talk to the rabbit who manages the databases). I would suggest someone else making a pull request to do this but I can do it in this one if you would prefer.

autumnull: Oh wait these are just really simple changes, i can do them in this one for sure. got confused.

internal/util/statustools.go
tsmethurst: Nah it's fine :)

PR #1411, opened 2023-02-02
[chore] reformat GetAccount() functionality, support updating accounts based on last_fetch

NyaaaWhatsUpDoc:


internal/id/ulid.go
```diff
@@ -35,13 +35,15 @@ const (
...
-func NewULID() (string, error) {
-	newUlid, err := ulid.New(ulid.Timestamp(time.Now()), rand.Reader)
+// NewULID returns a new ULID string using the current time.
+func NewULID() string {
+	ulid, err := ulid.New(
+		ulid.Timestamp(time.Now()), rand.Reader,
+	)
 	if err != nil {
-		return "", err
+		panic(err)
```
NyaaaWhatsUpDoc, line 44: If we're failing reads from `rand.Reader`, i.e. `/dev/urandom` (or `/random`) then there is something _very_ wrong. Most UUID libraries have a `.MustNew()` for this reason, and since we're making a wrapper library we don't need to bother with `Must___()` prefixed functions.

internal/federation/dereferencing/account.go
```diff
@@ -38,389 +36,201 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/id"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
 	"github.com/superseriousbusiness/gotosocial/internal/media"
-	"github.com/superseriousbusiness/gotosocial/internal/transport"
 )
 
-var webfingerInterval = -48 * time.Hour // 2 days in the past
```
NyaaaWhatsUpDoc, line 44: This change will be _much_ easier to read if you just select `...` and then `view file`. The changes like this make it so much harder to read, it's basically entirely new functions so probably easier to read it that way :p

internal/federation/dereferencing/account.go
```diff
@@ -38,389 +36,201 @@ import (
...
-				sharedInbox = sharedInboxURI.String()
-			}
-		}
+	// Ensure ID is set and update fetch time.
+	latestAcc.ID = account.ID
+	latestAcc.FetchedAt = time.Now()
 
-		sharedInboxChanged = true
-		foundAccount.SharedInboxURI = &sharedInbox
+	// Fetch latest account media (TODO: check for changed URI to previous).
```
NyaaaWhatsUpDoc, line 189: I plan on a later PR to correct this along with some other cleanup of the functions being called here. For now, it actually helps deal with any weird database model situations that people have (considering we _are_ in alpha).

testrig/config.go
```diff
@@ -34,7 +34,7 @@ func InitTestConfig() {
 }
 
 var testDefaults = config.Configuration{
-	LogLevel:        "trace",
+	LogLevel:        "info",
```
tsmethurst, line 37: RIP unreadable spam logs

internal/id/ulid.go
tsmethurst: Good change yeah

internal/federation/handshake.go
```diff
@@ -17,12 +17,3 @@
 */
 
 package federation
```
tsmethurst, line 19: might as well delete this file now since there's nothing in it

internal/federation/dereferencing/account.go
tsmethurst: Ack :D I often use the side-by-side diff thing rather than the one-column diff thing, which is also a bit easier in this case.

internal/federation/dereferencing/account.go
```diff
@@ -441,83 +251,23 @@ func (d *deref) dereferenceAccountable(ctx context.Context, username string, rem
 		return nil, fmt.Errorf("DereferenceAccountable: error resolving json into ap vocab type: %w", err)
 	}
 
+	//nolint shutup linter
```
tsmethurst, line 254: better to specify which linter you're shutting up with nolint comments imo, prevents other stuff from not being linted if this switch changes in future

internal/federation/dereferencing/account.go
```diff
@@ -530,39 +280,31 @@ func (d *deref) populateAccountFields(ctx context.Context, account *gtsmodel.Acc
...
-func (d *deref) fetchRemoteAccountMedia(ctx context.Context, targetAccount *gtsmodel.Account, requestingUsername string, blocking bool) (bool, error) {
-	var (
-		changed bool
-		t       transport.Transport
-	)
+func (d *deref) fetchRemoteAccountMedia(ctx context.Context, targetAccount *gtsmodel.Account, requestingUsername string, blocking bool) error {
+	// Fetch a transport beforehand for either(or both) avatar / header dereferencing.
+	tsport, err := d.transportController.NewTransportForUsername(ctx, requestingUsername)
+	if err != nil {
+		d.dereferencingAvatarsLock.Unlock()
```
tsmethurst, line 287: aren't you gonna be potentially unlocking an unlocked lock here, since you moved the transport creation up in the function?

tsmethurst (commented):
Looking good! Couple comments. Shall I tidy up the tests?

Issue #1230, opened 2022-12-08
[feature] Instrument GoToSocial with OpenTelemetry or similar tracing thingy


tsmethurst:
It would be handy if we could implement some kind of tracing thing so that incoming http requests can have an ID assigned to them, and we can keep track via the logs (or something) of what happens to requests. We can also enable bun's open telemetry doodad to get some more traceability on db requests.

Links of interest:

- https://opentelemetry.io/docs/instrumentation/go/
- https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gin-gonic/gin
-  https://github.com/uptrace/bun/tree/master/extra/bunotel

related to https://github.com/superseriousbusiness/gotosocial/pull/1145 and https://github.com/superseriousbusiness/gotosocial/issues/1218

NyaaaWhatsUpDoc:
is this worth adding to the metrics issue to better track it all together? i worry points raised in one may get missed when implementing the other

tsmethurst:
Mmm which issue was that? maybe a good idea

NyaaaWhatsUpDoc:
#1218

tsmethurst:
> #1218

Aha! I did update the checklist there. If it gets messy we can make a larger issue for both, a boss fight issue if you will

PR #1423, opened 2023-02-05
Plumb a request ID through HTTP and logger

daenney:
# Description

This is the revival of #1145. Irrespective of what else we may want to do as part of #1230, having a unique identifier per request will be necessary.

Getting this in place now would let us add Prometheus metrics on the HTTP side and tack on an Exemplar for the request ID. That gets us the first step between being able to correlate metrics + logs if you ingest them somewhere centrally without the cardinality explosion of the request ID being a label on the metric.

After that, we can take a look at OpenTelemetry tracing. That's a bit complicated and has its own TraceID (which is a `[16]byte`). I don't think there's any way for us to directly map the Request ID to the Trace ID and we can't use the Trace ID as the Request ID since tracing can be sampled (meaning not each request would have a Trace ID).

## Checklist

### DIFF
--- a/cmd/gotosocial/action/server/server.go
+++ b/cmd/gotosocial/action/server/server.go
@@ -154,6 +154,7 @@ var Start action.GTSAction = func(ctx context.Context) error {
 
 	// attach global middlewares which are used for every request
 	router.AttachGlobalMiddleware(
+		middleware.RequestID(),
 		middleware.Logger(),
 		middleware.UserAgent(),
 		middleware.CORS(),
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	github.com/disintegration/imaging v1.6.2
 	github.com/gin-contrib/cors v1.4.0
 	github.com/gin-contrib/gzip v0.0.6
+	github.com/gin-contrib/requestid v0.0.6
 	github.com/gin-contrib/sessions v0.0.5
 	github.com/gin-gonic/gin v1.8.2
 	github.com/go-fed/httpsig v1.1.0
--- a/go.sum
+++ b/go.sum
@@ -180,6 +180,8 @@ github.com/gin-contrib/cors v1.4.0 h1:oJ6gwtUl3lqV0WEIwM/LxPF1QZ5qe2lGWdY2+bz7y0
 github.com/gin-contrib/cors v1.4.0/go.mod h1:bs9pNM0x/UsmHPBWT2xZz9ROh8xYjYkiURUfmBoMlcs=
 github.com/gin-contrib/gzip v0.0.6 h1:NjcunTcGAj5CO1gn4N8jHOSIeRFHIbn51z6K+xaN4d4=
 github.com/gin-contrib/gzip v0.0.6/go.mod h1:QOJlmV2xmayAjkNS2Y8NQsMneuRShOU/kjovCXNuzzk=
+github.com/gin-contrib/requestid v0.0.6 h1:mGcxTnHQ45F6QU5HQRgQUDsAfHprD3P7g2uZ4cSZo9o=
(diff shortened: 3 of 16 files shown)


daenney:
The same issue as I originally had with #1145 remains, I need to figure out how to add this to the DB log lines

daenney:
Mmm, so the tests initially failed because I ran them against the wrong branch myself. The downside of plumbing the requestID in the JSON is that I have to update all the tests. However, since the X-Request-ID header would be returned in the HTTP response I'm inclined to say that's enough when debugging (you're already looking at JSON) so it doesn't have to be included in the body. I've kept it in the HTML templates though since that's user facing.

daenney:
So peeking around a bit, it seems we pass a `context.Context` to most DB functions. That seems to be the `c.Request().Context()`, which returns a `context.Background` if it's not set. Unfortunately things done to the `gin.Context` don't translate to a key set on the `context.Context` so that doesn't seem to directly help us.

The requestid middleware doesn't set that key if a context.Context exists, it just sets headers:
```go
c.Request.Header.Add(headerXRequestID, rid)
c.Header(headerXRequestID, rid)
```

If we can control the context somehow and set the request ID as a key on it too, then we'd have it in any of the DB functions and we could get the logger to conceivable include it?

I'm not familiar enough with Gin to know what's sensible to do here.

NyaaaWhatsUpDoc:
I have had quite a bit of experience with crafting request ID libraries . I won't go into it now but I'll help with this at some point in the week if that's good. I also want to take a look under the hood at how the gin request ID library is doing it just to make sure I think they're doing it the right way since I have opinions on it :p

daenney:
Oh yeah, take your time! I just raised it again now that 0.6.x is out so we could get started.

From my point of view, the library works in a sensible way:
* If your choice if request ID header is already set, use that. This is good, because it means an LB can set this so you can correlate those logs too. Especially useful in cloud-y environments
* If not, it calls `uuid.New()` resulting in a UUIDv4. I'm not really convinced a UUIDv4 is necessary

Both the used header and the ID generator are configurable and is why I currently have the thin wrapper. At the very least I think the header should be configurable through the config file so people can tweak that to their needs. The generator is probably not something that we should expose, but we can chose to plug in a different one.

daenney:
It feels like the middleware almost does what we need, it's just missing:

```go
ctx := context.WithValue(c.Request.Context(), "requestid", rid)
c.Request = c.Request.WithContext(ctx)
```

If it did that (too) then we'd be able to get at the requestid from any function that takes a context that got triggered from an HTTP endpoint?

daenney:
I'm going to keep dumping thoughts here, sorry @NyaaaWhatsUpDoc :sweat_smile:.

I took a look at how OpenTelemetry generates their Trace ID and it's basically a `[16]byte`. What they seem to do is use `math.Rand()` and read from that directly, but seed it with a value from `crypto/rand` first: https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/id_generator.go.

This is probably faster than the default generator in gin-contrib/requestid which uses a UUID instead and is probably something we could use. A somewhat interesting possibility would be that if we use the same technique then at least in theory there might be a way for us to provide the Request ID as the Trace ID later on.

However, requestid allows for using an incoming header to set the request ID and this is a very useful property when you have an HTTP proxy or load balancer in front of your system as it makes log correlation easy. In AWS for example you'll get `X-Amzn-Trace-Id` if you enable access logs for an ALB, which you'll want to have somewhere and put in your logs. But it has a completely different format that we probably can't reliable coerce into a TraceID. So even though we could use the same approach to generating a Request ID internally as what OpenTelemetry does for the TraceID, we should never use one as the other so long term traced requests would have both a RequestID and a TraceID.

daenney:
I created #1476 showcasing an OTEL-like implementation for the request ID so there's something more concrete to look at than my musings in prose.

PR #1437, opened 2023-02-06
[chore] Update bug report template

daenney:
# Description

> Beware. Controversy

There's been a recent increase in new ActivityPub clients. Unfortunately most clients are coded directly against the Mastodon API and make deep rooted assumptions about it in their implementations. This leads to those clients not working with GTS. Though GTS implements a good chunk of the Mastodon API, it's always in catch-up mode. The Mastodon Client API is not part of the ActivityPub spec.

This change aims to add some gentle language to the bug report template to direct folks that are reporting compatibility issues with closed source clients to go talk to those developers instead. GTS developers can reasonably help debug an issue if they have access to the source code on both ends, but when the client is closed source the burden of ensuring compatibility can't be on GTS developers.

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -9,6 +9,16 @@ body:
     attributes:
       value: |
         Thanks for taking the time to fill out this bug report! Please be cautious with the sensitive information/logs while filing the issue.
+  - type: markdown
+    attributes:
+      value: |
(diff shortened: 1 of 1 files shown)


daenney:
Just for context; I realise this PR could seem a bit weird. My only goal here is to capture what seemed to be the rough consensus around this when the discussion happened in the Matrix room. 

Whether this needs to be written down at all or if it should be surfaced in the bug report template I don't know.

tsmethurst:
That looks good, and makes sense, thanks :)

PR #1440, opened 2023-02-06
[bugfix] don't trash emoji in profile fields on edit

### LINKED ISSUE #1439: [bug] 0.7.0-rc1 Setting profile fields removes custom emoji from other profile fields

If you have custom emoji in either your display name or your profile note, and you use the GTS user settings interface to set a profile field _other_ than the one(s) containing custom emoji, custom emoji will vanish from those fields and be replaced by their shortcodes. This is a backend bug - the unchanged fields aren't transmitted by the frontend.

0.7.0-rc1

### GoToSocial Arch

amd64 Binary

### Browser version

_No response_

### What happened?

_No response_

_No response_

### How to reproduce it?

Set a custom emoji in your display name.
Verify this appears correctly.
Change your profile note (without including the same emoji as the display name), avatar, or something else.
Check that the display name emoji is now missing.

_No response_

Sentynel:
# Description

Fix emoji being removed from profile fields when other fields are edited.

There's no trivial way of identifying from the list of emoji in the Account object which ones are used in the DisplayName and which in the Note, as both support emoji. Therefore, this change reparses both of these fields for emoji on any change to the profile. If a new version has been submitted, this is parsed and applied to the account; otherwise we use the current version from the account.

closes #1439
closes #(another issue)

## Checklist

### DIFF
--- a/internal/processing/account/update.go
+++ b/internal/processing/account/update.go
@@ -53,12 +53,15 @@ func (p *processor) Update(ctx context.Context, account *gtsmodel.Account, form 
 			return nil, gtserror.NewErrorBadRequest(err)
 		}
 		account.DisplayName = text.SanitizePlaintext(*form.DisplayName)
+	}
 
-		formatResult := p.formatter.FromPlainEmojiOnly(ctx, p.parseMention, account.ID, "", account.DisplayName)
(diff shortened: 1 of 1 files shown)


tsmethurst:
Thanks!

PR #1461, opened 2023-02-08
[chore/bugfix] Serve + throttle publickey separately from rest of ActivityPub API

tsmethurst:
# Description

>

This pull request updates some of our routing code to serve the `/users/:username/main-key` ActivityPub endpoint separately from the other AP endpoints, in order to apply separate throttling + cache policies on it (public cache with lengthy max-age).

The reason for this change is that when a status is federated from GtS to a Mastodon server, Mastodon always makes a request to get the main-key again (https://github.com/mastodon/mastodon/issues/19217). If it can't get the main-key because of throttling or some other issue, the status won't be accepted by that Mastodon server, leading to spotty federation.

By throttling the main key separately from other AP endpoints, there should be less chance of a post not federating to mastodon due to server unavailability.

Finally, the PR also includes changes to the nginx advanced documentation, to provide the option of allowing nginx to cache main-key responses in addition to webfinger responses.

## Checklist

### DIFF
--- a/docs/installation_guide/advanced.md
+++ b/docs/installation_guide/advanced.md
@@ -279,36 +279,44 @@ This section contains a number of additional things for configuring nginx.
 
 If you want to harden up your NGINX deployment with advanced configuration options, there are many guides online for doing so ([for example](https://beaglesecurity.com/blog/article/nginx-server-security.html)). Try to find one that's up to date. Mozilla also publishes best-practice ssl configuration [here](https://ssl-config.mozilla.org/).
 
-### Caching Webfinger
+### Caching Webfinger and Public Key responses
 
-It's possible to use nginx to cache the webfinger responses. This may be useful in order to ensure clients still get a response on the webfinger endpoint even if GTS is (temporarily) down.
+It's possible to use nginx to cache webfinger and public key responses. This may be useful in order to ensure clients still get a response on these endpoints even if your GoToSocial instance is (temporarily) down, or requests are being throttled.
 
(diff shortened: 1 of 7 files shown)


docs/installation_guide/advanced.md
```diff
@@ -319,6 +327,25 @@ server {
...
 
+  location ~ ^\/users\/(?:[a-z0-9_\.]+)\/main-key$ {
+    proxy_set_header Host $host;
+    proxy_set_header X-Forwarded-For $remote_addr;
+    proxy_set_header X-Forwarded-Proto $scheme;
+
+    proxy_cache gotosocial_ap_public_responses;
+    proxy_cache_background_update on;
+    proxy_cache_key $scheme://$host$uri;
+    proxy_cache_valid 200 604800s;
```
daenney, line 338: Is it safe to cache this for a week? I'm unfamiliar with how, or if, key rotation works in the context of activitypub.

tsmethurst: Yes, it's safe :) Under normal circumstances those keys will never change.

NyaaaWhatsUpDoc:
We could still provide throttling for it, but give it a separate throttling middleware instance similar to how we handle federated + client APIs. What do you think?

tsmethurst:
> We could still provide throttling for it, but give it a separate throttling middleware instance similar to how we handle federated + client APIs. What do you think?

Yeah I'm up for that :) I'll change it.

PR #1466, opened 2023-02-09
[performance] remove throttling timers

NyaaaWhatsUpDoc:
We already set request read/write timeouts (which will in turn cancel the request contexts). Timers in the throttling middleware are un-needed, and under load will add *a lot* of running timers. This also adds:
- support custom setting for throttling retry-after
- use retry-after header in transport controller

Checklist:


NyaaaWhatsUpDoc:
Just need to update example configuration and the CLI/env parsing tests and will be ready.

internal/transport/transport.go
```diff
@@ -152,18 +156,26 @@ func (t *transport) do(r *http.Request, signer func(*http.Request) error, retryO
...
 				return rsp, nil
 			}
 
 			// Generate error from status code for logging
 			err = errors.New(`http response "` + rsp.Status + `"`)
+
+			// Search for a provided "Retry-After" header value.
+			if after := rsp.Header.Get("Retry-After"); after != "" {
+				// Attempt to parse this value as integer no. of seconds.
+				if u, _ := strconv.ParseUint(after, 10, 32); u != 0 {
```
tsmethurst, line 174: Retry-After can also be an HTTP Date, so it might also be worth using http.ParseTime() to try parsing this if the ParseUint fails. What do you think?

internal/transport/transport.go
```diff
@@ -152,18 +156,26 @@ func (t *transport) do(r *http.Request, signer func(*http.Request) error, retryO
...
+					backoff = time.Duration(u) * time.Second
```
tsmethurst, line 175: Probably worth comparing this to some maximum (5 minutes?) and just dropping the request if it's over that maximum, to prevent clients being held open by malicious Retry-After responses.

PR #1467, opened 2023-02-09
[chore] small changes missed in previous dereferencer.GetAccount() PRs

NyaaaWhatsUpDoc:


internal/federation/dereferencing/account.go
```diff
@@ -338,47 +334,108 @@ func (d *deref) dereferenceAccountable(ctx context.Context, transport transport.
...
+	processing, err := d.mediaManager.ProcessMedia(ctx, data, nil, accountID, &media.AdditionalMediaInfo{
+		Avatar:    func() *bool { v := false; return &v }(),
+		RemoteURL: &avatarURL,
+	})
+	if err != nil {
+		return "", err
+	}
+
+	// Store media in map to mark as processing.
+	id := processing.AttachmentID()
```
tsmethurst, line 368: You're checking further up the function whether this is already stored with the accountID, but then storing it with the ID of the attachment, so I think the first check will never be a hit (unless i'm missing something)

internal/federation/dereferencing/account.go
```diff
@@ -338,47 +334,108 @@ func (d *deref) dereferenceAccountable(ctx context.Context, transport transport.
...
+		return "", err
+	}
+
+	// Acquire lock for derefs map.
+	unlock := d.derefAvatarsMu.Lock()
+	defer unlock()
+
+	if processing, ok := d.derefAvatars[accountID]; ok {
+		// we're already dereferencing it, nothing to do.
+		return processing.AttachmentID(), nil
```
tsmethurst, line 350: Perhaps before returning, it's worth calling `processing.LoadAttachment` to make sure that the attachment will really be in the database for the caller to select with the ID; it's idempotent anyway I think so it should be fine.

internal/federation/dereferencing/account.go
```diff
@@ -338,47 +334,108 @@ func (d *deref) dereferenceAccountable(ctx context.Context, transport transport.
...
 		return "", err
 	}
 
-	data := func(innerCtx context.Context) (io.ReadCloser, int64, error) {
-		return transport.DereferenceMedia(innerCtx, avatarIRI)
+	// Acquire lock for derefs map.
+	unlock := d.derefHeadersMu.Lock()
+	defer unlock()
+
+	if processing, ok := d.derefHeaders[accountID]; ok {
```
tsmethurst, line 400: same comments as above re: map key and calling Load

PR #1476, opened 2023-02-11
[feat] Add a request ID and include it in log lines

daenney:
# Description

This adds the necessary code to:
* Extend the global logger to allow for hooks that can be used to modify what's getting logged
* Add a new middleware that:
  * Fetches or generates a Request ID as the first middleware in the chain 
  * Plonk the Request ID in both the Gin and stdlib context's
  * Add a hook to the logger to extract the RequestID and log it along with everything else
* Pull the context with our request ID through just about every `log` call so we can log it along with anything else

## Checklist


cmd/gotosocial/action/server/server.go
```diff
@@ -154,6 +154,7 @@ var Start action.GTSAction = func(ctx context.Context) error {
 
 	// attach global middlewares which are used for every request
 	router.AttachGlobalMiddleware(
+		middleware.RequestID("X-Request-Id"),
```
daenney, line 157: This should come from config instead, it's hardcoded for now.

daenney:
Despite the fact that the request ID is now set on the `context.Context` and that there's a way to retrieve it, a lot of functions that may get called along the way from a request handler simply do `log.Info()` etc. calling into the global log package. This results in a log line without fields being emitted, so those lines don't emit an entry with a request ID. 

If the `log.XXX` functions would take a `context.Context` as their first argument it becomes possible to extract a requestID in those functions and call `logf` with `fields` not set to `nil`.

daenney:
So golang-ci lint is a little unhappy with us due to the context changes I guess. I'm not super convinced it's rightfully complaining about all cases though:

```go
internal/media/cron.go:49:13: Non-inherited new context, use function like `context.WithXXX` instead (contextcheck)
			log.Error(nil, err)
			         ^
internal/media/processingmedia.go:251:33: Function `decodeVideoFrame->decodeVideoFrame$1` should pass the context parameter (contextcheck)
		video, err := decodeVideoFrame(rc)
		                              ^
internal/media/prune.go:89:14: Non-inherited new context, use function like `context.WithXXX` instead (contextcheck)
		if err := f(context.Background()); err != nil {
		           ^
internal/typeutils/astointernal.go:85:12: Non-inherited new context, use function like `context.WithXXX` instead (contextcheck)
		log.Infof(nil, "error extracting account emojis: %s", err)
		         ^
internal/typeutils/astointernal.go:260:43: Function `extractAttachments` should pass the context parameter (contextcheck)
	status.Attachments = c.extractAttachments(statusable)
	                                         ^
internal/config/validate.go:58:13: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
		log.Warnf(nil, "%s was set to 'http'; this should *only* be used for debugging and tests!", ProtocolFlag())
		          ^
internal/cache/util.go:38:13: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
	log.Panicf(nil, "failed %s after %d tries", msg, count)
	           ^
internal/concurrency/workers.go:69:12: SA1012: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (staticcheck)
	log.Infof(nil, "%s created with workers=%d queue=%d",
```

daenney:
Thanks to the awesome help of @NyaaaWhatsUpDoc this is now ready to go! Ish. Once we figure out what to do about the lint thing.

NyaaaWhatsUpDoc:
Definitely best to disable that linter imo. It's been annoying me for a while :p

daenney:
Weird extra thing, the following lint warnings are bogus:

```
internal/media/processingmedia.go:251:33: Function `decodeVideoFrame->decodeVideoFrame$1` should pass the context parameter (contextcheck)
		video, err := decodeVideoFrame(rc)
```

```
internal/typeutils/astointernal.go:260:43: Function `extractAttachments` should pass the context parameter (contextcheck)
	status.Attachments = c.extractAttachments(statusable)
```
Neither of those two functions take a context in their signature.

tsmethurst:
I'm OK with dropping the contextcheck linter altogether in a separate PR, afaik it hasn't found any actual hits yet, only false positives.

Anyway, reviewing this now :)

example/config.yaml
```diff
@@ -690,6 +690,15 @@ syslog-protocol: "udp"
...
 syslog-address: "localhost:514"
 
+##################################
+##### OBSERVABILITY SETTINGS #####
+##################################
+
+# String. Header name to use to extract a request or trace ID from. Typically set by a
+# loadbalancer or proxy.
+# Default: "X-Request-Id"
+request-id-header: "X-TraceID"
```
tsmethurst, line 700: This should be set to the default value

.golangci.yml
```diff
@@ -27,3 +26,5 @@ linters-settings:
   govet:
     disable:
       - composites
+  staticcheck:
+    checks: ["all", "-SA1012"]
```
tsmethurst, line 30: What's this do? A comment could be handy :)

.golangci.yml
```diff
@@ -14,7 +14,6 @@ run:
 linters:
   # enable some extra linters, see here for the list: https://golangci-lint.run/usage/linters/
   enable:
-    - contextcheck
```
tsmethurst, line 17: Ah OK you included it here, also fine!

docs/configuration/observability.md
```diff
@@ -0,0 +1,16 @@
+# Observability
```
tsmethurst, line 1: don't forget to add this to the `mkdocs.yml` file so it's included properly in the documentation :)

docs/configuration/observability.md
```diff
@@ -0,0 +1,16 @@
...
+
+```yaml
+##################################
+##### OBSERVABILITY SETTINGS #####
+##################################
+
+# String. Header name to use to extract a request or trace ID from. Typically set by a
+# loadbalancer or proxy.
+# Default: "X-Request-Id"
+request-id-header: "X-TraceID"
```
tsmethurst, line 15: same comment as above, make sure this default is consistent

internal/api/fileserver/servefile.go
```diff
@@ -125,15 +128,21 @@ func (m *Module) ServeFile(c *gin.Context) {
...
+		rng,
+		content.ContentLength,
+	)
 }
 
 // serveFileRange serves the range of a file from a given source reader, without the
 // need for implementation of io.Seeker. Instead we read the first 'start' many bytes
 // into a discard reader. Code is adapted from https://codeberg.org/gruf/simplehttp.
-func serveFileRange(rw http.ResponseWriter, src io.Reader, rng string, size int64) {
+func serveFileRange(rw http.ResponseWriter, r *http.Request, src io.Reader, rng string, size int64) {
```
tsmethurst, line 145: would it be better to just have this accept ctx as the first parameter rather than passing the request into it? afaict the request is only being used for the context anyways

internal/federation/dereferencing/account.go
```diff
@@ -114,7 +114,7 @@ func (d *deref) GetAccountByUsernameDomain(ctx context.Context, requestUser stri
 	// Try to update existing account model
 	enriched, err := d.enrichAccount(ctx, requestUser, nil, account, false, block)
 	if err != nil {
-		log.Errorf("GetAccountByUsernameDomain: error enriching account from remote: %v", err)
+		log.Errorf(ctx, "GetAccountByUsernameDomain: error enriching account from remote: %v", err)
```
tsmethurst, line 117: can remove the function name from this now i think

internal/federation/dereferencing/emoji.go
```diff
@@ -111,7 +111,7 @@ func (d *deref) populateEmojis(ctx context.Context, rawEmojis []*gtsmodel.Emoji,
 			// have to get it from the database again
 			gotEmoji = e
 		} else if gotEmoji, err = d.db.GetEmojiByShortcodeDomain(ctx, e.Shortcode, e.Domain); err != nil && err != db.ErrNoEntries {
-			log.Errorf("populateEmojis: error checking database for emoji %s: %s", shortcodeDomain, err)
+			log.Errorf(ctx, "populateEmojis: error checking database for emoji %s: %s", shortcodeDomain, err)
```
tsmethurst, line 114: function name in this file can be removed too i believe

internal/federation/federatingactor.go
```diff
@@ -56,7 +56,7 @@ func newFederatingActor(c pub.CommonBehavior, s2s pub.FederatingProtocol, db pub
 // method will guaranteed work for non-custom Actors. For custom actors,
 // care should be used to not call this method if only C2S is supported.
 func (f *federatingActor) Send(c context.Context, outbox *url.URL, t vocab.Type) (pub.Activity, error) {
-	log.Infof("federating actor: send activity %s via outbox %s", t.GetTypeName(), outbox)
+	log.Infof(c, "federating actor: send activity %s via outbox %s", t.GetTypeName(), outbox)
```
tsmethurst, line 59: we can probs drop this federating actor hint as well since the location will be logged anyways

internal/federation/federatingdb/util.go
```diff
@@ -304,7 +304,7 @@ func extractFromCtx(ctx context.Context) (receivingAccount, requestingAccount *g
 		var ok bool
 		receivingAccount, ok = receivingAccountI.(*gtsmodel.Account)
 		if !ok {
-			log.Panicf("extractFromCtx: context entry with key %s could not be asserted to *gtsmodel.Account", ap.ContextReceivingAccount)
+			log.Panicf(ctx, "extractFromCtx: context entry with key %s could not be asserted to *gtsmodel.Account", ap.ContextReceivingAccount)
```
tsmethurst, line 307: function name

internal/federation/federatingprotocol.go
```diff
@@ -232,7 +232,7 @@ func (f *federator) AuthenticatePostInbox(ctx context.Context, w http.ResponseWr
 // blocked must be false and error nil. The request will continue
 // to be processed.
 func (f *federator) Blocked(ctx context.Context, actorIRIs []*url.URL) (bool, error) {
-	log.Debugf("entering BLOCKED function with IRI list: %+v", actorIRIs)
+	log.Debugf(ctx, "entering BLOCKED function with IRI list: %+v", actorIRIs)
```
tsmethurst, line 235: think it's worth making this trace logging now rather than debug, while you're here :)

internal/middleware/requestid.go
```diff
@@ -0,0 +1,111 @@
...
+var (
+	// ridCtxKey is the key underwhich we store request IDs in a context.
+	ridCtxKey ctxType = "id"
+
+	// crand provides buffered reads of random input.
+	crand = bufio.NewReader(rand.Reader)
+	mrand sync.Mutex
+
+	// base32enc is a base 32 encoding based on a human-readable character set (no padding).
+	base32enc = base32.NewEncoding("0123456789abcdefghjkmnpqrstvwxyz").WithPadding(-1)
```
tsmethurst, line 47: nice, i like this a lot, thank you for excluding l and i :pray:

internal/middleware/requestid.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// RequestID fetches the stored request ID from context.
+func RequestID(ctx context.Context) string {
+	id, _ := ctx.Value(ridCtxKey).(string)
+	return id
+}
+
+// AddRequestID returns a gin middleware which adds a unique ID to each request (both response header and context).
+func AddRequestID(header string) gin.HandlerFunc {
```
tsmethurst, line 80: very cool

internal/middleware/requestid.go
```diff
@@ -0,0 +1,111 @@
...
+
+		// Store request ID in new request ctx and set new gin request obj.
+		ctx := context.WithValue(c.Request.Context(), ridCtxKey, id)
+		c.Request = c.Request.WithContext(ctx)
+
+		// Set the request ID in the rsp header.
+		c.Writer.Header().Set(header, id)
+
+		// Next handler.
+		c.Next()
```
tsmethurst, line 109: the call to `c.Next()` doesn't do anything here. If you want to write the header last (to avoid it being set in other handlers) you should call next first, and *then* call c.Writer.Header().Set

tsmethurst: ^^ at least, I think so

web/template/404.tmpl
```diff
@@ -29,7 +29,8 @@
 		</p>
 		<p>
 			If you believe this 404 was an error, you can contact
-			the instance admin.
+			the instance admin. Provide them with the following request
+			Request ID: <code>{{.requestID}}</code>.
```
tsmethurst, line 33: do you pass `requestID` in somewhere when rendering the template? I might have missed it

web/template/error.tmpl
```diff
@@ -20,6 +20,7 @@
     <main>
         <section class="error">
           <span>❌</span> <pre>{{.error}}</pre>
+		  <span>Request ID</span> <code>{{.requestID}}</code>
```
tsmethurst, line 23: same question re: passing `requestID` into the template

tsmethurst (commented):
looking good so far :)

internal/api/fileserver/servefile.go
NyaaaWhatsUpDoc: i thought about that, but ultimately i thought it a bit more idiomatic to fit closer to the http.HandlerFunc signature as that is what it is :p

internal/federation/dereferencing/account.go
NyaaaWhatsUpDoc: dammit i thought i'd caught all of these :laughing:

internal/federation/federatingprotocol.go
NyaaaWhatsUpDoc: good idea! i assume same for the other debug entries within this function?

internal/middleware/requestid.go
NyaaaWhatsUpDoc: :grin:

internal/federation/dereferencing/account.go
tsmethurst: :smiling_imp:

web/template/error.tmpl
NyaaaWhatsUpDoc: Ah that might have been my bad removing them. Though I'm not 100% sure they were in right place to go into the template in the places I removed (I may be wrong though). Either way, simple enough fix :p

web/template/404.tmpl
NyaaaWhatsUpDoc: See below

internal/middleware/requestid.go
NyaaaWhatsUpDoc: I was just following what some of the other middleware imports we have did tbh. Can remove that later after work, and I don't think we need to worry about the request ID header being over-written.

.golangci.yml
daenney: Right, I'll add a comment. Right now it enables all checks (the default for staticcheck in golangci-lint) but disables SA1012, the other context lint check.

internal/middleware/requestid.go
daenney: Took a look at some other middlewares. Turns out it's indeed only necessary if you want to do something in the return chain, like measure the time since the start of the request.

web/template/error.tmpl
daenney: I'm fairly certain we were passing them in at some point :smile:. I'll fix the indentation here while we're at it too.

internal/middleware/requestid.go
tsmethurst: Yep indeed! Next just makes all of the wrapped handlers process inside the handler you're calling c.Next() in, so it's useful for doing before+after stuff :)

docs/configuration/observability.md
daenney: Done!

web/template/404.tmpl
daenney: Fixed it in `internal/api/util/errorhandling.go`.

daenney:
The history is a bit messy now and you can't really roll back any individual commit so probably best to squash-merge this

tsmethurst:
> The history is a bit messy now and you can't really roll back any individual commit so probably best to squash-merge this

we always do, don't worry :)

NyaaaWhatsUpDoc:
we love squerges

daenney:
I smashed the history to make this easier to rebase. The only way I could find to retain @NyaaaWhatsUpDoc's authorship was the `Co-authored-by` tag in the commit. I hope that's OK.

daenney:
I rebased this on main, should be good to go with 0.7.0 being out.

tsmethurst:
Merged, thank you! :)

PR #1481, opened 2023-02-12
[chore] Do cache-control in a less silly way to avoid writing header twice

tsmethurst:
# Description

>

I was way too hasty merging https://github.com/superseriousbusiness/gotosocial/pull/1480.

This PR achieves the same effect, but without having to write the `Cache-Control` header twice for every request when using non-proxying s3; instead the storage method is just checked once at startup and the appropriate cache-control timeout is selected then.

## Checklist


internal/api/fileserver/servefile.go
```diff
@@ -88,12 +88,6 @@ func (m *Module) ServeFile(c *gin.Context) {
 	}
 
 	if content.URL != nil {
-		// This is a non-local S3 file we're redirecting to.
```
NyaaaWhatsUpDoc, line 91: I think it's at least worth keeping the first line of this comment here :)

tsmethurst: Ah thanks, that was an accident!

PR #1482, opened 2023-02-12
[performance] processing media and scheduled jobs improvements

NyaaaWhatsUpDoc:
Changes:
- use `runners.Processor{}` to ensure singular running instance of processing media/emoji
- retry processing media / emoji if error is due to context cancel during request callstacks
- use simply `runners.WorkerPool{}` instead of wrapping concurrency package for simplicity
- start moving some workers to state object
- emoji are only ever processed synchronously, remove separate emoji worker pool (also scales up media worker pool to accomodate)
- use `runners.Scheduler{}` instead of cron, and add to state workers for management
- testrig is now only included in binary if `DEBUG=1` is set during build

Checklist:


NyaaaWhatsUpDoc:
Okay I'll try get to that singular test failure tomorrow :thinking:

internal/workers/workers.go
```diff
@@ -0,0 +1,71 @@
+package workers
```
tsmethurst, line 1: license

tsmethurst (commented):
looks good to me! good work :) the retry of processing on canceled contexts is gonna be really handy i think

PR #1489, opened 2023-02-13
Support multiple subscriptions on single websocket connection

### LINKED ISSUE #1418: [bug] [Ice Cubes] Multiple streaming subscriptions on single websocket

The Ice Cubes client (https://github.com/Dimillian/IceCubesApp) fails to auto-update the timeline. 

I think this is because GoToSocial doesn't yet support multiplexing multiple subscriptions over a single websocket connection.

v0.6.0 git-3283900

### GoToSocial Arch

x86_64

### Browser version

_No response_

### What happened?

I took a look in the Xcode debugger. Ice Cubes is connecting to `/api/v1/streaming`, using a "Authorization" HTTP header with the usual bearer token. It includes no query parameters with the request, but rather sends websocket messages to subscribe once the connection is established.

Looking at the GoToSocial source:

- StreamGETHandler() fails quickly if the "stream" query key is missing. I think that is what is happening here.
- Next, the auth token is checked. The function looks for it in the query keys, and also the "Sec-Websocket-Protocol" header but does not appear to check the HTTP "Authorization" header.

GoToSocial should accept the "Authorization" header as an alternative to the "access_token" query parameter and "Sec-Websocket-Protocol" header.

GoToSocial should allow the websocket to be established even when there is no "stream" query parameter.

GoToSocial should process (un)subscribe messages sent by the client over the websocket stream.

### How to reproduce it?

Launch Ice Cubes and let it connect to GoToSocial and view your timeline.

If you run it in the Xcode debugger you can see it log "There was a bad response from the server.". It attempts to reconnect every few seconds. At the server end you can see it log the HTTP 400 status after each attempt.

This PR has a bit of a description on how things work in Mastodon (I found the API docs to be unclear). https://github.com/mastodon/mastodon/pull/14524

When Ice Cubes can establish the websocket (tested with Mastodon on mas.to), it sends two subscription messages:
```
{ "type": "subscribe", "stream": "user" }
{ "type": "subscribe", "stream": "direct" }
```

#1367 could be related to this issue.

darrinsmart:
I had a go at implementing a fix for #1418. This allows clients to subscribe to multiple event streams on a single [WebSocket connection.](https://docs.joinmastodon.org/methods/streaming/#websocket)

# Description

* Clients can now authenticate WebSocket connections using OAuth in addition to the existing token query parameter & `Sec-Websocket-Protocol` header.

* The stream.Stream struct now stores a set of timeline strings, rather than just a single one. Events matching any of these timelines are forwarded to the client over the WebSocket.

* JSON messages received from the client are processed for subscribing and unsubscribing.

I've tested this with Ice Cubes, which now gets live timeline updates. I also checked Pinafore and Metatext, which worked before, and still work.

Note that this is my first time writing Go code, and pretty much my first time reading it too. I've tried to follow all the guidelines but if I've got it wrong please let me know. I'm happy to make any suggested changes or rework the implementation as required.

closes #1418

## Checklist

### DIFF
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -20,14 +20,16 @@ package streaming
 
 import (
 	"context"
-	"errors"
-	"fmt"
 	"time"
 
 	"codeberg.org/gruf/go-kv"
 	apiutil "github.com/superseriousbusiness/gotosocial/internal/api/util"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
+	"github.com/superseriousbusiness/gotosocial/internal/oauth"
+	streampkg "github.com/superseriousbusiness/gotosocial/internal/stream"
+	"golang.org/x/exp/slices"
 
 	"github.com/gin-gonic/gin"
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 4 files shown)


NyaaaWhatsUpDoc:
@darrinsmart would you like me to rebase this and fix the conflicts?

darrinsmart:
@NyaaaWhatsUpDoc that's kind of you, thanks. But it was easy enough for me to rebase. I removed an unnecessary for-loop label while I was in there.

tsmethurst:
@NyaaaWhatsUpDoc lemme know if you wanna review this one again :)

tsmethurst:
I'm gonna merge this bad boy so we can try it out on main and see if any issues arise :) Thank you very much for your contribution! Congrats on the Go coding :D

PR #1492, opened 2023-02-13
[chore] Webkit frontend fixes

f0x52:
# Description

As reported:
1. There is a horizontal scrollbar when a vertical scrollbar is needed (firefox and safari on osx)
2. also in safari mobile the pofile card is unnecessarily wide and hides all the inputs:

This fix should address 2, as tested in Epiphany (Webkit), and might fix 1 too, but can't reproduce it here in the first place.

## Checklist

### DIFF
--- a/web/source/css/base.css
+++ b/web/source/css/base.css
@@ -47,7 +47,6 @@ html, body {
 
 body {
 	line-height: 1.5em;
-	min-height: 100vh;
 	position: relative;
 }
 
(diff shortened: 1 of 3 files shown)


tsmethurst:
This looks OK to me. Does it work alright on firefox and chrome(ium)?

f0x52:
Firefox yes, haven't tested chromium

tsmethurst:
Tested on chromium and it looks good! :)

Issue #881, opened 2022-10-03
[feature] Customizable instance pages


f0x52:
Allow adding and editing custom instance pages, with some defaults for ToS/CoC, optional public display of instance blocks etc. Settings to link them on the frontpage and/or in footer links

Super-issue for
- https://github.com/superseriousbusiness/gotosocial/issues/143
- https://github.com/superseriousbusiness/gotosocial/issues/378
- https://github.com/superseriousbusiness/gotosocial/issues/707
- https://github.com/superseriousbusiness/gotosocial/issues/729

PR #1495, opened 2023-02-13
[feature] About page

f0x52:
# Description
Basic /about page, using the full instance description and generating some more info from various settings. Also links to `/about/suspended` if enabled.

image

## Checklist

### DIFF
--- a/internal/web/about.go
+++ b/internal/web/about.go
@@ -19,49 +19,29 @@
 package web
 
 import (
-	"fmt"
 	"net/http"
 
 	"github.com/gin-gonic/gin"
 	apiutil "github.com/superseriousbusiness/gotosocial/internal/api/util"
 	"github.com/superseriousbusiness/gotosocial/internal/config"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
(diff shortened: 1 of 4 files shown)


f0x52:
I think this will serve as a decent base combining the editable long description with some auto-generated info about the settings, but should be override-able when we get more generic 'instance pages' support (https://github.com/superseriousbusiness/gotosocial/issues/881)

f0x52:
something to fix: custom emoji are displayed way too big in the contact card :")
image

tsmethurst:
Should i merge this now and we can fix the emoji stuff in a separate patch? or do you wanna do it in this PR?

f0x52:
might as well wait with merging till it's fixed I think, should be soon(tm)

xmgz:
Excuse me, may I propose an enhacement here? I think it's the right place.

wouldn't it be good have a direct link to "/about" in the profile/user-TL page footer? As it happens with "contact" and "instance version". Otherwise this **About** page is hidden without public links (as far as I'd seen) for casual visitiors.

Having **markdown** for the long instance description would also help, adding some formatting. In my opinion.

thank you.

PR #1496, opened 2023-02-13
[bugfix] Fix up `error getting account avatar/header` errors, other small fixes

tsmethurst:
# Description

>

This pull request:

1. Fixes a bug where avatars were being set as 'avatar=false' by accident, leading to them being pruned and then later not found.
2. Allows for situations where a remote header/avatar url has changed to empty, and should therefore be unset.
3. Fixes the `admin media prune remote` command to initialize db state and caches properly.
4. Updates media attachment select db logic to select a bunch of IDs and then select attachments from those IDs, rather than doing big ol' queries.
5. Removes some unused db relations that were causing us to make bigger queries than we needed to be.

Tested on goblin.technology and seems OK there :)

## Checklist

### DIFF
--- a/internal/federation/dereferencing/account.go
+++ b/internal/federation/dereferencing/account.go
@@ -366,40 +366,39 @@ func (d *deref) fetchRemoteAccountAvatar(ctx context.Context, tsport transport.T
 	unlock := d.derefAvatarsMu.Lock()
 	defer unlock()
 
-	if processing, ok := d.derefAvatars[avatarURL]; ok {
-		// we're already dereferencing it, just wait til it's loaded.
-		if _, err := processing.LoadAttachment(ctx); err != nil {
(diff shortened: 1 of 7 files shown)


internal/federation/dereferencing/account.go
```diff
@@ -357,7 +381,7 @@ func (d *deref) fetchRemoteAccountAvatar(ctx context.Context, tsport transport.T
 
 	// Create new media processing request from the media manager instance.
 	processing, err := d.mediaManager.ProcessMedia(ctx, data, nil, accountID, &media.AdditionalMediaInfo{
-		Avatar:    func() *bool { v := false; return &v }(),
+		Avatar:    func() *bool { v := true; return &v }(),
```
tsmethurst, line 384: this was the actual bug

Issue #1469, opened 2023-02-10
[chore] De-interface a bunch of things!


tsmethurst:
The code I originally put in place for early versions of GtS contained a wacky amount of interface abstractions, because I wanted to make sure we had enough elbow room for dependency injection later. However, a lot of these interfaces don't really add anything and are unlikely to ever need dependency injection.

To make the codebase more readable + easier to navigate, we should de-interface these and make them into simple struct pointers instead.

Candidates we should consider for this (not exhaustive, there are probably more):

- [ ] Federator
- [ ] Dereferencer
- [x] Media manager
- [ ] Oauth server
- [ ] OIDC IDP
- [x] Processor
- [x] Sub-processors
- [x] Visibility filter
- [x] Typeutils typeconverter

Places where interface abstraction are actually useful:

- Cache: might want to inject dependency on external cache later
- DB: might want to implement another db struct later
- Email: sender vs no-op sender
- Router: handy to expose only functions we want instead of passing the whole gin router around
- Storage: we already have a few different implementations
- Timeline / timeline manager: we might want to dependency inject external storage of timelines later

tsmethurst:
Gonna close this because we're sort of just doing this as we go along and I don't think we need a separate issue for it.

PR #1501, opened 2023-02-14
[chore] Deinterface processor and subprocessors

tsmethurst:
# Description

>

This PR refactors the processor (and sub processors) to deinterface it (#1469).

Also some general tidying up of subprocessor functions, with an eye towards naming consistency.

I considered moving some of the other files inside `internal/processing` into their own subfolders/subprocessors, but most of them are just two or three functions. If they grow, we should consider moving them into subfolders, but now it's probably not necessary!

- [x] Account subprocessor
- [x] Admin subprocessor
- [x] Federation subprocessor
- [x] Media subprocessor
- [x] Report subprocessor
- [x] Status subprocessor
- [x] Streaming subprocessor
- [x] User subprocessor
- [x] Processor itself

## Checklist


NyaaaWhatsUpDoc:
Instead of prefixing all of the function names by their processor name (it stutters a bit reading it internally), what if we were to just rename them to be say `processor.GetStatus()` and `processor.GetAccount()`? Given different modules handle logic for different areas, there shouldn't need to be any namespace collisions :thinking:. Or alternatively another might be to make each of the subprocessors available from the processor itself e.g. `processor.Accounts().Get()` kind of thing?

What do you think?

tsmethurst:
> another might be to make each of the subprocessors available from the processor itself e.g. processor.Accounts().Get() kind of thing?

I think this might be a better approach yeah. That also gives us the option in future of passing only the required subprocessor into those api route modules :eyes:

tsmethurst:
@NyaaaWhatsUpDoc have now updated this PR to expose subprocessors by name, and reduced some of the verbosity in their function names :) https://github.com/superseriousbusiness/gotosocial/pull/1501/commits/df2089218fb0f48af510a73c3ecb5eb34c19bc5d

PR #1506, opened 2023-02-14
[bug] Fix nginx fileserver caching example

daenney:
# Description

This updates the example to ensure the nginx proxies the request on to GTS if the file is not found on disk. This can happen due to media pruning.

## Checklist

### DIFF
--- a/docs/installation_guide/advanced.md
+++ b/docs/installation_guide/advanced.md
@@ -390,11 +390,21 @@ server {
     add_header Cache-Control "public";
   }
 
+  location @fileserver {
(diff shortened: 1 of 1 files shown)


daenney:
Actually no, this now goes through to the backend always, because the `$uri` contains `/fileserver`.

daenney:
> Actually no, this now goes through to the backend always, because the `$uri` contains `/fileserver`.

Never mind, it does work. Debugging this with headers gets confusing.

tsmethurst:
Thanks daenney!

PR #1507, opened 2023-02-15
Fix 410 Gone race on account deletes

### LINKED ISSUE #1313: [bug] 400 errors on delete requests

ActivityPub user delete messages cause 400 errors due to trying to webfinger the deleted user. Example request:

```
POST /users/sentynel/inbox HTTP/1.0
Host: social.sentynel.com
Connection: upgrade
X-Forwarded-For: REMOVED
X-Forwarded-Proto: https
Content-Length: 771
User-Agent: http.rb/5.1.0 (Mastodon/4.0.2; +https://mastodonapp.uk/)
Date: Thu, 05 Jan 2023 23:17:09 GMT
Accept-Encoding: gzip
Digest: SHA-256=wZOpZiEzBCDDrJ+WVQmzjsjfjsPd+N7gBn2ymmDrw6k=
Content-Type: application/activity+json
Signature: keyId="https://mastodonapp.uk/users/REMOVED#main-key",algorithm="rsa-sha256",headers="(request-target) host date digest content-type",signature="REMOVED"

{"@context":"https://www.w3.org/ns/activitystreams","id":"https://mastodonapp.uk/users/REMOVED#delete","type":"Delete","actor":"https://mastodonapp.uk/users/REMOVED","to":["https://www.w3.org/ns/activitystreams#Public"],"object":"https://mastodonapp.uk/users/REMOVED","signature":{"type":"RsaSignature2017","creator":"https://mastodonapp.uk/users/REMOVED#main-key","created":"2023-01-05T23:16:58Z","signatureValue":"REMOVED"}}
```

Error response (note also that this is contained in the HTML error template, but should probably be the JSON one for the API?).
```
1Bad Request: couldn't get requesting account https://mastodonapp.uk/users/REMOVED: item could not be retrieved: GetRemoteAccount: error while fingering: fingerRemoteAccount: error fingering @REMOVED@mastodonapp.uk: GET request to https://mastodonapp.uk/.well-known/webfinger?resource=acct:REMOVED@mastodonapp.uk failed (410): 410 Gone
```

We probably still need to do this request to verify the user is actually gone (since we can't check the message signature unless we already have their key), but 410 is the expected result.

v0.6.0

### GoToSocial Arch

_No response_

### Browser version

_No response_

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

Sentynel:
# Description

This handles a case where a 410 Gone could occur at an unexpected place in the processing of account deletion requests for accounts we were previously unaware of.

This is currently running on my live GTS instance, though the actual race hasn't yet happened so I can't strictly prove it's working yet.

closes #1313 for real

## Checklist

### DIFF
--- a/internal/federation/federatingprotocol.go
+++ b/internal/federation/federatingprotocol.go
@@ -209,6 +209,16 @@ func (f *federator) AuthenticatePostInbox(ctx context.Context, w http.ResponseWr
 		transport.WithFastfail(ctx), username, publicKeyOwnerURI, false,
 	)
 	if err != nil {
+		if errors.Is(err, transport.ErrGone) {
+			// This is the same case as the http.StatusGone check above.
+			// It can happen here and not there because there's a race
+			// where the sending server starts sending account deletion
+			// notifications out, we start processing, the request above
+			// succeeds, and *then* the profile is removed and starts
+			// returning 410 Gone, at which point _this_ request fails.
+			w.WriteHeader(http.StatusAccepted)
+			return ctx, false, nil
+		}
 		return nil, false, fmt.Errorf("couldn't get requesting account %s: %s", publicKeyOwnerURI, err)
 	}
 


tsmethurst:
Hell yeah, nice one! Will review rn :)

tsmethurst:
This seems like a good solution. Is there any risk we might end up not processing the delete correctly, but still return 202 to the caller?

Sentynel:
I don't believe so - if you look in `GetAccountByURI()` https://github.com/superseriousbusiness/gotosocial/blob/main/internal/federation/dereferencing/account.go#L63-L82, you can see that errors in `enrichAccount()` are non-fatal in the case that we already have the account in the database (it just falls back to using the one it already had and swallows the error). It's only in the case we didn't have the account already that failure to look it up is fatal. Therefore, we can only get a 410 Gone at this point in the code if we didn't already have the account in the database, and doing nothing is therefore safe.

tsmethurst:
Brilliant, thank you! Will merge before 0.7.0 so we can include this fix then :) Much appreciated

NyaaaWhatsUpDoc:
Nice fix! Well caught :)

PR #1525, opened 2023-02-18
[performance] cache media attachments

NyaaaWhatsUpDoc:
Changes:
- add media attachment caching
- slightly tweak default cache config values
- update all media attachment db calls to go through cache
- update envparsing test

Checklist:


NyaaaWhatsUpDoc:
The main remaining issue here is that different testing areas are using different state.State instances (and therefore cache instances) which is causing the failures. *I think*.

NyaaaWhatsUpDoc:
errors with "no space left on device" early on 🤔

internal/cache/gts.go
```diff
@@ -145,6 +153,7 @@ func (c *gtsCaches) Stop() {
 	tryUntil("stopping gtsmodel.DomainBlock cache", 5, c.domainBlock.Stop)
 	tryUntil("stopping gtsmodel.Emoji cache", 5, c.emoji.Stop)
 	tryUntil("stopping gtsmodel.EmojiCategory cache", 5, c.emojiCategory.Stop)
+	tryUntil("stopping gtsmodel.MediaAttention cache", 5, c.media.Stop)
```
tsmethurst, line 156: Media attention :D

internal/db/bundb/media.go
```diff
@@ -51,22 +51,52 @@ func (m *mediaDB) GetAttachmentByID(ctx context.Context, id string) (*gtsmodel.M
...
+func (m *mediaDB) PutAttachment(ctx context.Context, media *gtsmodel.MediaAttachment) error {
+	return m.state.Caches.GTS.Media().Store(media, func() error {
+		_, err := m.conn.NewInsert().Model(media).Exec(ctx)
+		return m.conn.ProcessError(err)
+	})
+}
+
+func (m *mediaDB) UpdateAttachment(ctx context.Context, media *gtsmodel.MediaAttachment, columns ...string) error {
+	// Update the media's last-updated
+	media.UpdatedAt = time.Now()
```
tsmethurst, line 76: It's possible here that if the caller provides columns which exclude last_updated, then the last updated value won't be set correctly. What I've been doing in other functions is checking if columns is set, and if so, adding 'last_updated' to the list by default. There's an example here: https://github.com/superseriousbusiness/gotosocial/blob/fe6c8b8152d723ff3788e35c79b2e840926d5b22/internal/db/bundb/status.go#L254-L258

There's likely a better way of doing this, however :P

internal/cache/gts.go
NyaaaWhatsUpDoc: ... :eyes:

internal/db/bundb/media.go
NyaaaWhatsUpDoc: ah gotcha shall handle this

PR #1527, opened 2023-02-18
[feature] Add metrics

daenney:
# Description

I'm opening this as a draft PR so I can tempt @NyaaaWhatsUpDoc :stuck_out_tongue:.

Current state:
* Instruments HTTP endpoints, haven't gotten around to the client yet
* I dislike magical global objects in prometheus like the default registry hence setting up an explicit one
* I've tried to find a way to make it as optional as possible and not load any machinery if metrics are not enabled
* There are probably better ways of doing and structuring this

## Checklist


NyaaaWhatsUpDoc:
You have successfully tempted me... When I get time I'll try bring across my existing metrics changes :')

daenney:
I put it "in production". You can see them at https://ap.dny.social/metrics.

I hadn't realised that `c.Request.Path` would include stuff like `/users/<username>`, that's going to do some cardinality explosion we want to avoid. I guess we could use `c.HandlerName()` instead? It would be nice if we could get a name like `/users/:id/inbox` instead.

daenney:
One thing that's probably worth thinking about is whether we want to partition by endpoint and method. This creates a ton of timeseries and histograms are generally expensive to ingest and do math over. It's possible to use summaries instead but that means the compute cost is now on the client instead of the probably beefier Prometheus server.

I'm thinking having separate subsystems, like `clientapi`, `federation` etc. + partitioning by status code might be enough? That should give you an indication of where a problem might be in general and trace the rest through logs or proper tracing once we add that.

daenney:
I took a peak at the Hachyderm public dashboard: https://grafana.hachyderm.io/public. One thing that's interesting there is they only surface HTTP GET/POST 200s for the federation endpoints. Maybe we actually want to do something similar since that's really the metric you're the most likely to be interested in, and bundle everything else in a histogram that doesn't split out by method and code?

daenney:
I'm going to axe this PR for a bit. I've been looking around and there's an alternative way I want to try and do this directly through the OpenTelemetry metrics. That avoids needing to potentially migrate those over later and would put us close to being able to wire up tracing too.

I'll be tinkering around with this for the weekend. I'll probably fax a PR sometime in the week if I can get a complete demo up.

PR #1534, opened 2023-02-19
Fix failure to look up remote profiles with duplicate emojis in some cases

### LINKED ISSUE #1532: [bug] Error when putting account with same emoji appeared multiple times on profile in database

When searching for some accounts, the following error will appear:
```
Error #01: error looking up account: enrichAccount: error putting in database: already exists
```
   
SQL statements executed just before the error message thrown out:
```sql
BEGIN;
INSERT INTO "account_to_emojis" ("account_id", "emoji_id") VALUES ('01GSMPNGHC04BXFECJH0VYRGXD', '01FNYQX4D65YVWRD5JXYS0NCJQ');
INSERT INTO "account_to_emojis" ("account_id", "emoji_id") VALUES ('01GSMPNGHC04BXFECJH0VYRGXD', '018BFQRPDTP5MPQSB314X34QF7');
INSERT INTO "account_to_emojis" ("account_id", "emoji_id") VALUES ('01GSMPNGHC04BXFECJH0VYRGXD', '01FNYQX4D65YVWRD5JXYS0NCJQ');
ROLLBACK;
```
    
So I think it's those emojis that appeared more than once on the profile that caused the error.
       
Further more, I created a account to verify my guess: `@test@nya.one`, and the same error occurred as expected.
    
Maybe there should have a filter to filter out those duplicated `account to emojis` pairs? I haven't used the orm before, so I am not sure.
https://github.com/superseriousbusiness/gotosocial/blob/70398891b838ae29e5f8e2ce76301a43995c0a5d/internal/gtsmodel/account.go#L49

v0.7.0

### GoToSocial Arch

_No response_

### What happened?

I was searching for some accounts to follow, but the server responded with a 500 error.

No error occurs when searching for those valid accounts.

### How to reproduce it?

Search `@Hokori@nya.one` or `@test@nya.one`.

_No response_

Sentynel:
# Description

Fix a bug where Misskey could send accounts with duplicate emojis, which we would then fail to process.

Also tidies up the local profile edit logic to make sure we don't set duplicate emojis that way, just in case.

closes #1532

## Checklist

### DIFF
--- a/internal/ap/extract.go
+++ b/internal/ap/extract.go
@@ -453,6 +453,7 @@ func ExtractHashtag(i Hashtaggable) (*gtsmodel.Tag, error) {
 // ExtractEmojis returns a slice of emojis on the interface.
 func ExtractEmojis(i WithTag) ([]*gtsmodel.Emoji, error) {
 	emojis := []*gtsmodel.Emoji{}
+	emojiMap := make(map[string]*gtsmodel.Emoji)
 	tagsProp := i.GetActivityStreamsTag()
 	if tagsProp == nil {
 		return emojis, nil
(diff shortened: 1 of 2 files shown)


tsmethurst:
Thank you!

Issue #947, opened 2022-11-02
[bug] 404 when trying to pin from pinafore


flo-dhalluin:
### Describe the bug with a clear and concise description of what the bug is.

Trying to pin a post from pinafore yields a 404 message flashing.

### What's your GoToSocial Version?

v0.5.2

### GoToSocial Arch

binary amd64

### What happened?

create a post from pinafore ( public ) -> pin -> 404 msg

pinafore appears to calls POST /api/v1/statuses/{status_id}/pin <- 404. 

Not sure if because pin is not implemented or it's actually a bug.. thanks !

tsmethurst:
Heya, it's just not implemented yet, so this is expected and normal :)

lzap:
Damn, I want to be able to pin. Let me implement this, is that a big feature or smallish? :-)

Can you direct me where to start digging? I am experienced software engineer, just basic information like "new column needs to be added, new endpoint, whatever". Cheers!

tsmethurst:
> Let me implement this, is that a big feature or smallish?

Just wait a bit, we're currently in the process of reordering our roadmap document to have a better idea of what needs to be implemented when, to stop things getting chaotic ;)

PR #1547, opened 2023-02-22
[feature] Client API endpoints + v. basic web view for pinned posts

tsmethurst:
# Description

>

This pull request implements client API endpoints to allow users to pin + unpin toots (maximum 10, currently), and to allow API callers to view pinned toots when getting an account's statuses.

It also makes some minor changes to our html templates and css to show a basic view of pinned toots at the top of a profile in the web view.

This PR does not yet include logic for federating pinned posts in and out of GoToSocial, that will be in a separate PR.

closes https://github.com/superseriousbusiness/gotosocial/issues/947

## Checklist


internal/api/client/accounts/statuses_test.go
```diff
@@ -115,7 +115,176 @@ func (suite *AccountStatusesTestSuite) TestGetStatusesPublicOnlyMediaOnly() {
 		suite.Equal(apimodel.VisibilityPublic, s.Visibility)
 	}
 
-	suite.Equal(`<http://localhost:8080/api/v1/accounts/01F8MH17FWEB39HZJ76B6VXSKF/statuses?limit=20&max_id=01F8MH75CBF9JFX4ZAD54N0W0R&exclude_replies=false&exclude_reblogs=false&pinned_only=false&only_media=true&only_public=true>; rel="next", <http://localhost:8080/api/v1/accounts/01F8MH17FWEB39HZJ76B6VXSKF/statuses?limit=20&min_id=01F8MH75CBF9JFX4ZAD54N0W0R&exclude_replies=false&exclude_reblogs=false&pinned_only=false&only_media=true&only_public=true>; rel="prev"`, result.Header.Get("link"))
+	suite.Equal(`<http://localhost:8080/api/v1/accounts/01F8MH17FWEB39HZJ76B6VXSKF/statuses?limit=20&max_id=01F8MH75CBF9JFX4ZAD54N0W0R&exclude_replies=false&exclude_reblogs=false&pinned=false&only_media=true&only_public=true>; rel="next", <http://localhost:8080/api/v1/accounts/01F8MH17FWEB39HZJ76B6VXSKF/statuses?limit=20&min_id=01F8MH75CBF9JFX4ZAD54N0W0R&exclude_replies=false&exclude_reblogs=false&pinned=false&only_media=true&only_public=true>; rel="prev"`, result.Header.Get("link"))
+}
+
+func (suite *AccountStatusesTestSuite) TestGetStatusesPinnedOnly1() {
```
NyaaaWhatsUpDoc, line 121: It's probably worth changing these test names to be descriptive of what they're doing instead of 1,2,3.

internal/api/client/accounts/statuses_test.go
```diff
@@ -115,7 +115,176 @@ func (suite *AccountStatusesTestSuite) TestGetStatusesPublicOnlyMediaOnly() {
...
+	apimodelStatuses := []*apimodel.Status{}
+	err = json.Unmarshal(b, &apimodelStatuses)
+	suite.NoError(err)
+	suite.Len(apimodelStatuses, 2)
+	suite.Empty(result.Header.Get("link"))
+
+	for _, s := range apimodelStatuses {
+		// Requesting account doesn't own these
+		// statuses, so pinned should be false.
+		suite.False(s.Pinned)
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: Okay wait I've realized this is still returning the pinned statuses it should do, but it seems a bit odd to only set this variable when it's the owner. I feel like this could end up a bit misleading to clients.

internal/gtsmodel/statuspin.go
```diff
@@ -0,0 +1,32 @@
...
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package gtsmodel
+
+import "time"
+
+// StatusPin refers to a status that an account has 'pinned' to the top of its profile.
```
NyaaaWhatsUpDoc, line 23: Do we need an entirely separate table for this? Would a `Pinned` field on the existing GTS status model not work for this? :thinking:. Pins should only be by account owner themselves so simply having `Pinned` on the status model and having the `AccountID` field would providing everything we have here bar the created/updated times.

NyaaaWhatsUpDoc:
Some questions about implementation details so far, but otherwise looking generally good. Once we've finished talking through these I'll finish reviewing :)

internal/gtsmodel/statuspin.go
daenney: In the case of potentially adding a pinned column instead, it might be interesting to use a nilable timestamp field instead of the boolean. It has the nice side effect of communicating not only if a post was pinned but also when that happened which can be useful when debugging issues.

NyaaaWhatsUpDoc: Oh that's a good idea. So something like `pinned_at`?

daenney: Yes, exactly like that. It's a handy trick to do on anything that you'd normally store as a boolean. It's a bit more expensive to store, 1 vs 8 bits typically, but you often end up wanting that information.

tsmethurst: Mmm, we don't need a separate table for it no, it just felt a bit cleaner to me. If we have a `pinned_at` or `pinned` field on the status--which is what Mastodon does--then it's the only field on the status of its kind. The rest of a status's fields pertain to properties *of* the status (like when it was created, content, visibility properties etc), rather than actions that have been performed *on* a status (like adding it to a collection). Conceptually, it also felt a bit more in-line with ActivityPub to have a separate table which encapsulates the idea of featured collections by gathering status pins.

By having it in a separate table, we also hold open the future possibility of allowing users to feature other people's (Public) statuses, though ofc we'd still need to discuss that and think about the implications of it.

That said, I'm not married to having it in a separate table, so happy to change it if you prefer :)

internal/api/client/accounts/statuses_test.go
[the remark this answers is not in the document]
tsmethurst: Yeah honestly it's a very strange part of the Mastodon API to have the `pinned` property behave in this way. I'm not really sure what to do about it without breaking existing client expectations though.

internal/api/client/accounts/statuses_test.go
tsmethurst: Will do!

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1547/commits/84b09e637291bcd1a9bc9786ef945f40e87f8c1a

internal/gtsmodel/statuspin.go
NyaaaWhatsUpDoc: It not only means one less table, it's one less cache to manage, one less cache cleaning operation that has to be performed every "sweep freq". And I'm not really a fan of the idea of being able to pin other people's statuses. Having people's statuses boosted is sort of to be expected, but literally pinning someone else's status feels a bit odd. I don't think that's something most people will expect in making a public status, that it could be pinned at the top of someone else's profile.

Sorry to be so opinionated on it :sweat_smile: though that's not to say I have "the one true, correct opinion" obviously

internal/api/client/accounts/statuses_test.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: Ah right that is a pain... Thanks again Mastodon

internal/gtsmodel/statuspin.go
tsmethurst: No worries! I think you're right, I'll make the changes tomorrow 😎

tsmethurst: Bam! https://github.com/superseriousbusiness/gotosocial/pull/1547/commits/efef51ba973a441aaebb664e2460c730498dc028

internal/web/profile.go
```diff
@@ -125,7 +139,8 @@ func (m *Module) profileGETHandler(c *gin.Context) {
 		"robotsMeta":       robotsMeta,
 		"statuses":         statusResp.Items,
 		"statuses_next":    statusResp.NextLink,
-		"show_back_to_top": showBackToTop,
+		"pinned_statuses":  pinnedResp.Items,
```
NyaaaWhatsUpDoc, line 142: Surely this will panic if !paging, i.e. pinnedResp == nil. Would it be better to instead do `var pinnedItems []any` and set these within the `!paging` clause i.e. `pinnedItem = pinnedResp`.

NyaaaWhatsUpDoc: Wait just realized that item is always allocated. But this way might be a lil' neater anywho :innocent:

internal/processing/status/pin.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Only pin status if it's not pinned already.
+	if !targetStatus.PinnedAt.IsZero() {
+		// Ensure we have enough slots to pin this.
+		pinnedCount, err := p.db.CountAccountPinned(ctx, requestingAccount.ID)
+		if err != nil {
+			return nil, gtserror.NewErrorInternalError(fmt.Errorf("error checking number of pinned statuses: %w", err))
+		}
+
+		if pinnedCount >= allowedPinnedCount {
```
NyaaaWhatsUpDoc, line 64: not an issue whatsover just a funny thought: this is gonna be the first thing goat hacks on her instance when she starts fucking around :p

internal/processing/status/pin.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("error fetching status %s: %w", targetStatusID, err))
+	}
+
+	if targetStatus.AccountID != requestingAccount.ID {
+		err = fmt.Errorf("status %s does not belong to account %s", targetStatusID, requestingAccount.ID)
+		return nil, gtserror.NewErrorUnprocessableEntity(err, err.Error())
+	}
+
+	if targetStatus.Visibility == gtsmodel.VisibilityDirect {
```
NyaaaWhatsUpDoc, line 46: how does it work out pinning unlisted and followers-only statuses? does it produce expected behaviour for those status privacy types?

NyaaaWhatsUpDoc:
2 final questions regarding expected behaviour just to double check this is expected behaviour that will work with existing clients. And one tiny nitpick.

But otherwise the code looks really good and once you have addressed those feel free to merge this :)

internal/processing/status/pin.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, gtserror.NewErrorNotFound(fmt.Errorf("status %s does not belong to account %s", targetStatusID, requestingAccount.ID))
+	}
+
+	if targetStatus.Visibility == gtsmodel.VisibilityDirect {
+		err := errors.New("cannot pin direct messages")
+		return nil, gtserror.NewErrorBadRequest(err, err.Error())
+	}
+
+	// Only unpin status if it's pinned.
+	if targetStatus.PinnedAt.IsZero() {
```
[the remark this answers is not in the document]
tsmethurst: Ah, I just tried it on mastodon and apparently not! It should return 422. For other endpoints for like, bookmarking and muting a status, it'll just return 200 if it's already bookmarked or muted, but not for this :') lord. Will fix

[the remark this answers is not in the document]
tsmethurst: Oh but unpinning twice just returns 200. Alright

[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1547/commits/0f6900a526351a1dc25fe970930243dab6ef9101 addressed here and added some tests :)

internal/processing/status/pin.go
tsmethurst: Yep yep! All selected pins are always run through visibility filters to make sure they can be seen, before they're returned to clients.

PR #1550, opened 2023-02-22
[chore] improve opengraph descripiton tag

### LINKED ISSUE #1549: [bug] og:description mangles escaped shell commands

When generating the og:description OpenGraph tag for a plaintext post, GtS strips escape `(\)` characters, which can result in the OpenGraph embed presenting a shell command or code that is not what was originally authored in the post.

v0.7.1 tag, source build

### Browser version

n/a

### What happened?

I authored a [post](https://electric.marf.space/@trysdyn/statuses/01GSWF656AV121H88Z28N2VSKA). In this post I include the following shell command as a recommendation for people to try.

`echo '\e]8;;http://example.com\e\This is a link\e]8;;\e'`

Someone then linked my post in Discord, which presented the unfurl as generated by the OpenGraph tags. There, the command was mangled to this:

`echo 'e]8;;http://example.comeThis is a linke]8;;e`

Note the missing backslashes. This significantly changes the impact of the command :)

I've verified the slashes are stripped on the og:description tag, so it's not Discord's doing.

The escape slashes to be left in the post text in og:description.

### How to reproduce it?

Post anything with the Public visibility level that contains any number of \ characters, then inspect the post in the frontend and view source to see the \s were stripped in og:description.

_No response_

daenney:
# Description

This changes parseDescription to properly encode things to be safe for usage without removing things like backslashes that may be relevant.

* text.SanitizePlaintext already calls html.UnescapeString so we don't have to do that
* Replace `\n` with space early
* Remove duplicate white-space by splitting on fields and joining
* HTML-escape the string we have
* For extra certainty, encode the backslash as `&bsol;`

Fixes #1549

## Checklist

### DIFF
--- a/internal/web/opengraph.go
+++ b/internal/web/opengraph.go
@@ -134,11 +134,11 @@ func parseTitle(account *apimodel.Account, accountDomain string) string {
 // parseDescription returns a string description which is
 // safe to use as a template.HTMLAttr inside templates.
 func parseDescription(in string) string {
-	i := html.UnescapeString(in)
-	i = text.SanitizePlaintext(i)
-	i = strings.ReplaceAll(i, "\"", "'")
-	i = strings.ReplaceAll(i, `\`, "")
+	i := text.SanitizePlaintext(in)
 	i = strings.ReplaceAll(i, "\n", " ")
+	i = strings.Join(strings.Fields(i), " ")
+	i = html.EscapeString(i)
+	i = strings.ReplaceAll(i, `\`, "&bsol;")
 	i = trim(i, maxOGDescriptionLength)
 	return `content="` + i + `"`
 }
--- a/internal/web/opengraph_test.go
+++ b/internal/web/opengraph_test.go
@@ -1,0 +1,50 @@
+/*
+   GoToSocial
+   Copyright (C) 2021-2023 GoToSocial Authors admin@gotosocial.org
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package web
+
+import (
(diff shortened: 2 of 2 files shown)


daenney:
If you now paste the resulting string in something like Codepen, you'll see it rendered like the original. Copy-pasting from there to your terminal has the backslashes the way you expect them to be.

The one thing I don't know if what the Discord preview will do, but I assume they run it through some HTML renderer that should be happy with the `&bsol;`

trysdyn:
> The one thing I don't know if what the Discord preview will do, but I assume they run it through some HTML renderer that should be happy with the `&bsol;`

Yeah. I blatted a quick test page onto my domain with a bunch of typical "fuzzer" strings in an og:description and linked it in Discord and a couple other places. `&bsol;` behaves as expected; should be fine for this.

daenney:
We could expand a bit on this, as https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#output-encoding-rules-summary says:

Encoding Type | Encoding Mechanism
-- | --
HTML Entity Encoding | Convert `&` to `&amp;`, Convert `<` to `&lt;`, Convert `>` to `&gt;`, Convert `"` to `&quot;`, Convert `'` to `&#x27`;
HTML Attribute Encoding | Except for alphanumeric characters, encode all characters with the HTML  Entity `&#xHH;` format, including spaces. (`HH` = Hex Value)

I think the second rule is a little extreme so encode everything that's non-alphanumeric, including spaces.

The `html.Escape()` is essentially doing the "HTML Entity Encoding" which is a subset of the "HTML Attribute Encoding". Since we're talking `meta property="og:description" content="CONTENT">` we should apply attribute encoding though most of the risk there only exists if you don't have balanced quotes.

> It’s critical to use quotation marks like " or ' to surround your variables. Quoting makes it difficult to change the context a variable operates in, which helps prevent XSS. Quoting also significantly reduces the characterset that you need to encode, making your application more reliable and the encoding easier to implement.

Since we have the quotes in `content=""` surrounding the output I think we're good, though I'd prefer to move the `content=""` bit itself into the template instead of in the output of `parseDescription`.

tsmethurst:
Nice! Could you add the license text to the top of the new file with the tests in? Then this should be good to merge :)

Thank you!

daenney:
Done! (Sorry I always forget about that one)

tsmethurst:
I think I originally tried adding content="" part to the template itself, but ran into some problem with it iirc (some consequence of go's template sanitization). Feel free to try it again though, if you wanna include that in this pr too that's fine.

daenney:
Ah interesting. I could see that being tricky. I'd say go ahead and merge it. I'll go duel `html/template` another day.

PR #1557, opened 2023-02-25
[chore] Inform user on how to get testrig

daenney:
# Description

This adds a dummy testrig subcommand when we're not building with DEBUG set. Now any invocation of gotosocial testrig, plain or with any kind of subcommands or arguments will result in that error message getting returned.

## Checklist

### DIFF
--- a/cmd/gotosocial/main.go
+++ b/cmd/gotosocial/main.go
@@ -20,6 +20,7 @@ package main
 
 import (
 	"log"
+	"os"
 	godebug "runtime/debug"
 	"strings"
 
@@ -66,6 +67,8 @@ func main() {
 	if debug.DEBUG {
 		// only add testrig if debug enabled.
 		rootCmd.AddCommand(testrigCommands())
+	} else if len(os.Args) > 1 && os.Args[1] == "testrig" {
+		log.Fatalln("gotosocial must be built and run with the DEBUG enviroment variable set to enable and access testrig")
 	}
 
 	// run


daenney:
Based on a suggestion from @f0x52!

daenney:
Before:
```
$ ./gotosocial testrig start
2023/02/25 19:16:42 error executing command: unknown command "testrig" for "gotosocial"
```

After:
```
$ ./gotosocial testrig start
2023/02/25 19:17:06 error executing command: gotosocial must be built and run with the DEBUG enviroment variable set to enable and access testrig
```

f0x52 (approved):

NyaaaWhatsUpDoc:
I think instead of adding testrig dummy commands, it might be worth instead just doing a `log.Fatal()` with help message. that way testrig help cmds are compiled into production builds

daenney:
Oh, that's an idea. So then I'm thinking we use `rootCmd.PreRunE()` instead and in there check for `debug.DEBUG` plus the args?

daenney:
Mmmm, that doesn't work because `rootCmd` doesn't have a `Run` of its own. It only dispatches. And patching that into the existing `PersistentPreRunE` feels a bit nasty. We could put it in the regular `PersistentPreRun` but then all sub-commands would inherit it which doesn't seem like a necessary thing for this one time check.

NyaaaWhatsUpDoc:
I mean you could just make that `if DEBUG` clause then do `else if len(os.Args) > 1 && os.Args[1] == "testrig" { ... }`

daenney:
> if len(os.Args) > 1 && os.Args[1] == "testrig" {

:facepalm: thank you for pointing that out. No idea why I was making it this hard.

NyaaaWhatsUpDoc:
😁

PR #1560, opened 2023-02-26
[feature] Federate pinned posts (aka `featuredCollection`) in and out

tsmethurst:
# Description

>

This pull request implements incoming and outgoing federation of featuredCollection aka pinned posts.

- A list of status URIs are now served at a user's featured collection URI for other servers to dereference.
- When a remote account is fetched, its featured collection will be dereferenced + relevant statuses pinned.
- Updated 'federating with gotosocial' docs to include featured collection stuff.
- Do some tidying up of duplicated code in the `internal/api/activitypub/users` package.
- Add tests for serving featuredCollection.

## Checklist


internal/federation/dereferencing/account.go
```diff
@@ -569,3 +578,150 @@ func (d *deref) fetchRemoteAccountEmojis(ctx context.Context, targetAccount *gts
 
 	return changed, nil
 }
+
+// fetchRemoteAccountFeatured dereferences an account's featuredCollectionURI (if not empty).
+// For each discovered status, this status will be dereferenced (if necessary) and marked as
+// pinned (if necessary). Then, old pins will be removed if they're not included in new pins.
+func (d *deref) fetchRemoteAccountFeatured(ctx context.Context, requestingUsername string, featuredCollectionURI string, accountID string) error {
+	if featuredCollectionURI == "" {
```
NyaaaWhatsUpDoc, line 586: i don't think this should be checked here, i think this should be checked by the caller of this function. otherwise i worry it could hide potential issues

internal/federation/dereferencing/account.go
```diff
@@ -569,3 +578,150 @@ func (d *deref) fetchRemoteAccountEmojis(ctx context.Context, targetAccount *gts
...
+			continue
+		}
+
+		statusURIs = append(statusURIs, statusURI)
+
+		status, _, err := d.GetStatus(ctx, requestingUsername, statusURI, false, false)
+		if err != nil {
+			// We couldn't get the status, bummer.
+			// Just log + move on, we can try later.
+			log.WithContext(ctx).Infof("error getting status from featured collection %s: %s", featuredCollectionURI, err)
```
NyaaaWhatsUpDoc, line 676: Instead of this you can just do `log.Infof(ctx, "message here ....", args...)` :). All of the base Info(), Warn(), Error() etc logging functions were updated to take a context as their first argument. Also I think we should  probably make this an error level log.

internal/federation/dereferencing/account.go
```diff
@@ -569,3 +578,150 @@ func (d *deref) fetchRemoteAccountEmojis(ctx context.Context, targetAccount *gts
...
+			// We got a whole Note. Extract the URI.
+			if note := iter.GetActivityStreamsNote(); note != nil {
+				if id := note.GetJSONLDId(); id != nil && id.IsIRI() {
+					statusURI = id.GetIRI()
+				}
+			}
+		case iter.IsActivityStreamsArticle():
+			// We got a whole Article. Extract the URI.
+			if article := iter.GetActivityStreamsArticle(); article != nil {
+				if id := article.GetJSONLDId(); id != nil && id.IsIRI() {
```
NyaaaWhatsUpDoc, line 654: you don't need to do the final `IsIRI()` check here, you can just straight-up do `id != nil { statusURI = id.GetIRI() } ` since we  do a final check on whether `statusURI` is nil below :) same goes for the switch case above.

internal/federation/dereferencing/account.go
NyaaaWhatsUpDoc: Same goes for the log entry below this on line ~700

internal/federation/dereferencing/account.go
```diff
@@ -569,3 +578,150 @@ func (d *deref) fetchRemoteAccountEmojis(ctx context.Context, targetAccount *gts
...
+	// Now that we know which statuses are pinned, we should
+	// *unpin* previous pinned statuses that weren't included.
+	for _, status := range wasPinned {
+		unpin := true // assume we have to unpin
+
+	innerLoop:
+		for _, statusURI := range statusURIs {
+			if status.URI == statusURI.String() {
+				unpin = false   // this status is included in most recent pinned uris
+				break innerLoop // no need to keep checking
```
NyaaaWhatsUpDoc, line 714: I think the logic you were probably going for here was to have the loop tag on the outer loop, and then here you'd put the `continue outerLoop`. This way you don't have to go past the `unpin` clause even when we know it is false.

internal/federation/dereferencing/account.go
```diff
@@ -569,3 +578,150 @@ func (d *deref) fetchRemoteAccountEmojis(ctx context.Context, targetAccount *gts
...
+			continue
+		}
+
+		if statusURI.Host != uri.Host {
+			// If this status doesn't share a host with its featured
+			// collection URI, we shouldn't trust it. Just move on.
+			continue
+		}
+
+		statusURIs = append(statusURIs, statusURI)
```
NyaaaWhatsUpDoc, line 670: I would move this line to be the final stage in this loop. Otherwise we'll end up attempting to unpin possibly non-extant (according to our DB) statuses.

NyaaaWhatsUpDoc:
I have *started* reviewing this and it looks really great so far :D. There's a few comments on things but otherwise logic is looking sound and code is very nice to read :). I'll try come back and finish this off later today, I'm currently 23 / 29 files through the review.

internal/federation/dereferencing/account.go
NyaaaWhatsUpDoc: Wait no we wouldn't... Ignore this, I see what you were going for. Though maybe worth adding a comment here to explain why we add this to the URI slice _before_ any later logic, just to make sure this doesn't get unpinned.

internal/federation/dereferencing/account.go
tsmethurst: Ahhh thank you, I always forget you can continue loop labels!

internal/federation/dereferencing/account.go
tsmethurst: Added a little comment :) https://github.com/superseriousbusiness/gotosocial/pull/1560/commits/cc079de0ae671f7ae3015d1b6493a1d7d3c0ad7b

internal/federation/dereferencing/account.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/1560/commits/106e7701884d98e3ca14f4abb60fdc793ffcb9b7

internal/federation/dereferencing/account.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/1560/commits/48e4677e9155801ef90fecbe94ccc5cfbac7b4d4

internal/federation/dereferencing/account.go
tsmethurst: Forgot about that, thank you :D https://github.com/superseriousbusiness/gotosocial/pull/1560/commits/a7e00d1b67f161f4b1dd74aaafbec38c229d59f5

internal/federation/dereferencing/account.go
tsmethurst: :+1: good point, changed it here: https://github.com/superseriousbusiness/gotosocial/pull/1560/commits/fa28f4dc759f7f3be11f8692c749b58a8d0a4f52

internal/typeutils/astointernal.go
```diff
@@ -181,9 +181,14 @@ func (c *converter) ASRepresentationToAccount(ctx context.Context, accountable a
...
 	}
 
-	// FeaturedURI
-	if accountable.GetTootFeatured() != nil && accountable.GetTootFeatured().GetIRI() != nil {
-		acct.FeaturedCollectionURI = accountable.GetTootFeatured().GetIRI().String()
+	// FeaturedURI aka pinned collection:
+	// Only trust featured URI if it has at least two domains,
+	// from the right, in common with the domain of the account
+	if featured := accountable.GetTootFeatured(); featured != nil && featured.IsIRI() {
+		if featuredURI := featured.GetIRI(); // nocollapse
```
NyaaaWhatsUpDoc, line 188: heh i like you used my horrible hack to force gofmt to not collapse the multi-line if statement :3 i mean you can use any comment string here but i always figure `nocollapse` looks most reasonable.

NyaaaWhatsUpDoc: (not an issue just interesting note :smile: )

internal/typeutils/internaltoas_test.go
```diff
@@ -544,6 +546,96 @@ func (suite *InternalToASTestSuite) TestReportToAS() {
...
+	suite.Equal(`{
+  "@context": "https://www.w3.org/ns/activitystreams",
+  "id": "http://localhost:8080/users/the_mighty_zork/collections/featured",
+  "orderedItems": [],
+  "totalItems": 0,
+  "type": "OrderedCollection"
+}`, string(bytes))
+}
+
+func (suite *InternalToASTestSuite) TestPinnedStatusesToAS3() {
```
NyaaaWhatsUpDoc, line 610: Probably worth having more description test names for these. E.g. MultipleItems, NoItems, SingleItem

internal/processing/fedi/collections.go
```diff
@@ -128,97 +153,30 @@ func (p *Processor) FollowingGet(ctx context.Context, requestedUsername string,
...
-	requestedAccount, err := p.db.GetAccountByUsernameDomain(ctx, requestedUsername, "")
-	if err != nil {
-		return nil, gtserror.NewErrorNotFound(fmt.Errorf("database error getting account with username %s: %s", requestedUsername, err))
-	}
-
-	// authenticate the request
-	requestingAccountURI, errWithCode := p.federator.AuthenticateFederatedRequest(ctx, requestedUsername)
+// FeaturedCollectionGet returns an ordered collection of the requested username's Pinned posts.
+// The returned collection have an `items` property which contains an ordered list of status URIs.
+func (p *Processor) FeaturedCollectionGet(ctx context.Context, requestedUsername string) (interface{}, gtserror.WithCode) {
```
NyaaaWhatsUpDoc, line 158: Are we missing visibility filtering in this function? :thinking:

NyaaaWhatsUpDoc:
Only 2 actionable comments, otherwise ready to go :)

internal/typeutils/astointernal.go
tsmethurst: ohh i actually thought it was really a hint to go fmt :') didn't realize it could be any arbitrary string, amazing

internal/processing/fedi/collections.go
tsmethurst: Yes! This was actually on purpose.

We return URIs regardless of who calls the endpoint (unless they're actually blocked ofc), and then the actual calls to dereference the statuses are where the visibility filtering comes in.

Reason for this: I wanted calls to the pinned endpoint to consistently return the same results so that remote instances aren't thinking constantly 'oh, the list has changed' just because one user has visibility on x status and another has visibility on y. This is also why we return just the URIs, to avoid serving statuses to people who oughtn't see them.

internal/typeutils/internaltoas_test.go
tsmethurst: Bugger, thought I'd tracked these down, will fix! Thanks!

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1560/commits/c7be42e30e4071f7f41b45fcb824b5a73a5aed91

internal/processing/fedi/collections.go
NyaaaWhatsUpDoc: ahhh okay. this indeed makes sense :)

PR #1554, opened 2023-02-25
[chore] Use latest containers when building

daenney:
# Description

This updates the Dockerfile to use the latest image versions for all the containers that are used when building the final GTS container.

## Checklist

Please put an x inside each checkbox to indicate that you've read and followed it: `[ ]` -> `[x]`

If this is a documentation change, only the first checkbox must be filled (you can delete the others if you want).

- [x] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [ ] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [ ] I/we have performed a self-review of added code.
- [ ] I/we have written code that is legible and maintainable by others.
- [ ] I/we have commented the added code, particularly in hard-to-understand areas.
- [x] I/we have made any necessary changes to documentation.
- [ ] I/we have added tests that cover new code.
- [x] I/we have run tests and they pass locally with the changes.
- [x] I/we have run `go fmt ./...` and `golangci-lint run`.

PR #1561, opened 2023-02-26
[feature] Provide .well-known/host-meta endpoint

### LINKED ISSUE #1558: [bug] host-meta endpoint needs to be implemented and used in search

It seems host-meta endpoint is used to find the direction for webfinger location. Tested this across different instances except pleroma lots of major systems have implemented it as shown in the screenshot.

image

we dont have this end-point right now but feels like this should not be a lot of work https://docs.joinmastodon.org/dev/routes/#host-meta is the reference i could find. 

0.7.0

### GoToSocial Arch

amd64

### What happened?

I was trying to follow someone. that person had host-meta as a url pointer for webfinger but didnt had webfinger redirection on website.

find the person and follow them

### How to reproduce it?

host which has following conditions met
1. web domain is different from user domain.
2. webfinger of user domain is not redirected to web domain
3.  host-meta on user domain points to correct webfinger spec url.

if we try to follow this person via gotosocial gts will not be able to find the user as it directly attempts to access webfinger end point.

https://mastodon.rfc1925.org/@ondrej/109926688806310155 this is the reference post for the same

daenney:
# Description

This is the first step to address #1558. This adds the host-meta endpoint on our side. It only implements the XML endpoint on /host-meta and not the JSON endpoint on /host-meta.json as nobody else seems to provide that either.

Yes, XML.

## Checklist

### DIFF
--- a/docs/configuration/general.md
+++ b/docs/configuration/general.md
@@ -43,6 +43,9 @@ host: "localhost"
 # to "gts.example.org/.well-known/webfinger" so that GtS can handle them properly.
 #
 # You should also redirect requests at "example.org/.well-known/nodeinfo" in the same way.
+#
+# You should also redirect requests at "example.org/.well-known/host-meta" in the same way.
+#
 # An empty string (ie., not set) means that the same value as 'host' will be used.
 #
 # DO NOT change this after your server has already run once, or you will break things!
--- a/docs/installation_guide/advanced.md
+++ b/docs/installation_guide/advanced.md
@@ -32,6 +32,9 @@ host: "localhost"
 # to "gts.example.org/.well-known/webfinger" so that GtS can handle them properly.
 #
 # You should also redirect requests at "example.org/.well-known/nodeinfo" in the same way.
+#
+# You should also redirect requests at "example.org/.well-known/host-meta" in the same way.
+#
 # An empty string (ie., not set) means that the same value as 'host' will be used.
 #
 # DO NOT change this after your server has already run once, or you will break things!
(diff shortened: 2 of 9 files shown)


daenney:
I've refrained from regenerating the swagger doc because based on #1554 it looks like we need to redo that in general?

daenney:
I think what's happening now is correct:

```
$ curl -X GET -H "Accept: application/xml" http://127.0.0.1:8080/.well-known/host-meta
<?xml version="1.0" encoding="UTF-8"?>
<XRD xmlns="http://docs.oasis-open.org/ns/xri/xrd-1.0"><Link rel="lrdd" type="application/xrd+xml" template="http://localhost:8080/.well-known/webfinger?resource={uri}"></Link></XRD>
```

daenney:
Discussing this a bit on Matrix, we've come to the conclusion we probably don't want to add this.

Our documentation already tells folks they have to proxy/redirect the webfinger endpoint if their host domain and account domain differ. Not redirecting webfinger is something we consider a misconfiguration that should be resolved by the admin and not something we want to add host-meta for.

Without host-meta people should always try to construct the webfinger URL based on account-domain/.well-known/webfinger which will respond as expected.

I'll leave this open for now in case folks want to chime in on why we may want to keep it.

tsmethurst:
I changed it to a draft, if we decide to keep it we can mark it as ready again :)

daenney:
I'm going to close this one. I really don't see the point of taking on this code when we already tell folks to properly redirect webfinger and that's what's generally expected nowadays. #1588 is around for host-meta based discovery of the webfinger endpoint.

PR #1576, opened 2023-02-28
[chore/performance] simplify storage driver to use storage.Storage directly

NyaaaWhatsUpDoc:
This simplifies our storage driver to directly use the interface type `storage.Storage` instead of the wrapping type `kv.KVStore{}`. We are using this is an adaptable media storage interface, but `kv.KVStore{}` is more intended for use as a key-value store, and as such uses a per-key mutex locking system which we don't need. This PR instead just updates us to use the underlying storage interface type directly, though using `kv.KVStore{}` like wrapper functions to prevent a huge number of required code changes.

### DIFF
--- a/internal/media/prune.go
+++ b/internal/media/prune.go
@@ -154,55 +154,41 @@ func (m *manager) PruneUnusedRemote(ctx context.Context, dry bool) (int, error) 
 }
 
 func (m *manager) PruneOrphaned(ctx context.Context, dry bool) (int, error) {
-	// keys in storage will look like the following:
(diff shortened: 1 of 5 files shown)


internal/media/prune.go
```diff
@@ -211,9 +197,8 @@ func (m *manager) PruneOrphaned(ctx context.Context, dry bool) (int, error) {
 		return totalPruned, nil
 	}
 
-	// This is not a drill!
-	// We have to delete stuff!
-	return totalPruned, m.removeFiles(ctx, orphanedKeys...)
+	// This is not a drill! We have to delete stuff!
+	return m.removeFiles(ctx, orphanedKeys...)
```
NyaaaWhatsUpDoc, line 201: this was just a quick little change i happened to pick up in this PR to return a more accurate count :p

Issue #1526, opened 2023-02-18
[bug] Status deletion fails to federate with an error if the parent status was deleted


Yuubari:
### Describe the bug with a clear and concise description of what the bug is.

When a status with a deleted parent is deleted, an error is logged. The status is deleted locally, however it does not get federated and remains on the remote follower's timeline.

The (already deleted) parent status can be a local one, or one from a remote instance.

### What's your GoToSocial Version?

0.7.0

### GoToSocial Arch

amd64 binary

### What happened?

The status was deleted locally, but the deletion was not federated. The following error is logged:
```
timestamp="18/02/2023 14:07:39.934" func=concurrency.(*WorkerPool).Queue.func1 level=ERROR type="worker.Worker[messages.FromClientAPI]" error="federateStatusDelete: error converting status to as format: StatusToAS: error getting replied to status <status id here>: no entries" msg="message processing error"
```

### What you expected to happen?

The status should have been deleted from the remote follower's timeline, like it happens with statuses that have existing, non-deleted parents, or stand-alone statuses without parents.

### How to reproduce it?

This was tested between GoToSocial 0.7.0 and Mastodon (looks like the remote runs 3.1.3).

1. Post a status (referred to as X); it can be either on the GoToSocial instance or on a remote one. Local and remote followers will see it in their timelines.
2. Post a reply on the GoToSocial instance to this status (referred to as Y). Local and remote followers will see it in their timelines.
3. Delete status X. It gets deleted; local and remote followers will no longer see it in their timelines.
4. Delete status Y. It gets deleted for you (and presumably for the local followers), but remains on remote followers' timelines.

tsmethurst:
Thank you! Will make a fix for this soon :)

PR #1584, opened 2023-03-03
[bugfix] Federate status Delete using just the URI, update Delete addressing

tsmethurst:
# Description

>

This pull request updates some of our status delete logic to federate the Delete using only the URI of the status.

It also updates the addressing of the Delete to always include the ActivityPub public URI in the `to` field unless the deleted post was a direct message, and also to always CC the replied-to user (if present) with the delete.

This should help to ensure that Deletes are federated a bit more widely. The trade off is that remote instances might become aware of the URI of now-deleted posts that were not originally addressed to them (but which they might have seen through some other method). Since it's just the URI and not the status itself, which can't possibly now be dereferenced, this seems a worthy tradeoff.

In future, we should also look at addressing deletes of non-direct messages to all known sharedInboxes, a little bit like mastodon does, but for now this is better than nothing :)

The PR also adds some documentation about how GtS handles deletes, under the 'federating with gotosocial' section.

closes https://github.com/superseriousbusiness/gotosocial/issues/1526

## Checklist

### DIFF
--- a/internal/typeutils/internaltoas.go
+++ b/internal/typeutils/internaltoas.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"crypto/x509"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"net/url"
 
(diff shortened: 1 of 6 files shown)


internal/typeutils/internaltoas.go
```diff
@@ -524,7 +514,7 @@ func (c *converter) StatusToAS(ctx context.Context, s *gtsmodel.Status) (vocab.A
 	switch s.Visibility {
 	case gtsmodel.VisibilityDirect:
 		// if DIRECT, then only mentioned users should be added to TO, and nothing to CC
-		for _, m := range s.Mentions {
+		for _, m := range mentions {
```
tsmethurst, line 517: Reason for these changes: earlier we were iterating through s.Mentions, but if `s.Mentions` wasn't populated before, then mentions are now in `mentions`, not `s.Mentions`!

PR #1586, opened 2023-03-04
[feature] Allow loading TLS certs from disk

daenney:
# Description

>

Currently, GtS only supports using the built-in LE client directly for TLS. However, admins may still want to use GtS directly (so without a reverse proxy) but with certificates provided through some other mechanism. They may have some centralised way of provisioning these things themselves, or simply prefer to use LE but with a different challenge like DNS-01 which is not supported by autocert.

This adds support for loading a public/private keypair from disk instead of using LE and reconfigures the server to use a TLS listener if we succeed in doing so.

Additionally, being able to load TLS keypair from disk opens up the path to using a custom CA for testing purposes avoinding the need for a constellation of containers and something like Pebble or Step CA to provide LE APIs.

## Checklist


CONTRIBUTING.md
```diff
@@ -418,6 +419,20 @@ In [./test/envparsing.sh](./test/envparsing.sh) there's a test for making sure t
...
 
+#### Federation
+
+By using the support for loading TLS files from disk it is possible to have two local instances with TLS to allow for (manually) testing federation.
+
+You'll need to set the following configuration options:
+* `GTS_TLS_CERTIFICATE_CHAIN`: poiting to a PEM-encoded certificate chain including the public certificate
+* `GTS_TLS_CERTIFICATE_KEY`: pointing to a PEM-encoded private key
+
+Additionally, for the Go HTTP client to recognise certificates issued by a custom CA as valid, you'll need to set one of:
```
daenney, line 430: Except on macOS or Windows: https://pkg.go.dev/crypto/x509#SystemCertPool

daenney:
Go has built-in support for generating a CA or certificates for you. Expanding a bit on that it should be fairly trivial to have a test helper script to generate the CA + issue certificates for a pair of named instances.

```
go run /usr/lib/go/src/crypto/tls/generate_cert.go -help
Usage of /run/user/1000/go-build2063357202/b001/exe/generate_cert:
  -ca
    	whether this cert should be its own Certificate Authority
  -duration duration
    	Duration that certificate is valid for (default 8760h0m0s)
  -ecdsa-curve string
    	ECDSA curve to use to generate a key. Valid values are P224, P256 (recommended), P384, P521
  -ed25519
    	Generate an Ed25519 key
  -host string
    	Comma-separated hostnames and IPs to generate a certificate for
  -rsa-bits int
    	Size of RSA key to generate. Ignored if --ecdsa-curve is set (default 2048)
  -start-date string
    	Creation date formatted as Jan 1 15:04:05 2011
```

NyaaaWhatsUpDoc:
> Go has built-in support for generating a CA or certificates for you. Expanding a bit on that it should be fairly trivial to have a test helper script to generate the CA + issue certificates for a pair of named instances. Or leverage something like [mkCert](https://github.com/FiloSottile/mkcert).
> 
> ```
> go run /usr/lib/go/src/crypto/tls/generate_cert.go -help
> Usage of /run/user/1000/go-build2063357202/b001/exe/generate_cert:
>   -ca
>     	whether this cert should be its own Certificate Authority
>   -duration duration
>     	Duration that certificate is valid for (default 8760h0m0s)
>   -ecdsa-curve string
>     	ECDSA curve to use to generate a key. Valid values are P224, P256 (recommended), P384, P521
>   -ed25519
>     	Generate an Ed25519 key
>   -host string
>     	Comma-separated hostnames and IPs to generate a certificate for
>   -rsa-bits int
>     	Size of RSA key to generate. Ignored if --ecdsa-curve is set (default 2048)
>   -start-date string
>     	Creation date formatted as Jan 1 15:04:05 2011
> ```

holy shit i can't believe i never knew this

CONTRIBUTING.md
NyaaaWhatsUpDoc: Fortunately we don't support either of these :p

NyaaaWhatsUpDoc:
nice work :)

tsmethurst:
Thanks for this!

PR #1588, opened 2023-03-04
[feature] Discover webfinger through host-meta

### LINKED ISSUE #1558: [bug] host-meta endpoint needs to be implemented and used in search

daenney:
# Description

>

I'm opening this as a draft because:
* The implementation makes me sad, please suggest improvements
* ~~I want to figure out how to properly test the fallback and ideally verify what gets set in the cache~~ 
* Arguably, I'm inclined to not want this code at all because the host-meta redirect goes against current best practices. But it's hard for end users to figure out why search isn't working and if/when they do manage to get another instance admin to potentially update their configuration. It's the fediverse so in this case I think it's worthwhile to try to be MAX_COMPATIBLE

---

This implements a fallback for discovering the webfinger endpoint in case the /.well-known/webfinger endpoint wasn't properly redirected. Some instances do this because the recommendation used to be to use host-meta for the webfinger redirect in the before times.

Closes #1558.

## Checklist


daenney:
I've added a test by redirecting to a different `webfinger` endpoint. Not sure if/how we want to test for the cached values?

internal/config/defaults.go
```diff
@@ -163,6 +163,10 @@ var Defaults = Configuration{
 			UserMaxSize:   100,
 			UserTTL:       time.Minute * 5,
 			UserSweepFreq: time.Second * 30,
+
+			WebfingerMaxSize:   100,
+			WebfingerTTL:       time.Hour * 24,
```
daenney, line 168: The webfinger endpoint would only ever change on an instance upgrade or reconfiguration of sorts, so it seems fine to cache it for a long time.

internal/config/defaults.go
```diff
@@ -163,6 +163,10 @@ var Defaults = Configuration{
...
+			WebfingerSweepFreq: time.Minute * 15,
```
daenney, line 169: Didn't know how much this needed to be so took a fraction of the cache time.

internal/config/defaults.go
```diff
@@ -163,6 +163,10 @@ var Defaults = Configuration{
 			UserMaxSize:   100,
 			UserTTL:       time.Minute * 5,
 			UserSweepFreq: time.Second * 30,
+
+			WebfingerMaxSize:   100,
```
daenney, line 167: Give we cache for 24hrs maybe this needs to be bigger?

internal/transport/finger.go
```diff
@@ -20,40 +20,167 @@ package transport
...
+	url, ok := t.controller.state.Caches.GTS.Webfinger().Get(targetDomain)
+	if !ok {
+		url = "https://" + targetDomain + "/.well-known/webfinger"
+	}
+	return url, ok
+}
 
-	// Generate new GET request from URL string
-	req, err := http.NewRequestWithContext(ctx, "GET", urlStr, nil)
+func prepWebfingerReq(url, domain, username string) (*http.Request, error) {
```
daenney, line 44: I pulled this one out of `Finger` since we end up needing it twice. It felt bad to have the same code twice in the body of that function.

internal/transport/finger.go
```diff
@@ -20,40 +20,167 @@ package transport
...
+	// Check if the request succeeded so we can bail out early
+	if rsp.StatusCode == http.StatusOK {
+		// Set the webfinger URL in the cache (again). If this was set this
+		// renews the TTL of the current entry
+		t.controller.state.Caches.GTS.Webfinger().Set(targetDomain, url)
+		return io.ReadAll(rsp.Body)
+	}
+
+	// Check for an unexpected status code
+	if (rsp.StatusCode != http.StatusOK) && cached {
```
daenney, line 87: I'm thinking this can actually just be `if cached` since we know we don't have a successful status code at this point anyway?

internal/config/defaults.go
tsmethurst: Agreed, 24 hours seems fine :)

internal/config/defaults.go
tsmethurst: Also seems good to me.

internal/config/defaults.go
tsmethurst: We could go 200 since the cached items are quite small. But there are also likely to be not that many endpoints configured with host-meta instead of a webfinger redirect.

daenney: I end up setting this entry for anything we finger though. I can improve that so we don't do that come to think of it, which would be better and make the 100 probably just fine.

daenney: There. Now it only ever stores URLs retrieved from host-meta, not a webfinger URL per targetDomain

internal/transport/finger.go
```diff
@@ -20,40 +20,174 @@ package transport
...
+	if !ok {
+		url = "https://" + targetDomain + "/.well-known/webfinger"
+	}
+	return url, ok
+}
 
-	// Generate new GET request from URL string
-	req, err := http.NewRequestWithContext(ctx, "GET", urlStr, nil)
+func prepWebfingerReq(url, domain, username string) (*http.Request, error) {
+	req, err := http.NewRequest("GET", url, nil)
```
tsmethurst, line 45: You can use http.GET instead of "GET" here (and elsewhere)

internal/transport/finger.go
```diff
@@ -20,40 +20,174 @@ package transport
...
+		}
+		return io.ReadAll(rsp.Body)
+	}
+
+	// From here on out, we're handling different failure scenarios and
+	// deciding whether we should do a host-meta based fallback or not
+
+	if (rsp.StatusCode >= 500 && rsp.StatusCode < 600) || cached {
+		// In case we got a 5xx, bail out irrespective of if the value
+		// was cached or not. The target is broken, so there's no point
```
tsmethurst, line 93: Not necessarily broken: we (and other servers) use 503 and 'retry-after' to signal that the server is currently overloaded and throttling.

internal/transport/finger.go
```diff
@@ -20,40 +20,174 @@ package transport
...
+		// If it's not a 5xx but the URL was cached, bail out too
+		return nil, fmt.Errorf("GET request to %s failed: %s", req.URL.String(), rsp.Status)
+	}
+
+	// So far we've failed to get a successful response from the expected
+	// webfinger endpoint. Lets try and discover the webfinger endpoint
+	// through /.well-known/host-meta
+	host, err := t.webfingerFromHostMeta(ctx, targetDomain)
+	if err != nil {
+		return nil, fmt.Errorf("failed to discover webfinger URL fallback for: %s through host-meta: %s", targetDomain, err)
```
tsmethurst, line 105: you might wanna wrap these errors -- `%w` instead of `%s` -- in case handlers wanna parse them out further down the line

tsmethurst (commented):
Looking good so far :) I'm wondering, do we need/want to cache cases where the webfinger URL didn't require any host-meta looking up, and just the initial constructed webfinger uri worked OK? I was thinking we could only cache cases where the host-meta lookup revealed a different uri from what we expected, if you see what I mean.

daenney:
> Looking good so far :) I'm wondering, do we need/want to cache cases where the webfinger URL didn't require any host-meta looking up, and just the initial constructed webfinger uri worked OK? I was thinking we could only cache cases where the host-meta lookup revealed a different uri from what we expected, if you see what I mean.

I've updated that now. Now only `webfingerFromHostMeta` can add new entry in the cache, and only if we got a cached entry + success do we renew the TTL. That should avoid us caching the default URL if that one worked.

internal/transport/finger.go
daenney: Done!

internal/transport/finger.go
daenney: Good point. Updated the comment to reflect that.

internal/transport/finger.go
daenney: Indeed

internal/transport/finger.go
```diff
@@ -20,40 +20,173 @@ package transport
...
-	// Prepare URL string
-	urlStr := "https://" +
-		targetDomain +
-		"/.well-known/webfinger?resource=acct:" +
-		targetUsername + "@" + targetDomain
+// webfingerURLFor returns the URL to try a webfinger request against, as
+// well as if the URL was retrieved from cache. When the URL is retrieved
+// from cache we don't have to try and do host-meta discovery
+func (t *transport) webfingerURLFor(targetDomain string) (string, bool) {
+	url, ok := t.controller.state.Caches.GTS.Webfinger().Get(targetDomain)
```
daenney, line 37: I've realised something here. The `Get` extends the item expiry time. I don't think we want that here. Just because we're fetching the value doesn't mean we want the TTL extended, only when we've done the request and the URL still succeeded we want to do that.

So it seems here we'd need to access the underlying `Cache`, `Lock()`/`Unlock()` ourselves and use `Cache.Get()` to avoid that update.

internal/transport/finger_test.go
```diff
@@ -0,0 +1,100 @@
...
+	_, err := suite.transport.Finger(context.TODO(), "someone", "misconfigured-instance.com")
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
+
+	suite.Equal(1, wc.Len(), "expect webfinger cache to hold one entry")
+	suite.True(wc.Has("misconfigured-instance.com"), "expect webfinger cache to have entry for misconfigured-instance.com")
+}
+
+func (suite *FingerTestSuite) TestFingerWithHostMetaCacheStrategy() {
```
daenney, line 39: This one's a bit long but it tests that things happen as currently intended:
* If we get a webfinger redirect, cache it
* Future successful requests extend the TTL
* Future failed requests don't touch the TTL so the entry will eventually expire

internal/transport/transport_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/media"
+	"github.com/superseriousbusiness/gotosocial/internal/processing"
+	"github.com/superseriousbusiness/gotosocial/internal/state"
+	"github.com/superseriousbusiness/gotosocial/internal/storage"
+	"github.com/superseriousbusiness/gotosocial/internal/transport"
+	"github.com/superseriousbusiness/gotosocial/testrig"
+)
+
+type TransportTestSuite struct {
```
daenney, line 19: I cribbed this from other test suites. I don't know if I need all this. But it works :sweat_smile:

internal/transport/transport_test.go
```diff
@@ -0,0 +1,83 @@
...
+	suite.mediaManager = testrig.NewTestMediaManager(&suite.state)
+	suite.federator = testrig.NewTestFederator(&suite.state, testrig.NewTestTransportController(&suite.state, testrig.NewMockHTTPClient(nil, "../../testrig/media")), suite.mediaManager)
+	suite.sentEmails = make(map[string]string)
+	suite.emailSender = testrig.NewEmailSender("../../web/template/", suite.sentEmails)
+	suite.processor = testrig.NewTestProcessor(&suite.state, suite.federator, suite.emailSender, suite.mediaManager)
+
+	testrig.StandardDBSetup(suite.db, nil)
+	testrig.StandardStorageSetup(suite.storage, "../../testrig/media")
+
+	ts, err := suite.federator.TransportController().NewTransportForUsername(context.TODO(), "")
```
daenney, line 70: Idk if this should be here/how to better do this. But I need a `transport.Transport` to call `Finger`

internal/config/defaults.go
NyaaaWhatsUpDoc: I'd maybe bump the max size of this given the long cache time and not expecting changes. How about 250?

internal/transport/controller.go
```diff
@@ -54,10 +55,12 @@ type controller struct {
 	trspCache cache.Cache[string, *transport]
 	badHosts  cache.Cache[string, struct{}]
 	userAgent string
+
+	state *state.State
 }
 
 // NewController returns an implementation of the Controller interface for creating new transports
-func NewController(db db.DB, federatingDB federatingdb.DB, clock pub.Clock, client pub.HttpClient) Controller {
+func NewController(db db.DB, federatingDB federatingdb.DB, clock pub.Clock, client pub.HttpClient, state *state.State) Controller {
```
NyaaaWhatsUpDoc, line 63: if we're accepting a state we don't also need to accept a separate db argument. i'd remove the db and put the state in first argument position imo

internal/transport/finger.go
```diff
@@ -20,40 +20,183 @@ package transport
...
+	if ok {
+		url = item.Value
+	}
 
-	// Generate new GET request from URL string
-	req, err := http.NewRequestWithContext(ctx, "GET", urlStr, nil)
+	return url, ok
+}
+
+func prepWebfingerReq(url, domain, username string) (*http.Request, error) {
```
NyaaaWhatsUpDoc, line 54: I would probably update this function signature to also take a context so we can instead do `http.NewRequestWithContext()`. Otherwise we keep end up calling `r.WithContext()` which afaik copies the request object.

internal/transport/finger.go
```diff
@@ -20,40 +20,183 @@ package transport
...
+	req, err := http.NewRequest(http.MethodGet, url, nil)
 	if err != nil {
 		return nil, err
 	}
+
+	q := req.URL.Query()
+
+	q.Add("resource", "acct:"+username+"@"+domain)
```
NyaaaWhatsUpDoc, line 62: Another little optimization we can do here to avoid allocating a query map for only a single key-value (though might need to change that incoming `url` argument variable name):
```
	value := url.QueryEscape("acct:" + username + "@" + domain)
	req.URL.RawQuery = "resource=" + value
```

NyaaaWhatsUpDoc:
Looks good! A few comments but then should be ready to go :)

internal/transport/controller.go
daenney: Done!

internal/transport/finger.go
daenney: Ya, I wasn't sure whether I should pull the context through here since we don't really need it and `r.WithContext` does a shallow copy. But might as well.

internal/transport/finger.go
daenney: Neat. Done!

internal/config/defaults.go
daenney: Done!

PR #1599, opened 2023-03-07
[chore] Refactor account deleting/block logic, tidy up some other processing things

tsmethurst:
# Description

>

This pull request does a whole bunch of refactoring and tidying up in the following areas:

- Tidy up a lot of logic in the processor around account deletion, and account block creation.
  - Remove duplicated logic.
  - Tidy up into smaller functions.
  - Remove more stuff when a block or delete happens.
- Add and use database interfaces for status faves and status bookmarks.
- Remove a lot of code duplication in the status processor.
- Delete follow request notifs on accepting/rejecting a follow request.
- Ensure that notification deletion also invalidates the cache entry for that notification (not the whole cache).
- Allow updateaccount to use specific columns.
- Allow processor workers to batch process related items.

## Checklist


internal/db/bundb/notification.go
```diff
@@ -105,15 +106,52 @@ func (n *notificationDB) GetNotifications(ctx context.Context, accountID string,
...
 		Exec(ctx); err != nil {
 		return n.conn.ProcessError(err)
 	}
 
-	n.state.Caches.GTS.Notification().Clear()
+	n.state.Caches.GTS.Notification().Invalidate("ID", id)
+	return nil
+}
+
+func (n *notificationDB) DeleteNotifications(ctx context.Context, targetAccountID string, originAccountID string, statusID string) db.Error {
```
NyaaaWhatsUpDoc, line 122: Might be a bit neater to have separate functions for this and one function that wraps it all. e.g. `n.deleteNotifications(ctx context.Context, ids []string) error` that called `n.DeleteNotification()` on each, and then just `DeleteNotificationsByTarget(()` `DeleteNotificationsByOrigin()` `DeleteNotificationsByStatus()`. Having 3 "nulable" function arguments feels a little messy to me imo.

What do you think?

tsmethurst: Check! I'll fix these up. I think there might be more so I'll have a look through and see what on earth I did yesterday.

internal/federation/federatingdb/followers.go
```diff
@@ -44,10 +44,17 @@ func (f *federatingDB) Followers(ctx context.Context, actorIRI *url.URL) (follow
 
 	iris := make([]*url.URL, 0, len(follows))
 	for _, follow := range follows {
+		if follow.Account == nil {
```
NyaaaWhatsUpDoc, line 47: Any idea what these kind of situations causing this could be? I'm not sure how I feel about just skipping over these without knowing the root cause.

NyaaaWhatsUpDoc: In the case of nil account in statuses I could understand because it could still be dereferencing, but this shouldn't be possible with a follow should it? Especially when it is the source account :thinking:

tsmethurst: > Any idea what these kind of situations causing this could be?

Yeah, I think there are a couple things that might necessitate doing a check like this:

1. Account gets deleted after the follows have already been selected, but before they've been iterated through.
2. Account gets partially deleted so the follow still remains but the account that created the follow is no longer in the db.

There are likely better ways of handling this, but in the interests of not hitting a nil pointer exception, this little nil guard check seemed like the simplest way to go about it.

NyaaaWhatsUpDoc: Ah yep that all makes sense! Is it worth adding a `log.Warn()` entry here just so we catch these?

internal/db/bundb/statusbookmark.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return s.conn.ProcessError(err)
+}
+
+func (s *statusBookmarkDB) DeleteStatusBookmarks(ctx context.Context, targetAccountID string, originAccountID string) db.Error {
+	if targetAccountID == "" && originAccountID == "" {
+		return errors.New("DeleteBookmarks: one of targetAccountID or originAccountID must be set")
+	}
+
+	// TODO: Capture bookmark IDs in a RETURNING
```
NyaaaWhatsUpDoc, line 159: I'll catch these in my refactor PR once we've got this one merged. As they will be necessary before release otherwise bookmarks will continue showing until cache eviction.

internal/db/bundb/statusbookmark.go
```diff
@@ -0,0 +1,197 @@
...
+
+	if _, err := q.Exec(ctx); err != nil {
+		return s.conn.ProcessError(err)
+	}
+
+	return nil
+}
+
+func (s *statusBookmarkDB) DeleteStatusBookmarksForStatus(ctx context.Context, statusID string) db.Error {
+	// TODO: Capture bookmark IDs in a RETURNING
```
NyaaaWhatsUpDoc, line 183: Same again.

internal/db/bundb/statusbookmark.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	bookmark.TargetAccount, err = s.state.DB.GetAccountByID(ctx, bookmark.TargetAccountID)
+	if err != nil {
+		log.Errorf(ctx, "error getting status bookmark target account %q: %v", bookmark.TargetAccountID, err)
+	}
+
+	bookmark.Status, err = s.state.DB.GetStatusByID(ctx, bookmark.StatusID)
+	if err != nil {
+		log.Errorf(ctx, "error getting status bookmark status %q: %v", bookmark.StatusID, err)
```
NyaaaWhatsUpDoc, line 60: I think we should probably make these core database model structures we're populating returnable errors instead of logging and continuing. Would definitely help narrow down an issue instead of needing to track how a status got further with nil models. And if we do happen to catch it while an account / status deletion is in process, I think a temporary 500 (especially given how unlikely it is) is a pretty reasonable error code. What do you think?

internal/db/bundb/statusfave.go
```diff
@@ -0,0 +1,171 @@
...
+		Model(fave).
+		Where("? = ?", bun.Ident("status_fave.ID"), id).
+		Scan(ctx)
+	if err != nil {
+		return nil, s.conn.ProcessError(err)
+	}
+
+	fave.Account, err = s.state.DB.GetAccountByID(ctx, fave.AccountID)
+	if err != nil {
+		log.Errorf(ctx, "error getting status fave account %q: %v", fave.AccountID, err)
```
NyaaaWhatsUpDoc, line 50: Same point as with the status bookmark model population errors.

internal/db/bundb/statusfave.go
```diff
@@ -0,0 +1,171 @@
...
+
+	fave.Status, err = s.state.DB.GetStatusByID(ctx, fave.StatusID)
+	if err != nil {
+		log.Errorf(ctx, "error getting status fave status %q: %v", fave.StatusID, err)
+	}
+
+	return fave, nil
+}
+
+func (s *statusFaveDB) GetStatusFaveByAccountID(ctx context.Context, accountID string, statusID string) (*gtsmodel.StatusFave, db.Error) {
```
NyaaaWhatsUpDoc, line 66: I think for this one it is worth having a separate cache lookup for it, so `"AcccountID.StatusID"`. And then do a similar unexported get function that takes an anonymous function a bit like `getStatus()` and getAccount()`. Otherwise in this case we'll _always_ be doing at least one database lookup for this model, and worst case two.

internal/db/bundb/statusfave.go
```diff
@@ -0,0 +1,171 @@
...
+func (s *statusFaveDB) PutStatusFave(ctx context.Context, statusFave *gtsmodel.StatusFave) db.Error {
+	_, err := s.conn.
+		NewInsert().
+		Model(statusFave).
+		Exec(ctx)
+
+	return s.conn.ProcessError(err)
+}
+
+func (s *statusFaveDB) DeleteStatusFave(ctx context.Context, id string) db.Error {
```
NyaaaWhatsUpDoc, line 118: You need to invalidate the favourite from the cache here.

internal/federation/federatingdb/followers.go
NyaaaWhatsUpDoc: (and same goes for the other cases of this you've introduced below)

internal/db/bundb/statusfave.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return s.conn.ProcessError(err)
+}
+
+func (s *statusFaveDB) DeleteStatusFaves(ctx context.Context, targetAccountID string, originAccountID string) db.Error {
+	if targetAccountID == "" && originAccountID == "" {
+		return errors.New("DeleteStatusFaves: one of targetAccountID or originAccountID must be set")
+	}
+
+	// TODO: Capture fave IDs in a RETURNING
```
NyaaaWhatsUpDoc, line 133: Again, I can catch these here in my PR

internal/db/bundb/statusfave.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if _, err := q.Exec(ctx); err != nil {
+		return s.conn.ProcessError(err)
+	}
+
+	return nil
+}
+
+func (s *statusFaveDB) DeleteStatusFavesForStatus(ctx context.Context, statusID string) db.Error {
+	// TODO: Capture fave IDs in a RETURNING
```
NyaaaWhatsUpDoc, line 157: And again.

internal/processing/account/block.go
```diff
@@ -34,158 +34,178 @@ import (
...
 	}
 
-	// if requestingAccount already blocks target account, we don't need to do anything
-	if blocked, err := p.state.DB.IsBlocked(ctx, requestingAccount.ID, targetAccountID, false); err != nil {
-		return nil, gtserror.NewErrorInternalError(fmt.Errorf("BlockCreate: error checking existence of block: %s", err))
-	} else if blocked {
+	if existingBlock != nil {
+		// Block already exists, nothing to do.
 		return p.RelationshipGet(ctx, requestingAccount, targetAccountID)
 	}
```
NyaaaWhatsUpDoc, line 46: When my PR gets merged in, do you think we should do a `visibility.AccountVisible(ctx, requestingAccount, targetAccount)` here?? Just so we don't let them create blocks against people they can't see? e.g. disabled users, suspended accounts, etc. We would also need to do this in the `BlockRemove()` function just to normalize the type of error return when the target is not visible. 

What do you think?

tsmethurst: I'm not sure, but maybe yes!

internal/processing/account/block.go
```diff
@@ -34,158 +34,178 @@ import (
...
-			TargetAccount: targetAccount,
-		})
-	}
-
-	// handle the rest of the block process asynchronously
-	p.state.Workers.EnqueueClientAPI(ctx, messages.FromClientAPI{
+		err = fmt.Errorf("BlockCreate: error creating block in db: %w", err)
+		return nil, gtserror.NewErrorInternalError(err)
+	}
+
```
NyaaaWhatsUpDoc, line 63: is it necessary to delete all of the mutual account notifications, bookmarks and favourites? in the cases where things get resolved and people undo blocks it can just sort of leave a hole in the the history so to speak. our status visibility and account visibility checks won't actually allow of them to be shown afaik.

NyaaaWhatsUpDoc: All of this is opinion though, do let me know your thoughts on this status of rambling :grin:

internal/processing/account/block.go
NyaaaWhatsUpDoc: :chef_kiss:

internal/federation/federatingdb/followers.go
tsmethurst: Yeah, a warn log is a good idea!

internal/db/bundb/statusbookmark.go
tsmethurst: Check! should I leave these TODOs in then, or take them out? Do you have a preference?

internal/db/bundb/statusbookmark.go
tsmethurst: That sounds reasonable!

internal/processing/account/block.go
tsmethurst: Mmmm, to me the behavior of deleting everything between the two accounts makes sense, but I see what you mean; if you just softblock someone, perhaps it's silly to delete everything. You might even then forget why you blocked them, and have nothing to help remind you. I'm OK with changing this behavior if you prefer!

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
+	// To prevent the account being created again,
+	// stubbify it and update it in the db.
+	// The account will not be deleted, but it
+	// will become completely unusable.
+	columns := stubbifyAccount(account, origin)
+	if err := p.state.DB.UpdateAccount(ctx, account, columns...); err != nil {
+		return gtserror.NewErrorInternalError(err)
+	}
+
+	l.Info("account deleted")
```
NyaaaWhatsUpDoc, line 83: just double checking was this intended to be info or trace? (not massively fussed either way)

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
-					if err := p.state.DB.DeleteWhere(ctx, []db.Where{{Key: "client_id", Value: t.ClientID}}, &gtsmodel.Application{}); err != nil {
-						l.Errorf("error deleting application: %s", err)
-					}
-					// delete the token itself
-					if err := p.state.DB.DeleteByID(ctx, t.ID, t); err != nil {
-						l.Errorf("error deleting oauth token: %s", err)
-					}
-				}
-			}
+	if err := p.deleteAccountBlocks(ctx, account); err != nil {
```
NyaaaWhatsUpDoc, line 58: Just opening a discussion here: do we definitely want to return on deletion of these account-related models? for similar reasons to the discussion we had above regarding blocks.

internal/processing/account/block.go
NyaaaWhatsUpDoc: I think as strictly a user, that's probably how I'd prefer things be. (don't feel like I'm pushing you into this though, only do if you agree!)

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
+		}
+
+		// now check that the password they supplied is correct
+		// make sure a password is actually set and bail if not
+		if user.EncryptedPassword == "" {
+			return gtserror.NewErrorForbidden(errors.New("user password was not set"))
 		}
+
+		// compare the provided password with the encrypted one from the db, bail if they don't match
+		if err := bcrypt.CompareHashAndPassword([]byte(user.EncryptedPassword), []byte(form.Password)); err != nil {
```
NyaaaWhatsUpDoc, line 110: Is authentication something we want to be doing in this layer? I just thought this would have already been handled in one of the higher-up layers nearer the HTTP handlers themselves :thinking:

internal/processing/account/delete.go
NyaaaWhatsUpDoc: Actually ignore me on this. In the case there is in error, this allows a retry to work successfully as the account stubbing is the last possible stage :)

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
 	}
 
-	var maxID string
+	// For each follow owned by this account, unfollow
+	// and process side effects (noop if remote account).
+	for _, follow := range following {
+		if uri, err := p.state.DB.Unfollow(ctx, account.ID, follow.TargetAccountID); err != nil {
+			return fmt.Errorf("deleteAccountFollows: db error unfollowing account: %w", err)
+		} else if uri == "" {
+			// There was no follow after all.
```
NyaaaWhatsUpDoc, line 224: definitely worth warn logging here imo with a dump of the follow structure

internal/federation/federatingdb/followers.go
NyaaaWhatsUpDoc: oh and to add to this, worth dumping the follow itself also for further debug (also the ID if it's a stray one to manually remove)

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
-			}
-			break
+			return fmt.Errorf("deleteAccountFollows: db error unfollowRequesting account: %w", err)
 		}
 
-		if len(statuses) == 0 {
-			break // reached end
+		if uri == "" {
+			// There was no follow request after all.
+			// Some race condition? Skip.
```
NyaaaWhatsUpDoc, line 251: same as above :innocent:

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
+	if !deletedAccount.IsLocal() {
+		// Don't try to process side effects
+		// for accounts that aren't local.
+		return func(account *gtsmodel.Account, follow *gtsmodel.Follow) *messages.FromClientAPI {
+			return nil // noop
+		}
+	}
+
+	return func(account *gtsmodel.Account, follow *gtsmodel.Follow) *messages.FromClientAPI {
+		if follow.TargetAccount == nil || follow.TargetAccount.IsLocal() {
```
NyaaaWhatsUpDoc, line 288: Would a better check for the target account be `if follow.TargetAccountID == ""  ...` ? :thinking: Just wondering why else the targetaccount would be nil.

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
+		}
+	}
+
+	// Process accreted messages asynchronously.
+	p.state.Workers.EnqueueClientAPI(ctx, msgs...)
+
+	return nil
+}
+
+func (p *Processor) unfollowSideEffectsFunc(deletedAccount *gtsmodel.Account) func(account *gtsmodel.Account, follow *gtsmodel.Follow) *messages.FromClientAPI {
```
NyaaaWhatsUpDoc, line 278: :woman_cook: :kissing_closed_eyes: :ok_hand:

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
+	// are all attached to statuses.
+
+	var (
+		statuses []*gtsmodel.Status
+		err      error
+		maxID    string
+		msgs     = []messages.FromClientAPI{}
+	)
+
+	for statuses, err = p.state.DB.GetAccountStatuses(ctx, account.ID, deleteSelectLimit, false, false, maxID, "", false, false); err == nil && len(statuses) != 0; statuses, err = p.state.DB.GetAccountStatuses(ctx, account.ID, deleteSelectLimit, false, false, maxID, "", false, false) {
```
NyaaaWhatsUpDoc, line 337: I think it might have been me that introduced this coding style into our codebase to begin with but might be worth instead just having a `for {}` loop that manually checks these things :p. Saves two separate instances of `GetAccountStatuses()` and a very long single-line

internal/processing/account/delete.go
```diff
@@ -33,153 +34,347 @@ import (
...
 			for _, boost := range boosts {
 				if boost.Account == nil {
-					// Fetch the relevant account for this status boost
+					// Fetch the relevant account for this status boost.
 					boostAcc, err := p.state.DB.GetAccountByID(ctx, boost.AccountID)
 					if err != nil {
-						l.Errorf("error fetching boosted status account for %q: %v", boost.AccountID, err)
-						continue
+						if errors.Is(err, db.ErrNoEntries) {
+							// We don't have an account for this boost
```
NyaaaWhatsUpDoc, line 365: Again worth logging on this one with a boost dump imo.

Considering all of these situations seem to be nil accounts it smells an awful like a bug we have yet to find :hankey:

internal/processing/account/delete.go
```diff
@@ -188,128 +383,125 @@ func (p *Processor) Delete(ctx context.Context, account *gtsmodel.Account, origi
...
-		// Update next maxID from last status
-		maxID = statuses[len(statuses)-1].ID
 	}
 
-	// 10. Delete account's notifications
-	l.Trace("deleting account notifications")
-	// first notifications created by account
-	if err := p.state.DB.DeleteWhere(ctx, []db.Where{{Key: "origin_account_id", Value: account.ID}}, &[]*gtsmodel.Notification{}); err != nil {
-		l.Errorf("error deleting notifications created by account: %s", err)
+	// Make sure we don't have a real error when we leave the loop.
```
NyaaaWhatsUpDoc, line 388: with the above `for {}` loop change we shouldn't need this extra check either :)

internal/processing/account/delete.go
```diff
@@ -188,128 +383,125 @@ func (p *Processor) Delete(ctx context.Context, account *gtsmodel.Account, origi
...
-	// 13. Delete account's mutes
-	l.Trace("deleting account mutes")
-	if err := p.state.DB.DeleteWhere(ctx, []db.Where{{Key: "account_id", Value: account.ID}}, &[]*gtsmodel.StatusMute{}); err != nil {
-		l.Errorf("error deleting status mutes created by account: %s", err)
+	return nil
+}
+
+func (p *Processor) deleteAccountPeripheral(ctx context.Context, account *gtsmodel.Account) error {
+	// Delete all bookmarks owned by given account.
+	if err := p.state.DB.DeleteStatusBookmarks(ctx, account.ID, ""); // nocollapse
```
NyaaaWhatsUpDoc, line 415: i love how prevalent my `nocollapse` bodge is becoming :woman_cook: :kissing_closed_eyes: :ok_hand:

internal/processing/account/delete.go
```diff
@@ -188,128 +383,125 @@ func (p *Processor) Delete(ctx context.Context, account *gtsmodel.Account, origi
...
-		}
+	// Delete all faves targeting given account.
+	if err := p.state.DB.DeleteStatusFaves(ctx, "", account.ID); // nocollapse
+	err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return err
 	}
 
-	// 17. Delete account's timeline
-	// TODO
+	// TODO: add status mutes here when they're implemented.
```
NyaaaWhatsUpDoc, line 438: is it worth adding a note here for account tags too since we remove the note above? i assume those would be tags found in the account bio :thinking:

NyaaaWhatsUpDoc, line 438: is it worth adding a note here for account tags too since we remove the note above? i assume those would be tags found in the account bio :thinking:

internal/processing/account/follow.go
```diff
@@ -33,172 +34,182 @@ import (
...
-	}
-	if form.Notify != nil {
-		fr.Notify = form.Notify
+		TargetAccount:   targetAccount,
+		ShowReblogs:     form.Reblogs,
+		Notify:          form.Notify,
 	}
 
-	// whack it in the database
 	if err := p.state.DB.Put(ctx, fr); err != nil {
```
NyaaaWhatsUpDoc, line 78: can't remember whether you added a `PutFollow()` function but if not there's definitely one coming in my refactor PR :p

internal/processing/account/relationships.go
```diff
@@ -138,3 +90,29 @@ func (p *Processor) RelationshipGet(ctx context.Context, requestingAccount *gtsm
 
 	return r, nil
 }
+
+func (p *Processor) accountsFromFollows(ctx context.Context, follows []*gtsmodel.Follow, requestingAccountID string) ([]apimodel.Account, gtserror.WithCode) {
+	accounts := make([]apimodel.Account, 0, len(follows))
+	for _, f := range follows {
+		if f.Account == nil {
+			// No account set for some reason; just skip.
```
NyaaaWhatsUpDoc, line 98: again, a warning log with follow model dump pls :innocent:

internal/processing/followrequest.go
```diff
@@ -29,21 +30,17 @@ import (
...
-	for _, fr := range frs {
+	accts := make([]apimodel.Account, 0, len(followRequests))
+	for _, fr := range followRequests {
 		if fr.Account == nil {
-			frAcct, err := p.state.DB.GetAccountByID(ctx, fr.AccountID)
-			if err != nil {
-				return nil, gtserror.NewErrorInternalError(err)
-			}
-			fr.Account = frAcct
+			// The creator of the follow doesn't exist,
```
NyaaaWhatsUpDoc, line 41: and again :innocent:

internal/processing/fromclientapi.go
```diff
@@ -636,13 +636,9 @@ func (p *Processor) federateAcceptFollowRequest(ctx context.Context, follow *gts
...
-	// if target account isn't from our domain we shouldn't do anything
-	if targetAccount.Domain != "" {
-		return nil
-	}
-
-	// if both accounts are local there's nothing to do here
-	if originAccount.Domain == "" && targetAccount.Domain == "" {
+	// Do nothing if target account *isn't* local,
+	// or both origin + target *are* local.
+	if targetAccount.IsRemote() || (originAccount.IsLocal() && targetAccount.IsLocal()) {
```
NyaaaWhatsUpDoc, line 641: you don't actually need the second `targetAccount.IsLocal()` here, that will be confirmed by this point :p

internal/processing/fromclientapi.go
```diff
@@ -709,13 +705,9 @@ func (p *Processor) federateRejectFollowRequest(ctx context.Context, followReque
...
```
NyaaaWhatsUpDoc, line 710: same again :innocent:

internal/processing/status/bookmark.go
```diff
@@ -31,91 +31,67 @@ import (
...
+	if err := p.state.DB.PutStatusBookmark(ctx, gtsBookmark); err != nil {
+		err = fmt.Errorf("BookmarkCreate: error putting bookmark in database: %w", err)
+		return nil, gtserror.NewErrorInternalError(err)
 	}
 
-	return apiStatus, nil
+	return p.apiStatus(ctx, targetStatus, requestingAccount)
 }
 
 // BookmarkRemove removes a bookmark for the requesting account, targeting the given status (no-op if bookmark doesn't exist).
```
NyaaaWhatsUpDoc, line 63: Just thinking regarding this BookmarkCreate and Remove functions; instead of passing a status ID and returning the status model (i assume to reduce database calls), it might make a little more sense to instead pass in a status model to begin with? Just for logic isolation

NyaaaWhatsUpDoc: I just looked up at the Follow/Unfollow methods above and they return the `Relationship{}` structure which does make more sense, but I'm not sure about the status here as we're not actually touching the status model itself at all.

internal/processing/status/fave.go
```diff
@@ -32,177 +32,134 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/uris"
 )
 
-// FaveCreate processes the faving of a given status, returning the updated status if the fave goes through.
+// FaveCreate adds a fave for the requestingAccount, targeting the given status (no-op if fave already exists).
 func (p *Processor) FaveCreate(ctx context.Context, requestingAccount *gtsmodel.Account, targetStatusID string) (*apimodel.Status, gtserror.WithCode) {
```
NyaaaWhatsUpDoc, line 36: Same thought goes with this as the bookmark function. If they're both going to cause too much code refactoring don't worry about it for now we can catch it in a later PR :p

internal/processing/status/fave.go
```diff
@@ -32,177 +32,134 @@ import (
...
+			err = fmt.Errorf("FavedBy: error checking blocks: %w", err)
+			return nil, gtserror.NewErrorInternalError(err)
+		} else if blocked {
+			continue
 		}
-		if !blocked {
-			filteredAccounts = append(filteredAccounts, fave.Account)
+
+		if fave.Account == nil {
+			// Account isn't set for some reason, just skip.
```
NyaaaWhatsUpDoc, line 132: log warning :innocent:

internal/processing/status/pin.go
```diff
@@ -89,7 +73,7 @@ func (p *Processor) getPinnableStatus(ctx context.Context, targetStatusID string
 //
 // If the conditions can't be met, then code 422 Unprocessable Entity will be returned.
 func (p *Processor) PinCreate(ctx context.Context, requestingAccount *gtsmodel.Account, targetStatusID string) (*apimodel.Status, gtserror.WithCode) {
```
NyaaaWhatsUpDoc, line 75: same as with the other status functions fave/bookmark above.

NyaaaWhatsUpDoc:
I FUMKIN' DID IT I FINISHED REVIEWING IT

internal/processing/processor.go
```diff
@@ -136,22 +136,26 @@ func NewProcessor(
...
-func (p *Processor) EnqueueFederator(ctx context.Context, msg messages.FromFederator) {
-	log.WithContext(ctx).WithField("msg", msg).Trace("enqueuing federator")
-	_ = p.state.Workers.Federator.MustEnqueueCtx(ctx, func(ctx context.Context) {
-		if err := p.ProcessFromFederator(ctx, msg); err != nil {
-			log.Errorf(ctx, "error processing federator message: %v", err)
-		}
-	})
+func (p *Processor) EnqueueFederator(ctx context.Context, msgs ...messages.FromFederator) {
+	log.Trace(ctx, "enqueuing federator")
+	for _, msg := range msgs {
```
NyaaaWhatsUpDoc, line 152: I think you meant to put the for loop inside the anonymous queue function :p. same goes for above. I know that's the original logic you intended from the reason you proposed the change to begin with :D

this should provide a nice performance boost :ok_hand:

internal/federation/federatingdb/followers.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/72eac6437d975850e7fd75e78f2d5b2e8ad4832b

internal/db/bundb/statusbookmark.go
NyaaaWhatsUpDoc: I think leaving them in is probably useful. Just to ensure no lost information and a little reminder, yaknow :)

internal/db/bundb/statusbookmark.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/4e2c30701fc4b4c38852a001e158febffbfda2e4

internal/db/bundb/statusfave.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/4e2c30701fc4b4c38852a001e158febffbfda2e4

internal/processing/account/delete.go
tsmethurst: This one was intended to be info :) Delete account events are unusual so I think it's not a bad idea to have them logged at a level where you'll see 'em

internal/processing/account/delete.go
tsmethurst: You're right! Tidied up that whole mess here: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/554aad8ebe98e90a785e69a5b43eb1d6f561e423

internal/processing/account/delete.go
tsmethurst: > it might have been me that introduced this coding style into our codebase to begin with

it was, absolutely self-busted, i'll change it here cuz it is a *little bit silly*

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/ee3465e0d1d9635cd92d3d3b8dc648b6b4040de2

NyaaaWhatsUpDoc: revealed myself as the stinkiest coder over them all

internal/processing/account/delete.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/e085890654167abc82fed7eebc8efb54249744f3

internal/processing/account/delete.go
tsmethurst: it's another one of those cases where maybe there was an error when selecting the TargetAccount from the db, some race condition or temporary outage, so it's worth checking imo

i added a commit to warn when it's not present, in line with the changes elsewhere

https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/e1cafb6b552e85f911545418b5fed8948d511f30

internal/processing/account/delete.go
tsmethurst: Yeah indeed. I think it might occur as a result of previous deletes failing, or some other fucked up db state (temporary outages when selecting account? it happens)

i hope adding these warnings will help us catch something, but with any luck they'll never actually be tripped :P 

https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/e8548f6fae4b347f39bed4a17011240a04ecee87

internal/processing/account/delete.go
tsmethurst: mm, i think when we implement accounts having tags in their bio properly, it'll be similar to how emojis work, so just a slice with tagIDs (comparable to emojiIDs) that gets stubbified along with everything else; tags aren't really owned by an account in the same way that other things are

internal/processing/account/follow.go
```diff
@@ -33,172 +34,182 @@ import (
...
 
-	// check for attempt to follow self
-	if requestingAccount.ID == targetAcct.ID {
-		return nil, gtserror.NewErrorNotAcceptable(fmt.Errorf("accountfollowcreate: account %s cannot follow itself", requestingAccount.ID))
-	}
-
-	// make the follow request
-	newFollowID, err := id.NewRandomULID()
+	// Create and store a new follow request.
+	followID, err := id.NewRandomULID()
```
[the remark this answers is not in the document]
tsmethurst: mmm i think it's this way because we use the id to generate the follow URI, and we don't necessarily want to give away to remotes when exactly a follow was created

internal/processing/account/relationships.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/7d633219f39a6c07ae7f9e3623b60b26344255a1

internal/processing/followrequest.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/e3b90cded5af6ca09b4c2afebbb3fbe4b93d1517

internal/processing/processor.go
tsmethurst: whoops, thank you! https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/d4bd3b3dfa8a062fd7e8268e8a8adb1aac6af584

internal/processing/fromclientapi.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1599/commits/02e05e73416c3fd4dd5d6c7ebf267041c7df2fa7

internal/processing/status/bookmark.go
tsmethurst: It's mostly structured this way because an api status (with bookmarked = true) is what the API expects in response to creating a bookmark. That's also why those follow/unfollow functions are like that: cuz the api wants a relationship to be served in response to doing a follow/unfollow.

PR #1604, opened 2023-03-09
[feature] Provide .well-known/host-meta endpoint

daenney:
# Description

>

Host-meta is back baby!

So it turns out, when the account and instance domain aren't the same (say pixie.town vs social.pixie.town), the official Mastodon Android client uses the `host-meta` endpoint to discover what the "actual" domain is. Currently this fallback isn't happening for GtS instances because we don't provide `host-meta` making a number of Mastodon apps not work.

The iOS client doesn't do this but (yet) but that's tracked in https://github.com/mastodon/mastodon-ios/issues/967 for now.

I suspect a number of other clients might do this too and having `host-meta` would make more things work with GtS than what we initially thought.

## Checklist

### DIFF
--- a/internal/api/wellknown/hostmeta/hostmetaget.go
+++ b/internal/api/wellknown/hostmeta/hostmetaget.go
@@ -1,0 +1,73 @@
+/*
+   GoToSocial
+   Copyright (C) 2021-2023 GoToSocial Authors admin@gotosocial.org
+
+   This program is free software: you can redistribute it and/or modify
+   it under the terms of the GNU Affero General Public License as published by
+   the Free Software Foundation, either version 3 of the License, or
+   (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Affero General Public License for more details.
+
+   You should have received a copy of the GNU Affero General Public License
+   along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package hostmeta
+
+import (
+	"bufio"
+	"bytes"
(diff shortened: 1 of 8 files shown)


internal/api/wellknown/hostmeta/hostmetaget.go
```diff
@@ -0,0 +1,73 @@
...
+		return
+	}
+
+	hostMeta := m.processor.Fedi().HostMetaGet()
+
+	// this setup with a separate buffer we encode into is used because
+	// xml.Marshal does not emit xml.Header by itself
+	var b bytes.Buffer
+	data := bufio.NewWriter(&b)
+	if _, err := data.Write([]byte(xml.Header)); err != nil {
```
NyaaaWhatsUpDoc, line 61: You can probably reformat this code section to be something like:
```
var buf bytes.Buffer

// Preallocate buffer of reasonable length.
buf.Grow(len(xml.Header)+64)

// No need to check for error on write to buffer.
_, _ = buf.WriteString(xml.Header)

// Encode host-meta as XML to in-memory buffer.
if err := xml.NewEncoder(&buf).Encode(hostMeta); err != nil {
    // etc etc.
}
```

Basically we don't need to use a `bufio.Writer` to write to a `bytes.Buffer` when it supports being written to directly, and using `WriteString(str)` instead of `Write([]byte(str))` saves some allocations as it doesn't need to create a byte-slice copy.

NyaaaWhatsUpDoc:
One small comment, but once that is sorted looks ready to go :)

internal/api/wellknown/hostmeta/hostmetaget.go
daenney: Done!

Thanks for giving me the code ready to copy-paste!

NyaaaWhatsUpDoc: hehe :grin:

Issue #1504, opened 2023-02-14
[feature] Test SMTP delivery


wioxjk:
### Is your feature request related to a problem ?

I just want a simple way to trigger a test-email.

### Describe the solution you'd like.

It would be nice if the admin-API could support such feature.
./gotosocial admin server test-smtp-delivery [address@example.net]

### Describe alternatives you've considered.

I am pretty sure that the GTS team have a better approach to this than the suggested one.

PR #1620, opened 2023-03-13
[feature] Allow admins to send test emails

tsmethurst:
# Description

>

This pull request adds a new admin client API endpoint `/api/v1/admin/email/test` which accepts POST requests with an `email` form field.

A test email will be sent to the given email address, so that admins can confirm their SMTP configuration is correct.

If the SMTP process is successful, a code 202 will be returned and admins can expect an email to be sent, which will look something like the following:

Screenshot from 2023-03-13 17-20-59

If the email sending fails during the SMTP stage, a code 422 will be returned, with an appropriate error message (example `{"error":"Unprocessable Entity: SMTP error: 535 Authentication failed"}`), which admins can use to help debug.

If the email sending fails for some other reason, a code 500 will be returned.

closes https://github.com/superseriousbusiness/gotosocial/issues/1504

## Checklist


internal/email/sender.go
```diff
@@ -25,13 +25,28 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/config"
 )
 
+// ErrorSMTP is returned from the email sender when
+// an smtp-specific error has occurred. The wrapped
+// error will always have been generated by a call
+// to the smtp.SendMail() function.
+type ErrorSMTP struct {
+	Wrapped error
+}
```
daenney, line 34: Should we also implement `Unwrap` on this so you could use `errors.Is` and `errors.As` on it?

tsmethurst:
Some questions for reviewers:

- is `/api/v1/admin/email/test` a decent endpoint for this?
- i wasn't sure about the code 422 but it seems to make some sort of sense; suggestions for a better code welcome, if one exists!
- will this work reasonably as an endpoint to call using the admin panel?

internal/email/sender.go
tsmethurst: Ah good point, thanks, I'll do that now

NyaaaWhatsUpDoc: Are we expecting to store other SMTP error data? I'm wondering whether it might be a little simpler if we added a helper method under `gtserror` like the following:
```
const (
	// error value keys.
	_ errkey = iota
	// ... other keys
	errorTypeKey
)

// Type checks error for a stored "type" value. For example
// an error from sending an email may set a value of "smtp"
// to indicate this was an SMTP error.
func Type(err error) string {
	s, _ := errors.Value(err, errorTypeKey).(string)
	return s
}

// SetType will wrap the given error to store a "type" value,
// returning wrapped error. See Type() for example use-cases.
func SetType(err error, errType string) error {
	return errors.WithValue(err, errorTypeKey, errType)
}
```

Then we can just do `gtserror.Type(err) == "smtp"` to check for smtp type errors. It also helps consolidate custom error code in a singular place. What do you think?

tsmethurst: Ah sure, I'm up for doing it that way instead :) I'll make the changes tonight or tomorrow!

tsmethurst: Changed it here: https://github.com/superseriousbusiness/gotosocial/pull/1620/commits/c906509718b4febcf217414a212f2be161bdae7f

Rather than just a string I figured we might as well use an error type thing so we can match against it a bit more easily :)

NyaaaWhatsUpDoc:
> will this work reasonably as an endpoint to call using the admin panel?

I think all 3 of these points seem reasonable. No disagreements from me :)

internal/gtserror/error.go
```diff
@@ -24,11 +24,19 @@ import (
 // package private error key type.
 type errkey int
 
+// ErrorType denotes the type of an error, if set.
+type ErrorType string
```
NyaaaWhatsUpDoc, line 28: Was there a reason to use a type alias here? I was thinking we probably don't need a separate type for this.

internal/gtserror/error.go
```diff
@@ -24,11 +24,19 @@ import (
...
 const (
 	// error value keys.
 	_ errkey = iota
 	statusCodeKey
 	notFoundKey
+	errorTypeKey
+
+	// error types
+	TypeSMTP    ErrorType = "smtp" // smtp (mail) error
+	TypeUnknown ErrorType = ""     // unknown or unset error type
```
NyaaaWhatsUpDoc, line 39: Do we need to set an unknown type? I can't imagine us checking against that :thinking: Though I suppose we could add it back in later if we do need it.

internal/processing/admin/email.go
```diff
@@ -0,0 +1,62 @@
...
+
+// EmailTest sends a generic test email to the given toAddress (which
+// should be a valid email address). To help callers differentiate between
+// proper errors and the smtp errors they're likely fishing for, will return
+// 422 + help text on an SMTP error, or error 500 otherwise.
+func (p *Processor) EmailTest(ctx context.Context, account *gtsmodel.Account, toAddress string) gtserror.WithCode {
+	// Pull our instance entry from the database,
+	// so we can greet the email recipient nicely.
+	instance := &gtsmodel.Instance{}
+	if err := p.state.DB.GetWhere(ctx, []db.Where{{Key: "domain", Value: config.GetHost()}}, instance); err != nil {
```
NyaaaWhatsUpDoc, line 39: Is it worth adding a `GetInstance()` function to the database? Just so we don't increase our reliance on using `GetWhere()` which will slow down further migrations from `bundb`.

NyaaaWhatsUpDoc: Obviously don't need to correct all uses in the codebase to use `GetInstance()`, that can be added to an issue or something :p

NyaaaWhatsUpDoc:
A few comments but otherwise looks really good :)

internal/gtserror/error.go
tsmethurst: I just find it a little bit cleaner + also provides better hints about what you should be expected to pass. I don't mind dropping it, but I do prefer using cheeky little type aliases for things like this.

internal/gtserror/error.go
tsmethurst: I'll take it out for now, since we don't use it!

internal/processing/admin/email.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1620/commits/761c34f1828e776c2a36e1874b87a98e13679eb1

internal/gtserror/error.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1620/commits/33edea359ba4f2ea00c35a4c9aa0076c7219e3ac

internal/gtserror/error.go
daenney: I kinda like those too. IDEs will also show that when you're inspecting things which can provide a useful hint as to what this thing actually is.

NyaaaWhatsUpDoc: Fair enough, leave as-is then :)

mirabilos:
Where in the dmin panel should show this up?

tsmethurst:
It's not in the admin panel yet, just in the API :) We'll add it soon (tm) so you can send test emails from the panel.

mirabilos:
tobi dixit:

>It's not in the admin panel yet, just in the API :) We'll add it
>soon(tm) so you can send test emails from the panel.

OK, thanks.

Enjoy,
//mirabilos
-- 
15:41⎜<Lo-lan-do:#fusionforge> Somebody write a testsuite for helloworld :-)

PR #1623, opened 2023-03-14
feat: initial tracing support

### LINKED ISSUE #1230: [feature] Instrument GoToSocial with OpenTelemetry or similar tracing thingy

theSuess:
# Description

This pull request implements basic tracing support for HTTP requests and database queries. It accomplishes this by utilizing the opentelemetry go sdk as well as pre-made integrations for [gin](]https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/github.com/gin-gonic/gin/otelgin) and [bun](https://github.com/uptrace/bun/tree/master/extra/bunotel).

To reduce binary size, tracing support can be disabled at build time by building GtS with the `notracing` build tag. This will reduce the final binary size by about 4 MB.

If you want to test this, the easiest way I found was to set up Grafana and Tempo locally using a compose file similar to the following:
```
---
volumes:
  grafana_data:
services:
  grafana:
    image: grafana/grafana:9.3.2
    volumes:
      - grafana_data:/var/lib/grafana
    environment:
      - GF_AUTH_ANONYMOUS_ENABLED=true
      - GF_AUTH_ANONYMOUS_ORG_ROLE=Admin
      - GF_AUTH_DISABLE_LOGIN_FORM=true
      - GF_FEATURE_TOGGLES_ENABLE=traceqlEditor
    ports:
      - "3001:3000"
  tempo:
    image: grafana/tempo:latest
    command:
      [
        "-storage.trace.backend=local",
        "-storage.trace.wal.path=/tmp/tempo/wal",
        "-storage.trace.local.path=/tmp/tempo/blocks",
        "-server.http-listen-port=3200",
      ]
    ports:
      - "14268:14268" # jaeger ingest
      - "3200:3200" # tempo
      - "4317:4317" # otlp grpc
```

Resulting traces look like this:
image

The tracing configuration currently supports the jaeger and otlp grpc protocols. IMHO these are enough for most users.

closes #1230 

## Checklist


theSuess:
I've intentionally left out the documentation as I'd like to get some feedback on the provided configuration options first

theSuess:
Also: if tracing is available, the requestId middleware will respect the trace id.
This should also work in distributed settings where GtS receives the parent trace id via the `traceparent` header

theSuess:
Also: if tracing is available, the requestId middleware will respect the trace id.
This should also work in distributed settings where GtS receives the parent trace id via the `traceparent` header

internal/config/config.go
```diff
@@ -151,7 +156,7 @@ type Configuration struct {
 	AdminTransPath        string `name:"path" usage:"the path of the file to import from/export to"`
 	AdminMediaPruneDryRun bool   `name:"dry-run" usage:"perform a dry run and only log number of items eligible for pruning"`
 
-	RequestIDHeader string `name:"request-id-header" usage:"Header to extract the Request ID from. Eg.,'X-Request-Id'"`
+	RequestIDHeader string `name:"request-id-header" usage:"Header to extract the Request ID from. Eg.,'X-Request-Id'. If tracing is enabled, this option has no effect as the trace ID is used as request id"`
```
daenney, line 159: I think this needs reconsidering. We currently pick up the Request-ID from a header like X-Request-ID which can be set by someone's LB/proxy. Having this ignored when tracing is active is not great, because we lose the link between the two.

I'm wondering if we can reverse it, whereby the RequestID is used as the TraceID

theSuess: To be compatible with the [trace context standard](https://w3c.github.io/trace-context/#trace-id), the format of the trace ID has to be very specific. We could however introduce a toggle to switch between logging the trace id and an external request id

daenney: Alternatively we could have both. That decouples the two systems, and then add the requestid as an attribute on the trace instead?

theSuess: Sure, that would work as well. However, I would still like to see the traceId in the logs at some point. This allows for correlation of logs (e.g errors or weird messages) and traces. At that point we'd have two unrelated IDs in a single log line which could cause confusion.

Just for consideration: if you generate a valid `traceparent` on your loadbalancer (or even enable native tracing), the requestId middleware would behave the same as before

daenney: If you look at the requestID middleware, you see it installs a log-hook. That's what adds a `requestid=xxxx` line to the logs currently. The tracing middleware can do the same

daenney: > Just for consideration: if you generate a valid traceparent on your loadbalancer (or even enable native tracing), the requestId middleware would behave the same as before

I suppose that's true but I'm not sure that'll help us in this case. Not every LB out there supports that feature, whereas the request ID thing has been around since the early 2000s. The request ID also has no particular requirements for the format, making it much easier to generate upstream than the specific traceparent format.

theSuess: I think keeping the systems separated makes a lot of sense, but then I'd also introduce a way to disable either system completely.

Adding the external request ID (if enabled) to the trace ID is something I can look into, however this could get complicated very quickly when taking the build-time opt-out into account.

daenney: > I think keeping the systems separated makes a lot of sense, but then I'd also introduce a way to disable either system completely.

Yeah, that makes sense. Even without tracing not everyone might want that so it seems reasonable to let folks disable it. Since it's barely any code and doesn't rely on external dependencies I think a config option is fine and we don't need a build-time option for this one.

> Adding the external request ID (if enabled) to the trace ID is something I can look into, however this could get complicated very quickly when taking the build-time opt-out into account.

I think this might be easier if you swap the middleware order around? If requestID goes first, then you can call https://github.com/superseriousbusiness/gotosocial/blob/196cd88b1c7c44a337ca12f6a804f1bb7fa83c4a/internal/middleware/requestid.go#L73 in the tracing middleware and add it to the span there. I think.

theSuess: Okay, so I've split up the two systems and extracted the requestID into the span. This is currently quite cumbersome and results in quite ugly middleware initialization.

I think I'll just implement the relevant parts of the otelgin library directly, which would allow reducing the number of middlewares needed.

tsmethurst: Sorry I'm coming in a bit late here, hadn't seen this discussion yet. I think having the requestid as a property on the trace id makes sense. In what situation would you want to log the trace id? In case of an error or something, so you can go look through your traces?

theSuess: Exactly - if some error occurs I could look up the trace ID and see the entire context (e.g source of the request, previous queries and timings etc.)

Most observability solutions also allow you to correlate logs with traces directly, which is an added bonus

internal/config/defaults.go
```diff
@@ -174,5 +179,6 @@ var Defaults = Configuration{
 
 	AdminMediaPruneDryRun: true,
 
-	RequestIDHeader: "X-Request-Id",
+	RequestIDEnabled: true,
```
tsmethurst, line 182: Could you explain in what circumstances would you want to set this to false? Previously we had this enabled by default (and not turn-offable) so I'm wondering what it is in the tracing stuff that means we might wanna turn this off. Not a barbed question, btw, I just don't understand :P

tsmethurst (commented):
Looking good so far :) Thanks for doing all this!

internal/config/defaults.go
theSuess: Sure! I'd disable this if tracing is enabled. My reverse proxy can collect trace data as well, so I'd have a full picture of the entire lifespan of the request without the requestId middleware. If tracing is desired, having a separate, unrelated requestID in the log line increases noise and makes it harder to pick out desired information.

Now if my reverse proxy would *not* support tracing, I'd leave this enabled to correlate between the proxy and GtS

theSuess:
I'm marking this as a draft and reopen when I've rewritten the middleware. Will keep the PR open however to keep the discussion around the specific approach taken

tsmethurst:
Hey @theSuess ! sorry for the delay in reviewing this, a few things got stacked up. Could you possibly rebase it? Then I can review it properly and hopefully merge it in :)

theSuess:
I've rebased the tracing stuff onto the latest changes

I noticed that the error page now reports the request ID so I'll remove the toggle for enabling/disabling the requestId middleware as it seems mandatory now

internal/config/defaults.go
```diff
@@ -102,12 +102,16 @@ var Defaults = Configuration{
...
-	SMTPUsername:           "",
-	SMTPPassword:           "",
-	SMTPFrom:               "GoToSocial",
-	SMTPDiscloseRecipients: false,
+	SMTPHost:     "",
+	SMTPPort:     0,
+	SMTPUsername: "",
+	SMTPPassword: "",
+	SMTPFrom:     "GoToSocial",
+
```
tsmethurst, line 110: Looks like you accidentally removed the disclose recipients config option

theSuess: Whoops, should be resolved now

NyaaaWhatsUpDoc:
Sorry for the lateness of getting to this PR (work has been a bitch recently lol), it looks good but I did have one question regarding the choice of metrics library itself. Would we be able to use a lighter alternative like: https://github.com/VictoriaMetrics/metrics

For an example of reduced binary sizes by library comparison: https://valyala.medium.com/stripping-dependency-bloat-in-victoriametrics-docker-image-983fb5912b0d

NyaaaWhatsUpDoc:
Ah finally my comment posts!! I have been trying for like 3hrs now lol

NyaaaWhatsUpDoc:
> > Sorry for the lateness of getting to this PR (work has been a bitch recently lol), it looks good but I did have one question regarding the choice of metrics library itself. Would we be able to use a lighter alternative like: [VictoriaMetrics/metrics](https://github.com/VictoriaMetrics/metrics)
> > For an example of reduced binary sizes by library comparison: [valyala.medium.com/stripping-dependency-bloat-in-victoriametrics-docker-image-983fb5912b0d](https://valyala.medium.com/stripping-dependency-bloat-in-victoriametrics-docker-image-983fb5912b0d)
> 
> Note that this PR is for tracing - not metrics. It just happens that OTEL can do both. I can take a look around if I can find a more lightweight tracing lib, but I'm not aware of one as of now.

Ah gotcha. I wasn't actually aware of any difference until now, I had always had metrics / tracing in the same meaning in my head! In that case I'm happy with this PR, all looks good :)

NyaaaWhatsUpDoc:
nice work! :rocket:

PR #1626, opened 2023-03-15
[chore] Update docs on how to login

### LINKED ISSUE #1609: App login when account and host domain differ
This is a write-up of an issue we've seen with a number of iOS and Android client apps that
* Don't manage to login to GtS at all
* Complete the login flow but then still sit there or display an empty timeline

This issue only applies to instances where the account domain is different from the API/host domain. For example instances where the users are `@account@instance.tld` but the instance server itself is hosted on `subdomain.instance.tld`.

The fundamental issue is: how does an app discover the difference between these two things:
* the API domain: `subdomain.instance.tld`
* the account domain: `instance.tld`

This is important for a very simple reason: the oauth login flow. If the instance lives on `subdomain.instance.tld` you have to start the oauth flow from there, and not `instance.tld`. If you have a redirect in place from `instance.tld` to `subdomain.instance.tld` you'll end up performing the whole oauth flow successfully and getting a token. But, then you'll start doing requests with an `Authorization` header to `instance.tld`, your server will issue a redirect to `subdomain.instance.tld` and the HTTP client will duly drop the `Authorization` header from future requests in order to avoid cross-domain credential leaks. This is sensible behaviour, but it means we need an accurate way of discovering what the actual domain the instance is running on is. This is what leads to an app looking like it has completed the login successfully but then just sitting there or displaying an entirely empty timeline.

## iOS Apps

These apps' behaviours were observed end-to-end with MITMproxy.

### Metatext
Doesn't seem to do any fallback discovery. You have to use the right domain from the start. It'll do an API request to `/api/v1/instance` though to figure out if there's something there.

With our current redirect recommendation, `instance.tld/api/v1/instance` will return a successful response (because clients typically follow redirects) but the app is unaware of that fact. This lets you start the oauth flow from the wrong domain.

### Ice Cubes

Doesn't do any fallback discovery. It'll try `/api/v1/instance`. If that redirects the oauth flow starts from the wrong domain. If it 404s the client won't let you proceed.

### Toot!

Toot! uses instances.social to search for the instance name you've put in. It doesn't appear to use this to discover any potential difference between API and account domain though and then tries a couple of endpoints to see if something responds:
* /api/v2/instance
* /api/v1/instance

Even when /api/v2/instance succeeds, which contains a `"domain": "subdomain.instance.tld"` in the response, Toot! appears to still check `/api/v1/instance`. When that fails it perform a `.well-known/host-meta` based fallback to pick out the right domain. It then initiates the oauth flow from the domain discovered through host-meta resulting in a successful login.

It's not entirely clear from the logs if the requests to `v2/instance` and `/v1/instance` happen sequentially or in parallel, but it seems like a failure on those triggers host-meta fallback discovery. Due to #1604 this now works. 

### Mammoth

Mammoth really struggles. It'll first search instances.social. If that doesn't return anything it'll try `/api/v1/apps` for some reason. If that 404s, it gives up and doesn't let you proceed.

If you redirect the root domain then the call to `/api/v1/apps` succeeds but then we're starting the oauth flow from the wrong domain so the app still ends up not working.

### Ivory

Ivory won't work with GtS at all because they've made an assumption that toot IDs are numerical. But we can still check the login behaviour.

Ivory has the same behaviour as Mammoth, going for `/api/v1/apps`. If that's redirected then it'll trigger the oauth flow from the wrong domain causing it to fail to login.

When 404ing `/api/v1/apps` on the main domain the app doesn't do any fallback checks and simply fails.

### Mastodon iOS

The official app currently doesn't implement the host-meta based fallback discovery, contrary to the official Android app. This results in much the same situation where if a redirect from the root domain is in place the oauth flow starts from the wrong domain and things don't work.

If you give the app the `subdomain.instance.tld` as a starting point you're able to login and use the app.

## Android

I have less visibility in the Android apps because:
* I can't make apps use the MITMproxy cert on new version of Android unless the app explicitly opts-in to "user certificates" which nobody does
* Android emulator segfaults on my desktop

### Tusky

Tusky seems to directly try the domain that it was given to login to, and call a POST on `/api/v1/apps`. If that fails the client won't continue and doesn't do a host-meta based discovery.

If `/api/v1/apps` etc. is redirected from the top domain though it'll proceed but as usual at the end we'll lose the authorization header.

### Fedilab

Fedilab seems to somehow be able to differentiate between the account and instance domain in certain circumstances. I can't quite see how it does that, but it's definitely not host-meta fallback based. I suspect it's using instances.social for this.

Same story as with everyone else, if `/api/v1/...` is redirected from the account to the instance domain then the oauth flow completes but you can't login.

### AndStatus

This one is slightly novel. It tries to query for `/.well-known/oauth-authorization-server` server first. If that fails it tries to post to `/api/v1/apps` a few times and then gives up. There's no host-meta based fallback.

Same story as with everyone else, if `/api/v1/...` is redirected from the account to the instance domain then the oauth flow completes but you can't login.

### Mastodon Android

The official app implements a `host-meta` fallback that's triggered when `/api/v1/instance` on the main domain 404s. Due to #1604 this now works.

## Conclusion

There basically seem to be 2 types of apps:
* Those that do host-meta based fallback discovery
* Those that don't, which are in the majority it seems

All apps that do host-meta based fallback discovery only do so when they receive a 404 on `/api/v1/instance` on the account domain. That's their heuristic for triggering it. If at any point you've configured your setup to redirect from `instance.tld` to `subdomain.instance.tld` this will cause apps to start the oauth flow from the wrong domain as `/api/v1/instance` will respond on the account domain.

For the apps that don't do fallback discovery, if at any point you've configured your setup to redirect from `instance.tld` to `subdomain.instance.tld` this will cause apps to start the oauth flow from the wrong domain as `/api/v1/instance` will respond. Some of these apps use `/api/v1/apps` instead.

None of the client apps seem to be querying `/.well-known/nodeinfo` for anything.

Some of these apps query instances.social as a way of displaying additional information about the instance in the app when searching for your instance. However, none seem to use the returned information to discover the difference between the account and instance domain.

We should check our docs to ensure that in the case of `account-domain` and `host` being different:
* Explicitly advise against redirecting or proxying anything other than the well-known endpoints
* Review our current reverse proxy configs to ensure our examples don't accidentally result in us redirecting or proxying anything other than the well-known endpoints

For apps without the fallback discovery, I'll try and raise that with developers if they have a way of reaching them. We shouldn't keep this issue open for that though. Once the docs are sorted we can close this as there's nothing more we can do from our side.

daenney:
# Description

This adds a section to the docs instructing how to ensure apps will be able to successfully login to an instance when host- and account-domain differ.

Resolves #1609

## Checklist

### DIFF
--- a/docs/installation_guide/advanced.md
+++ b/docs/installation_guide/advanced.md
@@ -8,6 +8,10 @@ Yes, you can! This is useful when you have something like a personal page or blo
 
 Please note that you need to do this *BEFORE RUNNING GOTOSOCIAL* for the first time, or things will likely break.
 
+An additional thing to keep in mind is that there is no good way for applications to detect if you're running this style of deployment. Therefor you should recommend that folks use `fedi.example.org` as the instance to login to in any client application.
+
+Some applications do have heuristics built-in to try and detect this situation and make login from either domain possible. That heuristic relies on `/api/v1/instance` or `/api/v1/apps` not responding on `example.org`. When that happens they'll do a fallback lookup by requesting `example.org/.well-known/host-meta`. You need to ensure that this endpoint is properly redirected to `fedi.example.org` as shown in our examples below. It is crucial you don't redirect `/api` or any of its subpaths from `example.org` to `fedi.example.org`, but only the well-known endpoints, to not break this heuristic.
+
 ### Step 1: Configure GoToSocial
 
 This step is easy.
@@ -59,7 +63,7 @@ The next step is more difficult: we need to ensure that when remote instances se
 
 Of course, we don't want to redirect *all* requests from `example.org` to `fedi.example.org` because that negates the purpose of having a separate domain in the first place, so we need to be specific.
 
-In the config.yaml above, there are two endpoints mentioned, both of which we need to redirect: `/.well-known/webfinger` and `/.well-known/nodeinfo`.
+In the config.yaml above, there are three endpoints mentioned, all of which we need to redirect: `/.well-known/webfinger`, `/.well-known/nodeinfo` and `/.well-known/host-meta`.
 
 Assuming we have an [nginx](https://nginx.org) reverse proxy running on `example.org`, we can get the redirect behavior we want by adding the following to the nginx config for `example.org`:
 
@@ -87,7 +91,7 @@ http {
 }
 ```
 
-The above configuration [rewrites](https://www.nginx.com/blog/creating-nginx-rewrite-rules/) queries to `example.org/.well-known/webfinger` and `example.org/.well-known/nodeinfo` to their `fedi.example.org` counterparts, which means that query information is preserved, making it easier to follow the redirect.
+The above configuration [rewrites](https://www.nginx.com/blog/creating-nginx-rewrite-rules/) queries to `example.org/.well-known/webfinger`, `example.org/.well-known/nodeinfo` and `example.org/.well-known/host-meta` to their `fedi.example.org` counterparts while preserving any query arguments, making it easier to follow the redirect.
 
 If `example.org` is running on [Traefik](https://doc.traefik.io/traefik/), we could use labels similar to the following to setup the redirect.
 


tsmethurst:
Thanks daenney!

PR #1640, opened 2023-03-22
[chore] Refactor timeline Get, other small fixes

tsmethurst:
# Description

>

This pull request does some v. necessary refactoring on our timeline code, mostly updating `internal/timeline/get.go` to make it more readable, remove some cruft, and add some comments that better explain what's actually happening, since it's not easy code to understand.

The PR also distinguishes properly between `minID` and `sinceID`, which have slightly different uses and indicate the direction that paging should be performed in. This is explained in the comments, but essentially sinceID means 'return from the top' and minID means 'return from the mark upwards'. It's likely that not distinguishing properly between these two things was causing users to miss statuses / not be able to fill gaps in their timeline properly.

We still ought to go through the rest of the `timeline` package and clean up other functions, but this is a start and I don't want to turn it into another massive, difficult-to-review refactor :')

## Checklist


tsmethurst:
Will open this one again. The scope has changed to a bigger refactor because.. well.. it wasn't the best code I've ever written :')

Issue #388, opened 2022-02-09
[feature] GtS CLI: list local users


igalic:
## problem

gotosocial cli currently has no way to list local users:

```
# gotosocial admin account --help
admin commands related to accounts

Usage:
  gotosocial admin account [command]

Available Commands:
  confirm     confirm an existing account manually, thereby skipping email confirmation
  create      create a new account
  demote      demote an account from admin to normal user
  disable     prevent an account from signing in or posting etc, but don't delete anything
  password    set a new password for the given account
  promote     promote an account to admin
  suspend     completely remove an account and all of its posts, media, etc

Flags:
  -h, --help              help for account
      --username string   the username to create/delete/etc
```

This might be fine for tiny instances, but anything can get pretty unwieldly pretty quickly

## Workaround

fiddle around in the database??

## Proposed solution:

add a sub-command that fiddles around in the database and lists local users:

```
gotosocial admin account list
```

A tabular display of username and capabilities and whether the account is disabled or suspended would be good, but having flags to select users based on these stati would be also very good!

```
gotosocial admin account list [--suspended|--unconfirmed|--disabled|--etc…]
```

tsmethurst:
Aha, this has been possible for a while now (see above commits) so I'll close this :)

PR #1648, opened 2023-03-26
[feature] Add list command to admin account

daenney:
# Description

>

A first pass at #388. Doesn't implement the filtering that was suggested there. I'm inclined to say we shouldn't have filter flags as that's a potentially never ending story. I'm thinking we might want to add a `--format=json` flag there instead and dump the object, so folks can slice, dice and filter with `jq` to their heart's content.

## Checklist

### DIFF
--- a/internal/db/bundb/user.go
+++ b/internal/db/bundb/user.go
@@ -122,6 +122,20 @@ func (u *userDB) GetUserByConfirmationToken(ctx context.Context, confirmationTok
 	}, confirmationToken)
 }
 
+func (u *userDB) GetAllUsers(ctx context.Context) ([]*gtsmodel.User, db.Error) {
+	var users []*gtsmodel.User
+	q := u.conn.
+		NewSelect().
+		Model(&users).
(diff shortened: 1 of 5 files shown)


daenney:
On my own instance this results in:

```
user  account                    approved admin moderator suspended confirmed
admin 01.......................H true     true  false     no        yes
ohno  01.......................P true     false false     no        yes

```

internal/db/bundb/user.go
```diff
@@ -122,6 +122,20 @@ func (u *userDB) GetUserByConfirmationToken(ctx context.Context, confirmationTok
 	}, confirmationToken)
 }
 
+func (u *userDB) GetAllUsers(ctx context.Context) ([]*gtsmodel.User, db.Error) {
```
daenney, line 125: I didn't plonk this into the cache because I wasn't really sure how to do that in this case (the list of users). Also (for now) it's only used by the CLI at which point there's isn't a ton of point to going through the cache.

PR #1652, opened 2023-03-28
[frontend] Settings navigation design

f0x52:
# Description

Headache refactor of the navigation menu structure, now featuring proper nesting and nicer design.
image

code to define routes/views
image

## Checklist


web/source/css/base.css
```diff
@@ -120,7 +120,7 @@ header a {
 }
 
 .excerpt-top {
-	margin-top: -1rem;
+	/* margin-top: -1rem; */
```
tsmethurst, line 123: might as well ditch commented out lines I think

web/source/settings/components/user-logout-card.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+"use strict";
+
+const React = require("react");
+
+const query = require("../lib/query");
+
+const Loading = require("./loading");
+
+module.exports = function UserLogoutCard() {
```
tsmethurst, line 28: Nice, this is cool :)

web/source/settings/style.css
```diff
@@ -79,77 +99,143 @@ section {
...
+				background: $settings-nav-bg-active;
+				font-weight: bold;
+				border-radius: 0;
 			}
 		}
+	}
 
+	li.nested { /* any deeper nesting, just has indent */
+		a.title {
+			/* padding-left: calc($fa-fw + 0.9rem); */
```
tsmethurst, line 217: same comment as above re: commented-out lines

web/source/settings/style.css
```diff
@@ -34,18 +34,38 @@ section {
...
+	border-radius: none;
+	border: none;
+}
+
 #root {
 	display: grid;
 	/* keep in sync with base.css .page {} */
-	grid-template-columns: auto minmax(auto, 50rem) auto;
-	grid-template-columns: auto min(92%, 50rem) auto;
+	/* grid-template-columns: auto min(92%, 50rem) 1fr; */
```
tsmethurst, line 63: another one here

web/source/settings/lib/navigation/util.js
```diff
@@ -20,15 +20,30 @@
...
+}
+
+function useHasPermission(permissions) {
+	const roles = React.useContext(RoleContext);
+	return checkPermission(permissions, roles);
+}
+
+function checkPermission(required, user) {
+	if (required === true) {
+		return true;
```
tsmethurst, line 37: Not sure I understand what's happening in this line, can you explain?

tsmethurst (commented):
This is looking good! Not very cursed at all!

web/source/settings/lib/navigation/util.js
f0x52: Added some comments; if the required permissions are `true`, no role check is needed, if it's an Array, check if one of the user's roles matches

tsmethurst: Ahh okay. I think it would make more sense to have:

```
if (requiredPermissions === false) {
  return true;
}
```

Then it more clearly indicates that permissions actually aren't required, so check permissions can always return OK. See what I mean?

PR #1656, opened 2023-03-30
[bugfix/chore] Refactor timeline code

### LINKED ISSUE #1600: [bug] Timeline just ends, prepareBehind: error preparing item with id

Load my home timeline on a non-caching front-end (like, e.g., elk or pinafore). The timeline just ends, sometimes, and GoToSocial then prints something like
```
timestamp="07/03/2023 11:51:40.134" func=timeline.(*manager).GetTimeline level=ERROR timelineAccountID=<accountId> msg="error getting statuses: prepareBehind: error preparing item with id <itemId>: statusPrepareFunction: error getting status with id <itemId>"
```

Given that it happens on at least two different clients, I think this might be an issue with GoToSocial. I do not know when it happens or how to trigger it, unfortunately.

 GoToSocial 0.7.0 git-b599309

### GoToSocial Arch

_No response_

### What happened?

The timeline ends, GoToSocial prints an error.

The timeline to go on.

### How to reproduce it?

Unknown

_No response_

tsmethurst:
# Description

>

This pull request refactors the timeline code for readability, bug fixes, and performance:

- Allow proper paging up using minID.
- Only use one linked list for timeline, instead of two separate ones.
- Only fail when preparing on proper database errors -- closes #1600 
- Reduce duplicated code.
- Remove unused functions.

## Checklist


internal/timeline/types.go
```diff
@@ -17,10 +17,18 @@
...
+// Timelineable represents any item that can be indexed in a timeline.
 type Timelineable interface {
 	GetID() string
 	GetAccountID() string
 	GetBoostOfID() string
 	GetBoostOfAccountID() string
 }
+
+// Preparable represents any item that can be prepared in a timeline.
+type Preparable interface {
```
NyaaaWhatsUpDoc, line 29: is it worth having a separate interface containing the same methods? maybe a slightly more generic interface name that would fit both object types :thinking:

tsmethurst: Yeah good point! Will change it

internal/processing/fromcommon.go
```diff
@@ -455,8 +455,8 @@ func (p *Processor) timelineStatusForAccount(ctx context.Context, account *gtsmo
 		return nil
 	}
 
-	// stick the status in the timeline for the account and then immediately prepare it so they can see it right away
-	if inserted, err := p.statusTimelines.IngestAndPrepare(ctx, status, account.ID); err != nil {
+	// stick the status in the timeline for the account
+	if inserted, err := p.statusTimelines.IngestOne(ctx, account.ID, status); err != nil {
```
tsmethurst, line 459: note for reviewer(s):

this now always prepares the entry as it's being inserted, since we weren't really using the other function anyways

internal/processing/statustimeline.go
```diff
@@ -174,37 +178,54 @@ func (p *Processor) HomeTimelineGet(ctx context.Context, authed *oauth.Auth, max
...
-	prevMinIDValue := ""
-	for i, item := range filtered {
+	var (
+		items          = make([]interface{}, 0, count)
+		nextMaxIDValue string
+		prevMinIDValue string
+	)
+
+	for i, s := range statuses {
+		// Set next + prev values before filtering and API
```
tsmethurst, line 202: note for reviewer(s):

this change ensures that the caller always uses the outside range of returned statuses, before they're filtered

previously, if the public timeline db call returned, say, 10 statuses, and 8 from the end got filtered out, the next maxID would be set to only the second status in the slice, meaning that the NEXT call to this function would then include the 8 statuses that were filtered out, causing unnecessary db calls + wasteful checks

the tradeoff here is that now we expose IDs of statuses in the Link headers that the caller doesn't want to see; but since these are public statuses, it's not really a big deal imo

internal/timeline/index.go
```diff
@@ -24,103 +24,205 @@ import (
...
 
-	// now check if the length of indexed items exceeds the amount of items required (position of itemID, plus amount of posts requested after that)
-	if t.indexedItems.data.Len() > position+amount {
-		// we have enough indexed behind already to satisfy amount, so don't need to make db calls
-		l.Trace("returning nil since we already have enough items indexed")
-		return nil
+	// We can now figure out if we need to make db calls.
+	var grabMore bool
+	switch {
+	case listLen < amount:
```
tsmethurst, line 103: note for reviewer(s):

There's likely a better way of doing this and/or more conditions that need to be checked.

internal/timeline/index.go
```diff
@@ -24,103 +24,205 @@ import (
...
+			// Page up.
+			minID = items[0].GetID()
+			if minID >= behindID {
+				// Can't go any further.
+				break
+			}
+		}
+
+		for _, item := range items {
+			ok, err := t.filterFunction(ctx, t.accountID, item)
```
tsmethurst, line 208: Here's where that pesky filter function is called (see above comment)

internal/timeline/manager.go
```diff
@@ -97,31 +100,41 @@ type manager struct {
...
-	// range through all timelines in the sync map once per hour + prune as necessary
+	// Start a background goroutine which iterates
+	// through all stored timelines once per hour,
+	// and cleans up old entries if that timeline
+	// hasn't been accessed in the last hour.
 	go func() {
 		for now := range time.NewTicker(1 * time.Hour).C {
-			m.accountTimelines.Range(func(key any, value any) bool {
-				timelineAccountID, ok := key.(string)
+			// Define the range function inside here,
```
tsmethurst, line 109: note for reviewer(s):

I wrote and rewrote this bit of code a couple of times to try to find the neatest way of doing it. I'm not 100% happy with the result; there's likely a nicer way of doing it.

internal/db/bundb/timeline.go
```diff
@@ -56,11 +59,9 @@ func (t *timelineDB) GetHomeTimeline(ctx context.Context, accountID string, maxI
 			bun.Ident("follow.target_account_id"),
 			bun.Ident("status.account_id"),
 			bun.Ident("follow.account_id"),
-			accountID).
-		// Sort by highest ID (newest) to lowest ID (oldest)
-		Order("status.id DESC")
+			accountID)
 
-	if maxID == "" {
+	if maxID == "" || maxID == id.Highest {
```
NyaaaWhatsUpDoc, line 64: I'd make this `|| maxID >= id.Highest`

NyaaaWhatsUpDoc: actually don't worry i'll catch this

Issue #1671, opened 2023-04-03
[feature] Mute User Boosts


ultramookie:
### Is your feature request related to a problem ?

There is a nice feature in Mastodon where one can mute a user's boosts. There are some users that have interesting posts that I'd like to follow, but they also boost a lot. This feature is nice since it allows users to only see original posts and not be flooded with boosts.

### Describe the solution you'd like.

A change on the backend to only show posts from a user. Clients like Elk and Semaphore already have the feature on the front-end.

### Describe alternatives you've considered.

NONE

tsmethurst:
Good idea :)

PR #1680, opened 2023-04-08
[feature] Receive notification when followed account posts (if desired)

tsmethurst:
# Description

>

This pull request implements logic for creating notifications for followers when an account posts something, under the following condition:

- Follower has 'notify' set to 'true' on their follow (default false, this is opt-in).
- Status is not a reblog or a reply, new top-level posts only.
- Status is HomeTimelineable, meaning the follower would have seen it in their home timeline anyway (so, all visibility checks pass).

To support this, this PR also adds logic for updating follows. Now (in-line with the mastodon API), when you POST to the `follow` endpoint for an account you already follow, that follow will be updated instead with the `notify` and `show_reblogs` values submitted in the form. This allows users to easily change their following settings for a given account. Same goes for follow requests that have not been approved yet.

Finally, the PR does a bunch of tidying up in `internal/processing/fromcommon.go` to reduce code duplication, and simplify creation + streaming of notifications.

EDIT: ALSO! I forgot! This PR should also make it so that 'show_reblogs' actually does something now; that is, if you have `show_reblogs` set to false for an account, reblogs by that account won't be inserted into the home timeline. So that's nice :) So it also closes https://github.com/superseriousbusiness/gotosocial/issues/1671

## Checklist


tsmethurst:
Opened this as a draft for now, we should merge + test/tweak it on main only once we've released 0.8.0.

internal/processing/account/follow.go
```diff
@@ -249,3 +249,91 @@ func (p *Processor) unfollow(ctx context.Context, requestingAccount *gtsmodel.Ac
 
 	return msgs, nil
 }
+
+func (p *Processor) updateFollow(ctx context.Context, requestingAccount *gtsmodel.Account, form *apimodel.AccountFollowRequest) (*apimodel.Relationship, gtserror.WithCode) {
```
tsmethurst, line 253: Note to reviewers:

There's probably a better way of doing this, with less code duplication, but it might involve creating a new interface or doing some type checking, which feels like overkill..

internal/processing/fromclientapi.go
```diff
@@ -160,11 +160,7 @@ func (p *Processor) processCreateStatusFromClientAPI(ctx context.Context, client
 		return errors.New("note was not parseable as *gtsmodel.Status")
 	}
 
-	if err := p.timelineStatus(ctx, status); err != nil {
-		return err
-	}
-
-	if err := p.notifyStatus(ctx, status); err != nil {
+	if err := p.timelineAndNotifyStatus(ctx, status); err != nil {
```
tsmethurst, line 163: Note to reviewers:

Changed this from separate timeline + notify functions to one function that handles both, to avoid having to iterate through the local followers slice twice.

internal/processing/fromcommon.go
```diff
@@ -25,296 +25,394 @@ import (
...
-	return nil
+	return p.stream.Delete(status.ID)
 }
 
-func (p *Processor) notifyReport(ctx context.Context, report *gtsmodel.Report) error {
+/*
+	EMAIL FUNCTIONS
+*/
+
+func (p *Processor) emailReport(ctx context.Context, report *gtsmodel.Report) error {
```
tsmethurst, line 412: Note to reviewers:

Renamed these 'notifyReport' functions to explicitly mark that they don't actually create *notifications* in the database, rather, they send *emails*

internal/processing/fromcommon.go
```diff
@@ -25,296 +25,394 @@ import (
...
+}
+
+func (p *Processor) timelineAndNotifyStatusForFollowers(ctx context.Context, status *gtsmodel.Status, follows []*gtsmodel.Follow) error {
+	var (
+		errs  = make(gtserror.MultiError, 0, len(follows))
+		boost = status.BoostOfID != ""
+		reply = status.InReplyToID != "" || status.InReplyToURI != ""
+	)
+
+	for _, follow := range follows {
```
tsmethurst, line 84: Note to reviewers:

This is where the magic happens :sparkles:

internal/processing/account/follow.go
tsmethurst: nvm!

internal/processing/account/follow.go
```diff
@@ -129,6 +142,70 @@ func (p *Processor) FollowRemove(ctx context.Context, requestingAccount *gtsmode
 	Utility functions.
 */
 
+// updateFollow is a utility function for updating an existing
+// follow or followRequest with the parameters provided in the
+// given form. If nothing changes, this function is a no-op and
+// will just return the existing relationship between follow
+// origin and follow target account.
+func (p *Processor) updateFollow(
```
NyaaaWhatsUpDoc, line 150: I do love this means of reducing code reuse! Quick (subjective) suggestion, what about updating this function signature to instead be 

```golang
func updateFollow(
	ctx context.Context,
	requestingAccount *gtsmodel.Account,
	form *apimodel.AccountFollowRequest,
	showReblogs, notify *bool, update func() error,
) (*apimodel.Relationship, gtserror.WithCode) {
}
```

that way you can keep all the same logic minus the need for type switch :)

internal/processing/account/follow.go
```diff
@@ -40,24 +42,35 @@ func (p *Processor) FollowCreate(ctx context.Context, requestingAccount *gtsmode
...
+		targetAccount.ID,
+	); err != nil && !errors.Is(err, db.ErrNoEntries) {
+		err = fmt.Errorf("FollowCreate: db error checking existing follow request: %w", err)
 		return nil, gtserror.NewErrorInternalError(err)
-	} else if followRequested {
-		// Already follow requested, just return current relationship.
-		return p.RelationshipGet(ctx, requestingAccount, form.ID)
+	} else if followRequest != nil {
+		// Already requested, update if necessary + return relationship.
+		followRequest.Account = requestingAccount
```
NyaaaWhatsUpDoc, line 68: do we need to set the account here since we're not actually using it in the function below?

NyaaaWhatsUpDoc: (same goes for above)

internal/processing/fromcommon.go
```diff
@@ -25,296 +25,394 @@ import (
...
+		status.AccountID,
+		status.ID,
+	)
 }
 
-func (p *Processor) notifyFave(ctx context.Context, fave *gtsmodel.StatusFave) error {
-	// ignore self-faves
-	if fave.TargetAccountID == fave.AccountID {
-		return nil
+func (p *Processor) notify(
```
NyaaaWhatsUpDoc, line 269: is it worth updating this function to instead take account objects instead of IDs? as most of the time they'll be populated in the originating database objects (unless barebones was set) and save a secondary fetch. And optionally you _could_ also do this for the status, just to ensure the status in question actually exists, but that's a very subjective one :p

internal/processing/fromcommon.go
```diff
@@ -25,296 +25,394 @@ import (
...
+			}
 		}
-		status.BoostOf = boostedStatus
 	}
 
-	if status.BoostOfAccount == nil {
-		boostedAcct, err := p.state.DB.GetAccountByID(ctx, status.BoostOfAccountID)
-		if err != nil {
-			return fmt.Errorf("notifyAnnounce: error getting account with id %s: %s", status.BoostOfAccountID, err)
+	// delete all mention entries generated by this status
```
NyaaaWhatsUpDoc, line 351: Optionally worth doing this in another PR, but is it worth creating a `DeleteMentionsForStatus()` function here and using the same `DELETE` query with returned IDs logic as we do in the other functions along these lines?

NyaaaWhatsUpDoc:
Just a few comments, then ready to go! Nice work this looks really great :D

internal/processing/account/follow.go
tsmethurst: You're right! Think that was left over from a previous version :) Will fix this.

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1680/commits/2bcbf9195146c6b103ee559772620e158ea19df8

internal/processing/account/follow.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1680/commits/2bcbf9195146c6b103ee559772620e158ea19df8

internal/processing/fromcommon.go
```diff
@@ -25,296 +25,394 @@ import (
...
-		}, &gtsmodel.Notification{}); err == nil {
-			// notification exists already so just continue
+	return nil
+}
+
+func (p *Processor) timelineAndNotifyStatusForFollowers(ctx context.Context, status *gtsmodel.Status, follows []*gtsmodel.Follow) error {
+	var (
+		errs  = make(gtserror.MultiError, 0, len(follows))
+		boost = status.BoostOfID != ""
+		reply = status.InReplyToID != "" || status.InReplyToURI != ""
```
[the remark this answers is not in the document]
tsmethurst: Right! Nice one, changed it here :) https://github.com/superseriousbusiness/gotosocial/pull/1680/commits/440c861d9f6d736619be7e5b5ff76752dc6430b1

internal/processing/fromcommon.go
tsmethurst: Mmm i'm not sure; it felt cleaner to do it this way since it's likely the account will already be in the cache anyway by this point, so it's just retrieving it from there. By just getting it like this we can remove some of those repetitive `if x.Account == nil { get account }` checks. I don't mind changing it if you like, but for me it's OK as is!

internal/processing/fromcommon.go
tsmethurst: Agreed! I added a couple `todo's` here so we don't forget (https://github.com/superseriousbusiness/gotosocial/pull/1680/commits/5a8f864295f540f669c9e2bfde0401e047c7c881); there's probably other places where we can use such functions as well. A separate little PR for that sounds good :) Happy to do it when I'm back!

internal/processing/fromcommon.go
NyaaaWhatsUpDoc: A cache lookup is cheaper than a database lookup for sure as it involves no syscalls, but it still involves a mutex lock, re-serializing a key, hashing that key value and then searching through the multiple slice buckets in the hashmap. So not cheap compared to passing in the account model. Besides if we know the account ID being passed in here is always set we shouldn't need to do that nil check :D

tsmethurst: Haha OK, I was just about to reply but old kim 'quick hands' got there first :D

ultramookie:
Thank you for implementing this!

Issue #1574, opened 2023-02-27
[feature] data protection: do not log IPs


margau:
### Is your feature request related to a problem ?

According to the principle of data economy, it would be nice to reduce the amount of personal data stored.
Currently, IPs are logged as part of the access log.

### Describe the solution you'd like.

Add an option to obfuscate/skip IPs, alternatively only log them related to error and/or warning logs.

### Describe alternatives you've considered.

NONE

daenney:
I think we should look at this but I'd like us to consider a couple of things. Whatever we do here, it's important to note that none of this matters if you don't trust your instance admin (who incidentally also has access to the DB, the host OS etc). There's also no way for a user to verify what the admin has configured or where else these logs may still go, what (other) data sets they may be combined with that would defeat not emitting the client IP on our side entirely etc.

With that said:
* I don't think we should obfuscate IPs. If the admin has the logging of IPs on we should assume they do so for good reason and not partially obstruct their view. If someone wishes to retain logs long term the responsible thing to do is redact that information for example during log ingestion. How to redact them (for example round them up to the nearest /24) is also very dependent on need and can vary based on jurisdiction. For those reasons I don't think redaction/obfuscation is something we should take on and instead have that handled through external tools that are built for this
* I don't think turning IP address logging on or off should be based on the log level. It should be a configuration value. If I chose to explicitly set it to off I might omit things during central log ingestion like obfuscating or redacting the IP believing that that information is never contained in the logs. Having a situation that implicitly results in the IP getting logged but only under certain conditions like the log level may result in this information getting retained against the knowledge and intent of the admin
* Since we have the request ID nowadays it's very simple for someone to correlate a request from a load balancer (where they may log the client IP still) and GtS on the backend if they have a need to do so. This means that turning off logging the client IP in GtS doesn't hinder day to day operation and debugging much but does result in one less source of logs that contain IPs to worry about. That's a good thing and why I think disabling client IP logging entirely is worthwhile to implement

daenney:
I would suggest we keep the current default. It's pretty normal to log the client IP, especially in the case where GtS might be running in non-proxied mode. We could infer it based on `trusted-proxies` but that's bound to be a mess. This feels like an appropriate thing to call out in our configuration guides.

[...]

PR #1687, opened 2023-04-10
[performance] improved request batching (removes need for queueing)

NyaaaWhatsUpDoc:
# Description
- moves http client retry logic to `internal/httpclient` as this makes more sense
- start adding support for #1574, the code is in place now only need to add config flag + docs
- redo batch delivery logic to handle at most $GOMAXPROCS simultaneously
- remove http request queueing logic now that we limit concurrent batch deliveries
- moves fast fail and request ID context keys to `internal/gtscontext` package, and moves adding necessary log hooks to a separate (and well indicated) file via init function under `internal/gtscontext/log_hooks.go`

Another added benefit of slowing down batch delivery in this manner instead of the old request queueing is that you don't receive 100s of instantaneous dereference / signing requests the second you push an outgoing AP activity :D

## Checklist


tsmethurst:
Looks great!!!

PR #1699, opened 2023-04-16
Explain that markdown is enabled on the user settings page.

technomancy:
# Description

The current wording explaining the markdown feature seems to imply that the input formatting can be set on a per-post basis, which no client currently supports. It's not clear that you have to go to the user settings page to change it.

Bonus change: a few items in the FAQ were outdated as they referred to missing features which now exist!

## Checklist


tsmethurst:
Thanks! :)

PR #1714, opened 2023-04-26
[performance] replace domain block cache with an in-memory radix trie

NyaaaWhatsUpDoc:
# Description
Replaces the original domain-block cache, which was an in-memory list backed up by a faster hashmap TTL cache. This replaces that entire complex system with a singular in-memory radix tree, that searches the individual children of a node using a binary search, since we can sort the tree beforehand.

Code and dependency-wise, this is a much simpler solution. It also has the added benefit that it relies only on atomic operations, no mutexes, so it should be a little faster! It may be a bit slower during the procedure of adding new domains, but considering this cache will be 95% of the time read-from, and 5% written-to, this is an acceptable trade-off.

See: https://en.wikipedia.org/wiki/Radix_tree

## Checklist

### DIFF
--- a/internal/cache/domain/domain.go
+++ b/internal/cache/domain/domain.go
@@ -19,151 +19,206 @@ package domain
 
 import (
 	"fmt"
-	"time"
+	"strings"
+	"sync/atomic"
+	"unsafe"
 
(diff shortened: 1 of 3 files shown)


tsmethurst:
Sick, this is really cool :heart_eyes:

PR #1717, opened 2023-04-28
add From to email header

### LINKED ISSUE #1716: [bug] test email fails to send using sendgrid error 550

Whenever I use the API `/api/v1/admin/email/test` to send a test email, I get returned that it failed to send via SendGrid and returns 550 because the "From:" header was not found in the message. Looking at the code, the `assembleMessage` function in `internal/email/common.go` seems to not be writing it in the final message. I don't *think* including this header will break any other smtp options. I am happy to submit a PR to fix this.

v0.8.1

### GoToSocial Arch

amd64 binary

### What happened?

When I send a test email via the API `POST /api/v1/admin/email/test` using sendgrid as my smtp service, I would get the following error in return:

```
<div class="content">

			<main>
				<section class="error">
					<h1>An error occured:</h1>
					<pre>Unprocessable Entity: 550 MIME message is missing &#39;From&#39; header</pre>

					<div>
						<span>Request ID:</span> <code>3jv7bjc704001et5n1qg</code>
					</div>

				</section>
			</main>

		</div>
```

I would expect to receive a 200

### How to reproduce it?

Send a test email via API. I do not know if other smtp providers require this header or not, but sendgrid definitely does.

_No response_

archangelic:
# Description

This fixes an issue with email headers that prevents sending

closes #1716

## Checklist

### DIFF
--- a/internal/email/common.go
+++ b/internal/email/common.go
@@ -103,6 +103,7 @@ func assembleMessage(mailSubject string, mailBody string, mailFrom string, mailT
 		// msg headers.'
 		msg.WriteString("To: Undisclosed Recipients:;" + CRLF)
 	}
+	msg.WriteString("From: " + mailFrom + CRLF)
 	msg.WriteString("Subject: " + mailSubject + CRLF)
 	msg.WriteString(CRLF)
 	msg.WriteString(mailBody)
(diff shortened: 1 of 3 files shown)


tsmethurst:
Whoops, thank you!

PR #1721, opened 2023-04-29
[bugfix] tweak httpclient error handling again ...

NyaaaWhatsUpDoc:


internal/httpclient/client.go
```diff
@@ -266,15 +265,13 @@ func (c *Client) DoSigned(r *http.Request, sign SignFunc) (rsp *http.Response, e
...
-			(*x509.HostnameError)(nil),
-			(*x509.UnknownAuthorityError)(nil),
-		) {
-			// Non-retryable TLS errors.
-			return nil, err
-		} else if strings.Contains(err.Error(), "stopped after 10 redirects") {
-			// Don't bother if net/http returned after too many redirects
+		} else if errstr := err.Error(); // nocollapse
+		strings.Contains(errstr, "stopped after 10 redirects") ||
+			strings.Contains(errstr, "tls: ") ||
```
NyaaaWhatsUpDoc, line 270: some context for this: the `crypto/tls` package doesn't have an exported error type and even when checking for `x509` error types you can't because `tls` does shit like `return errors.New("..." + err.Error())`!! :[

PR #1722, opened 2023-04-30
[feature] Enable federation in/out of profile PropertyValue fields

tsmethurst:
# Description

>

This pull request implements federation of PropertyValue fields on profiles (those cool little key/value things that Mastodon started using, and then everyone else followed suit because they're really useful).

It does this by pulling in a new version of our go-fed fork which understands how to use PropertyValue fields.

Now, updates going out from GtS serialize fields properly, and accounts coming in also have their fields fetched and stored as appropriate.

Major bits of work:

- Add RawFields value on Account model, for storing fields before parsing to html; add database migration for this.
- Update API endpoints to parse fields as HTML after initial sanitization (allows for sharing links, mentions, etc).
- Hook in custom serializer for Activities and Objects. This was necessary to normalize account `attachment` (where the PropertyValue fields go) as an array, which is what is expected by all other implementations. I tried doing this without, but Mastodon and Akkoma (and likely other servers) don't understand cases where `attachment` is one embedded PropertyValue object, rather than single-length array.
- Update docs, tests.

## Checklist


internal/ap/normalize.go
```diff
@@ -22,12 +22,21 @@ import (
 	"github.com/superseriousbusiness/activity/streams"
 )
 
-// NormalizeActivityObject normalizes the 'object'.'content' field of the given Activity.
+/*
```
tsmethurst, line 25: Renamed these because we might want some normalize outgoing functions as well at some point, and I want it to be clear that ones here are used only for things that are coming *in*.

internal/ap/serialize.go
```diff
@@ -18,13 +18,41 @@
...
+//
+// Callers should always call this function instead of streams.Serialize, unless there's a
+// very good reason to do otherwise.
+//
+// Currently, the following things will be custom serialized:
+//
+//   - OrderedCollection: 'orderedItems' property will always be made into an array.
+//   - Any Accountable type: 'attachment' property will always be made into an array.
+//   - Update: any Accountable 'object's set on an update will be custom serialized as above.
+func Serialize(t vocab.Type) (m map[string]interface{}, e error) {
```
tsmethurst, line 41: Here's the custom serializer. It only does custom actions for a few types now, but we can extend this later as we wish.

internal/ap/serialize.go
```diff
@@ -33,32 +61,146 @@ import (
 // See:
 //   - https://github.com/go-fed/activity/issues/139
 //   - https://github.com/mastodon/mastodon/issues/24225
-func SerializeOrderedCollection(orderedCollection vocab.ActivityStreamsOrderedCollection) (map[string]interface{}, error) {
```
tsmethurst, line 36: Possible todo for this and below functions would be to extract some of the logic and put it in separate NormalizeOutgoingXYZ functions

internal/federation/federatingactor.go
```diff
@@ -59,11 +59,11 @@ type federatingActor struct {
 // implements the pub.FederatingActor interface.
 func newFederatingActor(c pub.CommonBehavior, s2s pub.FederatingProtocol, db pub.Database, clock pub.Clock) pub.FederatingActor {
 	sideEffectActor := pub.NewSideEffectActor(c, s2s, nil, db, clock)
-	customActor := pub.NewCustomActor(sideEffectActor, false, true, clock)
+	sideEffectActor.Serialize = ap.Serialize // hook in our own custom Serialize function
```
tsmethurst, line 62: This is where the custom serializer function is passed; pretty neat I think!!

internal/ap/extract.go
```diff
@@ -266,6 +266,59 @@ func ExtractSummary(i WithSummary) string {
...
+			continue
+		}
+
+		propertyValue := iter.GetSchemaPropertyValue()
+		if propertyValue == nil {
+			continue
+		}
+
+		nameProp := propertyValue.GetActivityStreamsName()
+		if nameProp == nil || nameProp.Len() != 1 {
```
NyaaaWhatsUpDoc, line 294: Is it worth accepting lengths > 1 here? i.e. `nameProp.Len() > 0` but we always just take the first value?

tsmethurst: Mmm, in this case that would mean the json object looks something like this:

```json
{
  "name": ["array","of","names","pee","pee","poo","poo"],
  "value": "whatever",
  "type": "PropertyValue"
}
```

There's no harm in taking just the first one I guess, or even concatenating the values into a single string, but I've never seen this out in the wild yet.

internal/ap/normalize.go
```diff
@@ -22,14 +22,23 @@ import (
...
+// NormalizeIncomingActivityObject normalizes the 'object'.'content' field of the given Activity.
 //
 // The rawActivity map should the freshly deserialized json representation of the Activity.
 //
-// This function is a noop if the type passed in is anything except a Create with a Statusable as its Object.
-func NormalizeActivityObject(activity pub.Activity, rawJSON map[string]interface{}) {
-	if activity.GetTypeName() != ActivityCreate {
-		// Only interested in Create right now.
+// This function is a noop if the type passed in is anything except a Create or Update with a Statusable or Accountable as its Object.
+func NormalizeIncomingActivityObject(activity pub.Activity, rawJSON map[string]interface{}) {
```
tsmethurst, line 39: This can now normalize account Update activities as well :)

internal/api/model/field.go
```diff
@@ -29,5 +29,5 @@ type Field struct {
 	Value string `json:"value"`
 	// If this field has been verified, when did this occur? (ISO 8601 Datetime).
 	// example: 2021-07-30T09:20:25+00:00
-	VerifiedAt string `json:"verified_at,omitempty"`
+	VerifiedAt *string `json:"verified_at"`
```
tsmethurst, line 32: According to the mastodon API, this should be `null` rather than omitted when not set, so changed it here.

internal/processing/fromfederator.go
```diff
@@ -370,16 +366,28 @@ func (p *Processor) processCreateFlagFromFederator(ctx context.Context, federato
...
-	// we want to fetch new bio, avatar, header, etc.
-	if _, err := p.federator.UpdateAccount(ctx,
+	// Because this was an Update, the new AP Object should be set on the message.
+	incomingAccountable, ok := federatorMsg.APObjectModel.(ap.Accountable)
+	if !ok {
+		return errors.New("Accountable was not parseable on update account message")
+	}
+
+	// Call RefreshAccount to fetch up-to-date bio, avatar, header, etc.
+	updatedAccount, err := p.federator.RefreshAccount(
```
tsmethurst, line 379: This is where that updated RefreshAccount function comes into play.

internal/validate/formvalidation.go
```diff
@@ -233,19 +233,19 @@ func ULID(i string) bool {
 	return regexes.ULID.MatchString(i)
 }
 
-func ProfileFieldsCount(fields []apimodel.UpdateField) error {
-	if length := len(fields); length > maximumProfileFields {
+func ProfileFieldsCount(fieldsLen int) error {
```
NyaaaWhatsUpDoc, line 236: Given the simplicity of this function now it might as well just be remove and the logic inlined into where it is needed :p.

Alternatively you could wrap the ProfileField() and ProfileFieldsAccount() functionality together? So if beyond max. number of fields then it returns error, else it goes through and ensures all fields are of correct length?

internal/federation/dereferencing/account.go
```diff
@@ -120,12 +125,62 @@ func (d *deref) GetAccountByUsernameDomain(ctx context.Context, requestUser stri
...
+}
+
+// fetchLatest defines a function for using a transport and uri to fetch the fetchLatest
+// version of an account (and its AP representation) from a remote instance.
+type fetchLatest func(ctx context.Context, transport transport.Transport, uri *url.URL, accountDomain string) (ap.Accountable, *gtsmodel.Account, error)
+
+// defaultFetchLatest deduplicates latest fetching code that is used in several
+// different functions. It simply calls the remote uri using the given transport,
+// parses a returned AP representation into an account, and then returns both.
+func (d *deref) defaultFetchLatest(ctx context.Context, transport transport.Transport, uri *url.URL, accountDomain string) (ap.Accountable, *gtsmodel.Account, error) {
```
NyaaaWhatsUpDoc, line 157: i have some thoughts on this, but i have my status refetch support PR open which also makes changes in this area which i'll use to address these if that's all good

internal/validate/formvalidation.go
tsmethurst: Good idea! Done :) https://github.com/superseriousbusiness/gotosocial/pull/1722/commits/0daee80e76d6fc7f0a2219ada40af696c4a2cd21

NyaaaWhatsUpDoc:
Webfinger test failed again so I merged in main to trigger a rebuild

tsmethurst:
> Webfinger test failed again so I merged in main to trigger a rebuild

to the tune of 'goldfinger'

"WEBFINGER (da da da da); it's the test -- the test that will fuck your suite! so don't run eet! no don't run eet!"

NyaaaWhatsUpDoc:
okay now that is a profile field fail @tsmethurst :sweat_smile: maybe a result of the merge it looks like

Issue #582, opened 2022-05-17
[frontend] Admin panel: 'Moderation' section for admin actions on accounts


tsmethurst:
We need a 'Moderation' section on the admin panel that contains the existing instance blocks implementation, as well as a way of letting admins trigger the `/api/v1/admin/accounts/{id}/action` endpoint.

Currently only the Suspend action is allowed through this endpoint, but this is already useful for cleaning up old accounts or suspending specific remote accounts that are causing problems.

This is blocked by https://github.com/superseriousbusiness/gotosocial/issues/581 since we need to the implement the endpoint described there first in order to make `/api/v1/admin/accounts/{id}/action` useable.

f0x52:
Working on the reports section now, so we have *some* way to get to user accounts by id, and perform actions, but would definitely be useful to have a general listing route soon(tm) as well, to make the section more useable

Issue #581, opened 2022-05-17
[feature] Admin: Allow admins to search + filter accounts


tsmethurst:
We should implement `/api/v1/admin/accounts` as seen here: https://docs.joinmastodon.org/methods/admin/

This will allow admins to more easily select accounts to take admin actions on, or to view accounts they've already blocked etc.

PR #1728, opened 2023-05-01
[frontend] Basic user moderation actions

f0x52:
# Description

Adds the most basic pages for user moderation actions (only suspend is currently supported by backend api). 
Further work is mostly blocked on #581 and #582 

## Checklist

### DIFF
--- a/web/source/settings/admin/reports/username.jsx
+++ b/web/source/settings/admin/reports/username.jsx
@@ -43,7 +43,7 @@ module.exports = function Username({ user, link = true }) {
 
 	if (link) {
 		Element = Link;
-		href = `/settings/admin/users/${user.id}`;
+		href = `/settings/admin/accounts/${user.id}`;
 	}
 
 	return (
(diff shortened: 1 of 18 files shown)


web/source/settings/admin/reports/username.jsx
```diff
@@ -41,12 +42,12 @@ module.exports = function Username({ user, link = true }) {
 	let href = null;
 
 	if (link) {
-		Element = "a";
-		href = user.account.url;
+		Element = Link;
+		href = `/settings/admin/users/${user.id}`;
```
tsmethurst, line 46: I think the path here should be 'accounts' instead of 'users', since users + accounts are separate things (user = a person with one or more accounts).

web/source/settings/user/profile.js
```diff
@@ -91,7 +91,7 @@ function UserProfileForm({ data: profile }) {
 					header={form.header.previewValue ?? profile.header}
 					display_name={form.displayName.value ?? profile.username}
 					username={profile.username}
-					role={profile.role.name}
+					role={profile.role}
```
tsmethurst, line 94: Can you explain this change?

tsmethurst (commented):
This is looking great so far :) The only thing I think is to ensure that 'account' and 'user' are represented properly in the code as two different concepts. There's quite a bit of mixing of the two here, and while it works fine code-wise, it might become confusing to reason about further down the line.

PR #1729, opened 2023-05-02
[chore] Hide vendor from diffs

daenney:
# Description

>

Linguist already classifies `vendor`/ as vendored files, meaning they don't count towards things like a repo language stat, but are still shows in diffs.

By classifying them as generated instead they're not only excluded from language stats but also hidden in diffs.

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/vendor/ linguist-generated


daenney:
Linguist seems to already do this in a more limited fasion: https://github.com/github-linguist/linguist/blob/master/lib/linguist/generated.rb#L395-L397.  Weird tld-check there too.

NyaaaWhatsUpDoc:
oh my god. so useful! you angel!

daenney:
I kinda hope this works as expected. I have a mild suspicion that what they mean with "hidden in diff" is the file diff isn't loaded but still shows in the listing during reviews, which would be annoying. It's not super well documented.

PR #1740, opened 2023-05-05
[feature] Implement preferences client API

daenney:
# Description

>

This adds the preferences endpoint to our Mastodon Client API implementation. It's a read-only endpoint that returns a number of user preferences. Applications can query these settings when logging in a user (for the first time) to configure themselves.

This came up during a debug session with @natsukagami in the help channel. Elk.zone has started using this endpoint but wasn't handling the absence of it. Elk has been updated to deal with that, but this seems useful enough to expose in general.

Relates to: https://github.com/elk-zone/elk/pull/2069

## Checklist


daenney:
So the additional Swagger gen changes look a bit... odd? Like it doesn't feel those should be there but I added them in the PR in case someone knows what's up. But I think I should drop that commit?

tsmethurst:
> So the additional Swagger gen changes look a bit... odd? Like it doesn't feel those should be there but I added them in the PR in case someone knows what's up. But I think I should drop that commit?

Yeah, I'm not sure what's going on there.... some change in the latest version of go swagger?

here's the version the ci/cd is currently using:

```
version: v0.30.4
commit: (unknown, mod sum: "h1:cPrWLSXY6ZdcgfRicOj0lANg72TkTHz6uv/OlUdzO5U=")
```

and the command for generating the docs:

```
swagger generate spec --scan-models --exclude-deps -o docs/api/swagger.yaml
```

daenney:
Ah, I might've forgotten about `--exclude-deps`, that would explain the extra diff I think. I'll try and rerun that again.

daenney:
Yup so adding `--exclude-deps` caused the extra diff to disappear. I've dropped the commit now

Issue #1662, opened 2023-03-31
[bug] Client Compatibility : Ebou Mac Client


anantshri:
### Describe the bug with a clear and concise description of what the bug is.

Ebou client https://terhech.de/ebou/ when attempting to login performs following steps correction.
1. Identifies the URL as a valid fediverse instance, 
2. Opens the authentication prompt
3. GTS provides the out of band token
4. Ebou provides an option to provide out of band token and click connect 
5. when you click connect, following error is received.
![image](https://user-images.githubusercontent.com/236843/229166491-b8142e6b-deaf-4039-ab4f-32f7553a9dae.png)

### What's your GoToSocial Version?

0.7.1

### GoToSocial Arch

amd64

### What happened?

Ebou client https://terhech.de/ebou/ when attempting to login performs following steps correction.
1. Identifies the URL as a valid fediverse instance, 
2. Opens the authentication prompt
3. GTS provides the out of band token
4. Ebou provides an option to provide out of band token and click connect 
5. when you click connect, following error is recieved.

### What you expected to happen?

user login

### How to reproduce it?

Ebou client https://terhech.de/ebou/ when attempting to login performs following steps correction.
1. Identifies the URL as a valid fediverse instance, 
2. Opens the authentication prompt
3. GTS provides the out of band token
4. Ebou provides an option to provide out of band token and click connect 
5. when you click connect, error is recieved.

daenney:
The `languages` key exists in the Instance entity in the Mastodon API since version 4.0.0 only, which was released on November 14th 2022. I'm honing in on that one since that's the only endpoint it could sensibly be checking as part of the login flow.

Ebou is likely broken with plenty of other implementations too, since it seems they're parsing the response on `/api/v1/instance` with the assumption that `languages` is present, but that's only true on `/api/v2/instance best I can tell.

The Mastodon docs are a bit weird on this as they say /api/v2/instance returns a V1:Instance which has `languages` as mandatory, but notes it was only added in 4.0.0. I think this is a confusing representation in the docs, as you can't add mandatory fields like that to V1:Instance since that's a backwards incompatible change.

This is a bug in Ebou and not something we should fix in GtS. Best I can see our responses on /api/v1/instance and /api/v2/instance are what they should be. Ebou should more permissive when it comes to parsing the responses and probably handle the absence of keys it cares about by defaulting them to something, like `languages: ['en']`.

rjp:
[...]

PR #1741, opened 2023-05-05
[bugfix] Return languages in api/v1/instance

daenney:
# Description

It turns out that in Masto v2.3.0 the languages key was added to the V1 Instance and that it's effectively mandatory. Though in GtS we don't really have this concept yet, some apps will explode if the languages key is missing altogether. So at least return the empty array on V1 too in the hopes that it makes things work well enough.

For history's sake, you can see the attributes that will get serialised in
https://github.com/mastodon/mastodon/blob/f877aa9d70d0d600961989b8e97c0e0ce3ac1db6/app/serializers/rest/v1/instance_serializer.rb#L6-L9. Because the attribute does not have a conditional defined for it, there isn't a filter that optionally omits it, or a def languages to modify the behaviour the attribute is effectively always included and serialised.

Fixes: #1662

## Checklist


tsmethurst:
Thank you! ❤️

PR #1742, opened 2023-05-06
[chore] update readthedocs deps, use conda

tsmethurst:
Fix some issues with read the docs.

Now, python version and versions of other deps are pinned using conda environment.yml; should make docs building v reproducible!


daenney:
That's rather awesome. It seems to be having some dependency resolution issues in the RTD environment :confused:

daenney (approved):

Issue #1735, opened 2023-05-03
[bug] Can't find user on domain with ü in name


moan0s:
### Describe the bug with a clear and concise description of what the bug is.

I tried to search an account on bawü.social which returns no result even though the full handle is used. I suspect the ü in the domain is responsible.

### What's your GoToSocial Version?

GoToSocial 0.8.1 git-4dbf9a2

### GoToSocial Arch

docker

### What happened?

Tried to follow @stadtwerketuebingen@bawü.social  by searching for them. Returned no account

### What you expected to happen?

See the account

### How to reproduce it?

Search for @stadtwerketuebingen@bawü.social

tsmethurst:
Ahh thanks, I think we're storing the punycode version of domains with non-ascii characters, and likely searching against that. We should convert the search domain into punycode as well. Will fix :)

PR #1743, opened 2023-05-06
[bugfix] Punycode fixes

tsmethurst:
# Description

>

This pull request does some unicode domain name fixes:

- Allow users to search for remote accounts using either punycode or unicode domain name (closes https://github.com/superseriousbusiness/gotosocial/issues/1735)
- Return unicode more often via the API, rather than the punycode representation, so you can actually see the domains people use in their original language. Applies to: `acct` property on apimodel account; apimodel mentions; domain blocks; instance peers. 
- Tweak + tidy up our regular expressions a wee bit.
- Other small fixes / cleanup.

Before:

Screenshot from 2023-05-06 13-21-17

After:

Screenshot from 2023-05-06 16-20-51

Screenshot from 2023-05-06 13-22-05

## Checklist


internal/api/client/search/searchget_test.go
```diff
@@ -142,6 +142,30 @@ func (suite *SearchGetTestSuite) TestSearchRemoteAccountByNamestringNoResolve()
...
+func (suite *SearchGetTestSuite) TestSearchRemoteAccountByNamestringSpecialCharsPunycode() {
+	query := "@üser@xn--xample-ova.org"
+	resolve := false
+
+	searchResult, err := suite.testSearch(query, resolve, http.StatusOK)
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
+
+	suite.Len(searchResult.Accounts, 1)
```
daenney, line 166: Should we add a check here to ensure we get the de-punified result?

tsmethurst: Sure! Just a sec...

daenney: Ah wait, no, that's only in the frontend that we depunify isn't it?

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1743/commits/af99ee1f44c41bc5341db7f9aa4c4c242d245a4e

tsmethurst: > Ah wait, no, that's only in the frontend that we depunify isn't it?

The search result is an API response, so counts as frontend for this purpose :)

internal/regexes/regexes.go
```diff
@@ -39,15 +38,43 @@ const (
 	follow    = "follow"
 	blocks    = "blocks"
 	reports   = "reports"
-)
 
-const (
-	maximumUsernameLength       = 64
-	maximumEmojiShortcodeLength = 30
+	schemes                  = `(http|https)://`                                                  // Allowed URI protocols for parsing links in text.
+	scriptChar               = `\p{L}\p{M}`                                                       // A single script character in any language, including chars with accents. Use inside `[]`.
```
daenney, line 43: I think this one may be wrong, the expression `\p{L}\p{M}` doesn't match the letter `a`, since we expect one, and only one, combining character after it. It also doesn't match the `à` when encoded as `U+00E0` instead of `U+0061 U+0300`

I think what we want is `\p{L}\p{M}*+`, making the combining mark optional and allow multiple ones as that's potentially possible. The `*+` makes it a possessive quantifier for optimisation reasons. I'm unsure if the Go regex library supports that though. If not we can go with only `*` I think

daenney: Wait, this works fine if you wrap the `[]` around it... why. I don't get why making it a character class suddenly makes this fine, and still doesn't match an individual diacritic mark which `\p{L}` would suggest it would do.

tsmethurst: Aha thank you! 😍 I'll fix this tomorrow and add a few more tests.

tsmethurst: fixed! i think! https://github.com/superseriousbusiness/gotosocial/pull/1743/commits/61d78fc1364f0c8cb5a57eee85d2999a34919495

https://regex101.com/r/9tjNUy/1

daenney: Yup, I think that works as expected now.

The only case that potentially remains are other numerals. So for example the Roman 3, Ⅲ, or the Arabic 25, ٢٥, aren't caught by this. Idk if that's something we want though. In most logogram based alphabets this works because unicode classifies them as letters. If that is something we want to allow we can repress the `0-9` by `\p{N}`

NyaaaWhatsUpDoc: Added this now thank you @daenney :)

NyaaaWhatsUpDoc:
webfinger error ... :thinking:

daenney:
> webfinger error ... thinking

I've seen that test error out like that exactly once on my local system and then never again. I have no idea how that happens and rerunning it will clear it. Feels like it's running into some timing issue so I suspect I missed a lock somewhere but I'm unsure.

daenney:
It's also a new error somehow. That code's been in for months and I've not seen that test fail before except now twice in a week.

Issue #1663, opened 2023-03-31
[bug] Settings/admin page does not respond into vertical mode soon enough


trysdyn:
### Describe the bug with a clear and concise description of what the bug is. Please include screenshots of any visual issues.

On viewing /settings in a browser window constrained to 1080 pixel width, the content pane overflows the browser window and results in an awkward horizontal scroll situation for text fields; also the logged in user box in the nav bar is extremely crunched to my avatar and the first two letters of my name.

Shrinking the window to about 960 pixel width will cause the responsive design to kick in and move to a more compact presentation; but in between the two, the presentation is a little strange.

This is new behavior I believe may have been introduced by some padding out of the sidebar in 0746ef7

Super duper low severity nit-pick :)

### What's your GoToSocial Version?

v0.7.1 git-60639a6

### Browser version

Firefox 111.0.1 (64-bit)

### What happened?

I pulled in head to fix the pgsql prune bug (thank you BTW!) and navigated to /settings to see the changes brought in by 0746ef7

I view and administrate GtS from a system that has a flipped orientation display, so the browser window is 1080 pixels wide. The settings frontend changes look neat, but in this orientation the main content pane overflows the window horizontally about 15%, meaning all the text fields require horizontal scrolling to fully see and the "who is logged in" element in the nav bar is extremely crunched.

### What you expected to happen?

The page moves to its compact/mobile display mode before resorting to horizontal overflow in text fields and crunching elements.

It looks like the `@media` query sets the marker line between modes at 60rem but if we want to avoid that horizontal scroll it needs to be 72-75rem (or deeper layout changes happen like shrinking the max width of the content panel).

### How to reproduce it?

View /settings in a window somewhere between 960 and 1080 pixels wide.

tsmethurst:
Aha, thank you :)

Issue #1529, opened 2023-02-18
[frogend/bug] Header is cut small on UXP-based web browsers depending on image


EchedeyLR:
### Describe the bug with a clear and concise description of what the bug is. Please include screenshots of any visual issues.

Currently, the `headerimage` is separated in a different element and not an actual background of `basic`.

This leads to the issue in which `basic` size and `headerimage` may be different, but for which `height` is dependent on `headerimage` to view the header properly of each user.

This is fixed in some web browsers by using `aspect-ratio` CSS feature, which is not supported in UXP-based web browsers by now.

This leads to a lower size picture cause `headerimage` to be too small to the point that the header is as a cut view.

### What's your GoToSocial Version?

0.7.0

### Browser version

Pale Moon 32.0.0

### What happened?

Header is cut small on UXP-based web browsers if the header picture is too small.

### What you expected to happen?

Header is properly showed with a good size on UXP-based web browsers regardless of the header picture height.

### How to reproduce it?

Just load any a profile in any UXP-based web browsers with a picture with more length that height in an high proportioned way.

### Anything else we need to know?

This may be fixed by enforcing a minimum height as showed in the following pictures:

![2023-02-18-172533_1920x1080_scrot](https://user-images.githubusercontent.com/56733813/219880092-b1d10206-382a-41fe-91e6-582604ddd1e7.png)
This is a picture without the fix.

![2023-02-18-172432_1920x1080_scrot](https://user-images.githubusercontent.com/56733813/219880105-4d2e116a-97cf-4bea-bac5-7e7ea0861f21.png)
This is a picture with the fix. Enforcing a specific height is also valid but minimum height is good enough.

EchedeyLR:
Mew...

EchedeyLR:
In my example I have a very wide wide image and this just "zooms" it making it fit as expected which exactly gives the same effect of aspect-ratio...

Vertical images should fit too...

Aspect ratio is a CSS 4 property and things worked before this feature was a thing in a different way...

;-;

handlerug:
Note to implementers: Before `aspect-ratio` was a thing, web developers enforced a constant aspect ratio using a quirk of the `padding-top` property:

```html
<div class="header-wrapper">
  <div class="header">
    Here's some stuff
  </div>
</div>

<style>
.header-wrapper {
  position: relative;
  /* this is relative to the width, not height */
  padding-top: 56.25%; /* 9/16 * 100 */
}

.header {
  position: absolute;
[...]

Issue #714, opened 2022-07-18
[frontend] Accessibility updates


tsmethurst:
Some pointers we should keep in mind when updating our html and css, from @Ryuno-Ki (following things are copy-pasted from fedi messages):

---

## Colors + contrast

Just checked goblin.technology with WAVE, Nu Validator and Orca screenreader.

Looks about perfect.

## HTML

<div className="short-description"><p>Powerful and Dangerous Goblin Tech (GoToSocial semi-public instance).</p></div>

still contains className (Ctrl-U to check server response)

Plus, there's an <a name> in the footer pointing to the git source.

WAVE tool lists several images missing alt attribute (decorative ones have alt="")

Plus, https://html5.validator.nu/?doc=https%3A%2F%2Fgts.superseriousbusiness.org%2F%40gotosocial&showimagereport=yes reports duplicated ID values.

## Navigation

Those action icons aren't focusable with keyboard (tabindex="0" missing?).

I miss skip links to jump straight to the timeline:
https://webaim.org/techniques/skipnav/

## Screenreaders

Listing to it with Orca screenreader makes me notice, that replies/boosts/favourites just read out the number. There is context missing about what the value is about.

You can utilise a technique such as https://webaim.org/techniques/css/invisiblecontent/ to provide text to screenreaders only.

Thinking about the narrative order (= DOM one) it might make sense to have the date (for toots) read out earlier?

If you turn .toot.expanded into a Flexbox, you can use the order property to push the .info down (but have it appear before .contentgrid in the DOM).

Be aware, that visual order differing from the DOM one is debated in the accessibility community.

bremensaki:
All the project's web views are currently only presented in a light-on-dark theme, which is an accessibility problem I don't see mentioned here.

Is checking a client's prefers-color-scheme setting and switching to suit on the roadmap anywhere?

tsmethurst:
> Is checking a client's prefers-color-scheme setting and switching to suit on the roadmap anywhere?

Not yet, I'll make an issue for it

tsmethurst:
I think the points in the original issue have now been addressed, especially since #2480

PR #1764, opened 2023-05-11
[frontend] Profiles with fields & more

f0x52:
# Description

Redesigned the profile template, including showing profile fields, and re-organized header, bio, pinned posts etc.
More accessibility review for profile and status templates, resulting in more semantic use of `<details>/<summary>` and other tweaks. 

fixes https://github.com/superseriousbusiness/gotosocial/issues/1529
edit: fixes https://github.com/superseriousbusiness/gotosocial/issues/1663

and most of https://github.com/superseriousbusiness/gotosocial/issues/714 which could use an update / split into remaining issues

image


f0x52:
looking at https://github.com/superseriousbusiness/gotosocial/issues/1663 and probably gonna sneak in a fix for that too

web/template/profile.tmpl
```diff
@@ -18,80 +18,127 @@
...
+
+			<div class="bio">
+				{{ if .account.Note }}
+				{{emojify .account.Emojis (noescape .account.Note)}}
+				{{else}}
+				This GoToSocial user hasn't written a bio yet!
+				{{end}}
+			</div>
+
+			<div class="sr-only" role="group">
```
tsmethurst, line 81: Can you explain these two very similar divs? Not sure I understand what's going on here.

web/template/status.tmpl
```diff
@@ -17,71 +17,114 @@
...
-			<a href="{{.URL}}"
-				 target="_blank"
-				 {{if .Description}}title="{{.Description}}"{{end}}
-				 data-pswp-width="{{.Meta.Original.Width}}px"
-				 data-pswp-height="{{.Meta.Original.Height}}px"
-				 {{if eq .Type "video"}}data-pswp-type="video"{{end}}
-				 data-cropped="true">
-				<img src="{{.PreviewURL}}" {{if .Description}}alt="{{.Description}}"{{end}} data-blurhash="{{.Blurhash}}"/>
-			</a>
+			<!-- {{if not .Description}}
```
tsmethurst, line 58: This should probs be either uncommented or removed.

tsmethurst (commented):
Great!!!

web/template/profile.tmpl
f0x52: it's a bit of a hack to make that section read better, as visually you need the grid to have separate elements for each column entry, but that makes them all separately focussed/read. So the first div is only read for screen readers, second only shown visually

web/template/status.tmpl
f0x52: Removed for now. We might want to re-introduce the "this media is lacking a description" indicator at some point, but it doesn't really work well when there's multiple media on a post, and even worse on mobile where it obscures a large area

web/source/css/profile.css
```diff
@@ -16,274 +16,252 @@
...
-	a img {
-		height: 5rem;
-	}
-}
-
-main {
-	background: transparent;
-	padding-top: 0;
+.page {
+	grid-template-columns: 1fr minmax(auto, 60rem) 1fr;
```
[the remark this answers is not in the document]
f0x52: yep, added the comment

web/template/profile.tmpl
tsmethurst: Aha okay, i suspected something like that but wasn't sure!

PR #1771, opened 2023-05-12
[frogend/chore] fix dynamicSpoiler elements

f0x52:
# Description

Another small oversight.. made the dynamic renaming of "show more"/"show less" CW buttons not work, and similar for the media spoiler eye icon label


tsmethurst:
Well spotted :eyes:

PR #1773, opened 2023-05-12
[bugfix] Downstep otel to fix freebsd compile issue

tsmethurst:
https://github.com/open-telemetry/opentelemetry-go/issues/4076


NyaaaWhatsUpDoc:
Note: once they make a release with https://github.com/open-telemetry/opentelemetry-go/commit/7dea2225a218872e86d2f580e82c089b321617b0 we should step our version back up again. I'm hoping that will be very soon so we can get the updated version into the v0.9.0 release.

PR #1799, opened 2023-05-16
[feature] Make client IP logging configurable

daenney:
# Description

Admins can now chose whether client IPs should be present in the logs. By default this is enabled in order to retain the current behaviour and because when running in non-proxied mode this is something admins probably want.

When having an LB in front of GtS this can be disabled as requsets can still be correlated between LB and GtS using the Request ID or the tracing facilities.

Fixes: #1574

## Checklist

### DIFF
--- a/internal/gtscontext/log_hooks.go
+++ b/internal/gtscontext/log_hooks.go
@@ -34,7 +34,7 @@ func init() {
 		}
 		return kvs
 	})
-	// Client IP middleware hook.
+	// Public Key ID middleware hook.
 	log.Hook(func(ctx context.Context, kvs []kv.Field) []kv.Field {
 		if id := PublicKeyID(ctx); id != "" {
 			return append(kvs, kv.Field{K: "pubKeyID", V: id})
--- a/internal/middleware/logger.go
+++ b/internal/middleware/logger.go
@@ -31,7 +31,7 @@ import (
 )
 
 // Logger returns a gin middleware which provides request logging and panic recovery.
-func Logger() gin.HandlerFunc {
+func Logger(logClientIP bool) gin.HandlerFunc {
 	return func(c *gin.Context) {
 		// Initialize the logging fields
 		fields := make(kv.Fields, 5, 7)
@@ -72,10 +72,7 @@ func Logger() gin.HandlerFunc {
 			fields[2] = kv.Field{"method", c.Request.Method}
 			fields[3] = kv.Field{"statusCode", code}
 			fields[4] = kv.Field{"path", path}
-			if includeClientIP := true; includeClientIP {
-				// TODO: make this configurable.
-				//
-				// Include clientIP if enabled.
+			if logClientIP {
 				fields = append(fields, kv.Field{
 					"clientIP", c.ClientIP(),
 				})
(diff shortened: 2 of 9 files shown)


internal/middleware/logger.go
```diff
@@ -31,7 +31,7 @@ import (
 )
 
 // Logger returns a gin middleware which provides request logging and panic recovery.
-func Logger() gin.HandlerFunc {
+func Logger(logClientIP bool) gin.HandlerFunc {
```
daenney, line 34: I opted for passing this into the middleware instead of doing the `log.Hook` because `log.Hook` only takes a `context.Context`. That means I'd have to store the clientIP as a context key on the `gtscontext` instead and add methods to add/retrieve it there.

That seemed a bit silly since we already have it right here in the Gin context. I'm open to suggestions on how to change it to make it use the hooks system instead.

internal/middleware/logger.go
```diff
@@ -72,10 +72,7 @@ func Logger() gin.HandlerFunc {
 			fields[2] = kv.Field{"method", c.Request.Method}
 			fields[3] = kv.Field{"statusCode", code}
 			fields[4] = kv.Field{"path", path}
-			if includeClientIP := true; includeClientIP {
-				// TODO: make this configurable.
-				//
-				// Include clientIP if enabled.
+			if logClientIP {
```
tsmethurst, line 75: With this way of doing things, you now check every request whether logClientIP is set. For performance reasons, it's probably better to check this just once when creating the middleware, and return a different middleware depending on whether it's set or not, something like:

```
func Logger(logClientIP bool) gin.HandlerFunc {
  if logClientIP {
    return middleware that logs client ip
  }
  return middleware that doesn't log client ip
}
```

internal/middleware/logger.go
tsmethurst: that seems fine :) it's in line with what we do elsewhere in any case!

internal/middleware/logger.go
daenney: Mmm, that's a good point. I actually wonder if the Go compiler realises that the boolean is basically static and may compile it out entirely. But that's an investigation for another day.

daenney: I feel like if we do the check before returning the handler we're going to end up with tons and tons of duplication based on this one conditional. Maybe I'm missing something obvious? This almost gets me back to pushing the clientIP on the gtscontext instead as that would avoid this...

tsmethurst: I think you can probably skip a lot of the duplication by just passing in a function to the middleware instantiation that just creates those log fields, if you see what I mean

tsmethurst: (also I think you can hint to the compiler using `const` but check with @NyaaaWhatsUpDoc on that because she knows way more about such things than i do!)

internal/gtscontext/log_hooks.go
```diff
@@ -34,7 +34,7 @@ func init() {
 		}
 		return kvs
 	})
-	// Client IP middleware hook.
+	// Public Key ID middleware hook.
```
NyaaaWhatsUpDoc, line 37: you can tell what i had on my mind here :')

internal/middleware/logger.go
NyaaaWhatsUpDoc: Honestly I think the singular if statement here should be fine, the only loop it will be a part of is the main accept-request-loop which relatively will have much muuuuuuch more logic doing everything else. (also, alternatively by passing in a function argument it will then have to do an if check on that each time it calls it if you don't manually do it yourself, since the runtime has to do that check in order to catch and panic with a "nil ptr" message :p)

PR #1817, opened 2023-05-23
docs: code highlighting configuration

decentral1se:
# Description

Before:
before

After:
after

## Checklist

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -35,6 +35,13 @@ extra_css:
 
 markdown_extensions:
   - admonition
+  - pymdownx.highlight:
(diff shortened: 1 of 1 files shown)


decentral1se:
@tsmethurst no new deps! cool, didn't see `docs/environment.yml` but that's exactly what i installed to get things up and running. so, yeh, i think this is just out-of-the-box functionality from `mkdocs-material`.

daenney (approved):
mkdocs-material declares a dependency on pygments so we're good to go here. No other tweaking aside from the config options is necessary.

PR #1821, opened 2023-05-23
[bugfix/chore] Inbox post updates

tsmethurst:
# Description

>

This PR updates some of our InboxPost logic, with the following changes:

- If something goes wrong inside the federator's `PostInboxScheme` function, *always* return an errWithCode back up to `InboxPost` so that it can be handled appropriately.
- Ignore the bool return value of `PostInboxScheme`, since it's not really that helpful for us. Now we just assume if there's no error then it's fine and we should return Accepted; if there is an error, use that instead.
- Other small bits of tidying up, improve InboxPost tests (so we can more easily add more).

## Checklist


tsmethurst:
IMO this is another one to consider merging *after* 0.9.0 is out, so we can try it out a bit. It might need some tweaking, I don't wanna put it in a release just yet.

internal/federation/federatingactor.go
```diff
@@ -67,99 +70,180 @@ func newFederatingActor(c pub.CommonBehavior, s2s pub.FederatingProtocol, db pub
...
+	}
+
+	// Request is now undergoing processing. Caller
+	// of this function will handle writing Accepted.
+	return true, nil
+}
+
+// contentType is a util function for ensuring valid
+// content-type of incoming ActivityPub POST requests.
+func contentType(contentType string) gtserror.WithCode {
```
NyaaaWhatsUpDoc, line 190: I was going to suggest that renaming this function might make it a little more self-explanatory as to what it is doing, but I took a look at the `activityStreamsMediaTypes` and realised we're probably not doing this check too elegantly with the existing iterate through all possibilities in the slice :p

I hope you don't mind I quickly just mocked up a possible alternative to keeping that slice of all possible permutations around, see what you think :)
```golang
// isASMediaType will return whether the given content-type string
// matches one of the 2 possible ActivityStreams incoming content types:
// - application/activity+json
// - application/ld+json;profile=https://w3.org/ns/activitystreams
//
// Where for the above we are leniant with whitespace and quotes.
func IsASMediaType(ct string) bool {
	var (
		// First content-type part,
		// contains the application/...
		p1 string

		// Second content-type part,
		// contains AS IRI if provided
		p2 string
	)

	// Split content-type by semi-colon.
	sep := strings.IndexByte(ct, ';')
	p1 = ct[:sep]
	if sep >= 0 {
		p2 = ct[sep+1:]
	}

	// Trim any ending space from the
	// main content-type part of string.
	p1 = strings.TrimRight(p1, " ")

	switch p1 {
	case "application/activity+json":
		return p2 == ""

	case "application/ld+json":
		// Trim all start/end space.
		p2 = strings.Trim(p2, " ")

		// Drop any quotes around the URI str.
		p2 = strings.ReplaceAll(p2, "\"", "")

		// End part must be a ref to the main AS namespace IRI.
		return p2 == "profile=https://www.w3.org/ns/activitystreams"

	default:
		return false
	}
}
```

And if you do like that I wrote a test to ensure the old values are still accepted (maybe worth adding some negative results in there too)
```golang
func TestIsASMediaType(t *testing.T) {
	for _, test := range []struct {
		Input  string
		Expect bool
	}{
		{
			Input:  "application/activity+json",
			Expect: true,
		},
		{
			Input:  "application/ld+json;profile=https://www.w3.org/ns/activitystreams",
			Expect: true,
		},
		{
			Input:  "application/ld+json;profile=\"https://www.w3.org/ns/activitystreams\"",
			Expect: true,
		},
		{
			Input:  "application/ld+json ;profile=https://www.w3.org/ns/activitystreams",
			Expect: true,
		},
		{
			Input:  "application/ld+json ;profile=\"https://www.w3.org/ns/activitystreams\"",
			Expect: true,
		},
		{
			Input:  "application/ld+json ; profile=https://www.w3.org/ns/activitystreams",
			Expect: true,
		},
		{
			Input:  "application/ld+json ; profile=\"https://www.w3.org/ns/activitystreams\"",
			Expect: true,
		},
		{
			Input:  "application/ld+json; profile=https://www.w3.org/ns/activitystreams",
			Expect: true,
		},
		{
			Input:  "application/ld+json; profile=\"https://www.w3.org/ns/activitystreams\"",
			Expect: true,
		},
	} {
		if IsASMediaType(test.Input) != test.Expect {
			t.Errorf("did not get expected result %v for input: %s", test.Expect, test.Input)
		}
	}
}
```

NyaaaWhatsUpDoc: (I haven't actually ran this test yet so I apologize if this function fails straight off-the-bat :see_no_evil:)

internal/api/activitypub/users/inboxpost.go
```diff
@@ -19,32 +19,32 @@ package users
...
-		apiutil.ErrorHandler(c, gtserror.NewErrorBadRequest(err, err.Error()), m.processor.InstanceGetV1)
-		return
-	}
-
-	if posted, err := m.processor.Fedi().InboxPost(apiutil.TransferSignatureContext(c), c.Writer, c.Request); err != nil {
-		if withCode, ok := err.(gtserror.WithCode); ok {
-			apiutil.ErrorHandler(c, withCode, m.processor.InstanceGetV1)
+	_, err := m.processor.Fedi().InboxPost(apiutil.TransferSignatureContext(c), c.Writer, c.Request)
+	if err != nil {
+		if errWithCode := new(gtserror.WithCode); errors.As(err, errWithCode) {
```
NyaaaWhatsUpDoc, line 35: Might be a little neater to tidy up these if else clauses with
```golang
if err != nil {
    var errWithCode *gtserror.WithCode
    
    if !errors.As(err, errWithCode) {
        / Something else went wrong, and someone forgot to return
	// an errWithCode! It's chill though. Log the error but don't
	// return it to the caller, to avoid leaking internals.
	log.WithContext(c.Request.Context()).Errorf("returning Bad Request to caller, err was: %q", err)
	errWithCode = gtserror.NewErrorBadRequest(err)
    }
    
    // Pass along confirmed error with code to the main error handler
    apiutil.ErrorHandler(c, *errWithCode, m.processor.InstanceGetV1)
}

// Inbox POST body was Accepted for processing.
c.JSON(http.StatusAccepted, gin.H{"status": http.StatusText(http.StatusAccepted)})
```

Issue #1412, opened 2023-02-03
[bug] Clear-text direct messages are logged when log-level is either info or debug


nachtlied1:
### Describe the bug with a clear and concise description of what the bug is.

Direct message content is logged in clear-text to `GoToSocial` application logs when the log level is set to `INFO` or `DEBUG`. This appears to happen for the following `func` / `level` combinations at minimum (there may be more, but these are the ones I've observed so far):

```
func=processing.(*processor).ProcessFromClientAPI level=INFO

func=federation.(*federatingActor).Send level=INFO

func=federatingdb.(*federatingDB).NewID level=DEBUG

func=federatingdb.(*federatingDB).Create level=DEBUG

func=processing.(*processor).ProcessFromFederator level=INFO
```

On an instance with more than one user, this effectively means that anyone with access to the application logs could read other users' direct messages in clear-text just by accessing and reading the logs.

### What's your GoToSocial Version?

0.6.0 git-f9e5ec9

### GoToSocial Arch

arm64 Docker

### Browser version

All

### What happened?

While reviewing my server's GoToSocial application logs, I noticed that my direct messages (both sent and received) were logged in clear-text at different log levels (`INFO` and `DEBUG`).

Example log entry for a received DM (clear-text DM content replaced by `[REDACTED DIRECT MESSAGE TEXT]`):
```
[...]

PR #1829, opened 2023-05-25
[bugfix] don't INFO level log DMs

decentral1se:
# Description

Only log DMs on DEBUG log level. I've tested this locally following the steps to reproduce in https://github.com/superseriousbusiness/gotosocial/issues/1412 and see no other INFO level logging of DMs after this change. From my reading of #1412 this is deemed the appropriate solution for this issue. Closes https://github.com/superseriousbusiness/gotosocial/issues/1412.

## Checklist

### DIFF
--- a/internal/federation/federatingactor.go
+++ b/internal/federation/federatingactor.go
@@ -68,7 +68,7 @@ func newFederatingActor(c pub.CommonBehavior, s2s pub.FederatingProtocol, db pub
 }
 
 func (f *federatingActor) Send(c context.Context, outbox *url.URL, t vocab.Type) (pub.Activity, error) {
-	log.Infof(c, "send activity %s via outbox %s", t.GetTypeName(), outbox)
+	log.Debugf(c, "send activity %s via outbox %s", t.GetTypeName(), outbox)
 	return f.wrapped.Send(c, outbox, t)
 }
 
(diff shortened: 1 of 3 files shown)


internal/federation/federatingactor.go
```diff
@@ -68,7 +68,7 @@ func newFederatingActor(c pub.CommonBehavior, s2s pub.FederatingProtocol, db pub
 }
 
 func (f *federatingActor) Send(c context.Context, outbox *url.URL, t vocab.Type) (pub.Activity, error) {
-	log.Infof(c, "send activity %s via outbox %s", t.GetTypeName(), outbox)
+	log.Debugf(c, "send activity %s via outbox %s", t.GetTypeName(), outbox)
```
tsmethurst, line 71: I don't think this one needs to change, it's only logging the type of activity being sent (Create, Undo, Like, etc), not the activity itself.

tsmethurst:
Hmm. I read through the original issue again and it's already the case that we don't log detailed object model stuff when configured `log-level` is `info` or less (https://github.com/superseriousbusiness/gotosocial/issues/1412#issuecomment-1562920865). So this PR is unnecessary :) Thanks for drawing attention to the issue though!

PR #1836, opened 2023-05-28
[feature] Add partial text search for accounts + statuses

tsmethurst:
# Description

>

This pull request updates our search functionality to add partial text search using the `LIKE` functionality from SQLite and Postgres.

Previously, users could only search directly by URL or namestring.

With these changes, users can now also search using arbitrary text strings for:

- Accounts:
  - Accounts they follow (search in account username, account display name, account note/bio)
  - Accounts they may or may not follow (search in account username, account display name)
- Statuses (search in status content warning and body)
  - Statuses requesting account posted themself.
  - Direct replies to statuses posted by requesting account.

**This is absolutely NOT arbitrary full text search; it functions more like Mastodon currently does when you have elasticsearch enabled, though it doesn't include statuses from others that you've only liked or boosted. You can't use this to search for random accounts or statuses across the fediverse (or even on your instance) using keywords.**

Since we still have no good way of supporting offset, paging is supported using max_id and min_id parameters when a specific type of result is searched for.

Search tends to resolve fairly quickly; on goblin.technology when searching for 'test' or 'boobs' or whatever else, it usually resolves in less than a second. Quite useable, in other words!

API endpoints touched:

- `/api/v1/search`
- `/api/v2/search`

API endpoints added:

- `/api/v1/accounts/search` - https://docs.joinmastodon.org/methods/accounts/#search
- `/api/v1/accounts/lookup` - https://docs.joinmastodon.org/methods/accounts/#lookup

Closes https://github.com/superseriousbusiness/gotosocial/issues/1848

## Checklist


tsmethurst:
Gonna convert this to a draft for now while I add in some other useful things.

tsmethurst:
Alright, think it's good to go now :)

internal/processing/search/lookup.go
```diff
@@ -0,0 +1,113 @@
...
+// See https://docs.joinmastodon.org/methods/accounts/#lookup
+func (p *Processor) Lookup(
+	ctx context.Context,
+	requestingAccount *gtsmodel.Account,
+	query string,
+) (*apimodel.Account, gtserror.WithCode) {
+	// Validate query.
+	query = strings.TrimSpace(query)
+	if query == "" {
+		err := errors.New("search query was empty string after trimming space")
```
NyaaaWhatsUpDoc, line 51: `gtserror.New` ;)

internal/processing/search/lookup.go
```diff
@@ -0,0 +1,113 @@
...
+		WithContext(ctx).
+		WithFields(kv.Fields{
+			{"query", query},
+		}...).
+		Debugf("beginning search")
+
+	// See if we have something that looks like a namestring.
+	username, domain, err := util.ExtractNamestringParts(query)
+	if err != nil {
+		err := errors.New("bad search query, must in the form '[username]' or '[username]@[domain]")
```
NyaaaWhatsUpDoc, line 72: and again ;)

internal/processing/search/lookup.go
```diff
@@ -0,0 +1,113 @@
...
+
+	account, err := p.accountByUsernameDomain(
+		ctx,
+		requestingAccount,
+		username,
+		domain,
+		false, // never resolve!
+	)
+	if err != nil {
+		if errors.As(err, new(*dereferencing.ErrNotRetrievable)) {
```
NyaaaWhatsUpDoc, line 84: One thing you can do here instead is `errorsv2.Assignable(err, (*dereferencing.ErrNotRetrievable)(nil))` using the import  `errorsv2 "codeberg.org/gruf/go-errors/v2"`. That way there's no need for the error allocation

internal/processing/search/lookup.go
```diff
@@ -0,0 +1,113 @@
...
+	// may cause the account to be filtered out if it's not
+	// visible to the caller, so anticipate this.
+	accounts, errWithCode := p.packageAccounts(ctx, requestingAccount, []*gtsmodel.Account{account})
+	if errWithCode != nil {
+		return nil, errWithCode
+	}
+
+	if len(accounts) == 0 {
+		// Account was not visible to the requesting account.
+		err := fmt.Errorf("%s not found", query)
```
NyaaaWhatsUpDoc, line 107: `gtserror.Newf` ;)

internal/processing/search/get.go
```diff
@@ -0,0 +1,696 @@
...
+		offset    = req.Offset
+		query     = strings.TrimSpace(req.Query)                      // Trim trailing/leading whitespace.
+		queryType = strings.TrimSpace(strings.ToLower(req.QueryType)) // Trim trailing/leading whitespace; convert to lowercase.
+		resolve   = req.Resolve
+		following = req.Following
+	)
+
+	// Validate query.
+	if query == "" {
+		err := errors.New("search query was empty string after trimming space")
```
NyaaaWhatsUpDoc, line 79: `gtserror.New` ;)

internal/processing/search/get.go
```diff
@@ -0,0 +1,696 @@
...
+		return nil, gtserror.NewErrorBadRequest(err, err.Error())
+	}
+
+	// Validate query type.
+	switch queryType {
+	case queryTypeAny, queryTypeAccounts, queryTypeStatuses, queryTypeHashtags:
+		// No problem.
+	default:
+		err := fmt.Errorf(
```
NyaaaWhatsUpDoc, line 88: `gtserror.Newf` ;)

internal/db/bundb/search.go
```diff
@@ -0,0 +1,411 @@
...
+	statusText := s.conn.NewSelect()
+
+	// SQLite and Postgres use different
+	// syntaxes for concatenation.
+	switch s.conn.Dialect().Name() {
+
+	case dialect.SQLite:
+		statusText = statusText.ColumnExpr(
+			"LOWER(? || COALESCE(?, ?)) AS ?",
+			bun.Ident("status.content"), bun.Ident("status.content_warning"), "",
```
NyaaaWhatsUpDoc, line 397: Is this the most efficient way to do this concatenating the strings? and not instead doing searches over either status content warning _or_ content? Might be worth doing an explain over this query vs searching over each separately to see which produces the better execution plan.

internal/db/bundb/search.go
```diff
@@ -0,0 +1,411 @@
...
+const (
+	escapeChar    = "\\"
+	zeroOrMore    = "%"
+	exactlyOne    = "_"
+	escEscapeChar = escapeChar + escapeChar
+	escZeroOrMore = escapeChar + zeroOrMore
+	escExactlyOne = escapeChar + exactlyOne
+)
+
+var replacer = strings.NewReplacer(
```
NyaaaWhatsUpDoc, line 73: i'd maybe just make these the strings just string literals within this here instead of constants, it makes it easier to see what exactly is being replaced. also maybe worth moving the function that uses this replacer to be at the top here so you can know right off-the-bat what the function is replacing from within 1 glance

internal/db/bundb/search.go
```diff
@@ -0,0 +1,411 @@
...
+	if following {
+		// Select only from accounts followed by accountID.
+		q = q.Where(
+			"? IN (?)",
+			bun.Ident("account.id"),
+			s.followedAccountIDs(accountID),
+		)
+	}
+
+	// Concatenate account username, displayname,
```
NyaaaWhatsUpDoc, line 146: Same goes here as for the status text query. Is concatenating definitely producing the best execution plan here?

internal/db/bundb/search.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+
+	if len(accountIDs) == 0 {
+		return nil, nil
+	}
+
+	// If we're paging up, we still want accounts
+	// to be sorted by ID desc, so reverse ids slice.
+	// https://zchee.github.io/golang-wiki/SliceTricks/#reversing
+	if !frontToBack {
```
NyaaaWhatsUpDoc, line 180: Supporting both directions of paging in a singular function definitely sounds like it could lead to hard to issues in the future. Is paging in both directions definitely needed / something that the mastodon API supports?

tsmethurst: It is yeah unfortunately. It's something we weren't supporting properly before a couple of months ago, and it caused problems with clients like Tusky. When we started GtS I *think* the mastodon API was less clear what the difference between since_id and min_id was, but it's fairly unambiguous now that min_id means paging upwards.

internal/db/bundb/search.go
tsmethurst: This is what I get for this query with `.expert` set:

```
sqlite>  SELECT "account"."id" FROM "accounts" AS "account" WHERE (("account"."domain" IS NULL) OR ("account"."domain" != "account"."username")) AND ("account"."id" < 'ZZZZZZZZZZZZZZZZZZZZZZZZZZ') AND ((SELECT LOWER("account"."username" || COALESCE("account"."display_name", '')) AS "account_text") LIKE '%turtle%' ESCAPE '\') ORDER BY "account"."id" DESC LIMIT 10;
(no new indexes)

SEARCH account USING INDEX accounts_id_idx (id<?)
CORRELATED SCALAR SUBQUERY 1
SCAN CONSTANT ROW
```

and then here's from EXPLAIN QUERY PLAN:

Screenshot from 2023-06-17 17-59-43

internal/db/bundb/search.go
tsmethurst: Here's the `.expert` output for this query:

```
sqlite> SELECT "status"."id" FROM "statuses" AS "status" WHERE (("status"."account_id" = '01F8MH1H7YV1Z7D2C8K2730QBF') OR ("status"."in_reply_to_account_id" = '01F8MH1H7YV1Z7D2C8K2730QBF')) AND ("status"."boost_of_id" IS NULL) AND ("status"."id" < 'ZZZZZZZZZZZZZZZZZZZZZZZZZZ') AND ((SELECT LOWER("status"."content" || COALESCE("status"."content_warning", '')) AS "status_text") LIKE '%hello%' ESCAPE '\') ORDER BY "status"."id" DESC LIMIT 10;
CREATE INDEX statuses_idx_3baca464 ON statuses(in_reply_to_account_id, id DESC);

MULTI-INDEX OR
INDEX 1
SEARCH status USING INDEX statuses_account_id_id_idx (account_id=? AND id<?)
INDEX 2
SEARCH status USING INDEX statuses_idx_3baca464 (in_reply_to_account_id=? AND id<?)
CORRELATED SCALAR SUBQUERY 1
SCAN CONSTANT ROW
USE TEMP B-TREE FOR ORDER BY
```

So we might wanna consider rejigging either creating that autocreated index `(in_reply_to_account_id, id DESC)`, and i think that will get rid of the temp b-tree for order by...

internal/processing/search/lookup.go
tsmethurst: Marking this and the other gtserror comments as resolved; we talked about it in chat:

> ah, kimb I remember now (looking at it again) the reason why I use gtserror.New(f) in some places, and in others just normal errors: if an error is being inserted in the 'safe' part of a gtserror.WithCode it gets returned to the API caller, and I didn't necessarily want them to see the function name
> so i tend to use gtserror only when it's an internal error or when there's no 'safe' or 'help text' error being provided

internal/db/bundb/search.go
tsmethurst: To me this one at least looks OK, at least I don't see anything immediately bad in it, but I admin I'm not great at reading these outputs. @VyrCossont could you possibly give this a once over?

internal/processing/search/lookup.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/1836/commits/bce9a50cba24973934f55f83a2f2744e6d9d58ea

internal/db/bundb/search.go
tsmethurst: Alright! How's this? https://github.com/superseriousbusiness/gotosocial/pull/1836/commits/2f4ca54c8e13c6b539e87db7af5112812b778331

internal/db/bundb/search.go
tsmethurst: OK so the temp b-tree order by doesn't go away in this case, awkwardly, so I'll take a look-see at rejigging the query.

tsmethurst: Added the index here: https://github.com/superseriousbusiness/gotosocial/pull/1836/commits/789ec012b9ef00196edd4bcef8f39f0f3c9ccfca

I *think* maybe the temp b-tree for order by is not *too* bad, since we should only be ordering 10 statuses, but I'm really not sure tbh.

PR #1839, opened 2023-05-29
[bugfix] only attempt to populate account/statuses from DB if already exist

NyaaaWhatsUpDoc:
# Description
Only populate account / status models in the dereferencer when we *know* it is an *existing* up-to-date model, otherwise errors may be returned for related models not yet dereferenced.

## Checklist

### DIFF
--- a/internal/federation/dereferencing/status.go
+++ b/internal/federation/dereferencing/status.go
@@ -116,6 +116,15 @@ func (d *deref) getStatusByURI(ctx context.Context, requestUser string, uri *url
 		}, nil)
 	}
 
+	// Check whether needs update.
+	if statusUpToDate(status) {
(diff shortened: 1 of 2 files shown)


internal/federation/dereferencing/status.go
```diff
@@ -107,6 +116,15 @@ func (d *deref) getStatusByURI(ctx context.Context, requestUser string, uri *url
 		}, nil)
 	}
 
+	// Check whether needs update.
+	if statusUpToDate(status) {
```
NyaaaWhatsUpDoc, line 120: we should probably make a v0.9.1 hotfix release with this PR cherry-picked, on account of the fact that without this status up-to-date check here instances are very likely refetching status content far more often than they should :')

PR #1853, opened 2023-06-01
[bugfix/chore] oauth entropy fix + media cleanup tasks rewrite

NyaaaWhatsUpDoc:
## Description
- creates new `cleaner` package for filesystem / database cleaning / fixing tasks
- moves media cleaning to `cleaner` package, mostly rewritten in the process
- combines multiple of the pruning jobs as the logic for local + remote is  (near enough) the same
- adds media cleanup task to ensure media entry cache states are in sync with storage
- fixes entropy generation for new "router sessions" (oauth registration flow state key generation)
- much more thorough removal on delete of media attachments from database

## Still needed
~~- finish tests fixing move media cleaning tests (combine local + remote prunes)~~
~~- figure out the testsuite emoji files having a fixed instance ID unrelated to test suite generated one~~
~~- test function for the new media FixCacheStates() cleaner function~~
~~- migration script to regenerate "router session" entropy~~
~~- enable passing in activitypub raw object to enrichAccount again (? this may be source of the missing profile image bug)~~
~~- figure out actual cause of the missing profile pictures bug (maybe old media cleanup related)~~
~~- move the emoji refetch code over to the `cleaner` package also~~
~~- cleanup the db.Media{} GetAttachments___() code, the changes here haven't be fully code commented yet and there's probably some overlapping / now unused functions that can be removed~~

## Checklist


NyaaaWhatsUpDoc:
This is the code I was discussing in matrix @tsmethurst, both the general little entropy fix and at least a solution for the missing profile images bug, even if not necessarily addressing the root cause. If the root cause was the old media cleanup code then it might actually be fixed, if not then the code here will at least be a mitigation for it.

cmd/gotosocial/action/admin/media/prune/orphaned.go
```diff
@@ -33,14 +34,16 @@ var Orphaned action.GTSAction = func(ctx context.Context) error {
...
+	}
 
-	pruned, err := prune.manager.PruneOrphaned(ctx, dry)
+	pruned, err := prune.cleaner.Media().PruneOrphaned(ctx)
 	if err != nil {
 		return fmt.Errorf("error pruning: %s", err)
 	}
 
-	if dry /* dick heyyoooooo */ {
+	if config.GetAdminMediaPruneDryRun() /* dick heyyoooooo */ {
```
tsmethurst, line 46: might as well delete the comment since it doesn't make sense anymore :( rip

NyaaaWhatsUpDoc: we still do dry runs! it's passed in by context now :) and it's the lowest level cleaner functions that perform the dry run check

internal/cleaner/cleaner.go
```diff
@@ -0,0 +1,110 @@
...
+	return &c.media
+}
+
+func (c *Cleaner) removeFiles(ctx context.Context, files ...string) (int, error) {
+	if gtscontext.DryRun(ctx) {
+		// Dry run, do nothing.
+		return len(files), nil
+	}
+
+	errs := make(gtserror.MultiError, 0, len(files))
```
tsmethurst, line 63: This is a REALLY minor quibble, and I may be totally wrong, so do what you will with this: Since files is likely to be a big number, and we don't anticipate that many errors, it might be worth just letting Go decide how to allocate the slice, rather than setting the capacity ourselves. The price for growing it is probably cheaper than the price for allocating a big slice we're not gonna use.

NyaaaWhatsUpDoc: very true

internal/cleaner/media.go
```diff
@@ -0,0 +1,529 @@
...
+	t := time.Now().Add(-24 * time.Hour * time.Duration(maxRemoteDays))
+	m.LogUncacheRemote(ctx, t)
+	m.LogPruneOrphaned(ctx)
+	m.LogPruneUnused(ctx)
+	m.LogFixCacheStates(ctx)
+	_ = m.state.Storage.Storage.Clean(ctx)
+}
+
+// LogUncacheRemote performs Media.UncacheRemote(...), logging the start and outcome.
+func (m *Media) LogUncacheRemote(ctx context.Context, olderThan time.Time) {
```
tsmethurst, line 53: These function names are pretty misleading, it's doing far more than logging :P Maybe change them for clarity?

cmd/gotosocial/action/admin/media/prune/all.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"context"
+
+	"github.com/superseriousbusiness/gotosocial/cmd/gotosocial/action"
+	"github.com/superseriousbusiness/gotosocial/internal/config"
+	"github.com/superseriousbusiness/gotosocial/internal/gtscontext"
+)
+
+// All performs all media clean actions
+var All action.GTSAction = func(ctx context.Context) error {
```
NyaaaWhatsUpDoc, line 12: This absolutely needs more documentation etc doing. This was thrown together for my own use case :')

internal/cleaner/media.go
```diff
@@ -0,0 +1,529 @@
...
+	}
+}
+
+// PruneOrphaned will delete orphaned files from storage (i.e. media missing a database entry).
+// Context will be checked for `gtscontext.DryRun()` in order to actually perform the action.
+func (m *Media) PruneOrphaned(ctx context.Context) (int, error) {
+	var files []string
+
+	// All media files in storage will have path fitting: {$account}/{$type}/{$size}/{$id}.{$ext}
+	if err := m.state.Storage.WalkKeys(ctx, func(ctx context.Context, path string) error {
```
tsmethurst, line 98: Hey nice you kept my use of walkkeys :)

internal/cleaner/media.go
NyaaaWhatsUpDoc: Would updating the comment to say `see Media.UncacheRemote()` make more sense?  That's what I was trying to indicate with the main `media.UncacheRemote()` and then utility `media.LogUncacheRemote()` which handles the former with logging, and the comment for it. But maybe I wasn't clear enough with it 🤔

internal/cleaner/media.go
NyaaaWhatsUpDoc: None of it was started from scratch, it was all just modified as and when :)

cmd/gotosocial/action/admin/media/prune/orphaned.go
tsmethurst: Ah no I meant it doesn't read `dry /* dick heyyoooooo */` anymore, so it just looks like we're randomly writing `/* dick heyyoooooo */`. Which, I guess, isn't necessarily a bad thing.

NyaaaWhatsUpDoc:
this is ready now @tsmethurst 

the other remaining changes that i crossed out above (though not exactly) i'll make into a second PR

internal/cleaner/media.go
tsmethurst: moh it's fine, i'm just gonna resolve this one

testrig/config.go
```diff
@@ -33,7 +33,7 @@ func InitTestConfig() {
 }
 
 var testDefaults = config.Configuration{
-	LogLevel:        "trace",
+	LogLevel:        "info",
```
tsmethurst, line 36: the latest salvo is fired in the eternal battle over trace vs info

internal/cleaner/cleaner.go
```diff
@@ -0,0 +1,133 @@
...
+// Emoji returns the emoji set of cleaner utilities.
+func (c *Cleaner) Emoji() *Emoji {
+	return &c.emoji
+}
+
+// Media returns the media set of cleaner utilities.
+func (c *Cleaner) Media() *Media {
+	return &c.media
+}
+
```
tsmethurst, line 63: i think a comment here briefly explaining the return bool could be pretty useful

internal/cleaner/cleaner.go
```diff
@@ -0,0 +1,133 @@
...
+	// jobs restartable if we want to implement reloads in
+	// the future that make call to Workers.Stop() -> Workers.Start().
+
+	// Schedule the cleaning tasks to execute every day at midnight.
+	c.state.Workers.Scheduler.Schedule(sched.NewJob(func(start time.Time) {
+		log.Info(nil, "starting media clean")
+		c.Media().All(doneCtx, config.GetMediaRemoteCacheDays())
+		c.Emoji().All(doneCtx)
+		log.Infof(nil, "finished media clean after %s", time.Since(start))
+	}).EveryAt(midnight, day))
```
tsmethurst, line 132: The API of this Schedule.EveryAt function is a bit confusing, since it's called EveryAt, but you provide the At time and then the Every interval. Not something for this PR, but it would be cool to update this in the scheduler at some point.

internal/cleaner/emoji.go
```diff
@@ -0,0 +1,238 @@
...
+
+	if category == nil {
+		// Category is missing.
+		return nil, true, nil
+	}
+
+	return category, false, nil
+}
+
+func (e *Emoji) delete(ctx context.Context, emoji *gtsmodel.Emoji) error {
```
tsmethurst, line 217: When we delete emoji, should we update statuses that use that emoji? It's sort of a pain but might be worth doing in future to avoid making SELECTs in the db that we know won't succeed.

internal/cleaner/media.go
```diff
@@ -0,0 +1,546 @@
...
+	return total, nil
+}
+
+// UncacheRemote will uncache all remote media attachments older than given input time.
+// Context will be checked for `gtscontext.DryRun()` in order to actually perform the action.
+func (m *Media) UncacheRemote(ctx context.Context, olderThan time.Time) (int, error) {
+	var total int
+
+	// Drop time by an hour to improve search.
+	olderThan = olderThan.Add(-time.Hour)
```
tsmethurst, line 173: Could you explain this? I don't understand!

internal/cleaner/media_test.go
```diff
@@ -0,0 +1,427 @@
...
+	suite.transportController = testrig.NewTestTransportController(&suite.state, testrig.NewMockHTTPClient(nil, "../../testrig/media"))
+}
+
+func (suite *MediaTestSuite) TearDownTest() {
+	testrig.StandardDBTeardown(suite.db)
+	testrig.StandardStorageTeardown(suite.storage)
+	testrig.StopWorkers(&suite.state)
+}
+
+// func (suite *MediaTestSuite) TestPruneOrphanedDry() {
```
tsmethurst, line 94: you can delete the commented tests i reckon

internal/db/bundb/media.go
```diff
@@ -119,11 +120,115 @@ func (m *mediaDB) DeleteAttachment(ctx context.Context, id string) error {
...
+
+			// Get related account model.
+			if _, err := tx.NewSelect().
+				Model(&account).
+				Where("? = ?", bun.Ident("id"), media.AccountID).
+				Exec(ctx); err != nil && !errors.Is(err, db.ErrNoEntries) {
+				return gtserror.Newf("error selecting account: %w", err)
+			}
+
+			var set func(*bun.UpdateQuery) *bun.UpdateQuery
```
tsmethurst, line 141: god damn, i didn't know you could do this!

tsmethurst: oh wait no i misunderstood I think... nevermind! I thought you were setting a new function on *bun.UpdateQuery with a pointer receiver

internal/federation/dereferencing/thread.go
```diff
@@ -279,13 +279,20 @@ stackLoop:
...
 
-			// Get the "next" page property IRI
+			// Get the IRI of the "next" property.
 			pageNextIRI := pageNext.GetIRI()
-			if pageNextIRI == nil {
+
+			// Ensure this isn't a self-referencing page...
+			// We don't need to store / check against a map of IRIs
+			// as our getStatusByIRI() function above prevents iter'ing
+			// over statuses that have been dereferenced recently.
```
tsmethurst, line 292: not sure I fully understand this, can you explain where/how this prevention happens?

internal/gtscontext/context.go
```diff
@@ -41,8 +41,20 @@ const (
 	httpSigVerifierKey
 	httpSigKey
 	httpSigPubKeyIDKey
+	dryRunKey
 )
 
+// DryRun ...
```
tsmethurst, line 47: Could you fill out these function comments (here and elsewhere?). Doesn't have to be in-depth, just something to make it a bit easier for people who aren't familiar with the codebase.

internal/media/manager.go
```diff
@@ -214,7 +211,7 @@ func (m *Manager) ProcessMedia(ctx context.Context, data DataFunc, accountID str
 func (m *Manager) PreProcessEmoji(ctx context.Context, data DataFunc, shortcode string, emojiID string, uri string, ai *AdditionalEmojiInfo, refresh bool) (*ProcessingEmoji, error) {
 	instanceAccount, err := m.state.DB.GetInstanceAccount(ctx, "")
 	if err != nil {
-		return nil, fmt.Errorf("preProcessEmoji: error fetching this instance account from the db: %s", err)
+		return nil, gtserror.Newf("error fetching this instance account from the db: %s", err)
```
tsmethurst, line 214: Best to use `%w` for these errors so we can unwrap them later if need be, right? Or was this purposeful?

internal/transport/controller.go
```diff
@@ -76,7 +76,7 @@ func NewController(state *state.State, federatingDB federatingdb.DB, clock pub.C
 		fedDB:     federatingDB,
 		clock:     clock,
 		client:    client,
-		trspCache: cache.New[string, *transport](0, 100, 0),
+		trspCache: cache.New[string, *transport](0, 1, 0),
```
tsmethurst, line 79: Can you explain this change? Not in the code necessarily, can be just here so I understand it

tsmethurst (commented):
Excellent changes!!

testrig/config.go
NyaaaWhatsUpDoc: :D :D :D

internal/cleaner/cleaner.go
NyaaaWhatsUpDoc: that is a good point :thinking: 

i shall add a note to my todo list!

internal/cleaner/media_test.go
NyaaaWhatsUpDoc: i'm leaving them here for now as i do plan on bringing them back, albeit merged together and fixed in the next PR if that's all good :)

internal/gtscontext/context.go
NyaaaWhatsUpDoc: ah yep i did intend to, when i make a function `// Name ...` that's like an internal mental note to fill that out but sometimes i forgor :')

internal/media/manager.go
NyaaaWhatsUpDoc: these changes i just made by doing a find and replace on existing error statements, otherwise i would have updated them to use `%w`

internal/transport/controller.go
NyaaaWhatsUpDoc: oh right yep that's the capacity isn't it, i was being stinky brained

NyaaaWhatsUpDoc: i shall fix

internal/cleaner/media.go
NyaaaWhatsUpDoc: This is actually following the existing logic, just previously it wasn't commented. I'll add a comment though making it more clear here, assuming I'm right on what the original logic was :p

internal/federation/dereferencing/thread.go
NyaaaWhatsUpDoc: Updated the comment now :) it's due to the `fetched_at` field being updated on `getStatusByURI()` and prevent repeated frequent refetches, and the fact that we only append dereferenced statuses above when returned `statusable` is non-nil (i.e. `fetched_at` was sufficiently long ago that the status was re-dereferenced)

PR #1854, opened 2023-06-01
[bugfix] allow httpclient connection reuse

### LINKED ISSUE #1847: [bug] HTTP client connection reuse

Looking around at our code, whenever we're acting as an HTTP client we don't seem to drain the response body.

We correctly follow the Go documentation and seem to `defer` a `*http.Response.Body.Close()` everywhere, but if there's still data on the response and we don't read it the connection can't get reused. This can happen fairly easily when we get an unexpected status code. We often only check the status code, but in many cases there may be a body with an error message too. But without reading the body, the connection can't be reused in those cases. When doing multiple requests against the same server, this incurs the cost of additional TCP and TLS handshakes which would be nice to avoid. This seems to primarily hit us when we're trying to deference an instance or are webfingering them but I imagine there's other places this happens when we're federating.

To fix it, we'd have to do something like the following everywhere instead:

```go
defer func() {
  io.Copy(io.Discard, resp.Body)
  resp.Body.Close()
}()
```

This gets rather tedious to write and is easily forgotten. We could do a helper method, something like `DrainAndClose(*http.Response)` which you could then defer instead.

We can pretty safely do this because our httpclient already wraps the body in an `io.LimitReader` so we're not at risk of accidentally reading a huge response body from a (malicious) server unnecessarily. At that point we just pay the cost of not being able to reuse the connection.

In the HTTP server in Go this draining happens automatically after the handler returns. I've looked at if we can solve this in the httpclient or the transport in a similar fashion but I can't come up with a solution that I like.

v0.9.0

### GoToSocial Arch

amd64 binary

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

I might've missed some rather clever piece of code in the code where this is already handled in a transparent manner.

NyaaaWhatsUpDoc:
# Description
closes #1847 

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	codeberg.org/gruf/go-debug v1.3.0
 	codeberg.org/gruf/go-errors/v2 v2.2.0
 	codeberg.org/gruf/go-fastcopy v1.1.2
-	codeberg.org/gruf/go-iotools v0.0.0-20221224124424-3386841cb225
+	codeberg.org/gruf/go-iotools v0.0.0-20230601182242-d933b07dcbef
 	codeberg.org/gruf/go-kv v1.6.1
 	codeberg.org/gruf/go-logger/v2 v2.2.1
 	codeberg.org/gruf/go-mutexes v1.1.5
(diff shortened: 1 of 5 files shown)


tsmethurst:
Nice! :)

PR #1857, opened 2023-06-02
[bugfix] Overwrite API client closed errors with `499 - Client Closed Request`

tsmethurst:
# Description

>

This pull request updates error logging in aputil to check if the error came about because the requester hung up the request before it could be completed. If so, whatever errWithCode was given is replaced with a less alarming error, that indicates that the client already closed the request before we could process it.

Not a standard error code, but then again it doesn't really matter because the client isn't going to see it anyway...

This should make it easier to debug GtS, since the logs won't show so many superfluous error messages.

## Checklist

### DIFF
--- a/internal/api/util/errorhandling.go
+++ b/internal/api/util/errorhandling.go
@@ -19,7 +19,6 @@ package util
 
 import (
 	"context"
-	"errors"
 	"net/http"
 
 	"codeberg.org/gruf/go-kv"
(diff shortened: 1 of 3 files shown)


internal/gtserror/withcode.go
```diff
@@ -45,6 +48,10 @@ type withCode struct {
 	code     int
 }
 
+func (e withCode) Original() error {
```
NyaaaWhatsUpDoc, line 51: It would probably be here to instead implement this as `Unwrap() error` because then it will fit the `errors.Unwrap()` and `errors.Is()` function interface requirements :)

tsmethurst: Done! I hope i did it right anyways :thinking:

internal/gtserror/withcode.go
```diff
@@ -23,13 +23,21 @@ import (
 	"strings"
 )
 
+const (
+	StatusClientClosedRequest     = 499
```
NyaaaWhatsUpDoc, line 27: I'd maybe add a note here explaining these are only here because `net/http` doesn't contain them

internal/api/util/errorhandling.go
```diff
@@ -83,25 +84,48 @@ func genericErrorHandler(c *gin.Context, instanceGet func(ctx context.Context) (
...
+// See: https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#nginx.
 func ErrorHandler(c *gin.Context, errWithCode gtserror.WithCode, instanceGet func(ctx context.Context) (*apimodel.InstanceV1, gtserror.WithCode), offers ...MIME) {
-	// set the error on the gin context so that it can be logged
-	// in the gin logger middleware (internal/router/logger.go)
+	var (
+		ctxErr       = c.Request.Context().Err()
+		clientClosed = ctxErr != nil && (errors.Is(ctxErr, context.Canceled) || errors.Is(ctxErr, context.DeadlineExceeded))
+	)
+
+	if clientClosed {
```
NyaaaWhatsUpDoc, line 108: I think you could probably just do `if err := c.Request.Context().Err(); err != nil {` here, not needing the above. Any returned context error here will be the base context that was originally set on the request and will only ever be one of deadline exceeded / canceled. What do you think?

internal/gtserror/withcode.go
tsmethurst: Done!

internal/api/util/errorhandling.go
tsmethurst: Got it! Thanks :) Fixed!

PR #1858, opened 2023-06-02
[bugfix] Fix first item of thread dereferencing always being skipped

tsmethurst:
# Description

>

This pull request fixes a bug in our thread dereferencing where we were always skipping the first item in replies by calling Begin() on the iter and then calling Next() immediately, skipping to the second item (which is mostly empty).

## Checklist

### DIFF
--- a/internal/federation/dereferencing/thread.go
+++ b/internal/federation/dereferencing/thread.go
@@ -229,23 +229,24 @@ stackLoop:
 
 				// Start off the item iterator
 				current.itemIter = items.Begin()
-				if current.itemIter == nil {
-					continue stackLoop
-				}
-			} else {
-				// Get next item iterator object
-				current.itemIter = current.itemIter.Next()
 			}
 
 		itemLoop:
 			for {
+				// Check for remaining iter
 				if current.itemIter == nil {
 					break itemLoop
 				}
 
-				itemIRI, _ := pub.ToId(current.itemIter)
+				// Get current item iterator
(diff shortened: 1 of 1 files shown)


internal/federation/dereferencing/thread.go
```diff
@@ -230,28 +232,18 @@ stackLoop:
 				if current.itemIter == nil {
 					continue stackLoop
 				}
+			} else {
+				// Get next item iterator object
+				current.itemIter = current.itemIter.Next()
```
NyaaaWhatsUpDoc, line 237: I think this might still end up skipping over the iterator in some situations :thinking: 

What about doing something like:
```golang
	pageLoop:
		for {
			if current.itemIter == nil {
				// Get the items associated with this page
				items := current.page.GetActivityStreamsItems()
				if items == nil {
					continue stackLoop
				}

				// Start off the item iterator
				current.itemIter = items.Begin()
			}

		itemLoop:
			for {
				// Check for remaining iter
				if current.itemIter == nil {
					break itemLoop
				}

				// Get current item iterator
				itemIter := current.itemIter

				// Set the next available iterator
				current.itemIter = itemIter.Next()

				// Check for available IRI on item
				itemIRI, _ := pub.ToId(itemIter)
				if itemIRI == nil {
					continue itemLoop
				}

				if itemIRI.Host == config.GetHost() {
					// This child is one of ours,
					continue itemLoop
				}

```

PR #1864, opened 2023-06-04
[bugfix] AppArmor profile for SQLite

daenney:
# Description

Our default configuration places the SQLite DB in /gotosocial/, but the AppArmor profile doesn't allow us to write there. Instead of making the whole directory writable, add a writable area in /gotosocial/db/ instead and advise in the docs to move the DB there.

## Checklist

### DIFF
--- a/example/apparmor/gotosocial
+++ b/example/apparmor/gotosocial
@@ -9,6 +9,7 @@ profile gotosocial flags=(attach_disconnected, mediate_deleted) {
   /usr/local/bin/gotosocial mrix,
 
   owner /gotosocial/{,**} r,
+  owner /gotosocial/db/* wk,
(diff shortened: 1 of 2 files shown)


example/apparmor/gotosocial
```diff
@@ -9,6 +9,7 @@ profile gotosocial flags=(attach_disconnected, mediate_deleted) {
   /usr/local/bin/gotosocial mrix,
 
   owner /gotosocial/{,**} r,
+  owner /gotosocial/db/* wk,
```
daenney, line 12: I'm using a single `*` here since we only need one level. SQLite creates `sqlite.db  sqlite.db-shm  sqlite.db-wal` in that directory.

tsmethurst:
Thanks :)

PR #1874, opened 2023-06-06
[chore] Refactor AP authentication, other small bits of tidying up

tsmethurst:
# Description

>

This pull request shaves the following yaks:

- Use `gtscontext` package for setting and retrieving HTTP signature values in middleware and go-fed interface functions, instead of ersatz context functions in ap package (now deleted).
- Tidy up authenticate function in federation package: break it down into smaller parts, make it more readable, reduce nesting.
- When going through a POST request to one of our inboxes, check otherIRIs more carefully by iterating through objects on the Activity (previous behavior was broken since it was trying to cast an iterator).
- Instead of returning 403 when receiving an activity with that only tangentially relates to a URI that our instance (or receiving account) blocks, return and check the new error type `errOtherIRIBlocked`, to just return a 202 Accepted to the caller but do no further processing on the activity. This will reduce the frequency of seeing 403 errors in the logs for things that really shouldn't 403.
- Tests, other bits of tidying up.

## Checklist


internal/ap/extract.go
```diff
@@ -343,62 +344,86 @@ func ExtractURL(i WithURL) (*url.URL, error) {
...
 
-// ExtractPublicKeyForOwner extracts the public key from an interface, as long as it belongs to the specified owner.
-// It will return the public key itself, the id/URL of the public key, or an error if something goes wrong.
-func ExtractPublicKeyForOwner(i WithPublicKey, forOwner *url.URL) (*rsa.PublicKey, *url.URL, error) {
-	publicKeyProp := i.GetW3IDSecurityV1PublicKey()
-	if publicKeyProp == nil {
-		return nil, nil, errors.New("public key property was nil")
+// ExtractPublicKey extracts the public key, public key ID, and public
+// key owner ID from an interface, or an error if something goes wrong.
+func ExtractPublicKey(i WithPublicKey) (
```
tsmethurst, line 349: note for reviewer(s):

functionally this is prettymuch the same, it's just a bit less mental

internal/federation/authenticate.go
```diff
@@ -19,298 +19,297 @@ package federation
...
+		return nil, nil, err
+	}
+
+	withPublicKey, ok := t.(ap.WithPublicKey)
+	if !ok {
+		err = gtserror.Newf("resource at %s with type %T could not be converted to ap.WithPublicKey", pubKeyID, t)
+		return nil, nil, err
+	}
+
+	pubKey, _, pubKeyOwnerID, err := ap.ExtractPublicKey(withPublicKey)
```
tsmethurst, line 308: note for reviewer(s):

This now re-uses the extraction logic in our AP package instead of duplicating that code.

internal/federation/federatingactor.go
```diff
@@ -148,20 +150,38 @@ func (f *federatingActor) PostInboxScheme(ctx context.Context, w http.ResponseWr
...
 	ctx, err = f.sideEffectActor.PostInboxRequestBodyHook(ctx, r, activity)
 	if err != nil {
 		return false, gtserror.NewErrorInternalError(err)
 	}
 
-	// Check authorization of the activity.
+	// Check authorization of the activity; this will include blocks.
 	authorized, err := f.sideEffectActor.AuthorizePostInbox(ctx, w, activity)
 	if err != nil {
+		if errors.As(err, new(errOtherIRIBlocked)) {
```
tsmethurst, line 164: note for reviewer(s):

This is where that new error type is checked!!

internal/middleware/signaturecheck.go
```diff
@@ -19,95 +19,101 @@ package middleware
...
+		if signature == "" {
+			signature = c.GetHeader(authHeader)
 		}
 
-		// set the verifier and signature on the context here to save some work further down the line
-		c.Set(string(ap.ContextRequestingPublicKeyVerifier), verifier)
-		c.Set(string(ap.ContextRequestingPublicKeySignature), signature)
+		// Set relevant values on the gin context
+		// to save some work further down the line.
+		gtscontext.SetSignatureContext(c, verifier, signature, pubKeyID)
```
NyaaaWhatsUpDoc, line 117: Do we need to set these on the gin context here? What if we set them on the gin request context instead so we don't need to do the extraction / transfer of signature context values? i.e.
```golang
ctx := c.Request.Context()
ctx = gtscontext.Set___(...)
c.Request = c.Request.WithContext(ctx)
```

In that case we'd be able to get rid fo the gtscontext.{Extract,Set}SignatureContext() methods.

What doy outhink?

internal/middleware/signaturecheck.go
```diff
@@ -19,95 +19,101 @@ package middleware
...
-		requestingPublicKeyID, err := url.Parse(requestingPublicKeyIDString)
+		// The request was signed! The key ID should be given
+		// in the signature so that we know where to fetch it
+		// from the remote server. This will be something like:
+		// https://example.org/users/some_remote_user#main-key
+		pubKeyIDStr := verifier.KeyId()
+		pubKeyID, err := url.Parse(pubKeyIDStr)
 		if err != nil {
-			log.Debugf(ctx, "http signature requesting public key id %s could not be parsed as a url: %s", requestingPublicKeyIDString, err)
+			log.Debugf(ctx, "pubkey id %s could not be parsed as a url: %s", pubKeyIDStr, err)
```
NyaaaWhatsUpDoc, line 78: I'd maybe update this to a warn log. I don't think this should be common, but if it does turn out to be it's probably a useful one to find in non-debug environments :)

internal/middleware/signaturecheck.go
```diff
@@ -19,95 +19,101 @@ package middleware
...
-			// 'Trying to parse a hostname and path without a scheme is invalid but may not necessarily return an error, due to parsing ambiguities'
-			log.Debugf(ctx, "http signature requesting public key id %s was nil after parsing as a url", requestingPublicKeyIDString)
+		}
+
+		// Key can sometimes be nil, according to url parse
+		// func: 'Trying to parse a hostname and path without
+		// a scheme is invalid but may not necessarily return
+		// an error, due to parsing ambiguities'. Check this.
+		if pubKeyID == nil {
+			log.Debugf(ctx, "pubkey id %s was nil after parsing as a url", pubKeyIDStr)
```
NyaaaWhatsUpDoc, line 88: I'd maybe just catch this in the above clause, so `if err !=nil || pubKeyID == nil {`. that way we catch it with the warn log too

internal/federation/authenticate.go
```diff
@@ -19,298 +19,297 @@ package federation
...
-		err := errors.New("http request wasn't signed or http signature was invalid")
-		errWithCode := gtserror.NewErrorUnauthorized(err, err.Error())
+	// Thanks to the signature check middleware,
+	// we should already have an http signature
+	// verifier set on the context. If we don't,
+	// this is an unsigned request.
+	verifier := gtscontext.HTTPSignatureVerifier(ctx)
+	if verifier == nil {
+		errWithCode := gtserror.NewErrorUnauthorized(errUnsigned, errUnsigned.Error()+" (verifier)")
 		log.Debug(ctx, errWithCode)
```
NyaaaWhatsUpDoc, line 81: Is it worth getting rid of these debug log entries now? since the errors indicate where in this function they come from based on the unique string, and if you wanted could also wrap the `errUnsigned` with:
```golang
err := gtserror.Newf("%w", errUnsigned)
errWithCode := gtserror.NewErrorUnauthorized(err, errUnsigned + " (verifier)")
return nil, errWithCode
```

Same goes for the below error checking clauses too. What do you think?

internal/federation/authenticate.go
```diff
@@ -19,298 +19,297 @@ package federation
...
+		// (account deleted, moved, etc). Add a tombstone
+		// to our database so that we can avoid trying to
+		// dereference it in future.
+		if err := f.HandleGone(ctx, pubKeyID); err != nil {
+			err := gtserror.Newf("error marking public key %s as gone: %w", pubKeyID, err)
+			return nil, gtserror.NewErrorInternalError(err)
 		}
-		log.Tracef(ctx, "authentication for %s NOT PASSED with algorithm %s: %s", pkOwnerURI, algo, err)
+
+		err := fmt.Errorf("account with public key %s is gone", pubKeyID)
```
NyaaaWhatsUpDoc, line 274: same goes here for gtserror.Newf() here :)

internal/federation/authenticate.go
```diff
@@ -19,298 +19,297 @@ package federation
...
+	l := log.
+		WithContext(ctx).
+		WithFields(kv.Fields{
+			{"requestedUsername", requestedUsername},
+			{"pubKeyID", pubKeyIDStr},
+		}...)
+
+	// Try a database only deref first. We may already
+	// have the requesting account cached locally.
+	reqAcctURI, pubKey, errWithCode := f.derefDBOnly(ctx, pubKeyIDStr)
```
NyaaaWhatsUpDoc, line 199: Consider both these functions contain the same initial logic. Could it be worth just having a singular `f.deref()` function, but do something like: 
```golang
reqAcct, err := f.db.GetAccountByPubKeyID(...)
    // ...
}

if reAcct != nil {
    reqAcctURI, err := url.Parse(reqAcct.URI)
    if err != nil {
        // ....
    }
    
    return reqAcctURI, reqAcct.PublicKey, nil
}

if pubKeyID.Host == config.GetHost() {
    return nil, nil gtserror.New("local public key for non-existent user")
}
```

Just to minimize the amount of code / functions. What do you think? This is a very subjective one obviously :D

internal/federation/federatingprotocol.go
```diff
@@ -47,77 +80,104 @@ import (
...
+	otherIRIs = util.UniqueURIs(otherIRIs)
 
-	// clean any instances of the public URI since we don't care about that in this context
-	cleaned := []*url.URL{}
-	for _, u := range deduped {
-		if !pub.IsPublic(u.String()) {
-			cleaned = append(cleaned, u)
+	// Clean any instances of the public URI, since
+	// we don't care about that in this context.
+	otherIRIs = func(iris []*url.URL) []*url.URL {
```
NyaaaWhatsUpDoc, line 165: It's probably worth doing this check after the unique URIs call just to minimize the number of URIs that function needs to go through and search for :)

internal/middleware/signaturecheck.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1874/commits/a458e9908df4e4496ca9212842e564460b735e53 kerblammo!

internal/middleware/signaturecheck.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1874/commits/a458e9908df4e4496ca9212842e564460b735e53 kapow!

internal/federation/authenticate.go
tsmethurst: Oh yeah I like it!

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1874/commits/a458e9908df4e4496ca9212842e564460b735e53

internal/federation/authenticate.go
```diff
@@ -19,298 +19,297 @@ package federation
...
-		// the request is remote and we don't have the public key yet,
-		// so we need to authenticate the request properly by dereferencing the remote key
-		gone, err := f.CheckGone(ctx, requestingPublicKeyID)
-		if err != nil {
-			errWithCode := gtserror.NewErrorInternalError(fmt.Errorf("error checking for tombstone for %s: %s", requestingPublicKeyID, err))
-			log.Debug(ctx, errWithCode)
-			return nil, errWithCode
-		}
+	// At this point no algorithms passed.
+	err := fmt.Errorf(
```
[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/1874/commits/c1a65cfb4e3f323432107da0fd20dc5d1accd624

internal/federation/authenticate.go
tsmethurst: Right! One slipped through.

https://github.com/superseriousbusiness/gotosocial/pull/1874/commits/4f4a869edcebb9efb78f12d2f36ed9d40802d94e

internal/federation/authenticate.go
tsmethurst: mmm i'm gonna leave this one if that's OK, I think ultimately it's prettymuch the same either way and i find this current way just a little more readable

internal/middleware/signaturecheck.go
```diff
@@ -19,95 +19,101 @@ package middleware
...
-		c.Set(string(ap.ContextRequestingPublicKeyVerifier), verifier)
-		c.Set(string(ap.ContextRequestingPublicKeySignature), signature)
+		// Set relevant values on the request context
+		// to save some work further down the line.
+		ctx = gtscontext.SetHTTPSignatureVerifier(ctx, verifier)
+		ctx = gtscontext.SetHTTPSignature(ctx, signature)
+		ctx = gtscontext.SetHTTPSignaturePubKeyID(ctx, pubKeyID)
+
+		// Clone the request to replace the context.
+		c.Request = c.Request.Clone(ctx)
```
NyaaaWhatsUpDoc, line 117: i'd change this to `c.Request.WithContext()`. this way we don't need to do a deep clone of the entire request each time

Issue #1832, opened 2023-05-26
[bug] Gotosocial 0.9 responded with the incorrect reblogged status of posts in the home timeline.


veer66:
### Describe the bug with a clear and concise description of what the bug is.

https://github.com/superseriousbusiness/gotosocial/assets/370688/cb98e412-5efa-4c34-9cd3-347e2fcef7eb

### What's your GoToSocial Version?

0.0

### GoToSocial Arch

arm64

### What happened?

The reblogged status of the post in the home timeline was false, although I boosted the post.

### What you expected to happen?

The reblogged status of the post in the home timeline must be true after I boosted the post.

### How to reproduce it?

1. Boost my post
2. Reload the home timeline, which is /api/v1/timelines/home?limit=20
3. Check reblogged status of the post

tsmethurst:
This is related to https://github.com/superseriousbusiness/gotosocial/issues/1449, we'll fix it soon(tm)!

veer66:
@tsmethurst Thank you. I will keep [disabling the timeline cache](https://github.com/superseriousbusiness/gotosocial/commit/522c45ff1355d5dfeb68559a708ee51fd99a089b) until #1449 is fixed.

Issue #1449, opened 2023-02-07
[chore] Update prepared entries in the home timeline linked list with latest faves/boosts/replies


tsmethurst:
Currently we implement timelines as a linked list of 'indexed' entries and 'prepared' entries, where 'prepared' items are API models that are ready to be serialized on demand.

This is nice and speedy when fetching timelines. However, it also means that prepared items can get out of date. For example, say a user scrolls through their home timeline and faves a post. The fave is put in the database, updated in the cache, and propagated out to the remote instance as a Like with no problem. However, the prepared entry in the timeline has not been updated to reflect the fave. For the user, this means that faves seem to disappear after the refresh their home timeline (see https://github.com/superseriousbusiness/gotosocial/issues/1446).

One possible solution to this: when a user interacts with a post (replies, faves, boosts), we should then iterate through that user's home timeline, find the post they interacted with, and update the API model representation to include the interaction.

NyaaaWhatsUpDoc:
My plan is to eventually have the timeline cache stored under the main set of state caches, so whenever these types of actions happen across the codebase we can easily invalidate all caches for a status of given ID.

decentral1se:
Coooool, got any pointers for an approach on a fix? Assume nobody is working on this.

I ran into https://github.com/superseriousbusiness/gotosocial/issues/1446 myself and would like to help try to fix this. Will take a run at it next time.

NyaaaWhatsUpDoc:
It will need waiting until a current open PR changing a lot of timeline code is merged I'm afraid. Will also need discussing first in the matrix as there's a number of things to keep in mind :)

PR #1879, opened 2023-06-10
[bugfix] Invalidate timeline entries for status when stats change

tsmethurst:
# Description

>

This pull request adds timeline entry invalidation, which allows the prepared/cached version of a status in a timeline to be unprepared/uncached, so that it will be re-prepared next time the timeline owner calls Get.

Invalidation is triggered in the following circumstances:

- Status gets boosted by someone.
- Status gets unboosted by someone.
- Status gets faved by someone.
- Status gets unfaved by someone.
- Status gets replied to by someone.
- Status gets unreplied to by someone (ie., a reply to that status is deleted).
- Status gets pinned by owning account (invalidate pinning accountID timelines only).
- Status gets unpinned by owning account (invalidate unpinning accountID timelines only).
- Status gets bookmarked (invalidate bookmarking accountID timelines only).
- Status gets unbookmarked (invalidate unbookmarking accountID timelines only).

closes https://github.com/superseriousbusiness/gotosocial/issues/1832
closes https://github.com/superseriousbusiness/gotosocial/issues/1449

## Checklist


internal/processing/fromcommon.go
```diff
@@ -439,16 +439,28 @@ func (p *Processor) wipeStatus(ctx context.Context, statusToDelete *gtsmodel.Sta
...
 
-	return p.stream.Delete(status.ID)
+	return p.stream.Delete(statusID)
+}
+
+// invalidateStatusFromTimelines does cache invalidation on the given status by
+// unpreparing it from all timelines, forcing it to be prepared again (with updated
+// stats, boost counts, etc) next time it's fetched by the timeline owner. This goes
+// both for the status itself, and for any boosts of the status.
+func (p *Processor) invalidateStatusFromTimelines(ctx context.Context, statusID string) error {
```
NyaaaWhatsUpDoc, line 458: I don't think it's worth returning an error from this function as i don't think unpreparing timeline entries should really be a show-stopping error. What if we made this function instead just do like:
```golang
if err := p.state.Timelines.Home.UnprepareItemFromAllTimelines(ctx, statusID); err != nil {
    log.Errorf(ctx, "error unpreparing status from home timelines: %v")
}

if err := p.state.Timelines.List.UnprepareItemFromAllTimelines(ctx, statusID); err != nil {
    log.Errorf(ctx, "error unpreparing status from list timelines: %v")
}
```

NyaaaWhatsUpDoc: Either that or could just move the above logic ^ into the functions calling this one as it's a very small amount of code-reuse. 

What do you think?

internal/timeline/manager.go
```diff
@@ -204,6 +212,31 @@ func (m *manager) WipeItemsFromAccountID(ctx context.Context, timelineID string,
...
+		// nolint:forcetypeassert
+		if err := v.(Timeline).Unprepare(ctx, itemID); err != nil {
+			errors.Append(err)
+		}
+
+		return true // always continue range
+	})
+
+	if len(errors) > 0 {
+		return gtserror.Newf("one or more errors unpreparing status %s: %w", itemID, errors.Combine())
```
NyaaaWhatsUpDoc, line 230: a more concise way of writing this wrapper error string would be `"error(s) unpreparing status %s: %w"`, what do you think?

internal/processing/status/common.go
```diff
@@ -66,3 +66,27 @@ func (p *Processor) getVisibleStatus(ctx context.Context, requestingAccount *gts
 
 	return targetStatus, nil
 }
+
+// invalidateStatus is a shortcut function for invalidating the prepared/cached
+// representation one status in the home timeline and all list timelines of the
+// given accountID. It should only be called in cases where a status update
+// does *not* need to be passed into the processor via the worker queue, since
+// such invalidation will, in that case, be handled by the processor instead.
+func (p *Processor) invalidateStatus(ctx context.Context, accountID string, statusID string) error {
```
NyaaaWhatsUpDoc, line 75: Again, same as the other invalidate from timelines function, I don't think this should be a show-stopping error for this calling this. Though it's a fair amount more logic so maybe worth rethinking a bit, as you'd want to return early during in this function for `GetListsForAccountID()`, but not so much for the other unprepare calls.

NyaaaWhatsUpDoc: so just to be clear, we don't want the caller of this to return early on any of this contained logic, but also within this function we don't necessarily want to return early on failed timeline unprepares as we want to make sure the other cached timelines also have it unprepared at least.

internal/timeline/manager.go
tsmethurst: nice yeah, fixed! https://github.com/superseriousbusiness/gotosocial/pull/1879/commits/3bc19dc4cfe983d8102eb91a1f45bf535095c35d

internal/processing/fromcommon.go
tsmethurst: Agreed! Changed it here :) https://github.com/superseriousbusiness/gotosocial/pull/1879/commits/3fde859fedceeb9d5e2b3cc874c9834dd20b54c3

internal/processing/status/common.go
tsmethurst: Think I got it here: https://github.com/superseriousbusiness/gotosocial/pull/1879/commits/dc2bb8c5eb87acc80be5518954eb9b16366af3a9

Fail if we can't get lists since that indicates a real db error, otherwise just log unprepare errors.

PR #1880, opened 2023-06-10
[feature] use `:local_only:` emoji to mark status non-federated

### LINKED ISSUE #301: [feature] Allow users to set `federated=false` on new statuses using an emoji shortcode
In Glitchsoc you can make a post local-only by appending the `:eye:` emoji shortcode to the *end* of a post: https://glitch-soc.github.io/docs/features/local-only-toots/

In Hometown this is similar, but it uses the `:local_only:` emoji shortcode *anywhere* in the post: https://github.com/hometown-fork/hometown/wiki/Local-only-posting#the-local_only-emoji

We should, imo, use the `:local_only:` approach in GoToSocial, because it's more explicit.

noerw:
This pull request enables marking posts as local-only by using a `:local_only:` emoji.
fixes #301

2 open questions:
- The API (used for the [existing status create tests](https://github.com/superseriousbusiness/gotosocial/blob/1f9d901bc00e686342bb4733313e858c06b9b49e/internal/processing/status/create_test.go#L248)) does not expose the `Federated` flag on statuses - can you point me how to test this instead?
- The emoji needs to be added manually for this feature to work - not ideal i think, but it works. I'm not sure how to go about adding a default emoji. If you could guide me a bit, I'll implement that too. Spontaneous ideas:
  - hardcode it -  where? makes customization / removal of the emoji hard
  - add it via a DB migration - a bit hacky, but enables customization, deletion, without further work
- ~~Is the added documentation actually correct that this flag also disables sending the post to remote servers?~~

---

aside: In the long run, clients should have a proper GUI for setting flags like federated / boostable / ... - ie servers would expose a well specified™ interface to retrieve post options & how they map to API parameters. This will take a collaborative effort from at least 1-2 client devs, and 1-2 server projects - last i checked (6 months ago), nobody worked in that direction.
Implementing the :local_only: emoji seems like a good stopgap to enable posting things that you don't want to be sent all over the place.


tsmethurst:
Thanks for doing this work!

Since creating that issue a couple years ago I've sort of been having some second thoughts about such stop-gap measures, since they tend to proliferate rather hacky workarounds (in this case hardcoding an emoji shortcode to have magical side effects) rather than solving the issue properly. It's v. understandable why that measure came about in the first place -- essentially, to get around official mastodon's hegemony -- but I've got some reservations about duplicating it now.

So what I'm trying to say is, I'm not 100% sure about merging this right now, especially since local-only posting currently only has the bare minimum amount of support in GtS, so we need to take a fresh pass at it anyway to make sure it all works properly before we encourage people to use it.

If we do merge this, it should come with a big warning that this is only a temporary workaround, and people shouldn't rely on it to always work, because I'd like to remove it at some point when (as you mention in your aside) we get around to trying to formalize a proper method for setting this flag that multiple server implementations can agree on.

NyaaaWhatsUpDoc:
Yeah thank you for the input @noerw! Though I'm with @tsmethurst on this one. Hardcoding an emoji for this seems like it's going to tie us into not-ideal behaviour that we may have to support going forwards.

I'm definitely not against local-only posting, but I think there needs to be some thoughts about how we're going to do this and perhaps comms with popular client software to see how they would ideally want to consume whether a status is local-only.

noerw:
Thanks, your concerns are totally understandable. I don't have more input for a decision if you want to merge this - I'm fine with running this patch on my own instance only. 

I'd love to work on specifying an API for this & start coordination, but I'm sure that I'll overload myself with this - especially as I'm not well connected with fedi devs.

I need this feature as i'm not comfortable posting some stuff to the wider fediverse - can you clarify if `Federated: false` actually does what I think it does? ie. does it only apply to ACLs, or does it really stop putting the post in the non-local inboxes?

tsmethurst:
Currently if federated = false, then the status won't be sent out to remote inboxes, indeed:

https://github.com/superseriousbusiness/gotosocial/blob/5e2897e35cd2bea889fa37a2a857f4dcc076dafc/internal/processing/fromclientapi.go#L489-L492

It also won't be shown in the web view of your profile, even if it's public and would normally be shown:

https://github.com/superseriousbusiness/gotosocial/blob/5e2897e35cd2bea889fa37a2a857f4dcc076dafc/internal/db/bundb/account.go#L620-L621

However, we haven't yet tested properly how airtight this is, it could be that if someone knows the URI of the status, they could still dereference it. Like I said, we still gotta do another proper pass on this and make sure it works as expected.

tsmethurst:
Also, just wanna say thank you for being chill about the fact that we're not sure yet if we wanna merge this work or not. Oftentimes people (understandably) get peeved when they've done work with no prior indication that it won't be accepted, so it's really appreciated when folks are like 'okay no problem!' :heart:

VyrCossont:
> aside: In the long run, clients should have a proper GUI for setting flags like federated / boostable / ... - ie servers would expose a well specified™ interface to retrieve post options & how they map to API parameters. This will take a collaborative effort from at least 1-2 client devs, and 1-2 server projects - last i checked (6 months ago), nobody worked in that direction.

Further aside: I'd love to support per-post `federated`/`boostable`/`replyable`/`likeable` (and maybe a future `discoverable`, `indexable`, or `searchable`?) in Feditext. I could do this today, because they're just extra boolean parameters that only GtS has, and we know how to version-sniff GtS already.

(I will happily take this discussion to a new issue if it goes anywhere.)

## bikeshedding

But… going cross-platform gets interesting. Let's take `federated` as an example. GtS has it, Glitch doesn't support local-only posts except through the magic emoji scheme this PR uses, and [Hometown has a `local_only` bool param that's the opposite sense of `federated`](https://github.com/hometown-fork/hometown/blob/74c5cd1f6e14dc532adf7540c9dbc451da61486d/app/controllers/api/v1/statuses_controller.rb#L68), and it's not possible to know that any of these are available except through version sniffing.

### extended configuration

One route we could go *for this specific case* is extending `/api/vX/instance` to return additional fields in `configuration.statuses` dedicated to visibility options, similiarly to the way I got GtS (superseriousbusiness/gotosocial#1363) and Glitch (glitch-soc/mastodon#2091) to add one for supported MIME types, and the proposed way Mastodon will be reporting supported posting languages (mastodon/mastodon#24443):
```json
"configuration": {
  "statuses": {
    "max_characters": 5000,
    "max_media_attachments": 6,
    "characters_reserved_per_url": 25,
    "supported_mime_types": [
      "text/plain",
      "text/markdown"
    ],
    "supported_visibility_options": {
      "federated": [false, true],
      "boostable": [false, true],
      "replyable": [false, true],
      "likeable": [false, true],
      "searchable": ["everyone", "mufos", "no_one"],
      "quotable": ["everyone", "mufos", "no_one"]
    }
  }
}
```

But this only works for simple bool/enum params for post visibility, so it's kind of overengineered in the wrong direction. As a client dev, you still need to know what `supported_visibility_options` means, you still need to know that they map to extra params that you need to add to your client's new-status model, you still need to build UI and localized names for any params and values your client intends to support. Same way that `supported_mime_types` still requires you to know that it maps to a `content_type` param on the client.

### extensions

Since client devs have to do that work anyway, I wonder if it would make more sense to just start declaring versioned extensions to the Mastodon-ish API:

```json
"configuration": {
  "extensions": {
    "gotosocial:visibility:v1": {}
  }
}
```

If you see that `gotosocial:visibility:v1` extension in an instance API response, you know you have the optional `federated`, `likeable`, `replyable`, and `boostable` extra status params with bool values and nothing else, and can light up client UI for them. (This extension has no config parameters and is thus associated with a `{}` value in the extensions map.)

Extensions that GtS supports can be typed and documented as fields on the `InstanceV1Configuration` and `InstanceV2Configuration` structs, and we can annotate which API endpoints are affected by a given extension using [OpenAPI operation tags](https://swagger.io/specification/#tag-object), which we're already using. Extensions should also be documented with an overview readme on a https://docs.gotosocial.org/en/latest/api/extensions/ docs page (which doesn't exist yet).

Other instance servers can implement `gotosocial:visibility:v1`, and provided they stick to the same meaning as GtS, any client that supports `gotosocial:visibility:v1` will work. And sticking to the same meaning should be easier for other projects implementing GtS extensions because we have nice docs.

#### versioning

Want to add a `looking_for_a_fight` visibility option?

```json
"configuration": {
  "extensions": {
    "gotosocial:visibility:v1": {},
    "gotosocial:visibility:v2": {}
  }
}
```

`v2` adds that new option. Meanwhile, you're still compatible with clients that only know `v1`.

#### other implementations

This is the approach Pleroma and Akkoma take, more or less. Their `nodeinfo` API has a `metadata` field (also available from the `instance` API as `pleroma.metadata`) with a `features` list:

```json
"metadata": {
  "features": [
    "pleroma_api",
    "akkoma_api",
    "mastodon_api",
    "mastodon_api_streaming",
    "polls",
    "v2_suggestions",
    "pleroma_explicit_addressing",
    "shareable_emoji_packs",
    "multifetch",
    "pleroma:api/v1/notifications:include_types_filter",
    "quote_posting",
    "editing",
    "bubble_timeline",
    "pleroma_emoji_reactions",
    "exposable_reactions",
    "profile_directory",
    "akkoma:machine_translation",
    "custom_emoji_reactions",
    "pleroma:get:main/ostatus"
  ]
}
```

#### no but for real Vyr how do we get other projects to care about this

I don't know. It might be an easy sell for Akkoma specifically, since they can trivially translate their `features` list into an `extensions` map. But from the little experience I have trying to get useful features into Mastodon, Pleroma, Glitch, and now GtS, if someone manages to get a feature into one instance server project, it's easier to point to that one and say "hey those folks already have this".

Client devs, I think, will welcome a formal extension mechanism, because it's easier than version sniffing and doesn't break when someone decides to customize their software name or version or do an outright fork. Certainly I'm on board. The TootSDK people might be worth talking to as well — I know their flavor detection code involves a lot of version checks.

daenney:
Though the general ability is something we'd like, we're still not sure we want it in this form. So instead of having this hanging in limbo forever, I'm going to go ahead and close this for now.

Thanks again for taking the time to prepare and propose the change. Hopefully we'll get to something like this in the future, with some better all-round support for this in clients and servers.

Issue #1876, opened 2023-06-07
[feature/config] Make max amount of profile fields configurable


f0x52:
Currently hardcoded in https://github.com/superseriousbusiness/gotosocial/blob/b401bd1ccbce7fa547c75e9f0ae89e211112c1bb/internal/validate/formvalidation.go#L46

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/1876.

PR #1885, opened 2023-06-12
[Frontend] Settings for profile fields

f0x52:
# Description
Adds GUI to the user settings interface for setting the new profile fields.
image

Also exposes the allowed amount of form fields through the instance API, though currently hard-coded to 6 pending https://github.com/superseriousbusiness/gotosocial/issues/1876

### DIFF
--- a/web/source/settings/lib/form/field-array.jsx
+++ b/web/source/settings/lib/form/field-array.jsx
@@ -1,0 +1,51 @@
+/*
+	GoToSocial
+	Copyright (C) GoToSocial Authors admin@gotosocial.org
+	SPDX-License-Identifier: AGPL-3.0-or-later
+
+	This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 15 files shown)


web/source/settings/lib/form/field-array.jsx
```diff
@@ -0,0 +1,65 @@
...
+
+	const value = React.useMemo(() => parseFields(initialValue, length), [initialValue, length]);
+
+	return {
+		name,
+		value,
+		ctx: fields.current,
+		maxLength: length,
+		selectedValues() {
+			// if any form field changed, we need to re-send everything
```
tsmethurst, line 51: Does this include cases where the field key/values don't change, but the order of the fields does? I assume yes but just wanna check.

f0x52: yeah, the only way to re-order fields is to either delete or edit values around yourself, so that always results in at least one form field changing

Issue #1881, opened 2023-06-11
[bug] `error getting object from announce`


tsmethurst:
Full log:

```
Jun 11 10:33:42 goblin.technology gotosocial[98491]: timestamp="11/06/2023 10:33:42.421" func=middleware.Logger.func1.1 level=ERROR latency=2.657946ms userAgent="http.rb/5.1.0 (Mastodon/4.0.2+hometown-1.1.1; +https://rage.love/)" method=POST statusCode=500 path=/users/tobi/inbox clientIP=78.47.167.98 error="Error #01: PostInboxScheme: error calling sideEffectActor.PostInbox: Announce: error converting announce to boost: ASAnnounceToStatus: error getting object from announce: no iri found for object prop\n" requestID=2b8g9am804001jfvy4t0 msg="Internal Server Error: wrote 54B"
```

PR #1897, opened 2023-06-16
[chore/bugfix] Refactor `ap/extract.go` functions, return URIs more reliably

tsmethurst:
# Description

>

This pull request does some refactoring and enprettification of functions in extract.go file in the `ap` package, to make the naming and commenting more consistent. It also uses the `pub.ToId` function way more liberally in order to ensure that URIs are extracted not just from URI strings, but also from Types that have an ID URI set. This should fix issues where, for example, some remote server Announce's something, and the Object is not an IRI but a whole-ass Note or something, and then GoToSocial fails to process the announce. Now, the URI will be extracted from the note, and the boosted status dereferenced as normal from the horse's mouth.

There are probably more bits of tidying up I could have done here, but you gotta draw the line somewhere!

closes https://github.com/superseriousbusiness/gotosocial/issues/1881

## Checklist

### DIFF
--- a/internal/ap/extract.go
+++ b/internal/ap/extract.go
@@ -661,31 +661,41 @@ func ExtractMention(i Mentionable) (*gtsmodel.Mention, error) {
 	return mention, nil
 }
 
-// ExtractActor extracts the actor ID/IRI from an interface WithActor.
-func ExtractActor(i WithActor) (*url.URL, error) {
+// ExtractActorURI extracts the first Actor URI
+// it can find from a WithActor interface.
+func ExtractActorURI(i WithActor) (*url.URL, error) {
 	actorProp := i.GetActivityStreamsActor()
 	if actorProp == nil {
 		return nil, errors.New("actor property was nil")
 	}
+
 	for iter := actorProp.Begin(); iter != actorProp.End(); iter = iter.Next() {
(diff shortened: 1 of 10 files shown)


internal/ap/extract.go
```diff
@@ -505,274 +567,301 @@ func ExtractHashtags(i WithTag) ([]*gtsmodel.Tag, error) {
...
-		t := iter.GetType()
-		if t == nil {
-			continue
-		}
 
-		if t.GetTypeName() != "Emoji" {
+	var (
+		l      = tagsProp.Len()
+		emojis = make([]*gtsmodel.Emoji, 0, l)
+		keys   = make(map[string]any, l) // Use map to dedupe items.
```
NyaaaWhatsUpDoc, line 620: Is there a reason we need to de-duplicate emoji and mentions below?? :thinking:

tsmethurst: There was some reason yeah, I think at some point we were seeing multiple instances of one emoji showing up, and it caused problems. Can't remember the exact nature of the problems, but the deduplication code was definitely here for a reason. For the mentions, I added that because the emojis had it and it felt neater (same for tags), but can drop it from those funcs if you like, since I'm not aware they were causing any issues.

PR #1903, opened 2023-06-18
[bugfix] Parse POST-style forms in the list member removal endpoint

handlerug:
# Description

This pull request fixes form parsing in the list member removal endpoint. This fixes removing accounts from lists in Mastodon clients like [Tusky](https://tusky.app) and [Mona](https://apps.apple.com/app/id1659154653) which send the account IDs as an `application/x-www-form-urlencoded` form with a DELETE method, not in query params. 

## Checklist


tsmethurst:
Thanks! Good catch. The solution is a bit cursed BUT it's also well explained.

Issue #1889, opened 2023-06-13
[bug] Don't pass URIs from blocked domains into the dereferencer


tsmethurst:
When dereferencing a status with lots of replies and stuff, oftentimes many of those replies will be from blocked instances that my instance is not interested in. Rather than just quietly ignoring replies from blocked domains, I end up getting a bunch of messages in my logs that look like this:

```
func=dereferencing.(*deref).dereferenceStatusDescendants level=ERROR username=tobi statusIRI=https://example.org/some/status/uri msg="error dereferencing remote status https://blocked.instance/status/uri/from/some/blocked/instance: enrichStatus: blocked.instance is blocked"
```

It's good that my instance doesn't make calls to blocked domains, BUT we should probably tone down these error messages to debug level or something like that, or do a blocked domain check earlier and just skip even trying to deref statuses from blocked domains, rather than letting the attempt reach the dereferencer and erroring from there.

Issue #1901, opened 2023-06-18
[bug] Sometimes fetching the timeline fails


handlerug:
### Describe the bug with a clear and concise description of what the bug is.

`func=middleware.Logger.func1.1 level=ERROR latency=9.178766ms method=GET statusCode=500 path=/api/v1/timelines/home error="Error #01: HomeTimelineGet: error getting statuses: func2: db error while trying to prepare 01H368DSP8YY1T9QK7XZ1R2BF3: func1: error getting status with id 01H368DSP8YY1T9QK7XZ1R2BF3: \"error populating status parent: no entries\",\"error populating status parent: no entries\"\n" requestID=r5tmkkm8040015hpvp4g msg="Internal Server Error: wrote 54B"`

The status `01H368DSP8YY1T9QK7XZ1R2BF3` indeed has `in_reply_to_id` set to `01H368AXWRHFVK6DWYJ6Q45RTH` which does not exist in the database. The latter is never mentioned in the logs, the former is only mentioned in errors related to home timeline queries.

### What's your GoToSocial Version?

v0.10.0

### GoToSocial Arch

amd64 Binary (Alpine Linux)

### How to reproduce it?

Run the GoToSocial server for a few days, it's bound to happen.

### Anything else we need to know?

This all probably looks too vague to you, so if you need some more concrete data or more logs, let me know.

tsmethurst:
Ahh I've just been investigating an extremely similar problem, to do with replies sometimes getting lost and threads getting broken. This is almost certainly related...

handlerug:
I've been running main for some time with the fix applied. Haven't had any issues until now. I've seen a post in reply to another post, tried to favorite the reply, got the 404 error. Tried to refresh the home timeline, got the 500 error about the very post I tried to favorite. GtS couldn't fetch the parent status of that post; pretty sure the whole two-status thread was deleted. Maybe some mix of interacting and having the post be deleted is handled incorrectly?

handlerug:
Also happens with list timelines:

[...]

PR #1921, opened 2023-06-22
[bugfix] Ensure `InReplyToID` set properly, update dereference ancestors func

tsmethurst:
# Description

>

This pull request updates our ancestor dereferencing logic:

- Don't check blocked domains for ancestors.
- Mark statuses as orphaned when their parent can't be retrieved :crying_cat_face: 
- Comment the shit out of the function to make it clearer what's happening.

Also:

- Update dereferencing descendants to not try to get statuses from blocked domains (closes https://github.com/superseriousbusiness/gotosocial/issues/1889 )
- Dereference ancestors *before* trying to timeline a status, to ensure that its InReplyToID is set properly (should avoid situations like https://github.com/superseriousbusiness/gotosocial/issues/1901).
- Little bit of refactoring I did when I was just having a think.

## Checklist


internal/federation/dereferencing/thread.go
```diff
@@ -19,103 +19,221 @@ package dereferencing
...
+			current = parent
+			continue
+		}
+
+		// We could not fetch the parent, check if we can do anything
+		// useful with the error. For example, HTTP status code returned
+		// from remote may indicate that the parent has been deleted.
+		code := gtserror.StatusCode(err)
+
+		switch {
```
NyaaaWhatsUpDoc, line 199: very subjective but one thing you could do here is:
```golang
// ...
switch code := gtserror.StatusCode(err); {

case code == http.StatusGone || code == http.StatusNotFound:
    // ...
}
```

internal/federation/dereferencing/thread.go
```diff
@@ -255,6 +373,16 @@ stackLoop:
 					continue itemLoop
 				}
 
+				// Ensure item not on a blocked domain.
+				if blocked, err := d.state.DB.IsDomainBlocked(
```
NyaaaWhatsUpDoc, line 377: we shouldn't need to do this here, the `getStatusByURI()` function already handles thiis

internal/federation/dereferencing/thread.go
```diff
@@ -19,103 +19,221 @@ package dereferencing
...
+				return gtserror.Newf("db error updating status %s: %w", current.ID, err)
 			}
 
-			// Set the fetched status
-			status = remoteStatus
+		case code != 0:
+			// We had a code, but not one indicating deletion,
+			// log the code but don't return error or update the
+			// status; we can try again later.
+			l.Debugf("cannot dereference parent (call to parent returned %d)", code)
```
NyaaaWhatsUpDoc, line 218: i'd change this to an `l.Warn()` level log and instead of the code provide the error (which will include the code too)

internal/federation/dereferencing/thread.go
```diff
@@ -19,103 +19,221 @@ package dereferencing
...
-		if replyIRI.Host == config.GetHost() {
-			l.Tracef("following local status ancestors: %s", status.InReplyToURI)
+			// Parent isn't pinned to this status (yet), see
+			// if we can get it from the db (we should be
+			// able to, since it has an ID already).
+			parent, err := d.state.DB.GetStatusByID(
+				gtscontext.SetBarebones(ctx),
+				current.InReplyToID,
+			)
+			if err != nil && !errors.Is(err, db.ErrNoEntries) {
```
NyaaaWhatsUpDoc, line 107: I was thinking, instead of handling the case of orphaned statuses here, which gets quite complicated, what if instead we follow the lead of my `cleaner` package I added in my PR and instead make a regular nightly job that cleans up orphaned statuses? It just helps simplify the code here quite a lot

NyaaaWhatsUpDoc: (would also be worth having a database migration that handles this too)

internal/federation/dereferencing/thread.go
```diff
@@ -19,103 +19,221 @@ package dereferencing
...
+		// Parent URI is valid and not blocked, try to get it.
+		// Note: This will also work for local statuses, so no
+		// need to check the URI Host value.
+		parent, _, err := d.getStatusByURI(ctx, username, inReplyToURI)
+		if err == nil {
+			// We successfully fetched the parent.
+			// Update current status with new info.
+			current.InReplyToID = parent.ID
+			current.InReplyToAccountID = parent.AccountID
+			if err := d.state.DB.UpdateStatus(
```
NyaaaWhatsUpDoc, line 180: Again I think this feels like logic we could maybe move to the cleaner to reduce complexity here and silo cleanup logic to a (subjectively :p) more appropriate location.

internal/federation/dereferencing/thread.go
```diff
@@ -19,103 +19,221 @@ package dereferencing
...
+		case code != 0:
+			// We had a code, but not one indicating deletion,
+			// log the code but don't return error or update the
+			// status; we can try again later.
+			l.Debugf("cannot dereference parent (call to parent returned %d)", code)
+
+		case errorsv2.Assignable(err, (*ErrNotRetrievable)(nil)):
+			// Not retrievable for some other reason, so just
+			// bail; we can try again later if necessary.
+			l.Debugf("cannot dereference parent (ErrNotRetrievable %q)", err)
```
NyaaaWhatsUpDoc, line 223: did you mean to return nil here? :p

tsmethurst: yes :D

internal/federation/dereferencing/thread.go
tsmethurst: I think you maybe meant this comment to be further down; the logic in this err == nil is necessary for marking the status as being a reply to a known ID

internal/federation/dereferencing/thread.go
```diff
@@ -19,103 +19,193 @@ package dereferencing
...
+			// work on, and keep iterating.
+			current = parent
+			continue
+		}
+
+		// We could not fetch the parent, check if we can do anything
+		// useful with the error. For example, HTTP status code returned
+		// from remote may indicate that the parent has been deleted.
+		switch code := gtserror.StatusCode(err); {
+		case code == http.StatusGone || code == http.StatusNotFound:
```
tsmethurst, line 171: The update here is something we can only do if we've just done an http call and got back a status code, so this part I'd definitely like to leave in.

internal/federation/dereferencing/thread.go
tsmethurst: Yeah that's a solid idea. I added a TODO just below for this, and removed this particular orphaned update. The other one, where the status is marked as orphaned based on the returned http code, i'd rather keep (see comment below)

internal/federation/dereferencing/thread.go
tsmethurst: Nice one, didn't know you could do that (with the semicolon). Changed!

internal/federation/dereferencing/thread.go
tsmethurst: Ack! I made getStatusByURI return Unretrievable now when the domain is blocked, so we don't keep logging all those blocked domain errors (which aren't really errors).

internal/federation/dereferencing/thread.go
```diff
@@ -19,103 +19,221 @@ package dereferencing
...
 			}
 
-			// Set the fetched status
-			status = localStatus
+			return nil // Cannot continue.
+		}
 
-		} else {
-			l.Tracef("following remote status ancestors: %s", status.InReplyToURI)
+		// Ensure parent not on a blocked domain.
```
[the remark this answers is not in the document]
tsmethurst: Alright! Changed this (see above reply)

internal/federation/dereferencing/thread.go
NyaaaWhatsUpDoc: Yeah agreed :)

PR #1931, opened 2023-06-27
[bug] Try to fix the webfinger test, again

daenney:
# Description
For some reason we hit the case in CI where the `TestFingerWithHostMetaCacheStrategy` seems to experience some time dilation. It's possible this is a genuine bug, but I can't for the life of me reproduce it locally, even after having run this test thousands of times (-count=1000 when invoking go test etc.)

This changes the test to explicitly stop the webfinger cache, set TTL and Sweep frequency to something well beyond the lifetime of the cache during the test and then starts the cache again. Hopefully that does it, because the other option that remains is that for some reason timekeeping in CI/Docker is not as precise as when running the test on a host.

## Checklist


tsmethurst:
fingers crossed!

Issue #1875, opened 2023-06-06
[bug] `error reading from websocket: websocket: close 1001 (going away)` should not be logged at error level


tsmethurst:
We get these logged when clients close their browser or otherwise end their websocket connection. They're not actually errors though, so it's a bit misleading. We should try to check for these specific errors and just let them slide, or log them at another level.

PR #1932, opened 2023-06-27
[chore/bugfix] Break Websockets logic into smaller read/write functions, don't log expected errors

tsmethurst:
# Description

>

This pull request does a bit of tidying up on our streaming/websockets handler. It was getting a bit overwhelming in one big function with multiple `go func()` calls, so it's now split into discrete reading and writing functions. If either function encounters an error, the other will also stop.

Also, now it uses the gorilla library to check for expected errors and doesn't log them. Closes https://github.com/superseriousbusiness/gotosocial/issues/1875

## Checklist


internal/api/client/streaming/stream.go
```diff
@@ -149,186 +149,289 @@ import (
...
+			}
+
+			// Reset pinger on successful send, since
+			// we know the connection is still there.
+			pinger.Reset(m.dTicker)
+
+		case <-pinger.C:
+			// Time to send a keep-alive "ping".
+			l.Trace("pinging websocket ...")
+			if err := wsConn.WriteMessage(websocket.PingMessage, nil); err != nil {
```
daenney, line 429: Looking at the docs I get the impression control messages like the ping should be sent using `.WriteControlMessage`. It might be fine to use `.WriteMessage` but `WriteControlMessage` seems to be doing a bunch of additional things that I think we may want.

tsmethurst: nice spot, fixed!

darrinsmart:
Working great with my clients - Ice Cubes & Phanpy. GTS 0.10.0-rc1.

Issue #176, opened 2021-08-30
[frontend] Showing unlisted toots in webview


f0x52:
[to be discussed]

imo yes because it tracks with how I use unlisted posting myself, ie show to everyone but don't put it on Local + Federated TL directly. And it matches how other implementations handle this type of status (so you could view it through an other instance anyways)

tsmethurst:
My reason for not wanting to show unlisted toots in web view was that I wanted to provide users with a way of selecting whether or not they wanted to show a public/unauthenticated web view for a status, *without* having to mark that status as followers-only or more restrictive.

So I was initially leaning towards *not* showing unlisted toots in webview. But from a strictly activitypub point of view, that doesn't *really* make sense because unlisted toots are actually CC to public. Not that there's an obligation to show a web view for anything, according to AP, mind you.

So I'm in two minds here. I do want to make web views optional, but i also don't want us to add too many switches and toggles, but I *also* don't want to break expected behavior.

f0x52:
I think an account-wide toggle would be a good compromise, something like
"Show my public and unlisted toots on the web. Note that other instances might show these statuses regardless of the setting"

Same for showing (toots on) your profile page

tsmethurst:
This makes sense. I will close this one then :)

PR #1935, opened 2023-06-28
[feature] Treat unlisted as public for visibility

daenney:
# Description

Up to now, GtS has required authentication to view an unlisted/unlocked status. However, looking around other implementations unlisted/unlocked still means public, but with the caveat that it won't show up in local and federated timelines or in hashtag searches. It's very often used to reply to a thread to avoid filling the timelines with replies to a conversation someone might not be interested in following.

Changing this will now make unlisted replies to a status on this instance viewable, which means we get a more complete view in the thread UI in our web UI and should result in a more complete view of a thread in other client applications too.

## Checklist

### DIFF
--- a/internal/visibility/status.go
+++ b/internal/visibility/status.go
@@ -96,7 +96,7 @@ func (f *Filter) isStatusVisible(ctx context.Context, requester *gtsmodel.Accoun
 		return false, nil
 	}
 
-	if status.Visibility == gtsmodel.VisibilityPublic {
+	if status.Visibility == gtsmodel.VisibilityPublic || status.Visibility == gtsmodel.VisibilityUnlocked {
 		// This status will be visible to all.
 		return true, nil
 	}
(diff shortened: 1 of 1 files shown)


daenney:
I'm a bit worried that there might be other places in the code that are somehow implicitly depending on the fact that unlisted was gated behind auth. I'm not sure how to go about testing/validating that.

daenney:
One thing this doesn't do is update `GetAccountWebStatuses` to include unlisted posts in the profile view. I'm not sure if we should keep it that way or not?

tsmethurst:
If it's alright with you, I'd like to wait to merge this until after 0.10.0 is out, so we can play around with it on main for a bit and see how it feels. I think we're gonna need to fiddle with a few other code paths as well, and we could also tie this in to #176

daenney:
Yap, totally fine! I'll convert it to draft until we get 0.10 out in that case so we don't merge it in.

PR #1942, opened 2023-07-03
[docs] Rework backups a bit

daenney:
# Description

This changes the existing backup documentation to:

* Push a bit harder on people to perform backups, it's not really just a nice to have
* Removes the language about migrating to/from GoToSocial and a different ActivityPub server since that's really not supported
* Adds a section about using backup software and provides an example on how to do this using Borgmatic

## Checklist

Please put an x inside each checkbox to indicate that you've read and followed it: `[ ]` -> `[x]`

If this is a documentation change, only the first checkbox must be filled (you can delete the others if you want).

- [x] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [ ] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [ ] I/we have performed a self-review of added code.
- [ ] I/we have written code that is legible and maintainable by others.
- [ ] I/we have commented the added code, particularly in hard-to-understand areas.
- [ ] I/we have made any necessary changes to documentation.
- [ ] I/we have added tests that cover new code.
- [ ] I/we have run tests and they pass locally with the changes.
- [ ] I/we have run `go fmt ./...` and `golangci-lint run`.

Issue #1776, opened 2023-05-14
[feature] Differentiate between local + remote media in filesystem structure


tsmethurst:
Right now we put all local + remote media in prettymuch the same file structure, keyed by account id and media id.

In future, it would be nice if remote media could go specifically in some kind of `remote-cache` directory, which would allow for easier filesystem backups which exclude remote cache files.

Relates to https://github.com/superseriousbusiness/gotosocial/issues/1621 in that if we're going to do a migration for that issue, we might as well include a migration for this one as well, to avoid doing expensive filesystem migrations twice.

hikari-no-yume:
I accidentally duplicated this issue in https://github.com/superseriousbusiness/gotosocial/issues/1998 (I closed it now). One other reason it'd be good to have is to make it easier to monitor the size of the remote media cache, in case you need to clear it or adjust the retention interval (https://github.com/superseriousbusiness/gotosocial/issues/1997).

daenney:
You can already do this (with 0.10-rcX) to some degree now, but it requires a bit of scripting. `gotosocial admin media list-local` will list all local files. If you have a single or few-user instance, you'll notice they all share certain prefixes.

From there on out, you can get an estimate this way:
```sh
$ du -hs <storage-local-base-path>
$ du -hs <storage-local-base-path>/<prefix>
```

That gives you the total, as well as the total in MB for every local prefix. Subtract those, and you have the size of the remote media cache.

The [`media-to-borg-patterns`](https://github.com/superseriousbusiness/gotosocial/blob/main/example/borgmatic/media-to-borg-patterns.py) script has an example on how to collect the local prefixes. You could adapt that to then issue the `du` commands, instead of outputting the borg patterns.

It's definitely not admin friendly right now, but if you're feeling a little brave this can get you there in the mean time.

chenzhiwei:
Hello, do you have any plan on this enhancement?

tsmethurst:
No not really! It's pretty low priority. We'll get to it eventually probably.

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/1776.

PR #1943, opened 2023-07-03
[feature] Add media list command

daenney:
# Description

This was brought on by my attempt to set up backups for my own instance and the update to the docs in #1942.

This is an attempt to help alleviate #1776. Using `admin media list --local` the full path to each local media file will be printed, with a newline. The output of this should be feadable into backup tools in order to allow to backup local media too. Together with the database this should allow to fully recover from the loss of an instance.

The list command also gets a `--remote` flag for symmetry. In the case of `--remote` we print the `RemoteURL` instead, the location the asset can be retrieved from.
    
To get all media, you can run with `--local` and `--remote`.

## Checklist


test/envparsing.sh
```diff
@@ -105,6 +105,7 @@ EXPECT=$(cat <<"EOF"
     "letsencrypt-email-address": "",
     "letsencrypt-enabled": true,
     "letsencrypt-port": 80,
+    "local": false,
```
daenney, line 108: Due to how flags are declared through the config module, we now end up with these somewhat generic names in the config. It's not very terrible, but I could easily see this clashing with a different command at some point.

daenney: I could change the flags to have a more verbose and unique name, like `--list-local`, but it starts to stutter a bit with `admin media list --list-local` etc.

cmd/gotosocial/action/admin/media/list.go
```diff
@@ -0,0 +1,80 @@
...
+		return fmt.Errorf("failed to retrieve attachments: %w", err)
+	}
+
+	f := bufio.NewWriter(os.Stdout)
+
+	mediaPath := config.GetStorageLocalBasePath()
+
+	for _, a := range attachments {
+		if local && a.RemoteURL == "" {
+			if _, err := f.WriteString(path.Join(mediaPath, a.File.Path) + "\n"); err != nil {
```
daenney, line 64: Not sure if we should return the error here, or just ignore it entirely. Likelihood of this causing an error is essentially 0, unless `os.Stdout` gets banged up somehow. It does have the advantage that if we do run into an issue, the command will exit non-zero so it's possible to detect that generating the list experienced an error. That seems desirable as the output of this command is meant to feed into other systems.

cmd/gotosocial/action/admin/media/list.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Set the state DB connection
+	state.DB = dbConn
+
+	attachments, err := dbConn.GetAttachments(ctx, "", 0)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve attachments: %w", err)
+	}
+
+	f := bufio.NewWriter(os.Stdout)
```
NyaaaWhatsUpDoc, line 58: do we need to buffer stdout here?

cmd/gotosocial/action/admin/media/list.go
```diff
@@ -0,0 +1,80 @@
...
+
+	dbConn, err := bundb.NewBunDBService(ctx, &state)
+	if err != nil {
+		return fmt.Errorf("error creating dbservice: %s", err)
+	}
+
+	// Set the state DB connection
+	state.DB = dbConn
+
+	attachments, err := dbConn.GetAttachments(ctx, "", 0)
```
NyaaaWhatsUpDoc, line 53: should probably use one of the other paged functions here? e.g. if you look at how we iterate over media in `internal/cleaner`. on large instances this will be loading a lot of attachments into memory at once.

test/envparsing.sh
NyaaaWhatsUpDoc: we're going to be having a tidy-up of CLI and environment stuff in a little while anyway so i wouldn't worry about it too much until then. we're still only alpha too so we can afford to have _some_ breaking changes here and there when it comes to CLI usage

NyaaaWhatsUpDoc: Actually thinking about it. What if we had `admin media [list-local,list-remote]` as two separate commands? As I'm not sure I could see a situation where you would want both filepath outputs and remote URLs output at the same time :thinking:

cmd/gotosocial/action/admin/media/list.go
NyaaaWhatsUpDoc: honestly i would just drop the error here. a stdout error indicates a serious issue that is going to stop this entire process from working. chances are stderr would be fucked too which is why i wouldn't bother with a panic either.

NyaaaWhatsUpDoc: also makes the code here much more concise :)

cmd/gotosocial/action/admin/media/list.go
daenney: I usually grab a buffered one when I expect to be flushing a lot of output. I'm happy to change it though.

cmd/gotosocial/action/admin/media/list.go
daenney: Ah indeed. I'll take a look at that!

test/envparsing.sh
daenney: I've been thinking the same. I'll do that, that also avoids the flag issue while we're at it.

cmd/gotosocial/action/admin/media/list.go
daenney: Mmm, so looking at `internal/cleaner` it seems to iterate on the files in storage. But I can take the same approach with iterating using `db.GetAttachments` by starting without a `maxID` but with a limit and then go from there until I've made it through. That would let me do something like take 50 files at a time and not blow up memory usage.

NyaaaWhatsUpDoc: Yeah that's what I was thinking essentially :)

cmd/gotosocial/action/admin/media/list.go
```diff
@@ -0,0 +1,172 @@
...
+	dbService db.DB
+	state     *state.State
+	maxID     string
+	limit     int
+	out       *bufio.Writer
+}
+
+var errListDone = errors.New("no more")
+
+func (l *list) Next(ctx context.Context) ([]*gtsmodel.MediaAttachment, error) {
```
daenney, line 47: I'm not super stoked about this "iterator" thing, but I wanted to avoid having all this code duplicated between the ListLocal and ListRemote commands.

Very open to better suggestions.

daenney: Also, I've been thinking that maybe I could add a `local bool` to this and based on that do the writing to `l.out` directly. Only reason I didn't is because I thought it might be nicer to keep the fetching from the DB and the printing to stdout separate.

tsmethurst:
You nerds seem to have this covered pretty well already, but ping me if you want my input on anything :)

cmd/gotosocial/action/admin/media/list.go
daenney: I've done away with the iterator-like thing because it just didn't feel right. I like the new approach with passing in a filter func a lot more.

tsmethurst:
I'm happy to merge this if you're done tweaking, it looks good to me :)

daenney:
I think the docs was the last thing I wanted to add. So yap, I'm good at this point for a squerge.

tsmethurst:
ACTIVATING SQUERGE

Issue #1940, opened 2023-07-03
[bug] Servers behind NAT64 cannot reach v4 servers due to blocking


eeeeeta:
### Describe the bug with a clear and concise description of what the bug is.

The NAT64 range `64:ff9b::/96` cannot be used with GoToSocial, because it is [explicitly blocked](https://github.com/superseriousbusiness/gotosocial/blob/2a40c81f105d592ee474dd28fe8a7fdc31cb73f1/internal/netutil/validate.go#L31) as a 'reserved range'. 

This seems inadvisable, given servers behind a standards-compliant NAT64 (like my testing instance at `gotosocial.i.eta.st`) won't be able to federate at all with v4-only servers due to this.

### What's your GoToSocial Version?

0.9.0 git-282be6f

### GoToSocial Arch

amd64 binary

### What happened?

Interacting with some users fails, e.g.

```
Jul 03 00:17:10 gotosocial gotosocial[3114]: timestamp="03/07/2023 00:17:10.353" func=dereferencing.(*deref).fetchStatusMentions level=ERROR msg="failed to dereference account https://birb.site/users/charlag: item could not be retrieved: enrichAccount: error deferencing https://birb.site/users/charlag: Get \"https://birb.site/users/charlag\": dial tcp [64:ff9b::b01f:d5e7]:443: dial within blocked / reserved IP range"
```

### What you expected to happen?

The server federated as normal

tsmethurst:
We're cooking something up to add exceptions in such cases, so folks can override the reserved ranges :)

igalic:
I think, all the address ranges in this table https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml that are globally reachable should definitely not be on that list to begin with

tsmethurst:
Ack, will review the defaults as well in that case and make sure we're not blocking anything we oughtn't

Issue #1262, opened 2022-12-13
[feature ] Ability to disable ValidateIP (in config?)


doenietzomoeilijk:
I'm running a website and GtS on the same server, the latter in a container. I'm horsing around with doing AP-y things on the website, but GtS refuses to follow it. This makes sense, because, as far as GtS-inside-the-container knows, the website lives at 10.0.0.something, which [this validator](https://github.com/superseriousbusiness/gotosocial/blob/746f3fa4e65ac2806955a4be8e969dd8a2636ccf/internal/netutil/validate.go#L88) rejects due to it being in private range.

While I understand where the check is coming from, I'd love to be able to somehow get around it. I don't think I can change the network situation too much. The website isn't running in a container, I'm not at a point in my life where I want to horse around with hairpin NAT on the router and the split horizon DNS is the most sane solution for my needs here.

I would, however, love to be able to follow myself. ;)

Are you open to some thinking about how this could be handled, possibly, or is the validator holy and not something that's ever up for configuration?

tsmethurst:
So in other words yes, I think we'd be open to this, but better check with @NyaaaWhatsUpDoc

doenietzomoeilijk:
Personally, I'd opt for (configurably) disabling just the private prefixes for IPv4, _maybe_ the link-local for IPv6, but my knowledge of IPv6 is somewhat lacking, so I'd leave that particular decision to someone who actually knows what's what.

As for IPv4 private: that's used _widely_ in all kinds of hosting setups, I'm fairly confident in saying that being able to disable that particular check is not much of an issue. If people set up their hosting with reserved or testing ranges, well, that's on them.

NyaaaWhatsUpDoc:
I have some ideas for this :)

Was thinking we could remove the private ranges from being hard-coded in, but instead make them defaults. So you can get rid of them by providing an empty `allow-list` config variable.

doenietzomoeilijk:
Even better, that way you can specify which exact private prefix you want to allow (because who uses all three, right?)

igalic:
explicit `allow-list` is the way to go, given the funky stuff some cloud providers do with private IPv4 and link-local IPv6 addresses

rsolvang:
I have run into the same problem while hosting several instances on the same network. We use the IP ranges `172.17.0.0/16` and `fdd0:172:17::/48`. Currently these instances can't reach each other and we would love to have an allow-list in the config file!

rsolvang:
[...]

PR #1950, opened 2023-07-06
[feature] enable + document explicit IP dialer allowing/denying

tsmethurst:
# Description

>

This pull request implements explicit allowing/blocking of IP address ranges in outgoing requests, and also tidies up ipv6 checks (thanks @daenney) to deny requests outside the ipv6GlobalUnicast prefix by default.

closes https://github.com/superseriousbusiness/gotosocial/issues/1940
closes https://github.com/superseriousbusiness/gotosocial/issues/1262

Credit to @NyaaaWhatsUpDoc for the patch

## Checklist


daenney:
I have a slightly improved approach to SSRF for IPv6 in https://github.com/daenney/ssrf/blob/e23a15d4f01ef7b9e27547e0d4ebde830b676b4b/ssrf_gen.go#L63.

It came from the realisation that for IPv6, what you can actually do is:
* Block anything that's not in the IPv6 global unicast scope
* Block a few weird subnets from within the IPv6 global unicast scope

So you can do a check like this first: https://github.com/daenney/ssrf/blob/e23a15d4f01ef7b9e27547e0d4ebde830b676b4b/ssrf.go#L237-L239. That limits how big the additional blocked subnets lists for IPv6 ends up being

daenney:
I don't think the NAT64 range should be added, that sounds wrong to me. 64:ff9b::/96 is not globally routable, so we shouldn't connect to it?

eeeeeta:
> I don't think the NAT64 range should be added, that sounds wrong to me. 64:ff9b::/96 is not globally routable, so we shouldn't connect to it?

See #1940 -- if a GtS server is behind a NAT64 proxy, such as mine, you *have* to be able to connect to addresses in that range in order to speak IPv4 at all.

daenney:
Right. Is there any risk we could be tricked into connecting to a private IPv4 address in this case?

daenney:
> Right. Is there any risk we could be tricked into connecting to a private IPv4 address in this case?

Looks like this should be fine. `127.0.0.1` would map to `64:ff9b::7f00:1` but you can't reach that unless your client has a route for 64:ff9b::/96 configured. This isn't the case by default on Linux, so you'll simply get a network unreachable:

```
$ ip route get 64:ff9b::7f00:1
RTNETLINK answers: Network is unreachable
```

We can allow the NAT64 subnet by default without that creating any holes for dual-stack setups etc.

tsmethurst:
For consistency, I'm happy to add `64:ff9b::/96` back in to the reserved ranges, and then admins with instances like @eeeeeta's can just add one line to their config to get things working. What do you think?

eeeeeta:
@daenney Yeah, you can't reach NAT64 stuff without there being an actual NAT64 appliance/box doing the translation for you (like [tayga](http://www.litech.org/tayga/)) somewhere else inside the local network.

There's no technical reason why you couldn't connect to private v4 space through such an appliance -- it'd just look like `64:ff9b::c0a8:1` instead of `192.168.0.1`. That said, this seems like a pretty weird attack vector to me personally given that if you're inside an IPv6-only environment, there probably aren't any exploitable/interesting devices with v4 addresses nearby anyway...

Personally I don't mind the one-line config fix -- I don't think it's necessary to block the range in the first place, but I can understand if you'd rather be careful :)

daenney:
Honestly, I think it's fine to allow the NAT64 range by default. The only place it could maybe do weirdness is if you're transitioning from a dual-stack setup to something that's v6-only internally. But you can solve that in the NAT64 appliance too.

tsmethurst:
Ah, let's keep it this way for now with the 'if not in global unicast reject by default, if in global unicast, make sure not reserved'. It's neat and fairly easy to understand.

daenney (approved):

internal/httpclient/sanitizer_test.go
```diff
@@ -0,0 +1,152 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package httpclient
```
NyaaaWhatsUpDoc, line 18: I would make this into `httpclient_test` and then access the functions you need to test by exporting them, otherwise you end up with tests compiled into the production binary

tsmethurst: Ahh right, thanks!

Issue #1915, opened 2023-06-21
[feature] OIDC: Add required claim for account creation


xenrox:
### Is your feature request related to a problem ?

When having a huge number of users in the database of the OIDC provider, it makes sense to allow only a subset of users access to a service. Right now all OIDC users can become GoToSocial users.

### Describe the solution you'd like.

Add a required claim (or group of claims) for account creation to be successful.

### Describe alternatives you've considered.

NONE

### Additional context.

Depends on https://github.com/superseriousbusiness/gotosocial/issues/1914 since I would extract the necessary information from the same claim, that is used to determine if users should get admin permissions.

A lot of other services have similar settings, see e.g. Synapse `attribute_requirement`.

Issue #1914, opened 2023-06-21
[feature] OIDC: Make oidc-admin-groups "groups" claim configurable


xenrox:
### Is your feature request related to a problem ?

For at least keycloak "roles" are preferred over "groups" when it comes to permissions - "groups" are used for user management.

### Describe the solution you'd like.

Add a new setting to configure which claim should be checked.

### Describe alternatives you've considered.

NONE

### Additional context.

I could work on a patch for this feature request. Should `oidc-admin-groups` maybe be renamed to `oidc-admin-claim`? 
What is the policy of GoToSocial when it comes to breaking changes/backwards compatibility?

tsmethurst:
> What is the policy of GoToSocial when it comes to breaking changes/backwards compatibility?

We prefer keeping backwards compatibility where possible, but in this case I think it's fine to break this as long as we make very clear in our release notes what has changed, and what instance admins need to change in their config. Do you know what kind of terminology other softwares use for this (synapse for example)?

xenrox:
Synapse uses `attribute_requirements` ([synapse documentation]) which is a map of claim -> value.

Hedgedoc uses a combination of `rolesClaim` and `accessRole` ([hedgedoc documentation]).

Gitlab uses `groups_attribute` and `required_groups` ([gitlab documentation]).

If it were up to me, I would probably go away from either `groups` or `roles` since those seem pretty implementation/provider specific.

[synapse documentation]: https://matrix-org.github.io/synapse/v1.59/usage/configuration/config_documentation.html
[hedgedoc documentation]: https://docs.hedgedoc.org/configuration/#oauth2-login
[gitlab documentation]: https://docs.gitlab.com/ee/integration/saml.html

daenney:
This is an unfortunate thing with OIDC, the fact that there's no default scope and claims that return some form of membership information. It would make life so much easier if there was a `membership` scope with a well-defined set of claims/keys.

The `attribute_requirements` seems like a nice way to do it. The only challenge there is that our `Claims` struct has specific tags that it decodes into https://github.com/superseriousbusiness/gotosocial/blob/672386a1b9a194cfe7493fe33a9b8d2742cdcb53/internal/oidc/claims.go#L23-L30

So we'd get something like:
```go
type Claims struct {
	Sub               string `json:"sub"`
	Email             string `json:"email"`
	EmailVerified     bool   `json:"email_verified"`
	Name              string `json:"name"`
[...]

PR #1961, opened 2023-07-08
[feature] More flexible OIDC attribute matching

daenney:
# Description

In order to solve #1914 and #1915 we need to make our matching on OIDC claims more flexible. Unfortunately, there is no well-defined and supported scope amongst all IdPs to expose membership information using defined claims. So far we've expected a groups claim to exist, but this is implementation dependent.

By commit:
* This changes our claims implementation so that for standardised claims defined by a well-known scope we retain a struct field. Any other keys in the claims object are stored by their key in the new Attrs field instead, as a list of strings. A custom UnmarshalJSON is implemented for this purpose in order to ensure we have type safety once the decoding is done.

## Checklist

### DIFF
--- a/example/config.yaml
+++ b/example/config.yaml
@@ -689,14 +689,14 @@ oidc-client-secret: ""
 # Array of string. Scopes to request from the OIDC provider. The returned values will be used to
 # populate users created in GtS as a result of the authentication flow. 'openid' and 'email' are required.
 # 'profile' is used to extract a username for the newly created user.
-# 'groups' is optional and can be used to determine if a user is an admin based on oidc-admin-groups.
+# You can add any additional scopes, such as whichever scope returns the account or admin claims that are
+# used in oidc-admin-required-claims and oidc-account-required-claims.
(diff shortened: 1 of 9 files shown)


daenney:
I think the basic logic for this is correct. But I need to add a battery of tests before we merge this. I've opened this as a draft PR because I don't want this getting merged for 0.10 since we're already in RC and this probably needs to sit on main for a while and bake.

daenney:
> I think the basic logic for this is correct. But I need to add a battery of tests before we merge this. I've opened this as a draft PR because I don't want this getting merged for 0.10 since we're already in RC and this probably needs to sit on main for a while and bake.

I've extracted the claims validation into its own function, which now has a bunch of tests to ensure rules evaluate the way we intend them to. However, it's missing an end-to-end test for the full callback flow.

daenney:
I need to rework this one. There's been tons of changes and rebasing this is proving rather difficult. Will rev a new PR when it's ready.

PR #1968, opened 2023-07-09
[feature] Allow exposing lookup without auth

### LINKED ISSUE #1939: [feature] Add config toggle to expose `/api/v1/accounts/lookup` without auth

Followgraph ([website](https://followgraph.vercel.app/), [GitHub](https://github.com/gabipurcaru/followgraph)) is a tool that allows users to expand their connection graph and find new people to follow. It would be great if that worked with GoToSocial! Right now, when entering my handle, the app fails with a `401 Unauthorized` for `https://fedi.hi.ls/api/v1/accounts/lookup?acct=max`. 

https://docs.gotosocial.org/en/latest/configuration/instance/ lists some configuration knobs to expose more APIs to unauthorized users, but based on the docstrings `/v1/accounts/lookup` does not seem to be one of them. At least for my use case, I feel that `/v1/accounts/lookup` could just be public/unauthorized.

0.9.0 git-282be6f

### How to reproduce it?

1. Visit https://followgraph.vercel.app/
2. Enter your GTS handle (or use max@fedi.hi.ls)
3. Click "Search"

Thanks for building GoToSocial! 🍰 😃

daenney:
# Description

This lets an instance admin configure whether the
/api/v1/accounts/lookup endpoint requires authentication or not. Parts of this information can be gleamed through webfinger so it's conceivable people are OK with this being public.

The endpoint is documented as being public in the Masto API, so this allows an admin to make our implementation behave in the same manner should they want to.

The default remains false, it needs to be explicitly enabled.

Fixes #1939

## Checklist

### DIFF
--- a/example/config.yaml
+++ b/example/config.yaml
@@ -380,6 +380,11 @@ instance-expose-suspended: false
 # Default: false
 instance-expose-suspended-web: false
 
+# Bool. Allow unauthenticated users to make queries to /api/v1/lookup. This allows them to
+# determine if a username is available on this instance, or if we're aware of a remote
+# account.
+instance-expose-lookup: false
+
 # Bool. Allow unauthenticated users to make queries to /api/v1/timelines/public in order
 # to see a list of public posts on this server. Even if set to 'false', then authenticated
 # users (members of the instance) will still be able to query the endpoint.
--- a/internal/api/client/accounts/lookup.go
+++ b/internal/api/client/accounts/lookup.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/gin-gonic/gin"
 	apiutil "github.com/superseriousbusiness/gotosocial/internal/api/util"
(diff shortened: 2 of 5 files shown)


daenney:
Switched this to draft as we shouldn't merge it for 0.10 now that we're in RC-mode.

daenney:
I'm closing this PR. Looking at it, this endpoint is basically "webfinger + profile retrieval" with a shortcut of not doing the  webfinger lookup. That's nice, but it's hardly necessary. This also doesn't work for us as for profile retrieval we insist on authorisation and even though that would then still happen on the regular account endpoint, this one would suddenly let you skip that restriction.

Ultimately, there's also no guarantee that any AP implementation provides a Masto-compatible API. With the discovery being doable through webfinger queries and profile retrieval being doable by retrieving the actor that webfinger returns, that would be a reliable and Mastodon Client API independent way of achieving the same.

PR #1969, opened 2023-07-09
[feature] Use Read/Write lock for config

daenney:
# Description

The most common case for GtS is to read config at startup and never call the config.SetXXX methods after that. This means that in almsot all cases when GtS is running, the only thing we'll do is call config.GetXXX. However, since we currently use a Lock(), two config.GetXXX methods will have to wait on each other.

This changes the config state mutex to an RWMutex, allowing many goroutines to read the config at the same time. This should ensure that requests in HTTP handlers aren't blocked on eachother if they happen to call a config.GetXXX method.

The Set methods continue to call Lock() instead of RLock(), as they need to modify the config state, and the same goes for Viper() and Config().

## Checklist

### DIFF
--- a/internal/config/gen/gen.go
+++ b/internal/config/gen/gen.go
@@ -105,9 +105,9 @@ func generateFields(output io.Writer, prefixes []string, t reflect.Type) {
 		// ConfigState structure helper methods
 		fmt.Fprintf(output, "// Get%s safely fetches the Configuration value for state's '%s' field\n", name, fieldPath)
 		fmt.Fprintf(output, "func (st *ConfigState) Get%s() (v %s) {\n", name, fieldType)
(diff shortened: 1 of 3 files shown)


NyaaaWhatsUpDoc:
Have you done a performance comparison with this? RWMutex as far as I'm aware is generally lower performance than a regular Mutex given the extra complexity.

daenney:
Oh from my point-of-view it's ready. I just marked it as draft so we don't merge it during the RC-cycle. But we could.

NyaaaWhatsUpDoc:
the temptation ...

daenney:
I suppose we could merge it. It's more of a performance fix, it doesn't introduce any new feature.

PR #1971, opened 2023-07-09
[bugfix] Align default values in the configuration file with the code

### LINKED ISSUE #1964: [bug] Discrepancies in default values in config.yaml and in the code

There are few discrepancies in default values of parameters defined in [example/config.yaml](https://github.com/superseriousbusiness/gotosocial/blob/v0.10.0-rc1/example/config.yaml) and in [internal/config/defaults.go](https://github.com/superseriousbusiness/gotosocial/blob/v0.10.0-rc1/internal/config/defaults.go).

Here is the list of all discrepancies I've found:
| #  | `example/config.yaml` | `internal/config/defaults.go` |
| :-: | :------------------: | :---------------------------: |
|  1  | _absent_ | `AdminMediaPruneDryRun: true` |
|  2  | `advanced-throttling-retry-after: "30s"` | _absent_ |
|  3  | `block-max-size: 100` | `BlockMaxSize: 1000` |
|  4  | _absent but ok :)_ | `ConfigPath: ""` |
|  5  | `db-sqlite-busy-timeout: "5m"` | `DbSqliteBusyTimeout: time.Minute * 30` |
|  6  | `host: "localhost"` | `Host: ""` |
|  7  | `instance-expose-public-timeline: false` | _absent_ |
|  8  | _absent_ | `InstanceMaxSize:   2000` |
|  9  | _absent_ | `InstanceSweepFreq: time.Minute` |
|  10 | _absent_ | `InstanceTTL: time.Minute * 30` |
|  11 | `oidc-admin-groups: []` | _absent_ |
|  12 | `smtp-from: ""` | `SMTPFrom: "GoToSocial"` |
|  13 | `storage-s3-access-key: ""` | _absent_ |
|  14 | `storage-s3-bucket: ""` | _absent_ |
|  15 | `storage-s3-endpoint: ""` | _absent_ |
|  16 | `storage-s3-secret-key: ""` | _absent_ |
|  17 | `webfinger-sweep-freq: "1m"` | `WebfingerSweepFreq: time.Minute * 15` |

v0.10.0-rc1

### GoToSocial Arch

All are affected

### What happened?

Application may act differently for different users if they chose to configure it with or without config.yaml (e.g. using only environment variables or runtime parameters to overwrite hard-coded defaults). This is in regards to items 2, 3, 5, 12 and 17.

Other items may result in rather small inconvenience (user would need to add them to config.yaml to change their default value) - items 1, 8, 9, 10 - or may be perceived as inconsistency in the code - items 7, 11, 13, 14, 15, 16 (e.g. there are other variables declared with empty string value).

I think the default values should be the same no matter which way of configuration user chooses.

### How to reproduce it?

_No response_

I would gladly work on providing a patch to address these discrepancies.
Please let me know if you see a value in such contribution and if you think that all or maybe only some of them should be aligned :)

firescry:
# Description

>

This pull request aligns default values defined in `example/config.yaml` and in `internal/config/defaults.go` - in most cases changes are done to the config file.
There is one exception where default value is added to `defaults.go` - it is for `AdvancedThrottlingRetryAfter` variable introduced as part of #1466.
Also snippets of `config.yaml` file are updated in the documentation.

Closes #1964

## Checklist

### DIFF
--- a/docs/configuration/media.md
+++ b/docs/configuration/media.md
@@ -7,7 +7,7 @@
 ##### MEDIA CONFIG #####
 ########################
 
-# Config pertaining to user media uploads (videos, image, image descriptions).
+# Config pertaining to media uploads (videos, image, image descriptions, emoji).
 
 # Int. Maximum allowed image upload size in bytes.
 # Examples: [2097152, 10485760]
@@ -53,6 +53,6 @@ media-emoji-local-max-size: 51200
 # This strikes a good balance between decent interoperability with instances that have
 # higher emoji size limits, and not taking up too much space in storage.
 # Examples: [51200, 102400]
-# Default: 51200
+# Default: 102400
 media-emoji-remote-max-size: 102400
 ```
--- a/example/config.yaml
+++ b/example/config.yaml
@@ -231,8 +231,8 @@ db-sqlite-cache-size: "8MiB"
 # If set to empty string or zero, the sqlite default will be used.
 # See: https://www.sqlite.org/pragma.html#pragma_busy_timeout
 # Examples: ["0s", "1s", "30s", "1m", "5m"]
-# Default: "5s"
-db-sqlite-busy-timeout: "5m"
+# Default: "30m"
+db-sqlite-busy-timeout: "30m"
 
 cache:
   # Cache configuration options:
@@ -246,7 +246,7 @@ cache:
   #### VISIBILITY CACHES ######
   #############################
   #
-  # Configure Status and account 
+  # Configure Status and account
   # visibility cache.
 
   visibility-max-size: 2000
@@ -265,7 +265,7 @@ cache:
     account-ttl: "30m"
     account-sweep-freq: "1m"
 
-    block-max-size: 100
+    block-max-size: 1000
     block-ttl: "30m"
     block-sweep-freq: "1m"
 
@@ -288,6 +288,10 @@ cache:
     follow-request-max-size: 2000
     follow-request-ttl: "30m"
     follow-request-sweep-freq: "1m"
+
+    instance-max-size: 2000
+    instance-ttl: "30m"
+    instance-sweep-freq: "1m"
 
     list-max-size: 2000
     list-ttl: "30m"
@@ -331,7 +335,7 @@ cache:
 
     webfinger-max-size: 250
     webfinger-ttl: "24h"
-    webfinger-sweep-freq: "1m"
+    webfinger-sweep-freq: "15m"
 
 ######################
 ##### WEB CONFIG #####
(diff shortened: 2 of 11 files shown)


docs/configuration/media.md
```diff
@@ -53,6 +53,6 @@ media-emoji-local-max-size: 51200
 # This strikes a good balance between decent interoperability with instances that have
 # higher emoji size limits, and not taking up too much space in storage.
 # Examples: [51200, 102400]
-# Default: 51200
+# Default: 102400
```
daenney, line 56: This change isn't correct. `local-max-size` is `51200` in the code: https://github.com/superseriousbusiness/gotosocial/blob/65c89709bc6bf05a369a591d19c612417794f50c/internal/config/defaults.go#L75

daenney:
I think this looks pretty good, and it's honestly something I'd be fine with merging in the RC-cycle as it fixes a bunch of erroneous values in `example/config.yaml`. We tell people to review and adopt those values when upgrading, so right now a bunch of instances are probably slightly misconfigured.

@tsmethurst what do you think?

firescry:
Thanks for the review! 😊

I set it as "ready for review" in case you would like to merge it as it is.

docs/configuration/smtp.md
```diff
@@ -43,8 +43,8 @@ smtp-password: ""
 
 # String. 'From' address for sent emails.
 # Examples: ["mail@example.org"]
-# Default: ""
-smtp-from: ""
+# Default: "GoToSocial"
+smtp-from: "GoToSocial"
```
tsmethurst, line 47: Damn is this really our default? We should change this in the config in that case, since this isn't a real email address. Best to keep default as an empty string I think.

example/config.yaml
```diff
@@ -742,8 +749,8 @@ smtp-password: ""
 
 # String. 'From' address for sent emails.
 # Examples: ["mail@example.org"]
-# Default: ""
-smtp-from: ""
+# Default: "GoToSocial"
+smtp-from: "GoToSocial"
```
tsmethurst, line 753: Same comment as above

tsmethurst:
This is helpful, thank you!

docs/configuration/smtp.md
firescry: I set value of _SMTPFrom_ in `docs/configuration/advanced.md` to empty string and updated both `example/config.yaml` and snippet in documentation accordingly.

example/config.yaml
firescry: Done :)

tsmethurst:
Kapow!

Issue #1069, opened 2022-11-17
[Client Compatibility] Sengi


anantshri:
Hello Folks,

Just a note from my observations around sengi. (https://github.com/NicolasConstant/sengi)

This particular tool works great with GotoSocial except a few warnings for example it reverts to using v1 api endpoint for search which gives a 404. 

We do have the right api endpoint on v2 and it should work however digging deep in the code it seems the rely on mastodon version reported by the server https://github.com/NicolasConstant/sengi/search?q=instance.major+%3E%3D+3%29 and if version is less them 3 they use API v1. Not sure if we want to address this by switching around our reported version number for better client compatibilities but just wanted to put it out

1. Sengi good for daily use with occasional errors which doesnt hamper the working.
2. errors mostly coz we advertise a version which they interpret as very old version of mastodon.

screenshot of error message 
![image](https://user-images.githubusercontent.com/236843/202462503-a7e0e209-96d7-4808-8db7-a542c6c974a5.png)

whtiehack:
+1

![image](https://user-images.githubusercontent.com/12741016/202654868-1a1583f5-15de-463d-b8a7-88c877317610.png)

EchedeyLR:
@whtiehack I believe you should open a different issue instead of posting here. The client is neither the same nor the error is similar.

whtiehack:
> @whtiehack I believe you should open a different issue instead of posting here. The client is neither the same nor the error is similar.

You're right. I'm sorry.

anantshri:
I think with v0.10 i have not seen these errors. search is also working fantastic. we can close this.

thanks for reminder @tsmethurst

Issue #1953, opened 2023-07-06
[feature] SemVer version in /api/vX/instance and /nodeinfo/2.0


VyrCossont:
### Is your feature request related to a problem ?

As a client dev, I would like to know approximately which version of GotoSocial I'm talking to, so that I can enable or disable my client's features as necessary (search, lists, etc.). Right now, if I run my GtS instance off `main`, the version fields in `/api/v1/instance`, `/api/v2/instance`, and `/nodeinfo/2.0` are a build identifier like `git-2a40c81`. This doesn't tell my client much.

### Describe the solution you'd like.

I suggest changing this so that builds off `main` or other Git branches (as opposed to version tags) are marked as [SemVer pre-release versions](https://semver.org/#spec-item-9) of whatever version's next on the roadmap, with the Git revision attached as [SemVer build metadata](https://semver.org/#spec-item-10). This would make my example version into `0.10.0-git+2a40c81`.

I can parse that with any SemVer library and then compare it to a known version like `0.9` to decide whether to enable (for example) profile field editing, subscriptions, and extra trans rights.

### Describe alternatives you've considered.

One way to do feature detection is to just call an API and see if it works, but not all APIs can be called safely just to test them, and being able to `HEAD` or `GET` an HTTP API doesn't necessarily tell you what methods, options, inputs, etc. can be used.

Another way is to have feature detection flags with names and add those to `instance` or `nodeinfo` APIs, which is what Pleroma and Akkoma do, and to some extent what the Mastodon-compatible `instance` API's cross-platform `configuration` object is for, but then someone has to name all of the features explicitly.

tsmethurst:
Mm yes, good one. If you build with `goreleaser` it does present something closer to what you're looking for -- `0.9.1-SNAPSHOT git-df39a13` in the case of the current deployment of goblin.technology -- but indeed if you build using `go build` or `./scripts/build.sh` it doesn't work that way. Will try to find some time to address this once 0.10.0 is out.

tsmethurst:
Oh actually if you do `./scripts/build.sh` it should be injecting the version into the build via the following line:

`GO_LDFLAGS="${GO_LDFLAGS-} -s -w -extldflags '-static' -X 'main.Version=${VERSION:-$(git describe --tags --abbrev=0)}'"`

But that only uses the most recent git tag, it doesn't assume snapshot of the next version.

VyrCossont:
[...]

PR #1977, opened 2023-07-10
[feature] Report Masto version in /api/v1/instance

daenney:
# Description

A number of applications use the version reported by /api/v1/instance as a way to do feature detection for the Mastodon client API. Due to the current GtS versions in 0.x, this often means GtS gets determined to be non-compatible by clients and libraries that only ever test against Mastodon proper.

This adds a configuration value that allows the admin to decide to inject a Mastodon version in that response instead, in order to cajole clients into working correctly with GtS. This is only done for /api/v1/instance as that's the Masto client API, nodeinfo is left as it is as that should not be used by Masto clients for feature detection.

Since the clients we recommend don't need to be tricked this way, this feature defaults to being off. Admins can enable it if so desired, which may be useful for small communities where admins might have less say over which clients people use.

## Checklist

### DIFF
--- a/internal/typeutils/internaltofrontend.go
+++ b/internal/typeutils/internaltofrontend.go
@@ -45,11 +45,16 @@ const (
 	instanceAccountsMaxFeaturedTags             = 10
 	instanceAccountsMaxProfileFields            = 6 // FIXME: https://github.com/superseriousbusiness/gotosocial/issues/1876
 	instanceSourceURL                           = "https://github.com/superseriousbusiness/gotosocial"
+	instanceMastodonVersion                     = "3.5.3"
 )
 
 var instanceStatusesSupportedMimeTypes = []string{
 	string(apimodel.StatusContentTypePlain),
 	string(apimodel.StatusContentTypeMarkdown),
+}
+
+func toMastodonVersion(in string) string {
+	return instanceMastodonVersion + "+" + strings.ReplaceAll(in, " ", "-")
 }
 
 func (c *converter) AccountToAPIAccountSensitive(ctx context.Context, a *gtsmodel.Account) (*apimodel.Account, error) {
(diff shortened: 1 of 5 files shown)


daenney:
This should fix #1069, I think this also would've fixed #113 . I also seem to recall an issue where Tobi and Blackle were discussing this and the solution of only optionally reporting the Masto version on this endpoint came up, but for the life of me I can't find it now.

internal/typeutils/internaltofrontend.go
```diff
@@ -45,13 +45,18 @@ const (
...
+	instanceMastodonVersion                     = "3.5.3"
 )
 
 var instanceStatusesSupportedMimeTypes = []string{
 	string(apimodel.StatusContentTypePlain),
 	string(apimodel.StatusContentTypeMarkdown),
 }
 
+func toMastodonVersion(in string) string {
+	return instanceMastodonVersion + "+" + strings.ReplaceAll(in, " ", "-")
```
daenney, line 57: I went for `+` here in order to turn our version into "build metadata" according to SemVer, so it should parse correctly for any clients that look at it. SemVer doesn't seem to like spaces which we have between the version and the git hash so that's replaced by a hyphen.

tsmethurst:
For context, this is where we had the discussion: https://github.com/superseriousbusiness/gotosocial/pull/778

Since we resolved a while ago that we're only supporting the Mastodon API up to v4.1.x (https://gts.superseriousbusiness.org/@gotosocial/statuses/01GZEBDK1JGTNJZ8SCPXA274RY) we could use that version number instead of 3.5.3. Although we don't actually support all features from 4.1.x yet. But then again, we don't support all features from 3.5.3 yet either :D

tsmethurst:
Since version numbers are being fiddled with here, here's a link to a related issue: https://github.com/superseriousbusiness/gotosocial/issues/1953

Not that we have to include the above in this PR necessarily, but it's worth bearing in mind for another PR.

VyrCossont:
> nodeinfo is left as it is as that should not be used by Masto clients for feature detection

FWIW I'm now using nodeinfo in [Feditext](https://github.com/feditext/feditext) because it's the easier way to get the real server software name and version number, instead of a fake Mastodon version number that doesn't actually correspond to the features of that Mastodon version (see also: Pleroma, Akkoma). I'm in favor of this change, because as noted, not everybody is going to bother testing against non-Mastodon implementations, and a Mastodon 3 or 4 version number is a lie that's certainly closer to the truth. But smart cross-platform clients have to use nodeinfo software name and version *as well as* the instance API to get an idea of which APIs should be available.

Issue #1980, opened 2023-07-11
TestFingerWithHostMetaCacheStrategy is flaky


MaeIsBad:
### Describe the bug with a clear and concise description of what the bug is.
TestFingerWithHostMetaCacheStrategy randomly fails

With no code changes: 
- fail: https://drone.superseriousbusiness.org/superseriousbusiness/gotosocial/6608/1/3
- fail: https://drone.superseriousbusiness.org/superseriousbusiness/gotosocial/6607/1/3
- pass: https://drone.superseriousbusiness.org/superseriousbusiness/gotosocial/6609/1/3

### What's your GoToSocial Version?

ca5492b65f45c7db8a9cfb767b0b48aa6cf6fe24

tsmethurst:
I'm gonna close this for now since the test is now disabled!

PR #1982, opened 2023-07-12
Support setting private notes on accounts

VyrCossont:
# Description

This pull request adds the Mastodon API for setting private notes, and implements getting private notes as part of the existing Mastodon relationship API (they're already part of the relationship object, but GtS always returns an empty string).

This involved adding one API endpoint, one table, one model cache, and the plumbing between them. I've included `ID`, `CreatedAt`, and `UpdatedAt` columns in the `notes` table to match the style of other tables, but the note feature doesn't actually use them, although they might be useful for a later bulk note import/export tool.

## Checklist


testrig/testmodels.go
```diff
@@ -1891,6 +1891,18 @@ func NewTestFaves() map[string]*gtsmodel.StatusFave {
...
 }
 
+// NewTestNotes returns some notes for use in testing.
+func NewTestNotes() map[string]*gtsmodel.Note {
+	return map[string]*gtsmodel.Note{
+		"local_account_2_note_on_1": {
+			ID:              "01H53TM628GNC4ZDNRGQGPK8S0",
+			AccountID:       "01F8MH5NBDF2MV7CTC4Q5128HF",
+			TargetAccountID: "01F8MH1H7YV1Z7D2C8K2730QBF",
+			Comment:         "extremely average poster",
```
tsmethurst, line 1901: damn that's cold

internal/db/bundb/migrations/20230711214815_notes.go
```diff
@@ -0,0 +1,45 @@
+package migrations
```
tsmethurst, line 1: Could you add the license text header here?

internal/processing/account/note.go
```diff
@@ -0,0 +1,31 @@
+package account
```
tsmethurst, line 1: License text again please, sorry to be a pain about it

internal/gtsmodel/note.go
```diff
@@ -0,0 +1,32 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package gtsmodel
+
+import "time"
+
+// Note stores a private note from a local account related to any account.
+type Note struct {
```
tsmethurst, line 23: Because "Note" already has a defined meaning in ActivityStreams, I think it would be better to call this AccountNote in order to differentiate it from an ActivityStreams Note, just to avoid any semantic confusion.

tsmethurst:
Looks great so far, thank you! Couple of small comments and then a slightly bigger one about changing Note -> AccountNote, what do you think?

NyaaaWhatsUpDoc:
Agreed with @tsmethurst on everything, this looks great :). Also quick Q, how was it working with the model cache as someone that hasn't worked quite so much with the codebase? My cache library is something I'm always looking to improve if I can!

VyrCossont:
@NyaaaWhatsUpDoc pretty easy once I looked at existing code. There's some boilerplate cache config code that I'd be using generics or macros for in another language, but I don't know Go well enough to suggest improvements there.

internal/processing/account/note.go
VyrCossont: Done.

internal/db/bundb/migrations/20230711214815_notes.go
VyrCossont: Done.

internal/api/client/accounts/note.go
```diff
@@ -0,0 +1,110 @@
...
+		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	if _, err := apiutil.NegotiateAccept(c, apiutil.JSONAcceptHeaders...); err != nil {
+		apiutil.ErrorHandler(c, gtserror.NewErrorNotAcceptable(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	targetAcctID := c.Param(IDKey)
```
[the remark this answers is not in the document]
VyrCossont: Yay! Done.

internal/gtsmodel/note.go
VyrCossont: Done.

VyrCossont:
Ready to merge. Prefer a fixup-like merge that removes text from all commits after the first one since they're not relevant outside the context of review. I just learned about `fixup! ` commit message prefixes *after* pushing all of these changes and I'm not sure if GitHub supports them when it does a squash merge.

VyrCossont:
I don't think that last is a real error. Looks like #1980.

tsmethurst:
> I don't think that last is a real error. Looks like #1980.

Yeah! I merged main into the branch which includes disabling that test (til we figure out what's going on, at least) so should pass now :)

PR #1987, opened 2023-07-13
[feature/performance] support uncaching remote emoji + scheduled cleanup functions

NyaaaWhatsUpDoc:
# Description
- adds `cached` column to emoji table in similar vein to media attachments
- add emoji cleanup functions to: uncache remote, prune unused, fix cache states
- support recaching emojis not cached on disk
- add emoji cleaner tests. (each designed to take a dataset (emoji slice) to allow easy testing of multiple datasets in the future when we can more easily generate them)
- some small cleanups in the media manager package

## Checklist

### DIFF
--- a/internal/cleaner/emoji.go
+++ b/internal/cleaner/emoji.go
@@ -311,12 +311,12 @@ func (e *Emoji) fixCacheState(ctx context.Context, emoji *gtsmodel.Emoji) (bool,
 	switch {
 	case *emoji.Cached && !exist:
 		// Mark as uncached if expected files don't exist.
-		l.Debug("cached=true exists=false => uncaching")
+		l.Debug("cached=true exists=false => marking uncached")
 		return true, e.uncache(ctx, emoji)
 
 	case !*emoji.Cached && exist:
 		// Remove files if we don't expect them to exist.
-		l.Debug("cached=false exists=true => deleting")
+		l.Debug("cached=false exists=true => removing files")
 		_, err := e.removeFiles(ctx,
 			emoji.ImageStaticPath,
 			emoji.ImagePath,
--- a/internal/cleaner/emoji.go
+++ b/internal/cleaner/emoji.go
@@ -37,19 +37,13 @@ type Emoji struct {
 
 // All will execute all cleaner.Emoji utilities synchronously, including output logging.
 // Context will be checked for `gtscontext.DryRun()` in order to actually perform the action.
-func (e *Emoji) All(ctx context.Context) {
-	e.LogPruneMissing(ctx)
+func (e *Emoji) All(ctx context.Context, maxRemoteDays int) {
+	t := time.Now().Add(-24 * time.Hour * time.Duration(maxRemoteDays))
+	e.LogUncacheRemote(ctx, t)
 	e.LogFixBroken(ctx)
-}
-
-// LogPruneMissing performs Emoji.PruneMissing(...), logging the start and outcome.
-func (e *Emoji) LogPruneMissing(ctx context.Context) {
-	log.Info(ctx, "start")
-	if n, err := e.PruneMissing(ctx); err != nil {
-		log.Error(ctx, err)
-	} else {
-		log.Infof(ctx, "pruned: %d", n)
-	}
+	e.LogPruneUnused(ctx)
+	e.LogFixCacheStates(ctx)
+	_ = e.state.Storage.Storage.Clean(ctx)
 }
 
 // LogUncacheRemote performs Emoji.UncacheRemote(...), logging the start and outcome.
@@ -72,46 +66,24 @@ func (e *Emoji) LogFixBroken(ctx context.Context) {
 	}
 }
 
-// PruneMissing will delete emoji with missing files from the database and storage driver.
-// Context will be checked for `gtscontext.DryRun()` to perform the action. NOTE: this function
-// should be updated to match media.FixCacheStat() if we ever support emoji uncaching.
-func (e *Emoji) PruneMissing(ctx context.Context) (int, error) {
-	var (
-		total int
-		maxID string
-	)
-
-	for {
-		// Fetch the next batch of emoji media up to next ID.
-		emojis, err := e.state.DB.GetEmojis(ctx, maxID, selectLimit)
-		if err != nil && !errors.Is(err, db.ErrNoEntries) {
-			return total, gtserror.Newf("error getting emojis: %w", err)
-		}
-
-		if len(emojis) == 0 {
-			// reached end.
-			break
-		}
-
-		// Use last as the next 'maxID' value.
-		maxID = emojis[len(emojis)-1].ID
-
-		for _, emoji := range emojis {
-			// Check / fix missing emoji media.
-			fixed, err := e.pruneMissing(ctx, emoji)
-			if err != nil {
-				return total, err
-			}
-
-			if fixed {
-				// Update
-				// count.
-				total++
-			}
-		}
+// LogPruneUnused performs Emoji.PruneUnused(...), logging the start and outcome.
+func (e *Emoji) LogPruneUnused(ctx context.Context) {
+	log.Info(ctx, "start")
+	if n, err := e.PruneUnused(ctx); err != nil {
+		log.Error(ctx, err)
+	} else {
+		log.Infof(ctx, "pruned: %d", n)
 	}
+}
 
-	return total, nil
+// LogFixCacheStates performs Emoji.FixCacheStates(...), logging the start and outcome.
+func (e *Emoji) LogFixCacheStates(ctx context.Context) {
+	log.Info(ctx, "start")
+	if n, err := e.FixCacheStates(ctx); err != nil {
+		log.Error(ctx, err)
+	} else {
+		log.Infof(ctx, "fixed: %d", n)
+	}
 }
 
 // UncacheRemote will uncache all remote emoji older than given input time. Context
@@ -127,8 +99,8 @@ func (e *Emoji) UncacheRemote(ctx context.Context, olderThan time.Time) (int, er
 	mostRecent := olderThan
 
 	for {
-		// Fetch the next batch of emojis older than last-set time.
-		emojis, err := e.state.DB.GetRemoteEmojisOlderThan(ctx, olderThan, selectLimit)
+		// Fetch the next batch of cached emojis older than last-set time.
+		emojis, err := e.state.DB.GetCachedEmojisOlderThan(ctx, olderThan, selectLimit)
 		if err != nil && !errors.Is(err, db.ErrNoEntries) {
 			return total, gtserror.Newf("error getting remote emoji: %w", err)
 		}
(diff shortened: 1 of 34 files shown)


internal/regexes/regexes.go
```diff
@@ -70,7 +70,7 @@ const (
 	statusesPath   = userPathPrefix + `/` + statuses + `/(` + ulid + `)$`
 	blockPath      = userPathPrefix + `/` + blocks + `/(` + ulid + `)$`
 	reportPath     = `^/?` + reports + `/(` + ulid + `)$`
-	filePath       = `^/?(` + ulid + `)/([a-z]+)/([a-z]+)/(` + ulid + `)\.([a-z]+)$`
+	filePath       = `^/?(` + ulid + `)/([a-z]+)/([a-z]+)/(` + ulid + `)\.([a-z0-9]+)$`
```
tsmethurst, line 73: oh shit look at that

tsmethurst: wait... are there cases where we have numbers in the suffix of a file? oh... i guess for like, `mp3`.. hmm OH OR `MP4`, shit!

internal/db/bundb/search.go
```diff
@@ -72,26 +72,26 @@ var replacer = strings.NewReplacer(
...
 // whereSubqueryLike appends a WHERE clause to the
-// given SelectQuery q, which searches for matches
-// of searchQuery in the given subQuery using LIKE.
-func whereSubqueryLike(
-	q *bun.SelectQuery,
-	subQuery *bun.SelectQuery,
-	searchQuery string,
+// given SelectQuery, which searches for matches
+// of `search` in the given subQuery using LIKE.
+func whereLike(
```
tsmethurst, line 77: Perhaps worth moving this function into util.go if we're now using it for more than just searching statuses / accounts

internal/db/bundb/status.go
```diff
@@ -429,6 +429,17 @@ func (s *statusDB) DeleteStatusByID(ctx context.Context, id string) db.Error {
 	})
 }
 
+func (s *statusDB) GetStatusesUsingEmoji(ctx context.Context, emojiID string) ([]*gtsmodel.Status, error) {
+	var statusIDs []string
+	if _, err := whereLike(s.conn.NewSelect().
+		Table("statuses").
+		Column("id"), "emojis", emojiID).
```
tsmethurst, line 436: I had to read this line like 3 times to understand the syntax here, maybe it's worth unpacking these functions a bit rather than calling them in line?

tsmethurst: I also think it would be useful to write a comment explaining the query a bit, cuz it's not really clear what the end result of this is, and why we want it like this

internal/db/bundb/migrations/20230724100000_emoji_cleanup.go
```diff
@@ -0,0 +1,57 @@
...
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			_, err := tx.ExecContext(ctx, "ALTER TABLE emojis ADD COLUMN cached BOOLEAN DEFAULT false")
+
+			if err != nil && !(strings.Contains(err.Error(), "already exists") || strings.Contains(err.Error(), "duplicate column name") || strings.Contains(err.Error(), "SQLSTATE 42701")) {
+				return err
+			}
+
+			if _, err := tx.NewUpdate().
+				Table("emojis").
+				Where("disabled NOT NULL AND disabled = false").
```
tsmethurst, line 38: You probably don't need this AND here, since `disabled` should always be set. 

From my db:

```
sqlite> select count(*) from emojis where disabled = false;
31000
sqlite> select count(*) from emojis where disabled is null;
0
```

internal/db/bundb/media.go
```diff
@@ -295,8 +272,55 @@ func (m *mediaDB) GetAttachments(ctx context.Context, maxID string, limit int) (
...
+	if err := q.Scan(ctx, &attachmentIDs); err != nil {
+		return nil, m.conn.ProcessError(err)
+	}
+
+	return m.GetAttachmentsByIDs(ctx, attachmentIDs)
+}
+
 func (m *mediaDB) GetAvatarsAndHeaders(ctx context.Context, maxID string, limit int) ([]*gtsmodel.MediaAttachment, db.Error) {
-	attachmentIDs := []string{}
+	var attachmentIDs []string
```
tsmethurst, line 323: is it worth making this a specific size already based on `limit`?

internal/db/bundb/media.go
```diff
@@ -295,8 +272,55 @@ func (m *mediaDB) GetAttachments(ctx context.Context, maxID string, limit int) (
...
+
+	if err := q.Scan(ctx, &attachmentIDs); err != nil {
+		return nil, m.conn.ProcessError(err)
+	}
+
+	return m.GetAttachmentsByIDs(ctx, attachmentIDs)
+}
+
+func (m *mediaDB) GetCachedAttachmentsOlderThan(ctx context.Context, olderThan time.Time, limit int) ([]*gtsmodel.MediaAttachment, db.Error) {
+	var attachmentIDs []string
```
tsmethurst, line 300: same comment re `limit` size

internal/db/bundb/media.go
```diff
@@ -295,8 +272,55 @@ func (m *mediaDB) GetAttachments(ctx context.Context, maxID string, limit int) (
...
+
+	q := m.conn.
```
tsmethurst, line 302: better use `bun.Ident` here, as we do elsewhere, to ensure cross-db compatibility.

internal/db/bundb/media.go
```diff
@@ -295,8 +272,55 @@ func (m *mediaDB) GetAttachments(ctx context.Context, maxID string, limit int) (
 	return m.GetAttachmentsByIDs(ctx, attachmentIDs)
 }
 
+func (m *mediaDB) GetRemoteAttachments(ctx context.Context, maxID string, limit int) ([]*gtsmodel.MediaAttachment, error) {
+	var attachmentIDs []string
+
+	q := m.conn.NewSelect().
+		Table("media_attachments").
+		Column("id").
+		Where("remote_url IS NOT NULL").
```
tsmethurst, line 281: same comment re bun.Ident

internal/cleaner/media.go
```diff
@@ -397,15 +389,33 @@ func (m *Media) fixCacheState(ctx context.Context, media *gtsmodel.MediaAttachme
...
 	}
 
-	// So we know this a valid cached media entry.
-	// Check that we have the files on disk required....
-	return m.checkFiles(ctx, func() error {
-		l.Debug("uncaching due to missing media")
-		return m.uncache(ctx, media)
-	},
+	// Check whether files exist.
+	exist, err := m.haveFiles(ctx,
```
tsmethurst, line 393: extremely piffling minor comment here so feel free to disregard, but i'd probs call this variable `filesExist` instead for clarity, since otherwise i'd wanna call it `exists`

internal/cleaner/media.go
```diff
@@ -397,15 +389,33 @@ func (m *Media) fixCacheState(ctx context.Context, media *gtsmodel.MediaAttachme
...
 		media.Thumbnail.Path,
 		media.File.Path,
 	)
+	if err != nil {
+		return false, err
+	}
+
+	switch {
+	case *media.Cached && !exist:
```
tsmethurst, line 402: nice! like this

internal/cleaner/emoji.go
```diff
@@ -145,22 +260,112 @@ func (e *Emoji) FixBroken(ctx context.Context) (int, error) {
...
+		return false, err
+	}
+
+	for _, status := range statuses {
+		if status.FetchedAt.After(after) {
+			l.Debug("skipping due to recently fetched status")
+		}
+	}
+
+	// This emoji is too old, uncache it.
```
tsmethurst, line 366: mood

internal/cleaner/emoji.go
```diff
@@ -145,22 +260,112 @@ func (e *Emoji) FixBroken(ctx context.Context) (int, error) {
...
+
+	// Load any related statuses using this emoji.
+	statuses, err := e.getRelatedStatuses(ctx, emoji)
+	if err != nil {
+		return false, err
+	}
+
+	for _, status := range statuses {
+		if status.FetchedAt.After(after) {
+			l.Debug("skipping due to recently fetched status")
```
tsmethurst, line 362: missing a return statement here i think

tsmethurst (commented):
Looking great so far! I wonder if, since we now are doing more work during cleanup operations, it might be worth scheduling it to run every 48 hours @ midnight instead of every 24 hours @ midnight. What do you think?

internal/regexes/regexes.go
NyaaaWhatsUpDoc: hehe i had the same realization seeing our regex :D

internal/db/bundb/search.go
NyaaaWhatsUpDoc: 100%

internal/db/bundb/status.go
NyaaaWhatsUpDoc: very true!

NyaaaWhatsUpDoc:
> Looking great so far! I wonder if, since we now are doing more work during cleanup operations, it might be worth scheduling it to run every 48 hours @ midnight instead of every 24 hours @ midnight. What do you think?

how do you feel if we leave it as 24hrs during the v0.10.0 -> v0.11.0 testing period and see how it is? as in time i would like to make this configurable anyways so for now just keeps it simpler by trying what we got :)

internal/cleaner/media.go
NyaaaWhatsUpDoc: i might leave this if that's good, the comment line and code line being the same length looks better imo. in kim's incredibly subjective vision :p

internal/db/bundb/media.go
NyaaaWhatsUpDoc: for hard-coded columns i'm not sure if we do need to do this. even in their docs they don't always use it, it seems like it is passed in when the column is an actual variable e.g a function argument that isn't always known at calling time or it is using a reserved keyword (which none of our columns / tables do)

internal/db/bundb/media.go
NyaaaWhatsUpDoc: see above :)

internal/db/bundb/media.go
tsmethurst: Ah, I meant in the 'Where' call, not the 'Column' call!

tsmethurst:
> > Looking great so far! I wonder if, since we now are doing more work during cleanup operations, it might be worth scheduling it to run every 48 hours @ midnight instead of every 24 hours @ midnight. What do you think?
> 
> how do you feel if we leave it as 24hrs during the v0.10.0 -> v0.11.0 testing period and see how it is? as in time i would like to make this configurable anyways so for now just keeps it simpler by trying what we got :)

Yeah that's fine :)

PR #1988, opened 2023-07-13
[bugfix] Set Vary header correctly on cache-control

tsmethurst:
# Description

>

This PR updates the cache-control middleware to allow callers to the middleware to set the Vary header appropriately. This fixes an issue where it was possible to get nginx to cache a 303 response for the main-key endpoint of a GtS user, because it was cacheing the redirect regardless of the Accept header value.

Also updates AP endpoints to prefer serving ActivityPub.

## Checklist

### DIFF
--- a/internal/api/auth.go
+++ b/internal/api/auth.go
@@ -43,13 +43,16 @@ func (a *Auth) Route(r router.Router, m ...gin.HandlerFunc) {
 
 	// instantiate + attach shared, non-global middlewares to both of these groups
 	var (
-		cacheControlMiddleware = middleware.CacheControl("private", "max-age=120")
-		sessionMiddleware      = middleware.Session(a.sessionName, a.routerSession.Auth, a.routerSession.Crypt)
+		ccMiddleware = middleware.CacheControl(middleware.CacheControlConfig{
+			Directives: []string{"private", "max-age=120"},
+			Vary:       []string{"Accept-Encoding"},
+		})
+		sessionMiddleware = middleware.Session(a.sessionName, a.routerSession.Auth, a.routerSession.Crypt)
 	)
 	authGroup.Use(m...)
 	oauthGroup.Use(m...)
-	authGroup.Use(cacheControlMiddleware, sessionMiddleware)
-	oauthGroup.Use(cacheControlMiddleware, sessionMiddleware)
+	authGroup.Use(ccMiddleware, sessionMiddleware)
+	oauthGroup.Use(ccMiddleware, sessionMiddleware)
 
 	a.auth.RouteAuth(authGroup.Handle)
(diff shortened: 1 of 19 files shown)


internal/api/fileserver.go
```diff
@@ -47,7 +47,9 @@ func (f *Fileserver) Route(r router.Router, m ...gin.HandlerFunc) {
 	// it expires. This ensures that clients won't cache expired
 	// links. This is done within fileserver/servefile.go.
 	if config.GetStorageBackend() == "local" || config.GetStorageS3Proxy() {
-		fileserverGroup.Use(middleware.CacheControl("private", "max-age=604800")) // 7d
+		fileserverGroup.Use(middleware.CacheControl(middleware.CacheControlConfig{
+			Directives: []string{"private", "max-age=604800"},
```
daenney, line 51: You can change `max-age=...` for `immutable` in this particular case.

internal/api/auth.go
```diff
@@ -43,13 +43,16 @@ func (a *Auth) Route(r router.Router, m ...gin.HandlerFunc) {
 
 	// instantiate + attach shared, non-global middlewares to both of these groups
 	var (
-		cacheControlMiddleware = middleware.CacheControl("private", "max-age=120")
-		sessionMiddleware      = middleware.Session(a.sessionName, a.routerSession.Auth, a.routerSession.Crypt)
+		ccMiddleware = middleware.CacheControl(middleware.CacheControlConfig{
+			Directives: []string{"private", "max-age=120"},
+			Vary:       []string{"Accept-Encoding"},
```
daenney, line 48: I think this one needs `Accept` too, as we can respond with HTML or JSON on most of these endpoints

internal/api/fileserver.go
NyaaaWhatsUpDoc: I think it's probably worth keeping the max-age for now. We did have some discussions about possible changes on media endpoints in the future.

NyaaaWhatsUpDoc:
Once that missing "Accept" @daenney pointed out is added this looks ready to go imo :)

internal/api/fileserver.go
tsmethurst: Aha, it appears that the way to go here is probably to set both max-age *and* immutable, judging by the MDN docs anyways! I will set both :)

See:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#immutable
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#avoiding_revalidation

internal/api/auth.go
tsmethurst: Ack! Will fix now

PR #1995, opened 2023-07-17
[bugfix] more robust list timeline invalidation

NyaaaWhatsUpDoc:
# Description
- make list + list-entry cache invalidation more thorough
- allow list entry invalidation by ListID + FollowID
- allow emoji invalidation by CategoryID
- use `defaultValue` when a limit of `0` is provided
- update errors being logged to DEBUG, to log to ERROR

## Checklist

### DIFF
--- a/internal/db/bundb/list.go
+++ b/internal/db/bundb/list.go
@@ -181,9 +181,25 @@ func (l *listDB) UpdateList(ctx context.Context, list *gtsmodel.List, columns ..
 }
 
 func (l *listDB) DeleteListByID(ctx context.Context, id string) error {
+	// Load list by ID into cache to ensure we can perform
+	// all necessary cache invalidation hooks on removal.
+	_, err := l.GetListEntryByID(
+		// Don't populate the entry;
+		// we only want the list ID.
+		gtscontext.SetBarebones(ctx),
+		id,
+	)
+	if err != nil {
+		if errors.Is(err, db.ErrNoEntries) {
+			// Already gone.
+			return nil
+		}
+		return err
+	}
+
 	defer func() {
-		// Invalidate all entries for this list ID.
-		l.state.Caches.GTS.ListEntry().Invalidate("ListID", id)
+		// Invalidate this list from cache.
+		l.state.Caches.GTS.List().Invalidate("ID", id)
 
 		// Invalidate this entire list's timeline.
 		if err := l.state.Timelines.List.RemoveTimeline(ctx, id); err != nil {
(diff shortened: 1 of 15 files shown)


internal/processing/report/get.go
```diff
@@ -73,22 +73,14 @@ func (p *Processor) GetMultiple(
 
 	count := len(reports)
 	items := make([]interface{}, 0, count)
-	nextMaxIDValue := ""
-	prevMinIDValue := ""
-	for i, r := range reports {
+	nextMaxIDValue := reports[count-1].ID
```
tsmethurst, line 76: For safety here it might be worth getting the count, as above, and then checking if it's == 0 and returning empty pageable response if so. Btw, directly above we're still doing `if err == db.ErrNoEntries` which should probably be replaced with `errors.Is`

internal/db/bundb/list.go
```diff
@@ -328,22 +353,20 @@ func (l *listDB) GetListEntries(ctx context.Context,
...
 func (l *listDB) GetListEntriesForFollowID(ctx context.Context, followID string) ([]*gtsmodel.ListEntry, error) {
-	entryIDs := []string{}
+	var entryIDs []string
 
 	if err := l.conn.
 		NewSelect().
 		TableExpr("? AS ?", bun.Ident("list_entries"), bun.Ident("entry")).
 		// Select only IDs from table
-		Column("entry.id").
+		ColumnExpr("?", bun.Ident("entry.id")).
```
tsmethurst, line 369: I believe the Column function already parses + properly escapes whatever you put into it, so it's probs not likely to change Column to a ColumnExpr in this way

internal/db/bundb/list.go
```diff
@@ -387,33 +408,42 @@ func (l *listDB) PopulateListEntry(ctx context.Context, listEntry *gtsmodel.List
...
-			if _, err := tx.
-				NewInsert().
-				Model(listEntry).
-				Exec(ctx); err != nil {
-				return err
-			}
+func (l *listDB) PutListEntries(ctx context.Context, entries []*gtsmodel.ListEntry) error {
+	defer func() {
+		// Collect unique list IDs from the entries.
+		listIDs := collate(func(i int) string {
```
tsmethurst, line 414: Oh! What's this? Is collate a new thing?

internal/db/bundb/list.go
```diff
@@ -428,40 +458,64 @@ func (l *listDB) DeleteListEntry(ctx context.Context, id string) error {
...
 			return err
 		}
 	}
 
 	return nil
 }
+
+// collate will collect the values of type T from an expected slice of length 'len',
+// passing the expected index to each call of 'get' and deduplicating the end result.
+func collate[T comparable](get func(int) T, len int) []T {
```
tsmethurst, line 504: oh i see!

tsmethurst:
Feel free to squerge when you're happy with this :)

internal/processing/report/get.go
NyaaaWhatsUpDoc: Ah yep in this location i should copy over the count=0 check. Will do! (and the error check too)

PR #2007, opened 2023-07-20
[feature] Lower remote media cache config duration

daenney:
# Description

The old default of 30d can lead to a lot of media getting cached and significant disk usage, even on small or single person instances. A lot of deployments decrease this value, to 15 or even less. This is less of an issue when using object storage, but for local storage which is the more popular deployment option running out of disk space is unpleasant.

With GoToSocial's aim to fit in small places, this changes the default to a much more conservative 7 days. In all likelihood people aren't scrolling that far back in their timeline so this change shouldn't result in any excessive refetching of remote media. Existing deployments will only be affected by this change if the admin hasn't already configured this value, or didn't bootstrap from the example configuration.

closes #(issue)
closes #(another issue)

## Checklist

### DIFF
--- a/docs/configuration/media.md
+++ b/docs/configuration/media.md
@@ -37,8 +37,8 @@ media-description-max-chars: 500
 #
 # If this is set to 0, then media from remote instances will be cached indefinitely.
 # Examples: [30, 60, 7, 0]
-# Default: 30
-media-remote-cache-days: 30
+# Default: 7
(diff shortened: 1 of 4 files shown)


NyaaaWhatsUpDoc:
looks good to me :) once v0.10.0 final is release will merge this :+1:

tsmethurst:
Agreed!

Issue #2003, opened 2023-07-19
[bug] Passwords > 72 chars cause havoc that needs DB access to recover


mirabilos:
### Describe the bug with a clear and concise description of what the bug is.

I’ve attempted to `admin account create` a user account for a bot, with a loooong (autogenerated, random) password. I got:

    2023/07/19 19:01:24 error executing command: error hashing password: bcrypt: password length exceeds 72 bytes

Unfortunately, neither a subsequent `admin account create` with a shorter password…

    2023/07/19 19:02:05 error executing command: username «censored» is already in use

… nor `admin account password` to set/change it…

    2023/07/19 19:02:41 error executing command: no entries

… could succeed.

----

To allow a subsequent `admin account create` to proceed, I had to hack around this on the DB level:

    gotosocial=# SELECT * FROM accounts WHERE username='«censored»';
    gotosocial=# DELETE FROM accounts WHERE id='…';

Inspection had shown that the newly created user already had a row in the `accounts` table but not in the `user` table. I hope I did not miss any others to clean up.

### What's your GoToSocial Version?

0.9.0 git-282be6f

### GoToSocial Arch

amd64 binary

### What happened?

see above

### What you expected to happen?

password hashing can take arbitrarily long inputs

### How to reproduce it?

I *was* going to say…

    $ setuidgid _gts /opt/GtS/gotosocial --config-path /opt/GtS/config.yaml admin account create --username foo --email bar@example.com --password xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

… but that yielded a different error…

    2023/07/19 19:34:46 error executing command: password is only 15% strength, try including more special characters, using uppercase letters, using numbers or using a longer password

… so I guess choose a random long pw…

    $ setuidgid _gts /opt/GtS/gotosocial --config-path /opt/GtS/config.yaml admin account create --username foo --email bar@example.com --password 70F85RjI3DPWcteNF9ACV5UO2PhCguaXZiQDVhXfhRazUWXmpYrTVfvGJfFtUW0JUv8v3JoUXxpV8hgyMhm9CYXsEAoM6trY605

tsmethurst:
> password hashing can take arbitrarily long inputs

[...]

PR #2008, opened 2023-07-20
[chore] Admin CLI + new account creation refactoring

tsmethurst:
# Description

>

This pull request does the following:

1. ~~Update `maxPasswordLength` in validation to 72 bytes. This is the limit that the bcrypt algorithm can handle without throwing an error. See https://github.com/superseriousbusiness/gotosocial/issues/2003. Regardless of whether we want to move away from bcrypt to something else to allow for longer passwords than 72 bytes, this needed to be fixed.~~ Already done in https://github.com/superseriousbusiness/gotosocial/pull/2012
2. Refactor the NewSignup function, which had a million parameters, to take a struct with sensible zero values instead.
3. Refactor various other parts of the code to use this new struct.
4. Refactor the admin account CLI functions to reduce code duplication.
5. When creating a user + account using NewSignup, ensure User doesn't already exist for the account before trying to create another one. This also makes NewSignup sort of idempotent.

## Checklist


internal/validate/formvalidation.go
```diff
@@ -32,8 +31,8 @@ import (
 )
 
 const (
-	maximumPasswordLength         = 256
-	minimumPasswordEntropy        = 60 // dictates password strength. See https://github.com/wagslane/go-password-validator
+	maximumPasswordLength         = 72 // 72 bytes is the maximum length afforded by bcrypt. See https://pkg.go.dev/golang.org/x/crypto/bcrypt#GenerateFromPassword.
```
daenney, line 34: :facepalm: :+1:

daenney: One other thing that comes to mind; we use this in a check of `len([]rune(password))`, but a rune is 4 bytes I believe. So I don't think this works as expected, unless we change the password length check to operate on bytes instead of runes.

daenney: Ah nvm, that got changed where we do `len(password)` now which should count bytes.

tsmethurst:
> My only concern with this is that this is a lot of functionality change bundled in this late in the RC-cycle. I think maybe we should set the 72 byte limit but leave the rest for after 0.10?

Mm that's a good point, I don't mind pulling out the 72 byte limit into a separate PR for now, then we can merge the rest later

Issue #1407, opened 2023-02-01
[performance] Implement retry logic for `SQLITE_BUSY` errors generated in WAL mode


tsmethurst:
We've been playing around in the last few weeks with SQLite tuning and WAL mode without shared cache.

See:

- https://www.sqlite.org/wal.html
- https://activesphere.com/blog/2018/12/24/understanding-sqlite-busy

In WAL mode, when running with more than one DB connection, we sometimes get `SQLITE_BUSY` errors, which can cause all sorts of oddness.

To resolve this, we have to write some logic somewhere in our database package to retry queries when receiving an `SQLITE_BUSY` error -- the official SQLite docs even say:

> there are some obscure cases where a query against a WAL-mode database can return [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy), so applications should be prepared for that happenstance.

For now, a decent interim solution was to always tune SQLite to use only 1 database connection, regardless of CPU count and CPU multiplier, but in future for better performance we can change this :)

NyaaaWhatsUpDoc:
worth noting it would work quite well handling this while moving away from using bun (our ORM). bun is unfortunately quite inflexible in allowing us to handle these kinds of things.

tgulacsi:
The cleanest case would be to have one goroutine handle all the writes (INSERT/DELETE/UPDATE) on one dedicated db.Conn, and all other queries (SELECT) could use any more connections.

But this assumes we don't use transactions, or the whole transaction must be processed by the writer goroutine.

And all this is only for SQLite, as PostgreSQL has lesser isolation level (READ COMMITED), and the SERIALIZABLE is opt-in.

NyaaaWhatsUpDoc:
After current open PRs have been dealt with, I'm going to take a look at maybe opening some PRs to `uptrace/bun` to expand it more for our usecase here (or at least try to). And hopefully see if there's any other performance improvements I can make to their library as those are obviously wins for us too :)

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/pull/2025

PR #2025, opened 2023-07-24
[performance] retry db queries on busy errors

NyaaaWhatsUpDoc:
## Description
- rename DBConn to WrappedDB to better follow "sql" package name conventions
- wrap `bun.DB` to perform retry on returned busy-like-errors (SQLite only, for now(?))
- removes unnecessary `db.Error` type
- makes `db.ErrNoEntries` simply a ptr to `sqlite.ErrNoRows` since that is dialect agnostic

## Checklist

### DIFF
--- a/internal/db/bundb/wrap.go
+++ b/internal/db/bundb/wrap.go
@@ -28,73 +28,76 @@ import (
 	"github.com/uptrace/bun/dialect"
 )
 
-// DBConn wrapps a bun.DB conn to provide SQL-type specific additional functionality
-type DBConn struct {
-	errProc func(error) error // errProc is the SQL-type specific error processor
-	*bun.DB                   // DB is the underlying bun.DB connection
+// WrappedDB wraps a bun database instance
+// to provide common per-dialect SQL error
+// conversions to common types, and retries
+// on returned busy errors (SQLite only for now).
+type WrappedDB struct {
+	errHook func(error) error
+	*bun.DB // underlying conn
 }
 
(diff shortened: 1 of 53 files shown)


NyaaaWhatsUpDoc:
this closes https://github.com/superseriousbusiness/gotosocial/issues/1407

internal/db/bundb/wrap.go
```diff
@@ -0,0 +1,252 @@
...
+		maxRetries = 16
+
+		// base backoff duration multiplier.
+		baseBackoff = 2 * time.Millisecond
+
+		// maximum backoff duration possible.
+		maxBackoff = baseBackoff * (1 << maxRetries)
+	)
+
+	for i := 0; i < maxRetries; i += 2 {
```
tsmethurst, line 228: does this mean the max number of retries is actually only 8, since you're incrementing i by 2 every loop? 8 is fine I think, just wanted to check that i'm understanding it correctly

tsmethurst (commented):
Looks great! A nice neat solution!

internal/db/bundb/wrap.go
NyaaaWhatsUpDoc: yeah you're right this isn't the neatest solution :thinking: i might update it to be a little clearer

tsmethurst: that would be cool, if there's a way of doing it nicely :) if it's too much of a headache, just comment it imo

PR #2031, opened 2023-07-27
[chore] Update activity dependency

tsmethurst:
Bump superseriousbusiness/activity (go-fed fork) to version including hashtags (thanks @Gabek)

This build of the library drops unused forgefed structs and interfaces, hence all the deleted lines.


daenney (approved):
`+7,812 −86,270` nice!

Test pass, so :shipit:.

Issue #2019, opened 2023-07-23
[bug] Published toots have non-clickable hashtags


ezaquarii:
### Describe the bug with a clear and concise description of what the bug is.

Published toots have non-clicable hashtags, like in this one:
https://social.etacassiopeiae.net/@ezaquarii/statuses/01H62AEBWRP45XCMN77R3X09J4

It has 2 hashtags: #email and #fediverse. First is clickable, second is not.

### What's your GoToSocial Version?

0.9.0
0.10.0

### GoToSocial Arch

amd64

### What happened?

I published a toot with 2 hashtags. First hashtag is clickable. Second hashtag is not.

Generally hashtags are not working randomly. Sometimes all work. Sometimes few, sometimes none.

### What you expected to happen?

All hashtags should be clickable.

### How to reproduce it?

One reliable method I found:

1. publish a toot using officialy supporteclient
2. delete and redraft
3. publish again

Second method: just publish a toot with multiple hashtags. Or URL and hashtags.

tsmethurst:
Heya, this is normal: hashtags are not implemented yet. They will be in the next release though :)

ezaquarii:
Thanks, that makes sense.

Out of curiosity, why some are clickable and the effect is random-ish?

tsmethurst:
> why some are clickable and the effect is random-ish

Just some weird parsing logic on our part, also fixing it on a branch rn.

Issue #1822, opened 2023-05-24
[bug] Re-used and updated hashtags (same spelling, different case) are not rendered as clickable URLs in user profiles or posts


nachtlied1:
### Describe the bug with a clear and concise description of what the bug is.

I tried to update my profile and convert all of the lower case hashtags to CamelCase, e.g. updating `#somehashtag` to `#SomeHashtag`.

Saving the updated profile *technically* worked and the change propagated as expected.

However, the 'new' CamelCase hashtags were NOT rendered as clickable URLs (e.g. https://subdomain.tld/tags/SomeHashtag)  as expected when I reloaded the profile in both the web interface and in my phone app. Rather, they were rendered as just another string.

### What's your GoToSocial Version?

0.9.0-rc2

### GoToSocial Arch

arm64 Docker

### What happened?

I looked at my account's `note` content in the DB `accounts` table and noticed that:
* one completely new hashtag that I added had an HTML `a href` tag: `<a href="https://subdomain.tld/tags/NewTag" class="mention hashtag" rel="tag nofollow noreferrer noopener" target="_blank">#<span>NewTag</span></a>`
* the updated CamelCase hashtags (which were previously all lower case) did *not* have `a href` tags. Instead they showed up as plain text: `#UpdatedCamelCaseTag`

In my logs I then saw a number of `duplicate key value violates unique constraint "tags_pkey" (SQLSTATE 23505)` errors. The number of errors aligned with the number of hashtags that I converted to CamelCase and occurred at the same time as the profile update request, so I'm assuming the two things are related:

```
May 23 05:33:33 ba1c28bdc863 gotosocial[1]: timestamp="23/05/2023 05:33:33.930" func=text.(*customRenderer).replaceHashtag level=ERROR requestID=cmshjhw804000qzv3a90 msg="error putting tags in db: ERROR: duplicate key value violates unique constraint \"tags_pkey\" (SQLSTATE 23505)"
May 23 05:33:33 ba1c28bdc863 gotosocial[1]: timestamp="23/05/2023 05:33:33.931" func=text.(*customRenderer).replaceHashtag level=ERROR requestID=cmshjhw804000qzv3a90 msg="error putting tags in db: ERROR: duplicate key value violates unique constraint \"tags_pkey\" (SQLSTATE 23505)"
[...]

Issue #967, opened 2022-11-06
[bug] Tags don't seem to be propagated to/from federated instances


fwaggle:
### Describe the bug with a clear and concise description of what the bug is.

If you create a public status on a GtS instance, and it's delivered to another instance, it will show up in the federated timeline there, but it won't show up in the tags search for any hashtags present.

The same with incoming statuses from federated persons you follow, though this is less serious as GtS doesn't have a tags endpoint yet anyway, but they're not put into the database. Local statuses are added to tags in the database correctly.

### What's your GoToSocial Version?

git 15be35673e0766d6bee398bbecd5bd9a1bb5d3b0

### GoToSocial Arch

amd64 self-built

### How to reproduce it?

Have a follower on a Mastodon instance. Watch the federated timeline and the tags search for a given tag on that instance. Create a status on GtS with that same tag, it will appear in the federated timeline but not the tag search.

### Anything else we need to know?

Inspecting the database reveals that GtS has found+parsed the tags correctly. I think this is happening because GtS isn't setting the `tag` property for hashtags correctly, and Mastodon is not parsing the hashtags from federated posts, relying instead on the tag property in the ActivityStream, but I'm not 100% sure I understand the spec correctly. :)

fwaggle:
Oh my dog... I guess this is already known. I didn't see a bug for it already, but feel free to close this if it's a dupe: https://github.com/superseriousbusiness/gotosocial/blob/15be35673e0766d6bee398bbecd5bd9a1bb5d3b0/internal/typeutils/internaltoas.go#L242

Not sure I'm smart enough to fix this one either. :)

tsmethurst:
Heya! Yeah this is known, it's just not supported yet :)

BKLronin:
This would be very helpful to discover things.

PR #2032, opened 2023-07-27
[feature] Hashtag federation (in/out), hashtag client API endpoints

tsmethurst:
# Description

>

This pull request implements the following:

1. Federate hashtags in and and out of GoToSocial correctly.
2. Client API to view hashtag timeline.
3. Client API to search for hashtags.
4. Serve stub page at web timeline endpoint (pending discussion about what to actually show at these endpoints).
5. Document hashtag usage in the user guide, and add federation docs for hashtags.

closes https://github.com/superseriousbusiness/gotosocial/issues/2019
closes https://github.com/superseriousbusiness/gotosocial/issues/1822
closes https://github.com/superseriousbusiness/gotosocial/issues/967

## Checklist


internal/web/tag.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// We only serve text/html at this endpoint.
+	if _, err := apiutil.NegotiateAccept(c, []apiutil.MIME{apiutil.TextHTML}...); err != nil {
+		apiutil.WebErrorHandler(c, gtserror.NewErrorNotAcceptable(err, err.Error()), instanceGet)
+		return
+	}
+
+	tagName, errWithCode := apiutil.ParseTagName(c.Param(apiutil.TagNameKey))
+	if errWithCode != nil {
+		apiutil.WebErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
```
NyaaaWhatsUpDoc, line 55: did you mean to use the `instanceGet` variable here? :p

tsmethurst: blam! https://github.com/superseriousbusiness/gotosocial/pull/2032/commits/d2ba9d1cdaa5b8dcc994ed48c94c36fe1dfcd571

internal/visibility/tag_timeline.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// By default we assume no auth.
+	requesterID := noauth
+
+	if requester != nil {
+		// Use provided account ID.
+		requesterID = requester.ID
+	}
+
+	visibility, err := f.state.Caches.Visibility.Load("Type.RequesterID.ItemID", func() (*cache.CachedVisibility, error) {
```
NyaaaWhatsUpDoc, line 47: considering for the tag timelineable check here we're mostly just checking boostability, and other than that rely entirely on `statusVisible()` to do the heavy lifting -- i don't think we need to use a separate type in the cache like this. we can probably just do something similar to what the `StatusBoostable` function is doing. what do you think?

NyaaaWhatsUpDoc: (doing this would mean we'd also be able to rely more heavily on existing cached `StatusVisible()` checks, and reduce the amount of items needing caching)

internal/text/replace.go
```diff
@@ -90,52 +79,44 @@ func (r *customRenderer) replaceMention(text string) string {
...
-		if err != nil {
-			if !errors.Is(err, db.ErrAlreadyExists) {
-				log.Errorf(r.ctx, "error putting tags in db: %s", err)
-				return text
+	// Append tag to result if not done already.
+	//
+	// This prevents multiple uses of a tag in
+	// the same status generating multiple
+	// entries for the same tag in result.
+	func() {
```
NyaaaWhatsUpDoc, line 103: why doing this in a sub-function out of interest? it doesn't look like it should be namespacing annything differently doing this :thinking:

internal/gtsmodel/tag.go
```diff
@@ -21,13 +21,20 @@ import "time"
...
+
+	// Original URL of a partially-filled-out tag.
+	// This SHOULD NOT be inserted in the database,
+	// cached, or serialized towards a user.
+	//
+	// It is only to be used when parsing a hashtag
+	// from a remote instance, since it may be useful
+	// to dereference latest notes from the given
+	// remote instance every now and then.
+	URL string `validate:"-" bun:"-"`
```
NyaaaWhatsUpDoc, line 39: is it worth renaming this variable a bit to emphasise it's not the same sort of URL field we find on our other models? e.g. `PartialURL` to indicate it's only set when we have the partial hashtag or something?

internal/text/replace.go
tsmethurst: Oh it's purely stylistic, I just find it a bit neater to be able to `return` early if it's a no-op instead of setting + checking a bool. Happy to change it if you prefer it the other way :)

internal/db/bundb/util.go
```diff
@@ -58,6 +59,30 @@ func whereLike(
 	)
 }
 
+// whereStartsWith is like whereLike,
+// but only searches for strings that
+// START WITH `search`.
+func whereStartsWith(
```
NyaaaWhatsUpDoc, line 65: very nitpicky one here (feel free to ignore); i was thinking that `whereStartsLike()` might work a little better as a function name to emphasise that this is using a `LIKE` under the hood. just to keep it closer to existing database syntax, and to make it easier scanning over database code and internally building a model of the query.

what do you think?

tsmethurst: Oh sure :) Will change it

internal/ap/extracthashtags_test.go
```diff
@@ -0,0 +1,70 @@
...
+	suite.Equal(true, *hashtagFediverse.Useable)
+	suite.Equal(true, *hashtagFediverse.Listable)
+
+	hashtagGoToSocial := hashtags[1]
+	suite.Equal("https://mastodon.social/tags/gotosocial", hashtagGoToSocial.URL)
+	suite.Equal("gotosocial", hashtagGoToSocial.Name)
+	suite.Equal(true, *hashtagGoToSocial.Useable)
+	suite.Equal(true, *hashtagGoToSocial.Listable)
+
+	hashtagGrüvy := hashtags[2]
```
NyaaaWhatsUpDoc, line 55: huh, didn't realise you could use non-ascii in Go variable names, cool!

internal/ap/extracthashtags_test.go
```diff
@@ -0,0 +1,70 @@
...
+	suite.Equal("https://mastodon.social/tags/this_hashtag_will_be_included_correctly", hashtagGrüvy.URL)
+	suite.Equal("grüvy", hashtagGrüvy.Name)
+	suite.Equal(true, *hashtagGrüvy.Useable)
+	suite.Equal(true, *hashtagGrüvy.Listable)
+
+	hashtagAngle := hashtags[3]
```
NyaaaWhatsUpDoc, line 61: mornin' angle

internal/api/model/tag.go
```diff
@@ -27,4 +27,8 @@ type Tag struct {
 	// Web link to the hashtag.
 	// example: https://example.org/tags/helloworld
 	URL string `json:"url"`
+	// History of this hashtag's usage.
+	// Currently just a stub, if provided will always be an empty array.
+	// example: []
+	History *[]any `json:"history,omitempty"`
```
NyaaaWhatsUpDoc, line 33: i don't know if you need to make this a ptr to slice. i think if you set `omitempty` then a nil slice should just be omitted afaik. unless there was a different reason for making it a ptr ofc  in which case ignore this!

internal/db/bundb/tag.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, m.conn.ProcessError(err)
+		}
+
+		return &tag, nil
+	}, id)
+}
+
+func (m *tagDB) GetTagByName(ctx context.Context, name string) (*gtsmodel.Tag, error) {
+	// Normalize 'name' string.
+	name = strings.TrimSpace(name)
```
NyaaaWhatsUpDoc, line 57: for the name normalization here do we not want to  use the `util.Normalize...()` function we're using in other places for the hashtags?

internal/federation/dereferencing/status.go
```diff
@@ -400,6 +403,44 @@ func (d *deref) fetchStatusMentions(ctx context.Context, requestUser string, exi
...
+	status.TagIDs = make([]string, len(status.Tags))
+
+	for i := range status.Tags {
+		// Get or create tag with this name.
+		var (
+			tag = status.Tags[i]
+			err error
+		)
+
+		if tag, err = d.state.DB.GetOrCreateTag(ctx, tag.Name); err != nil {
```
NyaaaWhatsUpDoc, line 417: how do you feel about moving the create logic here instead? handling both a `Get` and `Create` in the same function in the database a bit like overcomplicating the database logic with code that should maybe happen further up. it would also more fit the logic in the other status `fetch___()` functions. what do you think?

docs/user_guide/posts.md
```diff
@@ -241,6 +241,26 @@ which will be rendered as:
...
+
+Most ActivityPub server implementations like Mastodon and similar only group together **Public** posts by the hashtags they use, but there is no guarantee about that. Generally speaking then, it is better to only use hashtags for Public visibility posts where you want the post to be able to spread more widely than it would otherwise. A good example of this is the `#introduction` hashtag, which tends to be used by new accounts who want to introduce themselves to the fediverse!
+
+Including hashtags in your post works like most other social media software: just add a `#` symbol before the word you want to use as a hashtag.
+
+Some examples:
+
+* `#introduction`
+* `#Mosstodon`
+* `#LichenSubscribe`
```
NyaaaWhatsUpDoc, line 256: ha!

internal/text/replace.go
NyaaaWhatsUpDoc: ah no i see what you mean now. got it :+1:

NyaaaWhatsUpDoc:
finished reviewing this now. looks really great!! once the comments have been addressed, jokey ones aside :p, this should be ready to go!

internal/ap/extracthashtags_test.go
tsmethurst: or ur devil

internal/gtsmodel/tag.go
tsmethurst: Since we weren't really using it anywhere, I just removed it: https://github.com/superseriousbusiness/gotosocial/pull/2032/commits/e8aae842a0e422f7867590d199b34a104d39deb9

We can always add it back in again later if we need it!

internal/db/bundb/util.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2032/commits/0dd992d9d80311836f8259a885f91ee0ba496d03

internal/api/model/tag.go
tsmethurst: Ah cool, I didn't know that! Will change it

tsmethurst: Ah wait no, I remember now, there is a reason to do it this way: we actually want to serialize the empty array when serializing the tag in a search result, so as not to break any client expectations, but to omit the empty array when serializing the tag as attached to statuses. So it's for mastodon API compliance reasons

NyaaaWhatsUpDoc: *robot house voice* masto-donnnn!

internal/db/bundb/tag.go
tsmethurst: You mean `text.NormalizeHashtag`? That function does only a very specific type of normalization: 1. removing the `#` character and 2. squashing characters and diacritics into single utf characters.

It doesn't touch casing, and there's a specific reason for that: when we insert a reference to the hashtag into the text when parsing new statuses created on the instance, we want to use the case the hashtag was posted in, because the author may have used upper camel case or some other significant casing to make it easier to read. So for example `#SomeHashtag` in the raw status text becomes `<a href="https://example.org/tags/somehashtag" class="mention hashtag" rel="tag">#<span>SomeHashtag</span></a>` in the finished HTML version of the status. If `text.NormalizeHashtag` also lowercased the tag, it would be `<a href="https://example.org/tags/somehashtag" class="mention hashtag" rel="tag">#<span>somehashtag</span></a>`.

When we select or insert the hashtag from/into the database though, we want to always use lowercase so that we don't end up with multiple differently-cased versions of that hashtag in the db.

So essentially we have two similarly-named normalization functions that do very different things. Perhaps this is a naming issue...

internal/federation/dereferencing/status.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/2032/commits/503a020d113655e1ccd79174d0d27ebc0cf2d169

internal/db/bundb/tag.go
NyaaaWhatsUpDoc: Ah nope that does make sense now, no worries :)

PR #2046, opened 2023-08-01
[feature] Configurable timezone in containers

### LINKED ISSUE #2045: [bug] Docker compose not support timezone env

Docker compose not support timezone env

0.10.1-SNAPSHOT 

### GoToSocial Arch

amd64 docker

### What happened?

I live in UTC+8, when i post a toot to Gotosocial through the app, time is always 8 hours slow.

Docker support zimezone env
We can set it in the `docker-compose.yml` or in the `Dockerfile`
There's a better way to use like some other containers:

```yml
    environment:
      - TZ=Asia/Shanghai
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - /etc/timezone:/etc/timezone:ro
```

### How to reproduce it?

Docker run an instance in other timezone.

I tried to set the local time of the server, but it didn't take effect for this container.

eallion:
# Description

This pull request fixes #2045 ,you can set timezone now.

There is a tested Docker image:
```bash
docker pull eallion/gotosocial:test
```

- https://hub.docker.com/r/eallion/gotosocial/tags

closes #2045 

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,8 @@ RUN yarn install --cwd web/source && \
 
 # stage 3: build the executor container
 FROM --platform=${TARGETPLATFORM} alpine:3.17.2 as executor
+RUN apk add --no-cache tzdata
+ENV TZ="UTC"
 
 # switch to non-root user:group for GtS
 USER 1000:1000
--- a/example/docker-compose/docker-compose.yaml
+++ b/example/docker-compose/docker-compose.yaml
@@ -13,6 +13,7 @@ services:
       GTS_DB_ADDRESS: /gotosocial/storage/sqlite.db
       GTS_LETSENCRYPT_ENABLED: "false"
       GTS_LETSENCRYPT_EMAIL_ADDRESS: ""
+      #TZ: "UTC"
       ## For reverse proxy setups:
       # GTS_TRUSTED_PROXIES: "172.x.x.x"
     ports:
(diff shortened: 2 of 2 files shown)


daenney (approved):
Hiya, thanks for this! If I remember right:

* You can only set environment variables through `-e VAR=value` if an equivalent `VAR` was defined in the Dockerfile, so we indeed need to add it
* Containers don't inherit the host timezone, so they would already be using UTC meaning that the default of `TZ=UTC` wouldn't suddenly cause the log timestamps to change if anyone pulled an updated container

(I reworded the issue subject to match our convention, as that ends up in the changelog).

tsmethurst:
To me this looks good! We should still make a few other changes around it but the core is good so I will merge it. Thank you!

tsmethurst:
Oh balls, I remember now why we *didn't* do stuff like this: we build docker containers for multiple architectures, using a single architecture. If we execute stuff inside containers during the build process, it will fail because we're trying to execute (for example) the 32 bit version of `apk` from a 64 bit machine. So I'm going to have to roll back this change until we can figure out an architecture-agnostic way of setting the timezone inside the container...

tsmethurst:
Opening this again because of https://github.com/superseriousbusiness/gotosocial/pull/2048

I'll investigate a way of doing this without having to call `apk` inside the container.

PR #2051, opened 2023-08-01
[feature] simpler cache size configuration

### LINKED ISSUE #2049: [feature] simpler cache configuration
currently our cache configuration is very complex. it requires twiddling the dials of maximum number of objects per object type, which unless you have some knowledge of our codebase must be quite intimidating ...

the proposed change is that instead we have a singular cache memory limit target variable, and we *aim* to keep our caches under this size using some estimates of average model size during runtime. we can have finer tuned dials that allow twiddling cache memory allocation of say accounts : statuses, but we should leave these out of the example config and have most people rely on the defaults of them. in time we can add docs for this. the cache memory limit target variable should be enough for most.

NyaaaWhatsUpDoc:
## Description

- updates caches to remove TTL checking + sweeping (still keeps to a maximum size)
- removes the many old, confusing cache configuration variables
- adds a singular `cache.memory-target` configuration variable to control cache sizes
- adds "hidden" mem ratio per cache configuration, not included in the example config

closes #2049 

## Checklist


NyaaaWhatsUpDoc:
Converting to draft as there's still some more specifics to figure out.

NyaaaWhatsUpDoc:
basically, when caches get full then every single write operation also involves an evict which is more expensive. i'm experimenting adding a regular sweep (not TTL, in fact this simpler method has some advantages over TTL) that checks caches are within 80% of their capacity once a minute, else truncating them. and this would all happen from a singular background goroutine which is a bonus too.

example/config.yaml
```diff
@@ -231,111 +231,13 @@ db-sqlite-cache-size: "8MiB"
...
-    webfinger-max-size: 250
-    webfinger-ttl: "24h"
-    webfinger-sweep-freq: "15m"
+  # cache.memory-target sets a target limit that
+  # the application will try to keep it's caches
+  # within. This is based on estimated sizes of
+  # in-memory objects, and so NOT AT ALL EXACT.
+  # Examples: ["100MiB", "200MiB", "500MiB", "1GiB"]
+  # Default: "200MiB"
+  memory-target: "200MiB"
```
tsmethurst, line 240: Damn so much cleaner!

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	codeberg.org/gruf/go-runners v1.6.1
 	codeberg.org/gruf/go-sched v1.2.3
 	codeberg.org/gruf/go-store/v2 v2.2.2
+	github.com/DmitriyVTitov/size v1.5.0
```
tsmethurst, line 19: Don't forget to add this new dependency to the readme with the others

example/config.yaml
NyaaaWhatsUpDoc: right!!

internal/config/defaults.go
```diff
@@ -126,111 +126,42 @@ var Defaults = Configuration{
...
+
+		// NOTE: these weighted by a totally
+		// assorted mixture of priority, and
+		// manual twiddling to get the generated
+		// cache capacity ratios within normal
+		// amounts dependent on relative sizes.
+		//
+		// when TODO items in the size.go source
+		// file have been addressed, these should
+		// be able to make some more sense :D
```
tsmethurst, line 139: Could you add a little more explanation somewhere of what these numbers mean? How they relate to each other etc?

cmd/gotosocial/action/server/server.go
```diff
@@ -117,6 +119,13 @@ var Start action.GTSAction = func(ctx context.Context) error {
 	state.Workers.Start()
 	defer state.Workers.Stop()
 
+	// Add a task to the scheduler to sweep caches.
+	// Frequency = 1 * minute
+	// Threshold = 80% capacity
+	sweep := func(time.Time) { state.Caches.Sweep(80) }
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: Yep! it basically does this internally:
```golang
		// Calculate number of cache items to truncate.
		max := (perc / 100) * float64(c.Cache.Cap())
		diff := c.Cache.Len() - int(max)
		if diff <= 0 {
			return
		}

		// Truncate by calculated length.
		items = c.truncate(diff)
```

internal/cache/size.go
```diff
@@ -0,0 +1,500 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/config"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/id"
+)
+
+const (
+	// example data values.
+	exampleID   = id.Highest
+	exampleURI  = "https://social.bbc/users/ItsMePrinceCharlesInit"
+	exampleText = `
```
tsmethurst, line 37: Having a normal one were we?

NyaaaWhatsUpDoc: so normal you wouldn't believe it

cmd/gotosocial/action/server/server.go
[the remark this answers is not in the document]
tsmethurst: Nice!

PR #2055, opened 2023-08-01
[bugfix] handle HEAD requests more elegantly

### LINKED ISSUE #2016: [bug] All HTTP HEAD requests get 404

GtS seems to respond with a 404 to all HEAD requests, including requests for resources that do exist. This is *technically* incorrect behavior and in theory (but not in practice in any way I am aware of) could break some monitoring systems, caching proxies, CDNs, etc.

The technically correct behavior if we're not handling HEADs is to respond with 405 to them.

v0.10.0

### GoToSocial Arch

amd64 source build

### What happened?

Someone joined the GtS help chat reporting their webfinger address was 404ing. While investigating I found their testing method was `curl -I` which uses a head request. The same URL returns a 200 OK if you use a GET request.

The HEAD request should return either a 200 OK with the response headers only, or a 405 Invalid Method to note clearly GtS does not service HEAD requests.

### How to reproduce it?

Hit GtS with any HEAD request, at either the SSL terminator or the service itself.

`curl -I https://gts.superseriousbusiness.org` will return a 404; dropping the `-I` will return a 200.

In my usual style this is a compliance/technicality nitpick, but it could become a problem one day, or is one in a way we're not yet aware of.

NyaaaWhatsUpDoc:
# Description

Updates our HTTP router handle requests with unsupported methods for that route, which means unless it's a HEAD request on a fileserver endpoint it returns a 405 Method Not Allowed.

closes #2016

## Checklist

### DIFF
--- a/internal/router/router.go
+++ b/internal/router/router.go
@@ -181,6 +181,7 @@ func New(ctx context.Context) (Router, error) {
 	// create the actual engine here -- this is the core request routing handler for gts
 	engine := gin.New()
 	engine.MaxMultipartMemory = maxMultipartMemory
+	engine.HandleMethodNotAllowed = true
 
 	// set up IP forwarding via x-forward-* headers.
 	trustedProxies := config.GetTrustedProxies()


tsmethurst:
Nice!

PR #2060, opened 2023-08-03
[performance] add caching of status fave, boost of, in reply to ID lists

NyaaaWhatsUpDoc:
# Description

- updates `go-cache` library to use more fine-grained zero value detection
- adds caching of status fave, boost of and in reply to ID slices
- adds caching of fetching a boosted status
- updates `wipeStatus()` to use a multierror, to ensure full wipe

The diff for this ended up shorter than I thought for the amount of work that went in... This PR is only as short as it is as a culmination of other efforts I put into #2049 #2027.

## Checklist


internal/gtserror/multi.go
```diff
@@ -19,16 +19,13 @@ package gtserror
...
 )
 
 // MultiError allows encapsulating multiple
 // errors under a singular instance, which
 // is useful when you only want to log on
 // errors, not return early / bubble up.
-type MultiError struct {
-	e []error
-}
+type MultiError []error
```
NyaaaWhatsUpDoc, line 28: i only did this so the multi error tests could be part of their own separate test package (otherwise they get compiled in even during production). if you'd prefer it be `type MultiError struct{ E []error }` in order to keep the tests as they are i can also do it that way instead. i don't really mind either way!

tsmethurst: No problem!

internal/db/bundb/status.go
```diff
@@ -96,6 +95,20 @@ func (s *statusDB) GetStatusByURL(ctx context.Context, url string) (*gtsmodel.St
 	)
 }
 
+func (s *statusDB) GetStatusBoost(ctx context.Context, boostOfID string, byAccountID string) (*gtsmodel.Status, error) {
```
tsmethurst, line 98: Is it possible for an account to boost a status multiple times? I think it *might* be, so there's a possibility this query could select multiple statuses. Could you double check this?

NyaaaWhatsUpDoc: I'll double check, but I always thought we were just unboosting and reboosting when that happened :thinking:

internal/db/bundb/status.go
```diff
@@ -245,11 +258,7 @@ func (s *statusDB) PopulateStatus(ctx context.Context, status *gtsmodel.Status)
 		}
 	}
 
-	if err := errs.Combine(); err != nil {
-		return gtserror.Newf("%w", err)
-	}
-
-	return nil
+	return errs.Combine()
```
tsmethurst, line 261: This squashes out the `getStatus: ` prefix in the returned error. Is this what we want? It's fine if so, I just wanna make sure.

internal/db/bundb/status.go
```diff
@@ -506,25 +515,17 @@ func (s *statusDB) GetStatusChildren(ctx context.Context, status *gtsmodel.Statu
...
-		NewSelect().
-		TableExpr("? AS ?", bun.Ident("statuses"), bun.Ident("status")).
-		Column("status.id").
-		Where("? = ?", bun.Ident("status.in_reply_to_id"), status.ID)
-	if minID != "" {
-		q = q.Where("? > ?", bun.Ident("status.id"), minID)
-	}
-
-	if err := q.Scan(ctx, &childIDs); err != nil {
-		if err != sql.ErrNoRows {
```
tsmethurst, line 521: Wow damn that's old code! Glad to see this changed :)

internal/db/bundb/status.go
NyaaaWhatsUpDoc: i updated the multierror to add the caller prefixes each time `.Append()` is called :)

internal/db/bundb/status.go
```diff
@@ -506,25 +515,17 @@ func (s *statusDB) GetStatusChildren(ctx context.Context, status *gtsmodel.Statu
...
-			log.Errorf(ctx, "error getting children for %q: %v", status.ID, err)
-		}
+	childIDs, err := s.getStatusReplyIDs(ctx, status.ID)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		log.Errorf(ctx, "error getting status %s children: %v", status.ID, err)
 		return
 	}
 
 	for _, id := range childIDs {
+		if id <= minID {
```
tsmethurst, line 525: So just for my own understanding: we now fetch (and cache) *all* status reply IDs first, and then page through them here, instead of doing paging at the db query level as before?

internal/db/bundb/status.go
```diff
@@ -553,48 +554,80 @@ func (s *statusDB) statusChildren(ctx context.Context, status *gtsmodel.Status,
...
 }
 
-func (s *statusDB) CountStatusReblogs(ctx context.Context, status *gtsmodel.Status) (int, error) {
-	return s.db.
-		NewSelect().
-		TableExpr("? AS ?", bun.Ident("statuses"), bun.Ident("status")).
-		Where("? = ?", bun.Ident("status.boost_of_id"), status.ID).
-		Count(ctx)
+func (s *statusDB) CountStatusReplies(ctx context.Context, statusID string) (int, error) {
+	statusIDs, err := s.getStatusReplyIDs(ctx, statusID)
```
tsmethurst, line 566: Nice, much cleaner :)

internal/db/bundb/status.go
NyaaaWhatsUpDoc: Yes, and remember that's top level status replies (i.e. only those with in_reply_to_id set to the current parent) not the entire chain

internal/db/bundb/status.go
```diff
@@ -553,48 +554,80 @@ func (s *statusDB) statusChildren(ctx context.Context, status *gtsmodel.Status,
...
+	statusIDs, err := s.getStatusBoostIDs(ctx, statusID)
+	return len(statusIDs), err
+}
 
-	return s.db.Exists(ctx, q)
+func (s *statusDB) getStatusBoostIDs(ctx context.Context, statusID string) ([]string, error) {
+	return s.state.Caches.GTS.BoostOfIDs().Load(statusID, func() ([]string, error) {
+		var statusIDs []string
+
+		// Status reply IDs not in cache, perform DB query!
```
tsmethurst, line 618: reply IDs -> boost IDs

(nitpick, sorry, i know)

internal/db/bundb/status.go
NyaaaWhatsUpDoc: :D

internal/db/bundb/status.go
NyaaaWhatsUpDoc: on it bawss

internal/db/bundb/status.go
NyaaaWhatsUpDoc: right it was a lil dusty!

tsmethurst (commented):
Looking good!

Just to check, when a new status is put in the database, does it also invalidate the boost IDs and reply IDs cache? Likewise for faves: does a new fave invalidate the faves/likes cache entries for that status ID?

I didn't see it in the code anywhere, so just wanted to verify that that's happening as expected.

internal/db/bundb/status.go
tsmethurst: Got it :) Thanks!

internal/db/bundb/status.go
tsmethurst: I think most clients force you to unboost before you can reboost, but if you do it via the API then all bets are off :P

NyaaaWhatsUpDoc: i guess in that case do we just want to return the latest of the boosts by this user?

NyaaaWhatsUpDoc:
> 

When we new status or favourite is put in the database, it's wrapped in the cache.Store() function, which calls all the required invalidate hooks :)

internal/db/bundb/status.go
NyaaaWhatsUpDoc: (i did just check and i think there are some cases of multiple boosts of a status by the same account ID)

NyaaaWhatsUpDoc: i.e. we could just do:
```
			return s.newStatusQ(status).
				Where("boost_of_id = ?", boostOfID).
				Where("account_id = ?", byAccountID).
				Order("created_at DESC").
				Limit(1).
				Scan(ctx)
```

tsmethurst: I think I'd prefer to have this be `GetStatusBoosts(ctx context.Context, boostOfID string, byAccountID string) ([]*gtsmodel.Status, error)` and leave it up to the caller to deal with the slice, what do you think?

tsmethurst: While we're on the topic, could you check this for faves as well? I have a suspicion it might be the same dealio.

NyaaaWhatsUpDoc: honestly being able to have multiple boosts / faves for the same status feels in contrary to how boosts / faves are presented client-side, and how people count them. we can just update our behaviour to do the same as we do for bookmarks / pinning

tsmethurst: Ah sure, that's also a fine option :) Let's just do the limit 1 thing for now then and make the fave/boost fix in another PR!

PR #2069, opened 2023-08-05
[chore] Remove go-playground/validator

daenney:
# Description

It turns out we're not actually using the validator code. This is a remnant from when we intended to use it, but the presence of it and its struct tags creates the illusion we're validating a lot of things we're not. It resulted in some confusion when we were trying to figure out language validation.

Remove all this code, so that only the validation functions from the validate package we actually use remain. I'm not touching the struct tags in the migrations in order to avoid things potentially thinking migrations need to be re-run.

## Checklist


internal/api/model/domain.go
```diff
@@ -25,7 +25,7 @@ import "mime/multipart"
 type Domain struct {
 	// The hostname of the domain.
 	// example: example.org
-	Domain string `form:"domain" json:"domain" validate:"required"`
+	Domain string `form:"domain" json:"domain"`
```
tsmethurst, line 28: This tag should stay, since it pertains to gin form parsing validation, not struct validation

tsmethurst:
Looks good, with that one comment aside, thanks for doing this :)

internal/api/model/domain.go
daenney: Ah, right. How nice of people to use the same struct tags :sweat_smile:

tsmethurst: Right!

PR #2071, opened 2023-08-06
[docs] fix postgres create database command

ekeih:
# Description

The current command `create database gotosocial with locale C.UTF-8 template template0;` fails because the locale has to be quoted:
```
postgres=# create database gotosocial with locale C.UTF-8 template template0;
ERROR:  syntax error at or near "."
LINE 1: create database gotosocial with locale C.UTF-8 template temp...
```

Tested manually that it works with the quotes and the official postgres examples also use quotes around the locales: https://www.postgresql.org/docs/current/sql-createdatabase.html

## Checklist

### DIFF
--- a/docs/configuration/database.md
+++ b/docs/configuration/database.md
@@ -37,7 +37,7 @@ Then you should have already created database `gotosocial` in Postgres, and give
 The psql commands to do this will look something like:
 
 ```psql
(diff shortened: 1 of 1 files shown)


tsmethurst:
Oh thanks :) Missed that completely

PR #2080, opened 2023-08-07
[chore] Use generic pointer function

daenney:
# Description

>

This replaces the different $TypePtr functions with a generic implementation. I would've kept it in testrig, but it turns out we have some stray `trueBool` and `falseBool` functions outside of tests too. So I opted for `util` instead.

## Checklist


testrig/testmodels.go
```diff
@@ -851,9 +827,9 @@ func NewTestAttachments() map[string]*gtsmodel.MediaAttachment {
 					Height:    404,
 					Size:      290880,
 					Aspect:    1.78217821782178,
-					Duration:  Float32Ptr(15.033334),
-					Framerate: Float32Ptr(30.0),
-					Bitrate:   Uint64Ptr(1206522),
+					Duration:  util.Ptr[float32](15.033334),
+					Framerate: util.Ptr[float32](30.0),
+					Bitrate:   util.Ptr[uint64](1206522),
```
daenney, line 832: These are the only ones where Go can't quite infer the right type, it considers them as float64 and int instead. It should be able to with some improvements in 1.21 I think.

daenney:
It's a minor cleanup, but it does have the advantage that in some places we no longer pull the testrig package in for tests, as they were only accessing the `TrueBool`/`FalseBool` functions.

PR #2081, opened 2023-08-07
[chore] Add Feditext as recommended client

daenney:
# Description

With Feditext now accepting beta users, this adds it as the third client to recommend so we have web and the dominant mobile platforms covered.

This also removes the screenshots from the README, because it became a mess trying to add a third one. Either the cells become very narrow, or the table doubles in height. As the UI may also change over time, it might be better to point folks at the apps instead who'll hopefully have up to date screenshots in their storefronts.

## Checklist


tsmethurst:
Nice one :)

Issue #1094, opened 2022-11-20
[performance] replace bun relational queries with separate db calls to better utilitise caching


NyaaaWhatsUpDoc:
An example of this can be seen here: https://github.com/superseriousbusiness/gotosocial/blob/9be16852f212f56637fd4a41f3361d30ce7deb5c/internal/db/bundb/account.go#L151

Database areas still requiring this
- [x] Accounts (partially done)
- [x] Emoji
- [x] Media
- [x] Mention
- [x] Relationships (partially done)
- [x] Statuses (partially done)
- [x] User

As part of this work you should also take into account the following issue #1095, so that when filling in related models you are only filling in the bare minimum instead of a full tree of populated models.

NyaaaWhatsUpDoc:
Account blocks (part of relationships) are currently being handled in #1085

tsmethurst:
We've sort of been working on this one again lately. Should we add it to the 0.11.0 milestone?

PR #2091, opened 2023-08-09
[performance] remove last of relational queries to instead rely on caches

NyaaaWhatsUpDoc:
# Description

- Removes relational part of queries for status selects and user selects
- Adds new separate `applicationDB{}` with its own caching

closes https://github.com/superseriousbusiness/gotosocial/issues/1094

## Checklist

### DIFF
--- a/internal/middleware/tokencheck.go
+++ b/internal/middleware/tokencheck.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/gin-gonic/gin"
 	"github.com/superseriousbusiness/gotosocial/internal/db"
-	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
 	"github.com/superseriousbusiness/gotosocial/internal/oauth"
 	"github.com/superseriousbusiness/oauth2/v4"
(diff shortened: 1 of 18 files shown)


internal/middleware/tokencheck.go
```diff
@@ -125,15 +124,16 @@ func TokenCheck(dbConn db.DB, validateBearerToken func(r *http.Request) (oauth2.
 			log.Tracef(ctx, "authenticated client %s with bearer token, scope is %s", clientID, ti.GetScope())
 
 			// fetch app for this token
-			app := &gtsmodel.Application{}
-			if err := dbConn.GetWhere(ctx, []db.Where{{Key: "client_id", Value: clientID}}, app); err != nil {
+			app, err := dbConn.GetApplicationByClientID(ctx, clientID)
```
tsmethurst, line 127: Much better, my goodness those old `GetWhere` things are clunky looking.

tsmethurst (approved):
Looks good to me! Feel free to squerge! :partying_face:

NyaaaWhatsUpDoc:
> Looks good to me! Feel free to squerge! partying_face

damn that was quick :O

tsmethurst:
> HOW IS THIS NOT PASSING I LITERALLY RAN THIS LOCALLY

stinky...

NyaaaWhatsUpDoc:
> > HOW IS THIS NOT PASSING I LITERALLY RAN THIS LOCALLY
> 
> stinky...

downvote

NyaaaWhatsUpDoc:
Right, so the issue this has revealed, is that when we delete statuses they are then not able to be included in the report after performing the suspension. (which i think is a similar case on Mastodon?). Previously we were saved by the test being just fast enough to beat-out generating the report before the enqueued status deletes, but now the test is failing the expected JSON.

So the solution here is either we update the test's expected JSON output, or we ignore deleting statuses if they're in a report.

PR #2109, opened 2023-08-13
[fix] Update CSP header for blob images (upload preview) and dev livereload

f0x52:
Updates the new CSP header to allow image blobs (used by the settings for previewing avatar/header/emoji uploads), and in debug mode, allowing the livereload websocket connection.
Also blocks `object-src` because that's a recommendation I found somewhere


internal/middleware/extraheaders.go
```diff
@@ -54,9 +54,11 @@ func BuildContentSecurityPolicy() string {
 		// Debug is enabled, allow
 		// serving things from localhost
 		// as well (regardless of port).
-		policy += " localhost:*"
+		policy += " localhost:* ws://localhost:*"
 	}
 
+	policy += "; img-src 'self' blob:; object-src 'none'"
```
daenney, line 60: I think this can break with the S3 policy, which also sets `img-src 'self'` plus the S3 endpoint. I think multiple `img-src` is valid, but idk what happens if we set `'self'` multiple times.

f0x52: hmmm having the duplicate is fine I think, as they are different policies (separated by `;`), but in this case we would have to add the `blob:` to line 93 as well, because otherwise that policy is more restrictive and thus no longer includes the `blob:` exception https://csplite.com/csp234/

due to the early returns it's a bit messy I think to totally do this without duplication, hm

f0x52: think I found a reasonably clean way to code it

daenney (approved):

PR #2120, opened 2023-08-15
[chore] ensure worker contexts have request ID

NyaaaWhatsUpDoc:
# Description
Ensures that worker contexts have request ID fields set.

## Checklist

### DIFF
--- a/internal/processing/workers/fromfediapi.go
+++ b/internal/processing/workers/fromfediapi.go
@@ -44,13 +44,15 @@ type fediAPI struct {
 	account    *account.Processor
 }
 
-func (p *Processor) EnqueueFediAPI(ctx context.Context, msgs ...messages.FromFediAPI) {
-	log.Trace(ctx, "enqueuing")
-	_ = p.workers.Federator.MustEnqueueCtx(ctx, func(ctx context.Context) {
+func (p *Processor) EnqueueFediAPI(cctx context.Context, msgs ...messages.FromFediAPI) {
+	_ = p.workers.Federator.MustEnqueueCtx(cctx, func(wctx context.Context) {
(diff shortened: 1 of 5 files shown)


internal/processing/workers/fromfediapi.go
```diff
@@ -44,13 +44,15 @@ type fediAPI struct {
 	account    *account.Processor
 }
 
-func (p *Processor) EnqueueFediAPI(ctx context.Context, msgs ...messages.FromFediAPI) {
-	log.Trace(ctx, "enqueuing")
```
NyaaaWhatsUpDoc, line 48: i removed the excess logging here (and in the ClientAPI enqueuer) as we already have a log message immediately in the directly called functions below

daenney:
This is cool and it works for the RequestID, but we'd lose the OpenTelemetry TraceID/SpanIDs etc. which would be nice to keep.

What would you say if instead we create a "value only" context to wrap it in instead. Something like:

```go
type WorkerContext struct {
	context.Context
}

func (WorkerContext) Deadline() (time.Time, bool) { return time.Time{}, false }
func (WorkerContext) Done() <-chan struct{}       { return nil }
func (WorkerContext) Err() error                  { return nil }

func WorkerValuesContext(ctx context.Context) context.Context {
	return WorkerContext{ctx}
}
```

daenney:
> How about [`175d6ea` (#2120)](https://github.com/superseriousbusiness/gotosocial/pull/2120/commits/175d6ea1932bf0099197846f74579b043080c6c4)? As we need to preserve the worker pool passed done/deadline/err, but the values of the original caller context are required

Nice!

daenney (approved):

PR #2121, opened 2023-08-15
[feature] list commands for both attachment and emojis

### LINKED ISSUE #2118: [feature] emojis are not included in admin media commands

When I run `gotosocial admin media list-local` my custom emojis are not listed at all. The same seems to happen with `gotosocial admin media list-remote`. This can be a problem for people performing backups using these commands.

v0.11.0-rc3

### GoToSocial Arch

arm64

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

rdelaage:
# Description

This pull request adds commands to list both attachment and emojis.

closes #2118 

## Checklist


tsmethurst:
Hiya! Sorry to keep you waiting, we've been trying to fix bugs from 0.11.0 so haven't been doing much reviewing of new PRs lately. I think in this case it's not so easy to add tests for our CLI tooling, so don't worry about it for now.

Would it be OK with you if I pushed a couple changes to your branch to fix the linter and make some small documentation changes?

daenney:
I took a look at this, and there's two things that we might want to consider:

* We're using media (can be either singular or plural) and emojis (plural), but then attachment (singular). We should be consistent here and since we expect more than 1, I think attachment**s** is the way to go
* The commands themselves are becoming a bit of a mouthful, "admin media list attachments remote" etc. I'm not sure what to do about that but I wonder if we should structure that differently

tsmethurst:
Hey @daenney and @rdelaage !

I pushed a couple commits to this branch to reduce the long command paths, and updated documentation for it. Lmk what you think :)

daenney (approved):

tsmethurst:
> Looks good to me, great work! The behavior for remote media changed, URL was returned before and local path are shown now. Could it be a problem?

Ah thanks, didn't realize that was the case. I wonder why we were ever returning remote URL :thinking: 

In any case, I think it makes more sense to return the storage path. We'll just have to remember to put a note about this when we release with this feature!

daenney:
> Looks good to me, great work! The behavior for remote media changed, URL was returned before and local path are shown now. Could it be a problem?

Good point. I think that's a good change actually. There's not a ton of point to knowing the remote URL, since if you want to do anything with that image it's probably deleting the locally cached copy, or using it as an exclusion list for backups. Not quite sure why I decided to return the URL originally for remote media.

Issue #1102, opened 2022-11-20
[enhancement/refactor] replace `Instance.Terms` with a list of InstanceRules


noerw:
### Motivation
I want to define rules for my instance and show them on the landing page.

### Current Status
As of 0.5.2 GtS stores an attribute `.Terms` for terms of service on instances. The attribute is never populated, as it does not exist in the federation API for remote instances, and neither in config file nor the `/admin` GUI for the local instance.

### Context
In v4.0 mastodon added a `rules` field to it's instance endpoint on the client API, serving the same purpose:
- https://docs.joinmastodon.org/entities/Instance/#rules
- https://docs.joinmastodon.org/entities/Rule/

### Proposal
In the spirit of keeping the APIs compatible-ish, it would make sense to adopt this and drop `.Terms`, which was never in any mastodon client API afaik.

Example use case for compatible APIs in this case: enable sites that scrape instances to be compared by their rules to work with GtS.

### Discussion

It should be discussed if GtS should implement this the same way as mastodon - they attach an ID to each rule, which is not used anywhere in the API surface and seems of little use. W.r.t the DB model, this would mean not having to work a separate table for instance rules, and simply doing array serialization with bun (`bun:",array"`) on the instances table. If not using a separate table, the API could still expose mastodon's model with the IDs being nulled.

The only case for storing rules with IDs is to future-proof GtS for compatibility with new mastodon APIs that need rule IDs as parameters. But make of that what you want, I'm new to gts development ;)

---

If you think this is a meaningful change, i'll be happy to work on implementing this + displaying rules on the landing page.

PR #2125, opened 2023-08-16
[feature] Instance rules

f0x52:
# Description

Milestone 10
c. Add rules table to database, update API to allow admins to set, update, delete individual rules
d. Page for showing instance rules / terms and conditions
e. Update admin panel -- allow admins to set, update, delete individual rules

resolves https://github.com/superseriousbusiness/gotosocial/issues/1102
tests/documentation to be added here or handled in a separate PR

## Checklist

### DIFF
--- a/internal/api/model/report.go
+++ b/internal/api/model/report.go
@@ -54,8 +54,8 @@ type Report struct {
 	StatusIDs []string `json:"status_ids"`
 	// Array of rule IDs that were submitted along with this report.
 	// Will be empty if no rule IDs were submitted.
-	// example: [1, 2]
-	RuleIDs []int `json:"rule_ids"`
+	// example: ["01GPBN5YDY6JKBWE44H7YQBDCQ","01GPBN65PDWSBPWVDD0SQCFFY3"]
+	RuleIDs []string `json:"rule_ids"`
 	// Account that was reported.
 	TargetAccount *Account `json:"target_account"`
 }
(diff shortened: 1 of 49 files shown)


internal/db/bundb/migrations/20230817174700_add_report_rule_ids.go
f0x52, line 1: not quite sure if this migration is correct, needs this specialcasing because sqlite does not have an array datatype where postgres does

internal/api/model/report.go
f0x52, line 1: still need to see how well this works with existing clients, the Mastodon API is a bit of a mess on rule ID's being strings or integers; https://docs.joinmastodon.org/entities/Rule/ says they are "Type: String (cast from integer, but not guaranteed to be a number)".
Meanwhile https://docs.joinmastodon.org/methods/reports/ mentions `rule_ids` to be "Array of Integer", except the Mastodon frontend sends ["1", "2"] etc, so just reusing the String datatypes returned by /api/v1/instance/rules.

Currently we use UUID's for rule IDs, which aren't parsable as Integers. We could use the `.Order` field instead though, which are guaranteed to be unique too

internal/db/bundb/rule.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if err := q.Scan(ctx); err != nil {
+		return nil, r.db.ProcessError(err)
+	}
+
+	return rules, nil
+}
+
+func (r *ruleDB) PutRule(ctx context.Context, rule *gtsmodel.Rule) error {
+	if r.db.Dialect().Name() == dialect.SQLite {
```
f0x52, line 91: again a bit of a weird workaround for sqlite limitations, sqlite has a [different implementation](https://www.sqlite.org/autoinc.html) of autoincrementing, and thus doesn't seem to be supported by the bundb `autoincrement` tag

internal/db/bundb/migrations/20230817174700_add_report_rule_ids.go
tsmethurst: It looks OK to me, and seems to work alright :)

internal/api/model/report.go
tsmethurst: Ack :+1: Let's run it for a while and see if we run into any issues.

Issue #2129, opened 2023-08-17
[bug] Ratelimit is not expressed as Mastodon, should be ISO8601 not Epoch


duponin:
### Describe the bug with a clear and concise description of what the bug is.

I'm trying to FediFetcher's, however it ends in ratelimit and gets fatal error because ratelimit isn't expressed as Mastodon.
It expects ratelimit to be ISO8601, but GoToSocial send it as Unix Epoch.

### What's your GoToSocial Version?

0.11.0

### GoToSocial Arch

arm64

### What happened?

GoToSocial ratelimit is a Unix epoch:

```
X-RateLimit-Reset: 1692259524
```

### What you expected to happen?

Mastodon ratelimit is ISO8601:

```
X-RateLimit-Reset: 2023-08-17T08:42:22+00:00
```

### How to reproduce it?

It's fully explained in FediFetcher's issue: https://github.com/nanos/FediFetcher/issues/78

NyaaaWhatsUpDoc:
Rate limit header in seconds is standard behaviour, if they don't support that then that's a bug on their end.

tsmethurst:
Mm, I think it would be less fragile behavior for a client app to be able to parse either epoch time or html datetime, which is indeed based on ISO8601 (as per https://developer.mozilla.org/en-US/docs/Web/HTML/Date_and_time_formats anyway).

A little search around for x-limit-reset shows that lots of applications seem to use one or the other format for this, with no clear standard saying which it ought to be -- at least, I can't find an RFC defining this. Some even provide number of seconds until the reset occurs, so that's a third format!

In this particular case, since we're trying to keep parity with the Mastodon API, we should probably try to return the same format, so I for one would be cool with changing this.

nanos:
Just to chime in here that I agree that FediFetcher should probably accept either, and I'll try to address that when I got a moment.

Just a minor correction: The format that Mastodon actually returns their `X-RateLimit-Reset` in is slightly different from what @duponin described (which is my fault: I had given him this incorrect info). Mastodon actually provides this without the time zone offset, but simply with `Z` instead:

```
X-RateLimit-Reset: 2023-08-17T08:42:22Z
```

mirabilos:
@nanos the `Z` there is the timezone offset

daenney:
[...]

PR #2151, opened 2023-08-23
[feature] Add rate limit exceptions option, use ISO8601 for rate limit reset

tsmethurst:
# Description

>

This pull request reworks our rate limiter a bit, bringing most of the logic from the previous middleware into our own code, and retaining just the limiter and storage, while adding comments to help explain what's happening, and TESTS!

Along the way, it also adds in a config option to allow rate limit exceptions, and changes the rate limit reset value to an ISO8601 timestamp, instead of unix seconds.

closes https://github.com/superseriousbusiness/gotosocial/issues/2129

## Checklist


NyaaaWhatsUpDoc:
This looks really great so far! Just one suggestion (which I am currently figuring out exactly myself right now...). I'll comment again when I have the suggestion properly put together, but to sum it up it's about using the faster+newer `net/netip` library instead of the older `net` library for address + prefix parsing :)

tsmethurst:
> using the faster+newer net/netip library instead of the older net library for address + prefix parsing

Right! At one point yesterday I had a crack at using that library instead, but there was some specific reason why it didn't work. I can't remember now what it was though :thinking:

internal/middleware/ratelimit_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+	type rlTest struct {
+		limit        int
+		exceptions   []string
+		clientIP     string
+		shouldPanic  bool
+		shouldExcept bool
+	}
+
+	for _, test := range []rlTest{
```
NyaaaWhatsUpDoc, line 56: love some good table driven tests :ok_hand:

NyaaaWhatsUpDoc:
Here's a diff :)

```
diff --git a/internal/middleware/ratelimit.go b/internal/middleware/ratelimit.go
index 602d46cf..a59a3e60 100644
--- a/internal/middleware/ratelimit.go
+++ b/internal/middleware/ratelimit.go
@@ -20,12 +20,12 @@ package middleware
 import (
 	"net"
 	"net/http"
+	"net/netip"
 	"strconv"
 	"time"
 
 	"github.com/gin-gonic/gin"
-	"github.com/superseriousbusiness/gotosocial/internal/log"
 	"github.com/ulule/limiter/v3"
 	"github.com/ulule/limiter/v3/drivers/store/memory"
@@ -61,15 +61,10 @@ func RateLimit(limit int, exceptions []string) gin.HandlerFunc {
 		},
 	)
 
-	// Parse rate limit exceptions into CIDRs.
-	exceptionCIDRs := make([]*net.IPNet, len(exceptions))
-	for i, e := range exceptions {
-		_, p, err := net.ParseCIDR(e)
-		if err != nil {
-			log.Panicf(nil, "could not parse rate limit exception %s as *net.IPNet: %q", e, err)
-		}
-
-		exceptionCIDRs[i] = p
+	// Convert exceptions IP ranges into prefixes.
+	exceptPrefs := make([]netip.Prefix, len(exceptions))
+	for i, str := range exceptions {
+		exceptPrefs[i] = netip.MustParsePrefix(str)
 	}
 
 	// It's prettymuch impossible to effectively
@@ -87,20 +82,26 @@ func RateLimit(limit int, exceptions []string) gin.HandlerFunc {
 		// Use Gin's heuristic for determining
 		// clientIP, which accounts for reverse
 		// proxies and trusted proxies setting.
-		clientIP := net.ParseIP(c.ClientIP())
+		clientIP := netip.MustParseAddr(c.ClientIP())
 
 		// Check if this IP is exempt from rate
 		// limits and skip further checks if so.
-		for _, cidr := range exceptionCIDRs {
-			if cidr.Contains(clientIP) {
+		for _, prefix := range exceptPrefs {
+			if prefix.Contains(clientIP) {
 				c.Next()
 				return
 			}
 		}
 
-		// Mask only IPv6 IPs (see above).
-		if len(clientIP) == net.IPv6len {
-			clientIP = clientIP.Mask(ipv6Mask)
+		if clientIP.Is6() {
+			// Convert to "net" package IP for mask.
+			asIP := net.IP(clientIP.AsSlice())
+
+			// Apply coarse IPv6 mask.
+			asIP = asIP.Mask(ipv6Mask)
+
+			// Convert back to netip.Addr from net.IP.
+			clientIP, _ = netip.AddrFromSlice(asIP)
 		}
 
 		// Fetch rate limit info for this (masked) clientIP.

```

example/config.yaml
```diff
@@ -848,6 +848,31 @@ advanced-cookies-samesite: "lax"
...
+#    and you're all using the same router/NAT, so you all have
+#    the same IP address, and you keep rate limiting each other.
+#
+# 3. You mostly use your own home internet to access your instance,
+#    and you want to exempt your home internet from rate limiting.
+#
+# You should be careful when adjusting this setting, since you
+# might inadvertently make rate limiting useless if you set too
+# wide a range. If in doubt, be too restrictive rather than too
+# lenient, and adjust as you go.
```
daenney, line 873: Both here and in `advanced.md`, can we include an example. Like `["192.168.0.0/16"]` etc, to give people an idea of what it should look like?

internal/middleware/ratelimit.go
```diff
@@ -20,47 +20,135 @@ package middleware
...
+	// It's prettymuch impossible to effectively
+	// rate limit the immense IPv6 address space
+	// unless we mask some of the bytes.
+	//
+	// This mask is pretty coarse, and puts IPv6
+	// blocking on more or less the same footing
+	// as IPv4 blocking in terms of how likely it
+	// is to prevent abuse while still allowing
+	// legit users access to the service.
+	ipv6Mask := net.CIDRMask(64, 128)
```
daenney, line 84: I think this is perfectly reasonable. A `/64` should be the smallest subnet that gets handed out to a customer, so realistically that always present 1 user/household for residential deployments. For server deployments you'd potentially rate limit all their instances in the same block, but that too seems fine. You're still rate limiting a single customer, not a whole ISP. You can easily get a parade of /64s for your server deployments.

example/config.yaml
tsmethurst: Sure! I'll add that :)

PR #2157, opened 2023-08-24
[chore] Add script for downloading latest snapshot tar.gz

tsmethurst:
# Description

>

This pull request adds a quick and fairly dirty bash script into the `scripts` directory that folks can use to pull the latest snapshot release tar.gz from our Minio S3 bucket.

I didn't advertise it anywhere in the docs because it's not something we should really 'recommend' to people, but we can at least point it to folks who regularly run snapshots and want a very slightly easier way of doing it.

## Checklist

### DIFF
--- a/scripts/get_latest_snapshot.sh
+++ b/scripts/get_latest_snapshot.sh
@@ -21,7 +21,7 @@ GITHUB_BRANCH="main"
 GITHUB_ENDPOINT="https://${GITHUB_API_HOST}/repos/${GITHUB_ORG}/${GITHUB_REPO}/commits/${GITHUB_BRANCH}"
 
 echo "fetching latest hash from endpoint '${GITHUB_ENDPOINT}'"
-LATEST_HASH="$(curl -s "${GITHUB_ENDPOINT}" | jq -r .sha)"
+LATEST_HASH="$(curl --silent --fail --retry 5 --retry-max-time 180 --max-time 30 "${GITHUB_ENDPOINT}" | jq -r .sha)"
 echo "got latest hash = ${LATEST_HASH}"
 
 MINIO_HOST="s3.superseriousbusiness.org"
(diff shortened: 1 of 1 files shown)


scripts/get_latest_snapshot.sh
```diff
@@ -0,0 +1,33 @@
...
+echo "fetching latest hash from endpoint '${GITHUB_ENDPOINT}'"
+LATEST_HASH="$(curl -s "${GITHUB_ENDPOINT}" | jq -r .sha)"
+echo "got latest hash = ${LATEST_HASH}"
+
+MINIO_HOST="s3.superseriousbusiness.org"
+MINIO_BUCKET="gotosocial-snapshots"
+MINIO_ENDPOINT="https://${MINIO_HOST}/${MINIO_BUCKET}/${LATEST_HASH}/${GTS_FILENAME}"
+
+echo "fetching latest snapshot tar.gz from endpoint '${MINIO_ENDPOINT}'"
+wget --no-verbose -O "./${GTS_FILENAME}" "${MINIO_ENDPOINT}"
```
NyaaaWhatsUpDoc, line 32: Why use `wget` here when we're already using `curl` above? If it's following redirects + setting output file you want then just do something like `curl -sLo "./${GTS_FILENAME}" "${MINIO_ENDPOINT}"`

daenney: It would be nice to add something like `--fail --retry 5 --retry-max-time 600` to "both" `curl` commands to. That should make it a bit more robust in case something temporarily blips.

tsmethurst: Thank you comrades! Will fix this up tomorrow most likely :)

daenney, line 32:
```suggestion
curl --silent --fail --retry 5 --retry-max-time 600 "${MINIO_ENDPOINT}" --output "./${GTS_FILENAME}"
```

daenney (approved):

PR #2167, opened 2023-08-29
[feature] Store admin actions in the db, prevent conflicting actions

tsmethurst:
# Description

>

This pull request updates our handing of admin actions (domain block, domain unblock, account suspend, etc) in the following ways:

#### When an admin action is performed on a domain or an account, store a record of that action in the database.

Previously, we only stored admin account actions.

Now, we can also store domain block / unblock actions (for starters), as well as completion time of said actions, and any errors that occurred while the action was being processed. In future this could be used to allow admins to review a timeline of actions. This is, imo, pretty sick :sunglasses: 

#### Don't allow multiple actions that target the same resource to run at the same time.

This is achieved by storing running admin actions in a map keyed by `[TargetCategory]/[TargetID]`, and removing them from the map when they're finished.

For example, a domain block for the domain `example.org` would have the key `domain/example.org`. A domain allow for the domain `example.org` would ALSO have the key `domain/example.org`. Same goes for another domain block for that domain. This ensures that these two actions cannot run at the same time. Now, an API request to perform a second action on an existing key will just get a 409 error back that looks something like:

`AccountAction: an action (account/01H90S1CXQ97J9625C5YBXZWGT) is currently running (duration 25.566746ms) which conflicts with the attempted action: wait until this action is complete and try again`

## Checklist


internal/processing/admin/actions.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/log"
+	"github.com/superseriousbusiness/gotosocial/internal/state"
+	"golang.org/x/exp/slices"
+)
+
+func errActionConflict(action *gtsmodel.AdminAction) gtserror.WithCode {
+	err := gtserror.NewfAt(
+		4, // Include caller's function name.
```
NyaaaWhatsUpDoc, line 34: i'm so glad the error package with caller name stuff is getting so well used :blush:

internal/processing/admin/util.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+}
+
+// rangeAccounts iterates through all accounts originating
+// from the given domain, and calls the provided range
+// function on each account.
+//
+// If an error is returned while selecting accounts, the
+// loop will stop and return the error.
+func (p *Processor) rangeAccounts(
```
NyaaaWhatsUpDoc, line 85: might be worth calling this one `rangeInstanceAccounts()` :thinking:

internal/processing/admin/actions.go
```diff
@@ -0,0 +1,156 @@
...
+) gtserror.WithCode {
+	key := action.Key()
+
+	// Check if an action with
+	// this key is already running.
+	a.m.Lock()
+	running, ok := a.r[key]
+	a.m.Unlock()
+
+	if ok {
```
NyaaaWhatsUpDoc, line 75: I think there's a possibility of a race condition here in that between this unlock, and the next lock, another goroutine may have quickly added a conflicting key to the map (however unlikely). I think the solution would be to move the `a.m.Unlock()` above into the `if ok { }` clause, and remove the `a.m.Lock()` below that (since there would be no need to relock).

internal/gtsmodel/user.go
```diff
@@ -57,3 +57,23 @@ type User struct {
 	ResetPasswordSentAt    time.Time    `bun:"type:timestamptz,nullzero"`                                   // When did we email the user their reset-password email?
 	ExternalID             string       `bun:",nullzero,unique"`                                            // If the login for the user is managed externally (e.g OIDC), we need to keep a stable reference to the external object (e.g OIDC sub claim)
 }
+
+// NewSignup models parameters for the creation
+// of a new user + account on this instance.
+//
+// Aside from username, email, and password, it is
+// fine to use zero values on fields of this struct.
+type NewSignup struct {
```
NyaaaWhatsUpDoc, line 66: :eyes:

internal/gtsmodel/adminaction.go
```diff
@@ -0,0 +1,79 @@
...
+type AdminActionCategory string
+
+const (
+	AdminActionCategoryAccount AdminActionCategory = "account"
+	AdminActionCategoryDomain  AdminActionCategory = "domain"
+)
+
+// AdminActionType describes a type of
+// action taken on an entity by an admin.
+type AdminActionType string
```
NyaaaWhatsUpDoc, line 36: It's not a _huge_ deal, but in cases like these where we're essentially dealing with enums it might be more space efficient in the database to use an int type for these. Then for more user friendly output in our code itself we could have a `func (t AdminActionType) String() string` that switches over all the possible action type ints.  Same goes for the category above. I know it's not really a big deal on the relatively small table that is admin actions, but I think it's probably just a good habit to get into (and may even save some disk-space on larger tables like `statuses` with visibility for e.g.). What do you think?

internal/processing/admin/domainblock.go
```diff
@@ -75,18 +78,100 @@ func (p *Processor) DomainBlockCreate(
...
+	}
+
+	actionID := id.NewULID()
+
+	// Process domain unblock side
+	// effects asynchronously.
+	if errWithCode := p.Actions.Run(
+		ctx,
+		&gtsmodel.AdminAction{
+			ID:             id.NewULID(),
```
NyaaaWhatsUpDoc, line 161: did you mean to set the `actionID` here? :p

internal/processing/admin/actions.go
```diff
@@ -0,0 +1,156 @@
...
+		running,
+		func(a *gtsmodel.AdminAction, b *gtsmodel.AdminAction) bool {
+			return a.ID > b.ID
+		},
+	)
+
+	return running
+}
+
+// TotalRunning is a sequel to the classic
```
NyaaaWhatsUpDoc, line 148: good post :+1:

internal/processing/admin/admin.go
```diff
@@ -33,6 +34,10 @@ type Processor struct {
 	mediaManager        *media.Manager
 	transportController transport.Controller
 	emailSender         email.Sender
+
+	// admin Actions currently
+	// undergoing processing
+	Actions *Actions
```
NyaaaWhatsUpDoc, line 40: Very subjective point here, feel free to ignore it. But to fit more with how all the other processors work I'd consider making this a private member and giving access to it via function, or internalize it and maybe make the `Action.Run()` command a subcommand of the `admin.Processor`. Again though, very subjective :p

internal/processing/admin/actions.go
tsmethurst: It is very handy!

internal/gtsmodel/user.go
tsmethurst: Ah this isn't new, just moved it into here from where it was before :P

internal/processing/admin/actions.go
tsmethurst: Good movie, too!

internal/processing/admin/domainblock.go
tsmethurst: I sure did :') Well spotted! https://github.com/superseriousbusiness/gotosocial/pull/2167/commits/6cb72207e77e5eba99c2846ae11f65cfb323511a

internal/processing/admin/util.go
tsmethurst: I changed it to `rangeDomainAccounts` because `InstanceAccount` is a term we use elsewhere to mean the instance service account, what do you think? https://github.com/superseriousbusiness/gotosocial/pull/2167/commits/900f38c966d052e76e444b6cb3c9bdbec283ad95

internal/gtsmodel/adminaction.go
tsmethurst: Ahh sure, I can change it. I sort of prefer having string constants in the db for this just to make it easier when trying to debug stuff by looking in the db directly, but indeed it's more performance/storage conscious to just use ints.

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2167/commits/a561e97fa77d0d2b63effc50299d6e81337dc61f Changed! uint8 and iota, I'm fancy :sunglasses:

internal/processing/admin/domainblock.go
```diff
@@ -415,18 +382,24 @@ func (p *Processor) domainBlockSideEffects(ctx context.Context, account *gtsmode
 //
 // It should be called asynchronously, since it can take a while when
 // there are many accounts present on the given domain.
-func (p *Processor) domainUnblockSideEffects(ctx context.Context, block *gtsmodel.DomainBlock) {
+func (p *Processor) domainUnblockSideEffects(
+	ctx context.Context,
+	adminAcct *gtsmodel.Account,
```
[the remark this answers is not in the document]
tsmethurst: Whoops we did not! Was just leftover from before I think. Changed :) https://github.com/superseriousbusiness/gotosocial/pull/2167/commits/edef481445c7dc92fa959c5702b206bdc02f4aa2

internal/processing/admin/admin.go
tsmethurst: Ahh sure why not, will change it :)

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2167/commits/e277460e28096b56d622514eeeb23e3475bfcdad

PR #2169, opened 2023-08-31
[bugfix] wrap bun.Tx to add our own error processing

NyaaaWhatsUpDoc:
# Description

Until now we have just been wrapping the bun database to add our own error processing hook, but we were forgetting the transactions! (agghhh!!!) Bun transactions are now properly wrapped to give our own error processing so that we may do nice, reasonable things like `if errors.Is(err, db.ErrAlreadyExists)` within transactions.

I desperately need to try adding SQLITE_BUSY retries into the modernc.org/sqlite package, and for the error processing, figuring out how to hook that in as a neater upstream fix to `uptrace/bun`... Because this wrapping situation is getting out of hand.

## Checklist


tsmethurst:
Nice nice nice :)

PR #2171, opened 2023-08-31
[bugfix] move SQLite pragmas into connection string

NyaaaWhatsUpDoc:
# Description

As the title says really! Moves the SQLite connection pragmas into the connection string, just to ensure that all our connection preferences are set on every created new connection, considering that it's a little unclear what data is shared across SQLite connections. These are set using the `_pragma` URI query key as specified in the `modernc.org/sqlite.Driver{}.Open()` documentation.

An example of a connection string with this:
`timestamp="01/09/2023 09:02:20.915" func=bundb.sqliteConn level=INFO msg="connected to SQLITE database with address file:/data/sqlite.db?_pragma=busy_timeout%28300000%29&_pragma=journal_mode%28WAL%29&_pragma=synchronous%28NORMAL%29&_pragma=cache_size%28-8192%29&_txlock=immediate"`

## Checklist

### DIFF
--- a/internal/db/bundb/bundb.go
+++ b/internal/db/bundb/bundb.go
@@ -423,6 +423,50 @@ func deriveBunDBPGOptions() (*pgx.ConnConfig, error) {
 // buildSQLiteAddress will build an SQLite address string from given config input,
 // appending user defined SQLite connection preferences (e.g. cache_size, journal_mode etc).
 func buildSQLiteAddress(addr string) string {
+	// Notes on SQLite preferences:
+	//
+	// - SQLite by itself supports setting a subset of its configuration options
+	//   via URI query arguments in the connection. Namely `mode` and `cache`.
+	//   This is the same situation for the directly transpiled C->Go code in
+	//   modernc.org/sqlite, i.e. modernc.org/sqlite/lib, NOT the Go SQL driver.
+	//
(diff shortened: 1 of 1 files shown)


internal/db/bundb/bundb.go
```diff
@@ -462,49 +420,107 @@ func deriveBunDBPGOptions() (*pgx.ConnConfig, error) {
...
+	//   in that it will attempt to retry the query until the `busy_timeout` time is
+	//   reached. And for whatever reason (:shrug:) SQLite is very particular about
+	//   setting this BEFORE the `journal_mode` is set, otherwise you can end up
+	//   running into more of these `SQLITE_BUSY` return codes than you might expect.
+	//
+	// - One final thing (I promise!): `SQLITE_BUSY` is only handled by the internal
+	//   `busy_timeout` handler in the case that a data race occurs contending for
+	//  table locks. THERE ARE STILL OTHER SITUATIONS IN WHICH THIS MAY BE RETURNED!
+	//  As such, we use our wrapping DB{} and Tx{} types (in "db.go") which make use
+	//  of our own retry-busy handler.
```
tsmethurst, line 469: excellent explanation, thanks for writing this up (and thanks from future you and me as well, reading this in two years, who would have had no idea why we did things this way if it wasn't for this comment)

PR #2172, opened 2023-08-31
[feature] Don't trace non-existing routes

daenney:
# Description

>

`c.FullPath()` is the empty string if a request doesn't match any route on our mux. In those cases, there's no value in emitting a trace. The trace will be empty, containing no other information beyond the fact that we didn't match a route. Since Gin breaks off the processing early we don't need to trace this request as it won't do anything and consumes no further resources.

The 404 will still be emitted by our logs and will be visible from a reverse proxy too.

## Checklist

### DIFF
--- a/internal/tracing/tracing.go
+++ b/internal/tracing/tracing.go
@@ -117,6 +117,15 @@ func InstrumentGin() gin.HandlerFunc {
 	)
 	propagator := otel.GetTextMapPropagator()
 	return func(c *gin.Context) {
+		spanName := c.FullPath()
+		// Do not trace a request if it didn't match a route. This doesn't omit
+		// all 404s as a request matching /:user for a user that doesn't exist
+		// still matches the route
+		if spanName == "" {
+			c.Next()
+			return
+		}
+
 		c.Set(tracerKey, tracer)
 		savedCtx := c.Request.Context()
 		defer func() {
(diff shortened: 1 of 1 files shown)


daenney:
One thing that this does change is that since we don't create a span anymore, the `log.Hook` from earlier won't be able to extract a `traceID` and set that in the logs. I think this is fine, since in this case we're not tracing anything. However, if we received the request from our reverse proxy and the proxy itself already created a span, there will be a traceID and it should still end up in our log lines.

daenney:
One thing I found a little surprising is that Gin doesn't immediately abort the request, despite there not being a path+method match. It seems to still execute the whole middleware chain. An alternative to doing it this way would be to have an abort middleware where we do the `c.FullPath() == ""` check and then `c.Abort(); return`.

daenney:
I've been running with this on my own instance for a few days, and it works as expected. There's no traces for "no route match" or "no method match" anymore. As a consequence, I get a bit more useful data now since more useful requests are now part of the sample.

tsmethurst:
> One thing I found a little surprising is that Gin doesn't immediately abort the request, despite there not being a path+method match. It seems to still execute the whole middleware chain. I suppose that can be useful for other things though, like ensuring rate limiting still takes place.

Yeah! I always found that behavior a little bit odd as well. There's probably some setting for it tucked into Gin somewhere...

tsmethurst:
Nice one!

PR #2181, opened 2023-09-04
[bugfix] fix checks for deref the same status descendants / ascendants

NyaaaWhatsUpDoc:
# Description

See title. Also some little refactoring to increase code reuse, reduce log string length and hopefully provide more useful information in the case of tracing.

## Checklist

### DIFF
--- a/internal/federation/dereferencing/thread.go
+++ b/internal/federation/dereferencing/thread.go
@@ -446,15 +446,14 @@ func getAttachedStatusCollection(status ap.Statusable) (page ap.CollectionPageab
 		return nil, ""
 	}
 
-	// Get the JSONLD ID of collection
-	pageID := page.GetJSONLDId()
-	if pageID == nil {
-		// MUST have an ID.
+	if pageID := page.GetJSONLDId(); pageID != nil {
+		// By default use collection JSONLD ID
+		return page, pageID.Get().String()
+	} else if statusID := status.GetJSONLDId(); statusID != nil {
+		// Else, if possible use status JSONLD ID
+		return page, statusID.Get().String()
+	} else {
+		// MUST have some kind of ID
 		return nil, ""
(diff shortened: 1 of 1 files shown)


internal/federation/dereferencing/thread.go
```diff
@@ -274,141 +301,139 @@ stackLoop:
...
 					if !gtserror.Unretrievable(err) {
 						l.Errorf("error dereferencing remote status %s: %v", itemIRI, err)
 					}
-
 					continue itemLoop
 				}
 
 				if statusable == nil {
-					// Already up-to-date.
+					// already up-to-date.
```
tsmethurst, line 357: Already up to date because it's been previously recently dereferenced right (hence why statusable is nil)? Probs worth noting that here, since it's not immediately obvious unless you already know how getStatusByURI works

NyaaaWhatsUpDoc: Ah yep will fill that out :+1:

internal/federation/dereferencing/thread.go
```diff
@@ -274,141 +301,139 @@ stackLoop:
...
+				page := getAttachedStatusCollection(statusable)
+				if page == nil {
 					continue itemLoop
 				}
 
 				// Put current and next frame at top of stack
 				stack = append(stack, current, &frame{
-					statusIRI:  itemIRI,
-					statusable: statusable,
+					pageURI: status.URI,
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: i mean technically we are extracting a first page of collection items (according to the library we're using). i'm not really sure what else i'd use for this since it is the `frame.pageURI` is the URI of the currently set `frame.page`  which explains it well in the context of the frame :thinking: 

any other ideas??

PR #2182, opened 2023-09-05
[chore] much improved paging package

NyaaaWhatsUpDoc:
# Description

- updates paging package to contain response packaging
- updates paging package to be much more flexible for use on any paged endpoint
- updates account blocks endpoint + relevant functions to use updated paging objects/funcs

## Checklist


internal/paging/parse.go
```diff
@@ -0,0 +1,111 @@
...
+	return &Page[string]{
+		Min:   MinShortcodeDomain(min),
+		Max:   MaxShortcodeDomain(max),
+		Limit: limit,
+	}, nil
+}
+
+// ParseLimit parses the limit query parameter from a request context, returning BadRequest on error parsing and _default if zero limit given.
+func ParseLimit(c *gin.Context, _default int) (int, gtserror.WithCode) {
+	const min, max = 2, 100
```
tsmethurst, line 89: Not sure about using these min max values here rather than passing them in depending on endpoint as we were before. The issue is that different endpoints have different max values (according to the mastodon API). So i'd rather it was possible to pass these in still.

internal/paging/parse.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+)
+
+// ParseIDPage parses an ID Page from a request context, returning BadRequest on error parsing,
+// and a nil page if no request parameters are set couplied with defaultlimit=0. i.e. setting a
+// defaultlimit value will enforce paging for the endpoint upon which this request is incoming from.
+func ParseIDPage(c *gin.Context, defaultlimit int) (*Page[string], gtserror.WithCode) {
+	// Extract request query params.
+	sinceID := c.Query("since_id")
```
tsmethurst, line 32: These query param consts are already defined in internal/api/utils, so probably best to reuse them from there, I think.

internal/paging/page.go
```diff
@@ -0,0 +1,299 @@
...
+
+import (
+	"net/url"
+	"strconv"
+	"strings"
+
+	"golang.org/x/exp/slices"
+)
+
+type Page[T comparable] struct {
```
tsmethurst, line 28: Are there any cases yet where this isn't `Page[string]`? If not, I'd tend to drop the generics to keep things a bit simpler.

internal/paging/page.go
```diff
@@ -0,0 +1,299 @@
...
+	Max Boundary[T]
+
+	// Limit will limit the returned
+	// page of items to at most 'limit'.
+	Limit int
+}
+
+// GetMin is a small helper function to return minimum boundary value (checking for nil page).
+func (p *Page[T]) GetMin() T {
+	if p == nil {
```
tsmethurst, line 42: I'm a bit confused by these nil checks. I'd assume if p (as a pointer to a Page) is nil, then this function call would return a nil pointer exception. Is that not the case?

internal/paging/boundary.go
```diff
@@ -0,0 +1,133 @@
...
+// using either the `since_id`,"DESC" name,ordering or
+// `min_id`,"ASC" name,ordering depending on which is set.
+func MinID(minID, sinceID string) Boundary[string] {
+	switch {
+	case minID != "":
+		return Boundary[string]{
+			Name:  "min_id",
+			Value: minID,
+
+			// with "min_id", return
```
tsmethurst, line 32: This isn't quite correct:

The ordering of items in a page should *always* be descending (newest -> oldest).

The difference between min_id and since_id in queries is that with since_id, you're requesting whatever from the top as long as it's older than since_id, and you want the first page of results. This is like pulling up on a timeline to refresh it; you see if anything new has popped in by getting the first page. If nothing new has popped in since since_id, it's an empty result.

With min_id, on the other hand, you're explicitly skipping a bunch of pages by requesting the page immediately *before* / *above* min_id. This is useful in cases where you were paging up through your timeline but didn't hit the top yet (newest posts).

In both cases, though, you want results descending, same as you'd want the order of words on a page to still be descending (left-to-right, top-to-bottom -- in english at least) if you were flicking through a book.

internal/paging/page.go
```diff
@@ -0,0 +1,299 @@
...
+	default:
+		return 0
+	}
+}
+
+// PageAsc will page the given slice of input according
+// to the receiving Page's minimum, maximum and limit.
+// NOTE THE INPUT SLICE MUST BE SORTED IN ASCENDING ORDER
+// (I.E. OLDEST ITEMS AT LOWEST INDICES, NEWER AT HIGHER).
+func (p *Page[T]) PageAsc(in []T) []T {
```
tsmethurst, line 105: Given my comment about always wanting pages to be returned in descending order, I don't think this function is necessary. Unless there's something in the masto API where something is ordered oldest-newest instead of newest-oldest, but I don't *think* that's the case.

internal/paging/page.go
NyaaaWhatsUpDoc: Nope you can call the function associated with ptr type even if it is nil :)

`func (t *Type) DoSomething(i int)`

is essentially just doing

`func DoSomething(t *Type, i int)`

so it's only if the ptr type is dereferenced does it actually do something to it

see: https://go.dev/play/p/_sEPQjlA1E0

tsmethurst: Dang :O Cool, thanks!

internal/paging/boundary.go
NyaaaWhatsUpDoc: I think I perhaps explained this in the comments correctly, or didn't update it after actually getting the image of it correct in my head.

In the code where I end up using this (in the branch this was pulled from), the order only determines how we _select_ the items from the database. Though clearly I need to update these code comments here. (and the function paging functions as in the comment below)

internal/paging/page.go
NyaaaWhatsUpDoc: Good point. I did go in thinking there may be an int type or something that required paging but as you say everything we're paging over at the moment is strings.

internal/paging/parse.go
NyaaaWhatsUpDoc: I find that when it comes to small easy-to-check strings like these, using constants for them makes them both longer and harder to read (you then have to follow the definition to actually see what it's providing, plus they don't get syntax color coded in the same way). If this becomes the defacto package for paging that we use, and these constants are only used for paging, surely  only having them here achieves the same thing :thinking: since they'd only be defined / used once, in this location

internal/paging/page.go
NyaaaWhatsUpDoc: this basically allows us to use the same code across the place without endless nil checks and these functions will take into account zero paging being provided. an example being a utility function i made for query selects:

```golang
// scanQueryPage appends paging parameters to a SELECT query based on given `page` and scanned `col` name. This automatically handles
// the case of nil `page, adding` min, max, limit and order to the query. Default ordering is DESC, and the returned slice is ALWAYS
// returned in descending order, even if the database query itself was ascending, as that's how our codebase handles model slices.
func scanQueryPage[T comparable](ctx context.Context, q *bun.SelectQuery, page *paging.Page[T], col bun.Ident) ([]T, error) {
	// Zero value of T.
	var zero T

	// Get paging parameters.
	min := page.GetMin()
	max := page.GetMax()
	limit := page.GetLimit()
	order := page.GetOrder()

	// Preallocate a destination slice.
	slice := make([]T, 0, limit)

	if min != zero {
		// Set page min column value.
		q = q.Where("? > ?", col, min)
	}

	if max != zero {
		// Set page max column value.
		q = q.Where("? < ?", col, max)
	}

	if limit > 0 {
		// Set page limit.
		q = q.Limit(limit)
	}

	if !order.Ascending() {
		// Default is descending.
		q = q.OrderExpr("? DESC", col)
	} else {
		// Page requires ascending.
		q = q.OrderExpr("? ASC", col)
	}

	// Perform the query, scanning result into slice.
	if err := q.Scan(ctx, &slice); err != nil {
		return nil, err
	}

	if order.Ascending() {
		// If we're paging up, we still want objects
		// to be sorted by col DESC, so reverse slice.
		// https://zchee.github.io/golang-wiki/SliceTricks/#reversing
		for l, r := 0, len(slice)-1; l < r; l, r = l+1, r-1 {
			slice[l], slice[r] = slice[r], slice[l]
		}
	}

	return slice, nil
}
```

internal/paging/parse.go
NyaaaWhatsUpDoc: Fixed that now :)

internal/paging/page.go
NyaaaWhatsUpDoc: Fixed that now :)

internal/paging/boundary.go
NyaaaWhatsUpDoc: Fixed this now. The order type now more accurately explains its usage, and the `MinID()` function inside gives an ASCII representation of how the `since_id` vs `min_id` page ordering works.

internal/paging/page.go
NyaaaWhatsUpDoc: Aaaaaand fixed along with a more robust series of tests with randomly generated input

internal/paging/boundary.go
```diff
@@ -0,0 +1,135 @@
...
+
+// MinID returns an ID boundary with given min ID value,
+// using either the `since_id`,"DESC" name,ordering or
+// `min_id`,"ASC" name,ordering depending on which is set.
+func MinID(minID, sinceID string) Boundary {
+	/*
+
+				Paging with `since_id` vs `min_id`:
+
+					 limit = 4       limit = 4
```
NyaaaWhatsUpDoc, line 28: note this looks much nicer in an actual IDE which shows tabs according to how Go likes them :')

internal/paging/parse.go
tsmethurst: True true true, okay, let's leave it for now and see later then :)

internal/paging/order.go
```diff
@@ -0,0 +1,57 @@
...
+package paging
+
+// Order represents the order an input
+// page should be sorted and paged in.
+//
+// NOTE: this does not effect the order
+// of returned API results, which must
+// always be descending order.
+//
+// Damn you Mastodon API for this
```
tsmethurst, line 27: I would tend to leave this comment out :P Or just change it to 'this is in line with Mastodon API expectations' or something, since that's a bit more informative

Issue #2170, opened 2023-08-31
[feature/bugfix] Admin action for remote public key rotation


tsmethurst:
Right now, GoToSocial isn't able to support cases where remote instances have to rotate their private + public keys (using `tootctl accounts rotate`, for Mastodon) due to data leakage, bugs, data loss, other issues.

This manifests as GoToSocial returning 401s to requests from remote instances in cases where the remote instance admin has had to rotate their keys. GtS keeps the old key cached, and tries to verify the incoming signature against the old key. Since the signature has been created with the new key, GtS rejects the request because it looks like a forgery.

It would probably be a good idea to support key rotation in a basic way for now to alleviate cases like this. Perhaps we could allow an admin, via the API, to wipe all keys for accounts from a specific instance, and code some logic to refetch them again when each account is next seen. At least then, admins could try to fix federation with specific instances by wiping their keys.

Whatever logic we end up using for this could be the basis for configurable automated key rotation, but that's way out of scope for this issue.

See this thread for context: https://queer.af/@erincandescent/110983340023366569

NyaaaWhatsUpDoc:
How does Mastodon handle key rotation with secure fetch enabled? Is there an AP activity sent out indicating signature invalidation?

erincandescent:
> How does Mastodon handle key rotation with secure fetch enabled? Is there an AP activity sent out indicating signature invalidation?

Mastodon sends out an update activity for the actor containing the new key, signed with the old key

erincandescent:
The key rotation logic was [implemented here](https://github.com/mastodon/mastodon/pull/8466/files). It may have changed since

tsmethurst:
> Mastodon sends out an update activity for the actor containing the new key, signed with the old key

Interesting! OK, we should try to handle that as well then :)

hikari-no-yume:
Hi everyone. I just YOLO-upgraded my server to 4b594516ec5fe6d849663d877db5a0614de03089 and used the new UI added in #2183 to expire all the keys for @erincandescent's instance, and I am happy to confirm that this seems to have fixed the federation between my instance and theirs; users on their instance now receive new posts from me again. Thanks!

PR #2183, opened 2023-09-06
[feature] Allow admins to expire remote public keys; refetch expired keys on demand

tsmethurst:
# Description

>

This pull request adds a new field to our account model, `public_key_expires_at`.

Using this field, it's now possible to manually expire a remote account's public key, which means that the next time your server receives a signed request from that account, it will attempt to fetch a new public key by calling the remote instance, and replace the public key in the database with the new one.

Refetching looks a bit like this in the logs:

```
timestamp="06/09/2023 17:25:36.832" func=federation.(*federator).derefPubKey level=INFO requestedUsername=tobi pubKeyID=https://queer.af/actor#main-key requestID=qsa1jtwa04001ycjm0pg msg="public key was cached, but expired at 2023-09-06 15:08:42.872193 +0000 +0000, trying dereference of new public key"
timestamp="06/09/2023 17:25:36.857" func=httpclient.(*Client).DoSigned level=INFO method=GET url=https://queer.af/actor#main-key requestID=qsa1jtwa04001ycjm0pg pubKeyID=https://goblin.technology/users/tobi/main-key msg="performing request"
timestamp="06/09/2023 17:25:36.897" func=federation.(*federator).derefPubKey level=INFO requestedUsername=tobi pubKeyID=https://queer.af/actor#main-key requestID=qsa1jtwa04001ycjm0pg msg="obtained a new public key to replace expired key, caching now; authorization for this request will be attempted with both old and new keys"
```

The PR also adds a route to the settings panel that allows admins to submit an action to mark all keys from a domain as expired, as a treat.

Screenshot of the settings panel, showing the new 'keys' expiry action.

closes https://github.com/superseriousbusiness/gotosocial/issues/2170

## Checklist


internal/federation/authenticate.go
```diff
@@ -45,11 +48,45 @@ var (
 	}
 )
 
+type PubKeyResponse struct {
```
NyaaaWhatsUpDoc, line 51: Quite a subjective note here; I think ts name sort of indicates this is the response we'd get back from an endpoint which is a little confusing from the AuthenticateFederatedRequest() function.

What do you think of changing the name to something along the lines of `AuthOwner`, and remove the `PubKeyExpired` boolean in favour of an account receiver function along the lines of `func (a *Account) PubKeyExpired()`. Just helps to simplify the code here a little and reduce doubled-up data

internal/federation/authenticate.go
```diff
@@ -196,42 +265,93 @@ func (f *federator) deref(
...
-	if errWithCode == nil {
-		l.Trace("public key cached, no dereference needed")
-		return reqAcctURI, pubKey, nil
+	pubKeyResponse, errWithCode := f.derefPubKeyDBOnly(ctx, pubKeyIDStr)
+	if errWithCode != nil {
+		return nil, errWithCode
 	}
 
-	l.Trace("public key not cached, trying dereference")
+	if pubKeyResponse != nil && !pubKeyResponse.PubKeyExpired {
```
NyaaaWhatsUpDoc, line 273: Following the if else flow along here is a little confusing, what do you think of a switch instead along the lines of:
```

expired := (pubKeyRsp != nil && pubKeyRsp.Expired)

switch {
// Yet-to-be cached public key
case pubKeyRsp == nil:
    log.Trace(...)

// Existing account with useable public key
case !expired:
    log.Trace(...)
    return pubKeyRsp, nil

// Existing account with an expired public key
case expired:
    log.Infof(...)
}

// ...

pubKey, pubKeyOwner, err := parsePubKeyBytes(...)
if err != nil {
    return nil, err
}

if !expired {
    // This is a new (to us) public key and owner URI
    return PublicKeyResponse{...}, nil
}

```

Obviously with var names changed a little if you make a change with the above struct ^

internal/federation/authenticate.go
tsmethurst: Good idea! I fixed 'er up :)

internal/federation/authenticate.go
tsmethurst: Honestly I wrote this logic like 4 different ways originally and wasn't super happy with any of them :') But the switch statement does seem neater yeah, so I fixed it up :)

NyaaaWhatsUpDoc:
nice work :cowboy_hat_face:

PR #2184, opened 2023-09-06
[feature] Support OTLP HTTP, drop Jaeger

daenney:
# Description

Jaeger supports ingesting traces using the OpenTelemetry gRPC or HTTP methods. The Jaeger project has deprecated the old jaeger transport.
    
* Add support for submitting traces over HTTP
* Drop support for the old Jaeger protocol
* Upgrade the trace libraries to v1.17
    
Fixes: #2176
Fixes: #2179

## Checklist


tsmethurst:
Code looks good to me. I don't use tracing myself, but if you've tested it and it works OK for you then feel free to squerge :)

daenney:
I use the gRPC one, which continues to work. I'll see if I can find an easy way to test the HTTP transport locally.

daenney:
Alright, everything works but had to slightly fix the docs. We shouldn't be included the protocol even for the HTTP endpoint.

daenney:
One thing is that even though the transport is called `http`, it'll use HTTPS unless `tracing-insecure-transport` is `true`. We could in theory have http and https transports, or parse the endpoint instead and figure it out from there. But this keeps the symmetry with the gRPC endpoint.

PR #2186, opened 2023-09-08
[feature] add paging to account follows, followers and follow requests endpoints

NyaaaWhatsUpDoc:
# Description

- adds paging to /api/v1/accounts/{id}/follow(s|ers) endpoints
- adds paging to /api/v1/follow_requests endpoint
- consolidates a bunch of the paged ID slice fetching functions in the DB
- adds `internal/processor/common` to contain shared processor logic
- updates some of the follow functions to use this new shared processor logic
- moves block fetcher and follow requests fetcher to accounts processor (this doesn't match how  the API endpoints map-out, but it makes more sense from a processor standpoint)
- fixes the broken /api/v1/blocks endpoint :woman_facepalming: 

## Checklist


internal/api/client/accounts/followers.go
```diff
@@ -87,11 +129,25 @@ func (m *Module) AccountFollowersGETHandler(c *gin.Context) {
 		return
 	}
 
-	followers, errWithCode := m.processor.Account().FollowersGet(c.Request.Context(), authed.Account, targetAcctID)
+	page, errWithCode := paging.ParseIDPage(c,
+		1,  // min limit
+		80, // max limit
+		40, // default limit
```
daenney, line 135: These 3 values show up in a few places. Should we turn these into `paging.DefaultMinLimit`, `.DefaultMaxLimit`, `.DefaultLimit` or something along those lines? It makes it easier to change the defaults if we want to, and would avoid needing to document it at the call sites.

internal/api/client/accounts/followers.go
```diff
@@ -45,6 +53,40 @@ import (
...
+//			Return only follower accounts *IMMEDIATELY NEWER* than the given min ID.
+//			The follower account with the specified ID will not be included in the response.
+//			NOTE: the ID is of the internal follow, NOT any of the returned accounts.
+//		in: query
+//		required: false
+//	-
+//		name: limit
+//		type: integer
+//		description: Number of follower accounts to return.
+//		default: 40
```
tsmethurst, line 87: You might also considering adding the min and max values here too. Not urgent, but nice-to-have, and we should eventually do this on all endpoints if we can.

internal/db/bundb/relationship.go
```diff
@@ -365,14 +360,14 @@ func newSelectLocalFollowers(db *DB, accountID string) *bun.SelectQuery {
...
 		).
-		OrderExpr("? DESC", bun.Ident("updated_at"))
+		OrderExpr("? DESC", bun.Ident("id"))
 }
 
 // newSelectBlocks returns a new select query for all rows in the blocks table with account_id = accountID.
 func newSelectBlocks(db *DB, accountID string) *bun.SelectQuery {
 	return db.NewSelect().
 		TableExpr("?", bun.Ident("blocks")).
-		ColumnExpr("?", bun.Ident("?")).
```
tsmethurst, line 375: gonna miss this bug :bug:

internal/db/bundb/relationship.go
```diff
@@ -311,7 +306,7 @@ func newSelectFollowRequesting(db *DB, accountID string) *bun.SelectQuery {
 		TableExpr("?", bun.Ident("follow_requests")).
 		ColumnExpr("?", bun.Ident("id")).
 		Where("? = ?", bun.Ident("target_account_id"), accountID).
-		OrderExpr("? DESC", bun.Ident("updated_at"))
```
tsmethurst, line 314: Presumably replacing with ID is faster here, nice. Does it work the same? We ID follows etc using id.NewUlid or whatever it's called?

internal/db/bundb/timeline.go
```diff
@@ -114,6 +114,7 @@ func (t *timelineDB) GetHomeTimeline(ctx context.Context, accountID string, maxI
 	follows, err := t.state.DB.GetAccountFollows(
 		gtscontext.SetBarebones(ctx),
 		accountID,
+		nil, // select all
```
tsmethurst, line 117: Cool, I like this syntax

internal/db/bundb/util.go
```diff
@@ -83,6 +85,29 @@ func whereStartsLike(
...
+// loadPagedIDs loads a page of IDs from given SliceCache by `key`, resorting to `load` function if required. Uses `page` to sort + page resulting IDs.
+// NOTE: IDs returned from `cache` / `load` MUST be in descending order, otherwise paging will not work correctly / return things out of order.
+func loadPagedIDs(cache *cache.SliceCache[string], key string, page *paging.Page, load func() ([]string, error)) ([]string, error) {
+	// Check cache for IDs, else load.
+	ids, err := cache.Load(key, load)
+	if err != nil {
+		return nil, err
+	}
+
+	// Our cached / selected IDs are
```
tsmethurst, line 97: Does this mean that the SliceCache always stores things in descending order? Just checking.

internal/processing/followrequest_test.go
```diff
@@ -30,35 +30,54 @@ import (
...
 
+// TODO: move this to the "internal/processing/account" pkg
 type FollowRequestTestSuite struct {
 	ProcessingStandardTestSuite
 }
 
 func (suite *FollowRequestTestSuite) TestFollowRequestAccept() {
-	requestingAccount := suite.testAccounts["remote_account_2"]
-	targetAccount := suite.testAccounts["local_account_1"]
+	requestingAccount := suite.testAccounts["local_account_1"]
```
tsmethurst, line 39: This name change is very confusing, especially in the subsequent FollowRequest instantiation. I'd tend to change them back since I think it was clearer before (though still not VERY clear)

tsmethurst (approved):
Looks good, great work! Some comments mostly for my own understanding, but I'd be happy to merge this as-is if, after this huge effort, you don't feel like getting stuck into it again!

internal/api/client/accounts/followers.go
NyaaaWhatsUpDoc: done :)

internal/api/client/accounts/followers.go
NyaaaWhatsUpDoc: these aren't always consistent per endpoint, we're basically following Mastodon's lead here. i _think_ they generally use the same limit values for endpoints returning the same types so they could be put into one place then, but i think until we can do that with https://github.com/superseriousbusiness/gotosocial/pull/2186#discussion_r1322859056 this too we'll end up forgetting to update the docs if we update them all at once. What do you think?

internal/db/bundb/relationship.go
NyaaaWhatsUpDoc: RIP formatted format operator

internal/db/bundb/relationship.go
NyaaaWhatsUpDoc: We do indeed!

internal/federation/federatingdb/following_test.go
```diff
@@ -47,8 +47,8 @@ func (suite *FollowingTestSuite) TestGetFollowing() {
 	suite.Equal(`{
   "@context": "https://www.w3.org/ns/activitystreams",
   "items": [
-    "http://localhost:8080/users/admin",
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: I have a feeling due to paging by IDs instead of `updated_at` :thinking:

internal/db/bundb/util.go
NyaaaWhatsUpDoc: done :)

internal/processing/followrequest_test.go
NyaaaWhatsUpDoc: I was basically updating this to match the terminology we use in all other endpoints, that is the requesting account is always that of the authorized request user. Otherwise it gets a little confusing when the terminology suddenly switches for this test :thinking:

NyaaaWhatsUpDoc: I added some code comments, is that any better?

tsmethurst: Ahh okay.... Damn, naming things really is hard innit. Alright! All fine then :)

internal/api/client/accounts/followers.go
NyaaaWhatsUpDoc: I'll add an issue for this at some point today to consolidate the paging returned type constants.

PR #2187, opened 2023-09-10
[feature] Support Actor URIs for webfinger queries

daenney:
# Description

>

It's now possible to pass an Actor URI as the resource to query for when doing a webfinger query. The code now extracts the username and domain from the URI. The URI needs to be fully qualified, including having a scheme of http or https to be recognised as such.

The acct scheme is handled as we used to, including dealing with an erroneous leading @ on the username. We retain the ability to handle resources without a scheme by parsing them again with the acct scheme if the original parse failed. This can happen due to parsing ambiguities when dealing with a string like `user@domain.tld:port`.

## Checklist

### DIFF
--- a/internal/util/namestring.go
+++ b/internal/util/namestring.go
@@ -79,12 +79,8 @@ func ExtractWebfingerParts(webfinger string) (username, host string, err error) 
 		return "", "", fmt.Errorf("unsupported scheme: %s for resource: %s", u.Scheme, orig)
 	}
 
-	userDomain := strings.Split(
-		strings.TrimPrefix(
-			u.Opaque,
-			"@",
-		),
-		"@")
+	stripped := strings.TrimPrefix(u.Opaque, "@")
+	userDomain := strings.Split(stripped, "@")
 	if len(userDomain) != 2 {
 		return "", "", fmt.Errorf("failed to extract user and domain from: %s", orig)
 	}
--- a/internal/util/namestring.go
+++ b/internal/util/namestring.go
@@ -19,6 +19,7 @@ package util
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/superseriousbusiness/gotosocial/internal/regexes"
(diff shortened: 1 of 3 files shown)


internal/util/namestring.go
```diff
@@ -40,19 +41,87 @@ func ExtractNamestringParts(mention string) (username, host string, err error) {
...
+
+	if u.Scheme == "http" || u.Scheme == "https" {
+		return ExtractWebfingerPartsFromURI(u)
+	}
+
+	if u.Scheme != "acct" {
+		return "", "", fmt.Errorf("unsupported scheme: %s for resource: %s", u.Scheme, orig)
+	}
+
+	userDomain := strings.Split(
```
tsmethurst, line 82: for readability it might be better to separate this call into two separate ones, one doing the trim prefix, and one doing the split

tsmethurst:
This is looking really good! A few things that occur to me:

1. Is it possible to webfinger remote accounts too, now, using this logic, or is that (still) guarded against? I expect we guard against it but just wanna check because I can't remember.
2. Might we want to reuse some of the util functions in `internal/uris` to parse out usernames using our regexes, instead of relying on strings.Split? I'm not really that fussed myself but it may be a little bit neater and possibly faster.

daenney:
> This is looking really good! A few things that occur to me:
> 
>     1. Is it possible to webfinger remote accounts too, now, using this logic, or is that (still) guarded against? I expect we guard against it but just wanna check because I can't remember.

There's two places where this gets used. One is when we receive a webfinger request: https://github.com/superseriousbusiness/gotosocial/blob/7011f57b09fc03bde9d48fef191e904110c09c4b/internal/api/wellknown/webfinger/webfingerget.go#L71-L88

That one we check the result against Host and AccountDomain before we then try to retrieve the user by passing only the username to WebfingerGet on the processor.

The other place is in https://github.com/superseriousbusiness/gotosocial/blob/7011f57b09fc03bde9d48fef191e904110c09c4b/internal/federation/dereferencing/finger.go#L56-L59

Here we only extract the account domain from it and that one is only ever parsing `resp.Subject`, after we've already fingered the remote account. No fingering happens afterwards. The `Subject` always uses the `acct` scheme.

>     2. Might we want to reuse some of the util functions in `internal/uris` to parse out usernames using our regexes, instead of relying on strings.Split? I'm not really that fussed myself but it may be a little bit neater and possibly faster.

I went with split because I don't like using regexes unless absolutely necessary. They're useful for extracting things from more free-form text, but since the format at this point very regular and predictable, I find split ends up being easier to understand.

internal/util/namestring.go
daenney: Done!

tsmethurst:
Ahh thank you for explaining! Alright, then I think this is ready to squerge! Nice work :sunglasses:

PR #2189, opened 2023-09-10
[docs] add fail2ban example regex in the doc

OniriCorpe:
# Description

This pull request add an example of regex for fail2ban, to make it easier for novices to configure it
This regex ignores failed API attempts, as it's not relevant

I don't know if it's also applicable to SSHGuard because I don't know it
That's explain why I didn't made a SSHGuard example too

## Checklist


tsmethurst:
Oh neat, thank you! :)

Issue #2213, opened 2023-09-19
[bug] Inconsistent paging through local timeline; local timeline appears to end prematurely


tsmethurst:
Sometimes while scrolling through the local timeline, a response will be returned from `api/v1/timelines/public?local=true` that contains an empty JSON array and no `Link` header, which causes clients to assume the end of the timeline has been reached, and stop trying to page through it (and indeed they *can't* page through it, because there's no link header to use).

To users, this gives the appearance that the local timeline just ends, or that it's being cleared beyond a certain time, but this is not the case.

I think the problem lies somewhere between the public timeline function in `internal/db/bundb/timeline.go` and the code for paging + filtering that timeline in `internal/processing/timeline/public.go`.

Interesting to note that this issue has probably been around a while, but we didn't notice it yet because nobody was really using the local timeline much, since almost all GtS instances have been single-user or very-few-users until now!

benjaminbhollon:
Yeah, I figure me trying to start a relatively large server with GtS (currently alpha.polymaths.social, will move to the main domain once alpha ends) will probably uncover myriad bugs. ;)

Happy to look through any logs or do any diagnosing on this, by the way.

NyaaaWhatsUpDoc:
this bug might be a good chance to update to the latest paging code :)

benjaminbhollon:
Timeline "cleared" again. This time I can confirm there were zero posts, not just a limit to how many we could scroll through. There's one now, but for a good bit there were zero.

benjaminbhollon:
The timeline is currently empty. `https://alpha.polymaths.social/api/v1/timelines/public?limit=20&local=true` returns `[]`

tsmethurst:
^^ Found a fix :)

Turned out to be rather a simple bug -- classic oversight on my part.

benjaminbhollon:
@tsmethurst We're still getting the timeline clearing after upgrading to rc1. :(

Any diagnostic I should look into?

benjaminbhollon:
I can also open a new issue if you'd prefer that.

benjaminbhollon:
It looks like even though the `link` header is being provided correctly, clients are ignoring it and just re-requesting the same link, setting the `maxid` parameter based on the last loaded status. Semaphore, for example, is just requesting the same URL over and over, even though the `next` value is present and points to a page with content. :P

benjaminbhollon:
[...]

PR #2236, opened 2023-09-29
[bugfix] Fix paging for empty items

tsmethurst:
# Description

>

This pull request fixes a bug where paging would stop when all items in a timeline response ended up being filtered out for various reasons (visibility, blocks, etc).

Basically, we were deriving paging headers properly, but then ignoring them in the PackagePageableResponse function when empty items were passed in.

As a fun little bonus, this PR also fixes paging for the public/local timeline to use minID properly. I don't think this was causing any real issues yet, but you never know.

All of the above is more or less just a quick fix 'for now' until we convert everything to the new paging code; but since that process might take a while, it's a worthy fix to do now imo.

closes https://github.com/superseriousbusiness/gotosocial/issues/2213

## Checklist

### DIFF
--- a/internal/util/paging.go
+++ b/internal/util/paging.go
@@ -48,11 +48,6 @@ type PageableResponseParams struct {
 // a bunch of pageable items (notifications, statuses, etc), as well
 // as a Link header to inform callers of where to find next/prev items.
 func PackagePageableResponse(params PageableResponseParams) (*apimodel.PageableResponse, gtserror.WithCode) {
-	if len(params.Items) == 0 {
-		// No items to page through.
-		return EmptyPageableResponse(), nil
-	}
-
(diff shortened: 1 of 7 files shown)


internal/util/paging.go
```diff
@@ -48,11 +48,6 @@ type PageableResponseParams struct {
 // a bunch of pageable items (notifications, statuses, etc), as well
 // as a Link header to inform callers of where to find next/prev items.
 func PackagePageableResponse(params PageableResponseParams) (*apimodel.PageableResponse, gtserror.WithCode) {
-	if len(params.Items) == 0 {
```
tsmethurst, line 51: Here's the fix: nuking this check so that paging doesn't get ignored for empty items.

PR #2248, opened 2023-10-03
[chore] internal/ap: add pollable AS types, code reformatting, general niceties

NyaaaWhatsUpDoc:

### DIFF
--- a/internal/ap/activitystreams.go
+++ b/internal/ap/activitystreams.go
@@ -78,3 +78,49 @@ const (
 	// and https://www.w3.org/TR/activitystreams-vocabulary/#dfn-tag
 	TagHashtag = "Hashtag"
 )
+
+// isActivity returns whether AS type name is of an Activity (NOT IntransitiveActivity).
+func isActivity(typeName string) bool {
+	switch typeName {
+	case ActivityAccept,
+		ActivityTentativeAccept,
+		ActivityAdd,
+		ActivityCreate,
+		ActivityDelete,
+		ActivityFollow,
+		ActivityIgnore,
+		ActivityJoin,
+		ActivityLeave,
+		ActivityLike,
+		ActivityOffer,
+		ActivityInvite,
+		ActivityReject,
+		ActivityTentativeReject,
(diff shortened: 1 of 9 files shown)


internal/ap/activitystreams.go
```diff
@@ -78,3 +78,49 @@ const (
...
+		ActivityFlag,
+		ActivityDislike:
+		return true
+	default:
+		return false
+	}
+}
+
+// isIntransitiveActivity returns whether AS type name is of an IntransitiveActivity.
+func isIntransitiveActivity(typeName string) bool {
```
daenney, line 117: Can this become `return !isActivity` instead? Or is there stuff in between that doesn't fall into either category?

NyaaaWhatsUpDoc: normally i'd agree, but given how absurdly flexible ActivityStreams vocab can be i think it's worth specifically testing for types they mark as extending `IntransitiveActivity`

internal/ap/extract.go
```diff
@@ -22,18 +22,60 @@ import (
...
 	"github.com/superseriousbusiness/activity/pub"
+	"github.com/superseriousbusiness/activity/streams/vocab"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/text"
 	"github.com/superseriousbusiness/gotosocial/internal/util"
 )
 
+// ExtractObject will extract an object vocab.Type from given implementing interface.
+func ExtractObject(with WithObject) vocab.Type {
```
tsmethurst, line 39: We'll have to be wary using this, because it will return nil (i think) for object entries that are just a URL.

NyaaaWhatsUpDoc: Yes I was thinking of adding a matching `ExtractObjectIRI()` function that can be used for when this returns nil

tsmethurst (approved):
lgtm! :rainbow:

PR #2249, opened 2023-10-03
[feature] tentatively start adding polls support

NyaaaWhatsUpDoc:
# Description
- add support for Question activity in federatingdb
- update internal/ap to handle multiple activity objects
- update federatingdb code to use new ap.ExtractObjects()
- freshen the processing/status.Processor{}.Create() function ready for polls support
- support status update activities, at a very simple level
- start adding in code for some poll support in type converter
- reduces number of allocations during text formatting

this PR is a bit of an odd one of changes here-and-there. it's essentially breaking off most of the remaining poll logic that I can, without introducing the polls themselves, into a separate PR

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,7 +14,6 @@ run:
 linters:
   # enable some extra linters, see here for the list: https://golangci-lint.run/usage/linters/
   enable:
-    - forcetypeassert
     - goconst
     - gocritic
     - gofmt
--- a/internal/ap/extract.go
+++ b/internal/ap/extract.go
@@ -35,39 +35,93 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/util"
 )
 
-// ExtractObject will extract an object vocab.Type from given implementing interface.
-func ExtractObject(with WithObject) vocab.Type {
+// ExtractObjects will extract object vocab.Types from given implementing interface.
+func ExtractObjects(with WithObject) []vocab.Type {
 	// Extract the attached object (if any).
-	obj := with.GetActivityStreamsObject()
-	if obj == nil {
+	objProp := with.GetActivityStreamsObject()
+	if objProp == nil {
 		return nil
 	}
 
-	// Only support single
-	// objects (for now...)
-	if obj.Len() != 1 {
+	// Check for zero len.
+	if objProp.Len() == 0 {
 		return nil
 	}
 
-	// Extract object vocab.Type.
-	return obj.At(0).GetType()
+	// Accumulate all of the objects into a slice.
+	objTypes := make([]vocab.Type, objProp.Len())
+	for i := 0; i < objProp.Len(); i++ {
+
+		// Get object vocab type at index.
+		objType := objProp.At(i).GetType()
+		if objType == nil {
+			return nil
+		}
+
+		// Set vocab type in slice.
+		objTypes[i] = objType
+	}
+
+	return objTypes
+}
+
+// ExtractObjectIRIs will extract object IRIs from given implementing interface.
+func ExtractObjectIRIs(with WithObject) []*url.URL {
+	// Extract the attached object (if any).
+	objProp := with.GetActivityStreamsObject()
+	if objProp == nil {
+		return nil
+	}
+
+	// Check for zero len.
+	if objProp.Len() == 0 {
+		return nil
+	}
+
+	// Accumulate all of the IRIs into a slice.
+	objIRIs := make([]*url.URL, objProp.Len())
+	for i := 0; i < objProp.Len(); i++ {
+
+		// Get iter at index.
+		at := objProp.At(i)
+		if !at.IsIRI() {
+			return nil
+		}
+
+		// Set vocab type in slice.
+		objIRIs[i] = at.GetIRI()
+	}
+
+	return objIRIs
 }
 
 // ExtractActivityData will extract the usable data type (e.g. Note, Question, etc) and corresponding JSON, from activity.
-func ExtractActivityData(activity pub.Activity, rawJSON map[string]any) (vocab.Type, map[string]any, bool) {
+func ExtractActivityData(activity pub.Activity, rawJSON map[string]any) ([]vocab.Type, []any, bool) {
 	switch typeName := activity.GetTypeName(); {
 	// Activity (has "object").
 	case isActivity(typeName):
-		objType := ExtractObject(activity)
-		if objType == nil {
+		objTypes := ExtractObjects(activity)
+		if len(objTypes) == 0 {
 			return nil, nil, false
 		}
-		objJSON, _ := rawJSON["object"].(map[string]any)
-		return objType, objJSON, true
+
+		var objJSON []any
+		switch json := rawJSON["object"].(type) {
+		case nil:
+			// do nothing
+		case map[string]any:
+			// Wrap map in slice.
+			objJSON = []any{json}
+		case []any:
+			// Use existing slice.
+			objJSON = json
+		}
+
+		return objTypes, objJSON, true
 
 	// IntransitiveAcitivity (no "object").
 	case isIntransitiveActivity(typeName):
-		return activity, rawJSON, false
+		return []vocab.Type{activity}, []any{rawJSON}, true
 
 	// Unknown.
 	default:
--- a/internal/ap/extract.go
+++ b/internal/ap/extract.go
@@ -36,7 +36,7 @@ import (
 )
 
 // ExtractObjects will extract object vocab.Types from given implementing interface.
-func ExtractObjects(with WithObject) []vocab.Type {
+func ExtractObjects(with WithObject) []TypeOrIRI {
 	// Extract the attached object (if any).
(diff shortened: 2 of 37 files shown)


.golangci.yml
```diff
@@ -14,7 +14,6 @@ run:
 linters:
   # enable some extra linters, see here for the list: https://golangci-lint.run/usage/linters/
   enable:
-    - forcetypeassert
```
tsmethurst, line 17: Mm, good idea to remove this. It was getting pretty annoying and not really helping us much.

internal/ap/extract.go
```diff
@@ -35,39 +35,93 @@ import (
...
-	// Extract object vocab.Type.
-	return obj.At(0).GetType()
+	// Accumulate all of the objects into a slice.
+	objTypes := make([]vocab.Type, objProp.Len())
+	for i := 0; i < objProp.Len(); i++ {
+
+		// Get object vocab type at index.
+		objType := objProp.At(i).GetType()
+		if objType == nil {
+			return nil
```
tsmethurst, line 58: Hmm, does this mean you're returning nil if even one item in the slice is nil? Might be worth commenting on why.

internal/ap/extract.go
```diff
@@ -35,39 +35,93 @@ import (
...
+	}
+
+	// Accumulate all of the IRIs into a slice.
+	objIRIs := make([]*url.URL, objProp.Len())
+	for i := 0; i < objProp.Len(); i++ {
+
+		// Get iter at index.
+		at := objProp.At(i)
+		if !at.IsIRI() {
+			return nil
```
tsmethurst, line 88: Similar to the above: if you have an objects array where the first entry is a Type and the second entry is an IRI (valid ActivityStreams representation), then you're gonna miss out on the IRI here because you've returned early.

internal/ap/normalize.go
```diff
@@ -39,60 +39,36 @@ import (
 // This function is a noop if the type passed in is anything except a Create or Update with a Statusable or Accountable as its Object.
 func NormalizeIncomingActivity(activity pub.Activity, rawJSON map[string]interface{}) {
 	// From the activity extract the data vocab.Type + its "raw" JSON.
-	dataType, rawData, ok := ExtractActivityData(activity, rawJSON)
-	if !ok {
+	dataTypes, rawData, ok := ExtractActivityData(activity, rawJSON)
+	if !ok || len(dataTypes) != len(rawData) {
```
tsmethurst, line 43: In which cases would these two things not be the same? Not asking for changes, just wanna understand.

.golangci.yml
NyaaaWhatsUpDoc: this was a commit of pure anger :D

internal/federation/federatingdb/accept.go
```diff
@@ -46,62 +46,22 @@ func (f *federatingDB) Accept(ctx context.Context, accept vocab.ActivityStreamsA
...
-					ReceivingAccount: receivingAccount,
-				})
-
-				return nil
-			}
-		}
+	for _, object := range ap.ExtractObjects(accept) {
+		switch object.GetTypeName() { //nolint:gocritic
+		case ap.ActivityFollow:
+			asFollow := object.(vocab.ActivityStreamsFollow) //nolint:forcetypeassert
```
tsmethurst, line 52: might as well do a find and replace for ` //nolint:forcetypeassert` since we're not using that linter as of this PR

internal/federation/federatingdb/update.go
```diff
@@ -103,3 +100,43 @@ func (f *federatingDB) updateAccountable(ctx context.Context, receivingAcct *gts
...
+	idProp := statusable.GetJSONLDId()
+	if idProp == nil || !idProp.IsIRI() {
+		return gtserror.New("invalid id prop")
+	}
+
+	// Get the status URI string for lookups.
+	statusURI := idProp.GetIRI()
+	statusURIStr := statusURI.String()
+
+	// Don't try to update local accounts.
```
tsmethurst, line 115: local statuses you mean right?

internal/processing/status/create.go
```diff
@@ -39,65 +39,69 @@ import (
...
+	// Get current time.
+	now := time.Now()
+
+	status := &gtsmodel.Status{
+		ID:                       statusID,
+		URI:                      accountURIs.StatusesURI + "/" + statusID,
+		URL:                      accountURIs.StatusesURL + "/" + statusID,
+		CreatedAt:                now,
+		UpdatedAt:                now,
+		Local:                    func() *bool { b := true; return &b }(),
```
tsmethurst, line 58: Don't forget we have that `util.Ptr` function now for such cases as this :)

internal/federation/federatingdb/update.go
NyaaaWhatsUpDoc: yep :')

internal/processing/workers/federate.go
```diff
@@ -226,6 +226,50 @@ func (f *federate) DeleteStatus(ctx context.Context, status *gtsmodel.Status) er
...
+	}
+
+	// Use ActivityStreams Statusable type as Object of Update.
+	update, err := f.converter.WrapStatusableInUpdate(statusable, false)
+	if err != nil {
+		return gtserror.Newf("error wrapping Statusable in Update: %w", err)
+	}
+
+	// Send the Update activity with Statusable via the Actor's outbox.
+	if _, err := f.FederatingActor().Send(ctx, outboxIRI, update); err != nil {
```
tsmethurst, line 266: In future we might want to add some functionality here to send updates to everyone who's liked/boosted the status as well. Not sure exactly how we'd do that, and definitely not saying it needs to be done here. Just a vague sort of reminder to us to do that at some point.

internal/ap/normalize.go
NyaaaWhatsUpDoc: it's more just like an integrity check, i should maybe add a code comment regarding this,

internal/ap/extract.go
NyaaaWhatsUpDoc: Ah I was assuming that wouldn't be valid, that an array must all be the same time. Would it perhaps be worth replacing both these functions with a singular one that returns a slice of `TypeOrIRI` ? :thinking:

internal/processing/workers/fromfediapi.go
```diff
@@ -509,6 +513,34 @@ func (p *fediAPI) UpdateAccount(ctx context.Context, fMsg messages.FromFediAPI)
 	return nil
 }
 
+func (p *fediAPI) UpdateStatus(ctx context.Context, fMsg messages.FromFediAPI) error {
```
tsmethurst, line 516: Good stuff!

internal/text/plain.go
```diff
@@ -103,11 +104,11 @@ func (f *Formatter) FromPlainEmojiOnly(
 	statusID string,
 	input string,
 ) *FormatResult {
-	// Initialize standard block parser
-	// that wraps result in <p> tags.
+	// Initialize block parser that
+	// doesn't wrap result in <p> tags.
 	plainTextParser := parser.NewParser(
```
tsmethurst, line 109: Why the change here you rascal? Doesn't this wildly change how our statuses are serialized?

tsmethurst (commented):
Coming along great :) Good work

internal/ap/extract.go
tsmethurst: Probably yeah! Mixed arrays are definitely something we need to take account of. They trip me up in JS all the time.

internal/text/plain.go
NyaaaWhatsUpDoc: This is specifically only in the `FromPlainEmojiOnly()` part, which we only use for pulling emoji out of updated account fields + display names, and formatting status spoiler text. So definitely not places we wanted automatic paragraph HTML tags anywho :p

internal/ap/extract.go
NyaaaWhatsUpDoc: argh what a bastard data type!!

internal/text/plain.go
tsmethurst: Ahh okay okay thanks, I didn't spot that.

.golangci.yml
tsmethurst: you and the linter have such a history; enemies to lovers arc when?

internal/processing/workers/federate.go
NyaaaWhatsUpDoc: Just thinking, for sending updates to likers / boosters, would we want to send the full updated object or just the updated URI so they can mark as invalidated?

NyaaaWhatsUpDoc: actually wait no AP says we should only use full objects. i'll see if it's a very quick thing i can add in here because if so that would be a useful one to add

NyaaaWhatsUpDoc: i started work on this but stashed the changes as i don't think i can write code i'll be happy with in looking neat + being in the right location if i do it quickly like thiss. :p

PR #2261, opened 2023-10-11
[feature] Show all follow's replies in home timeline. Resolves #2254

### LINKED ISSUE #2254: [bug] Follow's replies to unfollowed post not showing in home timeline

This may well be expected/desirable behaviour, or I may have configured something somewhere but I've noticed I don't see replies by people I follow on posts from someone I don't follow in the home or federated timeline. If I view the user's profile via my own instance frontend (Semaphore), I see their reply under "Toots and Replies", but I cannot see it anywhere else.
Example in the "What happened?" section

0.11.1-SNAPSHOT-git-d173fcd

### GoToSocial Arch

amd64 Container (Podman)

### What happened?

1. User X, who I don't follow, posts a message.
    - I'm not expecting to see this one in my home timeline
2. User Y, who I do follow, replies to User X's post
3. User Y's reply to User X's post is not visible in home or federated timeline, but is visible under that User Y's profile via local frontend

User X and User Y are both on remote instances

1. User X posts a message - I don't expect to see this
2. User Y replies to User X's message.
3. User X's post shows in my home and federated timelines.

### How to reproduce it?

1. Follow a popular account
2. See posts they make and posts that are replies to others you follow in the home timeline
3. Don't see posts thats are replies to people you don't follow in the home timeline

I love you guys! You're doing a great job. Seriously, best instance software out there (and I've used a quite a few at this point!)

pswilde:
# Description

>

This pull request implements an account configurable option to show all follow's replies in their home timeline.
GoToSocial by default does not show replies made by someone the account follow's if that reply does not relate to the account i.e. it's a direct reply to an unfollowed account, or does not contain mentioned of anyone else the account follows.
This is a perfectly reasonable default, however, it does limit how someone can organically "explore" the fediverse by seeing what people they follow are replying to. 
I have added a setting available in /settings/user/profile where someone can enable the visibility of these replies in their home timeline if they want them.
This option is then checked in `/internal/visibility/home_timeline.go` when checking the visibility of a status and skips the block that ignores the conversation "irrelevant to owner"

closes #2254 

## Checklist

### DIFF
--- a/docs/api/swagger.yaml
+++ b/docs/api/swagger.yaml
@@ -204,6 +204,10 @@ definitions:
                 description: Account has enabled RSS feed.
                 type: boolean
                 x-go-name: EnableRSS
+            show_all_replies:
+                description: Account has enabled Show All Replies
+                type: boolean
+                x-go-name: ShowAllReplies
             fields:
                 description: Additional metadata attached to this account's profile.
                 items:
@@ -3627,6 +3631,10 @@ paths:
                 - description: Enable RSS feed for this account's Public posts at `/[username]/feed.rss`
                   in: formData
                   name: enable_rss
+                  type: boolean
+                - description: Enable showing all follow's replies in home timeline
+                  in: formData
+                  name: show_all_replies
                   type: boolean
                 - description: Profile fields to be added to this account's profile
                   in: formData
--- a/docs/user_guide/settings.md
+++ b/docs/user_guide/settings.md
@@ -105,6 +105,16 @@ Turning on the discoverable flag may take a week or more to propagate; your acco
 !!! info
     The discoverable setting is about **discoverability of your account**, not searchability of your posts. It has nothing to do with indexing of your posts for search by Mastodon instances, or other federated instances that use full text search!
 
+### Timelines
+
+#### Show All Replies
+
+Some replies that accounts you follow are ignored by default. This is when this reply is not directly to you, or does not mention other people you follow.  
+However, if you wish to make these replies visible, you can do so by enabling this option.  
+
+!!! info
+    Enabling this option will increase the amount of status visible in your home timeline
+
 ### Advanced
 
 #### Custom CSS
--- a/internal/api/client/accounts/accountupdate.go
+++ b/internal/api/client/accounts/accountupdate.go
@@ -261,7 +261,7 @@ func parseUpdateAccountForm(c *gin.Context) (*apimodel.UpdateCredentialsRequest,
 			form.FieldsAttributes == nil &&
 			form.CustomCSS == nil &&
 			form.EnableRSS == nil &&
-            form.ShowAllReplies == nil) {
+			form.ShowAllReplies == nil) {
 		return nil, errors.New("empty form submitted")
 	}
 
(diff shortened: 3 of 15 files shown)


pswilde:
I'll be working on the tests ASAP

pswilde:
Testrig adjusted to allow for the new code so tests now pass. 
Tests already in place were not passing previously due to this setting being enabled for "HomeTimelineable" tests - set to false so all default tests will pass successfully.

tsmethurst:
Hmm, I'm not so sure about this functionality. We already make it possible (via the account view) to see replies made by a followed account even if the reply is to accounts you don't follow. For example, on pinafore when looking at an account, you can click 'toots and replies' and see all the replies there. And you can also discover other accounts in your home timeline via boosts and whatnot ofc. So personally I don't *really* see the benefits of also including replies to unfollowed accounts on the home timeline in this way.

tsmethurst:
Thinking about it a bit more, I think if someone enables this setting in the best intention ("I wanna find interesting people to follow"), it also might end up encouraging dogpiling / making dogpiling easier and amplifying outrage ("Who's this asshole with the bad take everyone is replying to?")

pswilde:
Hey,
OK fair enough.

I'm aware the posts are available under the profile, but for me this doesn't work.

For example, say someone I follow replied to somebody I don't follow saying something like "Hey, that feature you added to the software product you make has really helped me out!" - that's interesting to me and I'd like to see it in my home timeline, I wouldn't go round all my follows and check their profiles to see if they've said something like this... that's what the home timeline is for in my opinion, and it's how other federated software works.

It's fine not having it as a default, but I still this - it's important for a person to choose if they want this functionality or not.

I mean, I do only follow people I trust to say reasonable, non-assholey things ( most of the time) , so maybe I'm optimistic here... I dunno 😂

tsmethurst:
> it's how other federated software works

Is it? I thought our current default was more consistent with how Mastodon etc present their home timeline, but perhaps I missed something that was added recently. I wasn't aware that Mastodon shows replies to people you don't follow, and that's the only fedi software (apart from GtS obviously) that I've used extensively myself.

> For example, say someone I follow replied to somebody I don't follow saying something like "Hey, that feature you added to the software product you make has really helped me out!" - that's interesting to me and I'd like to see it in my home timeline

I do get where you're coming from, but the use case you describe here seems to be covered by boosts already. If the person you follow replies to the post but also boosts the post they just replied to, you'll see the original post (albeit not the reply, unless you click on the boost to open the context). Folks can also boost the reply they just made, if they want to, to ensure folks they follow see it on their home timeline. That feels less invasive to me than a setting where you can see all replies to sort of snoop on conversations you're not really involved in.

pswilde:
To be fair I haven't used Mastodon for a few years now - maybe they've changed from when I last used it or my memory of that time is wrong, but definitely Pleroma (and forks) show all replies in the home timeline like this.... maybe that's why everyone hates it :laughing: 
I appreciate you understand my point of view though, thanks. Personally, I don't really want to rely on someone I follow boosting something for visibility... they should be, yes, but it isn't guaranteed. 

Anyway, I totally respect your choice in this, I won't push the matter, I can just run my extra bit out to one side.

Thanks for all your hard work!

tsmethurst:
> Thanks for all your hard work!

No problem!

I'll close this PR then, but thank you again for opening it and for writing it so nicely!

PR #2267, opened 2023-10-16
[feature] Add db models + functions for ID-ing status/conversation threads

tsmethurst:
# Description

>

Initial pull request for adding a thread ID to statuses, which will support status thread muting in an upcoming PR (models and DB functions for status mutes also added). 

Top-level local statuses are now given a thread ID when they're created, and subsequent replies to that status, and any children of that status, inherit the thread ID of their parent. This includes replies made by both local and remote accounts. This provides an easy, good-enough way of recognizing whether a status is part of a thread, without having to navigate up a whole chain of statuses looking for a relevant parent.

Note: threads always start being tracked from a *local* user's status onwards; purely remote threads are not given a thread ID, since there's no point us keeping track of such things.

DB functions have been added to select the IDs of all statuses in a thread. Although these aren't currently used for anything, they might come in handy in the future. Related: this PR doesn't replace (or intend to replace) the existing logic for modeling a status's context by looking up and down through ancestors and descendants. Maybe we could use thread IDs like that in the future, but I don't think so and it's not of interest right now since our existing system works rather well.

## Checklist


tsmethurst:
I'm gonna close this for now and reopen it when mutes are included as well; since we're currently in a release cycle it's a bit awkward to merge this now anyways, so might as well just bundle status mute functionality in as well.

PR #2278, opened 2023-10-19
[feature] Status thread mute/unmute functionality

tsmethurst:
# Description

>

This pull request implements support for muting/unmuting a status thread in order to ignore notifications on that thread for replies, boosts, and likes. Useful when you've posted a banger and the notifications on it become overwhelming, or when you've been mentioned in a thread you're not interested in.

To enable this functionality, the concept of a 'ThreadID' has been added to the codebase:

### What's a ThreadID

Not much, what's a ThreadID with you?

Local statuses are now assigned a thread ID when they're created, and subsequent replies to that status, and any children of that status, inherit the thread ID of their parent. This includes replies made by both local and remote accounts. This provides an easy, good-enough way of recognizing whether a status is part of a thread, without having to navigate up a whole chain of statuses looking for a relevant parent.

New remote statuses (either top-level or replies) either inherit the threadID of their parent, or get a new threadID created for them if they mention a local account. If they don't mention or reply to a local status, they don't get a threadID since there's no point keeping track of that.

~~DB functions have been added to select the IDs of all statuses in a thread. Although these aren't currently used for anything, they might come in handy in the future.~~ Related: this PR doesn't replace (or intend to replace) the existing logic for modeling a status's context by looking up and down through ancestors and descendants. Maybe we could use thread IDs like that in the future, but I don't think so and it's not of interest right now since our existing system works rather well.

### How do mutes work

When you mute a "status" you're actually creating a mute that targets the thread ID of that status. When a notification is about to be generated for a reply/like/boost, the notification function will first check if there's a mute in place of the thread to which the notification-creating activity belongs. Only if there's no mute will the notification be created.

## Checklist


iliana:
(Apologies for a drive-by question — I'm looking at a few different ActivityPub implementations to see what might meet my needs in the near future. Thanks for your consideration!)

If I'm reading the PR description properly: each status can only belong to a single ThreadID, correct? A feature I would find helpful is the ability to mute a _subthread_, for situations where some people decide to have a conversation in my replies that I no longer wish to observe, but still see notifications for the rest of the thread.

Being able to keep track of multiple ThreadIDs would be useful for that, but I'm not sure how that would look in either the database or the API.

tsmethurst:
> If I'm reading the PR description properly: each status can only belong to a single ThreadID, correct

That's right yes!

> A feature I would find helpful is the ability to mute a subthread, for situations where some people decide to have a conversation in my replies that I no longer wish to observe, but still see notifications for the rest of the thread.

That makes sense. Unfortunately it's a bit of a pain in the arse because every reply can technically be a subthread, so in deeply nested threads with lots of replies you're gonna end up keeping track of a bananas amount of thread IDs.

Not something to be fixed for this PR, I think, but perhaps in future we can fiddle with this and see if there's a way to mute from a specific status ID downwards, instead of muting a thread ID. Not sure though!

> Apologies for a drive-by question

No worries :)

internal/db/bundb/migrations/20231016113235_mute_status_thread.go
```diff
@@ -0,0 +1,148 @@
...
+			if err != nil && !(strings.Contains(err.Error(), "already exists") ||
+				strings.Contains(err.Error(), "duplicate column name") ||
+				strings.Contains(err.Error(), "SQLSTATE 42701")) {
+				return err
+			}
+
+			// Index new + existing tables properly.
+			for table, indexes := range map[string]map[string][]string{
+				"threads": {
+					"threads_id_idx": {"id"},
```
NyaaaWhatsUpDoc, line 111: do we need to create indices for the primary columns? (genuine question, i don't know the answer here)

internal/db/bundb/thread.go
```diff
@@ -0,0 +1,197 @@
...
+
+		// Update statuses that use the given thread id.
+		//
+		// Note: deleting a thread (which is just a grouping)
+		// doesn't delete statuses in the thread, it just
+		// de-threads them.
+		_, err = tx.
+			NewUpdate().
+			Table("statuses").
+			Set("? = ?", bun.Ident("thread_id"), (*string)(nil)).
```
NyaaaWhatsUpDoc, line 100: it might look a little more SQL-like if you make this a `Set("? = NULL")`, it took me a little to understand what was going on here and I'm pretty sure the null syntax is dialect agnostic

internal/db/bundb/thread.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/uptrace/bun"
+)
+
+type threadDB struct {
+	db    *DB
+	state *state.State
+}
+
+func (t *threadDB) GetThread(ctx context.Context, id string) (*gtsmodel.Thread, error) {
+	thread, err := t.state.Caches.GTS.Thread().Load("ID", func() (*gtsmodel.Thread, error) {
```
NyaaaWhatsUpDoc, line 36: considering the thread object itself is empty minus the lookup ID itself, i'm thinking it might be worth just getting rid of the `Thread{}` model cache and instead have a `ThreadStatusIDs() *SliceCache[string]` that caches the contents of `thread_to_statuses` by their `thread_id` storing a slice of `status_id`.

it would then raise the question, do we bother checking the `threads` table for an entry, or instead just go straight for the (now cached) call to `getThreadStatusIDs(...)` and wrap the result in a `Thread{}` on success...

we'd then need to update the cache hooks to invalid the `ThreadStatusIDs()` cache on status invalidation when `.ThreadID != ""`, and in delete / update thread functions below too.

what do you think?

internal/gtsmodel/statusmute.go
```diff
@@ -19,7 +19,7 @@ package gtsmodel
 
 import "time"
 
-// StatusMute refers to one account having muted the status of another account or its own.
+// StatusMute IS DEPRECATED -- USE THREADMUTE INSTEAD NOW! THIS TABLE DOESN'T EXIST ANYMORE!
```
NyaaaWhatsUpDoc, line 22: do you know when we can delete this model? :thinking:

internal/processing/workers/fromclientapi_test.go
```diff
@@ -324,6 +325,114 @@ func (suite *FromClientAPITestSuite) TestProcessCreateStatusReply() {
...
 
+func (suite *FromClientAPITestSuite) TestProcessCreateStatusReplyMuted() {
+	var (
+		ctx              = context.Background()
+		postingAccount   = suite.testAccounts["admin_account"]
+		receivingAccount = suite.testAccounts["local_account_1"]
+
+		// Admin account posts a reply to zork.
+		// Normally zork would get a notification
+		// for this, but zork mutes this thread.
```
NyaaaWhatsUpDoc, line 336: oh saucy zork, just far too popular

internal/processing/workers/surfacenotify.go
```diff
@@ -114,6 +138,24 @@ func (s *surface) notifyFave(
 		return nil
 	}
 
+	// Ensure boostee hasn't
```
NyaaaWhatsUpDoc, line 141: i assume you meant "favee" :p

internal/db/bundb/status.go
```diff
@@ -634,14 +679,19 @@ func (s *statusDB) getStatusBoostIDs(ctx context.Context, statusID string) ([]st
 	})
 }
 
-func (s *statusDB) IsStatusMutedBy(ctx context.Context, status *gtsmodel.Status, accountID string) (bool, error) {
```
NyaaaWhatsUpDoc, line 637: Most of our database functions simply accept IDs, it might fit in more if it were to accept a `threadID` here. What do you think?

internal/processing/status/mute.go
```diff
@@ -0,0 +1,150 @@
...
+	requestingAccount *gtsmodel.Account,
+	targetStatusID string,
+) (*gtsmodel.Status, gtserror.WithCode) {
+	targetStatus, err := p.state.DB.GetStatusByID(ctx, targetStatusID)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return nil, gtserror.NewErrorInternalError(err)
+	}
+
+	if targetStatus == nil {
+		err := gtserror.Newf("status %s not found in the db", targetStatusID)
```
NyaaaWhatsUpDoc, line 49: for these error strings it's probably worth normalizing these to follow the style in `processor/common.Processor{}.GetTargetStatusByID()` (or just using that function). also by using `gtserror.New()` like functions in the error help text fields below we'll be leaking the function name in responses (not a massive deal, but it's not really information the client needs to know).

NyaaaWhatsUpDoc:
A few points worth addressing, but otherwise this looks great :D

internal/db/bundb/migrations/20231016113235_mute_status_thread.go
tsmethurst: Not *really* but it doesn't do any harm afaik

internal/db/bundb/status.go
```diff
@@ -390,6 +407,23 @@ func (s *statusDB) UpdateStatus(ctx context.Context, status *gtsmodel.Status, co
 				}
 			}
 
+			// If the status is threaded, create
+			// link between thread and status.
+			if status.ThreadID != "" {
+				if _, err := tx.
+					NewInsert().
```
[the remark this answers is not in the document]
tsmethurst: I think it could change if a remote status was edited to mention a local user, yeah.

internal/db/bundb/status.go
tsmethurst: Ah sure I can change that!

internal/db/bundb/thread.go
tsmethurst: Ah sure, I'll change that.

internal/gtsmodel/statusmute.go
tsmethurst: No not really. Probably when we do a first beta release we can clean up migrations and stuff and make a fresh start, but I haven't really given it much thought yet.

internal/processing/status/mute.go
tsmethurst: Since these error messages aren't send along to the client I think it's OK (they only get serialized to the client if the second argument, `helpText` or whatever it's called, also has something in it). Will normalize the message format though!

internal/processing/workers/fromclientapi_test.go
tsmethurst: cursed to write banger posts in hell for eternity

internal/db/bundb/thread.go
NyaaaWhatsUpDoc: I think the `ThreadStatusIDs()` will still be worth doing a cache for, even if the `Thread{}` model cache gets left. Threading is pretty heavily integrated through statuses logic so that's potentially a lot of extra database lookups

internal/db/bundb/status.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: i think how i worded this didn't accurately get across what i was asking :p, i meant like do we possibly need to handle the case where threading info has changed i.e. `tx.NewUpdate()...` for the thread link?

internal/gtsmodel/statusmute.go
NyaaaWhatsUpDoc: ah yeah that's true it is still required for old migrations and bits. was just wondering what in the codebase was still relying on it!

internal/db/bundb/status.go
tsmethurst: Shuffled it around here: https://github.com/superseriousbusiness/gotosocial/pull/2278/commits/c0fe5a99e1e3fe9efa23d68f9b28769bf4e4eee4

internal/processing/status/mute.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2278/commits/29c0d947bce916a0ebffba722c358358981fa6d2

internal/db/bundb/status.go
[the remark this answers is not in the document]
tsmethurst: I'm not really sure what you mean... could you explain?

internal/db/bundb/thread.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2278/commits/163e5368fc4ff9d8badd088d6c89b5bfb1905c2f

internal/processing/workers/surfacenotify.go
tsmethurst: Ahh, copy paste error :facepalm: https://github.com/superseriousbusiness/gotosocial/pull/2278/commits/906a2b0f8f7459ef9b1c10bfa5d46fe685ebd684

internal/db/bundb/thread.go
tsmethurst: Hmmm I'm not sure what you mean, I'm afraid. Do we mean when building a thread context (getting status parents and descendants)? Because this logic doesn't replace that, that's still handled by iterating up and down a thread. That GetThread() function (which selects status IDs from the database) is actually not called by anything at this point, except in tests. Maybe it would be clearer if I just removed those functions for now, what do you think?

internal/db/bundb/status.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: ah don't worry about it i'm clearly not very good at explaining this one. it was only a double checking thing anyways :)

internal/db/bundb/thread.go
NyaaaWhatsUpDoc: i think maybe my brain was a little scattered yesterday reading through this... i see what you mean now it isn't used anywhere yet. in which case maybe it is worth removing GetThread() and DeleteThread() functions for now, and probably the cache for them too since they're never fetched. the PutThread() function can just do a direct db.NewInsert().Model(thread). when more functionality is added that requires on threading it might become obvious then which caching solution (or, both!) works best.

tsmethurst: Ack! Got it, I'll remove the unused functions and caches :)

internal/db/bundb/status.go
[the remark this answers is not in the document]
tsmethurst: Ahh no worries!

PR #2287, opened 2023-10-23
[chore] bump go version -> 1.21.x

tsmethurst:
# Description

>

Bump go version from 1.20.x to 1.21.x.

This will let us use the fancy new Go language features, if we like.

Note: we were already building using go 1.21.x, so no difference there, we're just switching the language version properly now.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -12,7 +12,7 @@ steps:
   # We use golangci-lint for linting.
   # See: https://golangci-lint.run/
   - name: lint
-    image: golangci/golangci-lint:v1.53.1
+    image: golangci/golangci-lint:v1.55.0
     volumes:
       - name: go-build-cache
         path: /root/.cache/go-build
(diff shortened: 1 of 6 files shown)


.drone.yml
```diff
@@ -28,7 +28,7 @@ steps:
           - pull_request
 
   - name: test
-    image: golang:1.20.4-alpine
+    image: golang:1.21.3-alpine
```
NyaaaWhatsUpDoc, line 31: if you just set this to `golang:1.21-alpine` it will always use the latest minor release (which are always bugfixes, never api changes)

NyaaaWhatsUpDoc: also i think you need to do this same thing in our Dockerfile

.golangci.yml
```diff
@@ -14,7 +14,6 @@ run:
 linters:
   # enable some extra linters, see here for the list: https://golangci-lint.run/usage/linters/
   enable:
-    - goconst
```
tsmethurst, line 17: Not sure why, but this started flagging more stuff and tbh it doesn't really help our "code quality" any so I just turned it off.

PR #2290, opened 2023-10-23
[bugfix/frontend] Add `nosubmit` option to form fields; use it when instance custom CSS disabled

tsmethurst:
# Description

>

When custom CSS is not enabled on an instance, submitting any custom CSS update (even if it's just an empty string) returns a 400 error since custom CSS is not allowed.

Previously, we were getting around this by just not generating a form field at all for custom CSS when it was disabled on the instance. When I refactored some of the frontend stuff, I made it so the form field was shown, but was not enabled and couldn't be updated by users. This causes empty `custom_css` to be sent along to the instance when doing profile updates, meaning the update always fails.

To get around this, this PR adds a new `nosubmit` option to form hooks. If set to `true`, then the hook fields become 'internal' and won't be submitted as part of mutation data. This means they can still be shown, but become essentially "read only".

With this change, it's possible to update profiles on instances that don't have custom CSS enabled, again.

Change also fixes a couple very small bugs that cropped up while I was looking for the issue.

## Checklist

### DIFF
--- a/web/source/settings/lib/form/submit.ts
+++ b/web/source/settings/lib/form/submit.ts
@@ -87,10 +87,10 @@ export default function useFormSubmit(
 			if (e.nativeEvent.submitter) {
 				// We want the name of the element that was invoked to submit this form,
 				// which will be something that extends HTMLElement, though we don't know
-				// what at this point.
+				// what at this point. If it's an empty string, fall back to undefined.
 				// 
 				// See: https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter
-				action = (e.nativeEvent.submitter as Object as { name: string }).name;
+				action = (e.nativeEvent.submitter as Object as { name: string }).name || undefined;
 			} else {
 				// No submitter defined. Fall back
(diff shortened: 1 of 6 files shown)


web/source/settings/lib/form/submit.ts
```diff
@@ -87,10 +87,10 @@ export default function useFormSubmit(
 			if (e.nativeEvent.submitter) {
 				// We want the name of the element that was invoked to submit this form,
 				// which will be something that extends HTMLElement, though we don't know
-				// what at this point.
+				// what at this point. If it's an empty string, fall back to undefined.
 				// 
 				// See: https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent/submitter
-				action = (e.nativeEvent.submitter as Object as { name: string }).name;
+				action = (e.nativeEvent.submitter as Object as { name: string }).name || undefined;
```
tsmethurst, line 93: This little bugger was causing form update state to not get back to the button that submitted it properly, meaning errors were being swallowed.

PR #2291, opened 2023-10-24
[feature] add per-uri dereferencer locks

NyaaaWhatsUpDoc:
# Description
Updates the codebase to share the `mutexes.MutexMap` with the `Dereferencer{}` in order to perform dereferences with per-URI mutex locks. This has been on the cards as "TODOs" in the codebase for a while, but only now getting around to it... This should handle race conditions inserting / updating accounts / statuses much better, and hopefully reduce the chances of these situations happening to begin with!

This also updates the `go-mutexes` library to the latest version which allows for more of the unlock-mutex code to be inlined.

## Checklist


internal/federation/federatingdb/lock.go
```diff
@@ -35,9 +34,5 @@ import (
 //
 // Used to ensure race conditions in multiple requests do not occur.
 func (f *federatingDB) Lock(c context.Context, id *url.URL) (func(), error) {
-	if id == nil {
-		return nil, errors.New("Lock: id was nil")
-	}
-	unlock := f.locks.Lock(id.String())
-	return unlock, nil
+	return f.state.FedLocks.Lock(id.String()), nil // id should NEVER be nil.
```
tsmethurst, line 37: lovely!

internal/federation/dereferencing/util.go
```diff
@@ -0,0 +1,29 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package dereferencing
+
+// doOnce wraps a function to only perform it once.
+func doOnce(fn func()) func() {
```
tsmethurst, line 21: Did you consider using the newly-added https://pkg.go.dev/sync#OnceFunc for this? Not sure how it weighs up against this solution, just thought I'd mention it!

tsmethurst (commented):
Great job!

internal/db/bundb/migrations/20220612091800_duplicated_media_cleanup.go
```diff
@@ -32,14 +31,14 @@ import (
 )
 
 func init() {
-	deleteAttachment := func(ctx context.Context, l log.Entry, a *gtsmodel.MediaAttachment, s *kv.KVStore, tx bun.Tx) {
-		if err := s.Delete(ctx, a.File.Path); err != nil && err != storage.ErrNotFound {
+	deleteAttachment := func(ctx context.Context, l log.Entry, a *gtsmodel.MediaAttachment, s storage.Storage, tx bun.Tx) {
```
NyaaaWhatsUpDoc, line 34: this change updates this old migration to use the storage.Storage type, instead of the kv.KVStore type (which we don't use anymore). this allows us to remove that dependency causing zero logic changes

tsmethurst:
Feel free to merge this when you're happy with it :)

Issue #2212, opened 2023-09-19
[feature] Make maintenance task time configurable


mirabilos:
### Is your feature request related to a problem ?

I think at midnight server time (UTC in my case), some maintenance tasks run, making the system otherwise sluggish.

### Describe the solution you'd like.

I’d like to be able to configure the time this runs, to not coincide with either personal usage or backup times.

### Describe alternatives you've considered.

NONE

### Additional context.

Also, how do I figure out which tasks are run and when they end(ed), to have a rough idea of how long a timeslot I need to reserve for them? That is, for what do I need to grep in the logs?

tsmethurst:
Good idea yeah; I think someone else recently requested the same thing in chat but we forgot to make an issue for it.

> Also, how do I figure out which tasks are run and when they end(ed), to have a rough idea of how long a timeslot I need to reserve for them? That is, for what do I need to grep in the logs?

I think if you grep 'prune' you should see entries in the logs for the tasks.

PR #2304, opened 2023-10-26
[feature] Customizable media cleaner schedule

tsmethurst:
# Description

>

This pull request implements customizable start time + interval period for media cleanup jobs.

Adds `media-cleanup-from` and `media-cleanup-every` config fields, which keep the current default values of `00:00:00` (midnight) and `86400000000000` or `24h` (24 hours).

These fields allow admins to run their cleanup at different times and rates. For example, say one wanted to run a cleanup every 3 days at 4.30am, one could set `media-cleanup-from` to `04:30:00` and `media-cleanup-every` to `72h`.

To make this work, instantiation of the cleaner has been pushed all the way back to the server process. This is a bit more complex, code-wise, but has the advantage of the scheduler not being run by default when just doing CLI commands and whatnot.

~~**Still a draft because I want to add some documentation for this.**~~

closes https://github.com/superseriousbusiness/gotosocial/issues/2212

## Checklist

### DIFF
--- a/docs/configuration/media.md
+++ b/docs/configuration/media.md
@@ -29,17 +29,6 @@ media-description-min-chars: 0
 # Default: 500
 media-description-max-chars: 500
 
-# Int. Number of days to cache media from remote instances before they are removed from the cache.
-# A job will run every day at midnight to clean up any remote media older than the given amount of days.
-#
-# When remote media is removed from the cache, it is deleted from storage but the database entries for the media
-# are kept so that it can be fetched again if requested by a user.
-#
-# If this is set to 0, then media from remote instances will be cached indefinitely.
-# Examples: [30, 60, 7, 0]
-# Default: 7
-media-remote-cache-days: 7
-
 # Int. Max size in bytes of emojis uploaded to this instance via the admin API.
 # The default is the same as the Mastodon size limit for emojis (50kb), which allows
 # for good interoperability. Raising this limit may cause issues with federation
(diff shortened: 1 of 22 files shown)


daenney (approved):

docs/configuration/media.md
```diff
@@ -55,4 +44,35 @@ media-emoji-local-max-size: 51200
...
+# If this is set to 0, then media from remote instances will be cached indefinitely.
+#
+# Examples: [30, 60, 7, 0]
+# Default: 7
+media-remote-cache-days: 7
+
+# String. 24hr time of day formatted as hh:mm:ss.
+# Examples: ["14:30:00", "00:00:00", "04:00:00"]
+# Default: "00:00:00" (midnight). 
+media-cleanup-from: "00:00:00"
```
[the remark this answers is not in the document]
tsmethurst: Good point! Changed here: https://github.com/superseriousbusiness/gotosocial/pull/2304/commits/fe1fe95a0a6e0589890b2f0bafe1bc4ca0b9f9ca

internal/cleaner/cleaner.go
```diff
@@ -109,16 +108,46 @@ func (c *Cleaner) removeFiles(ctx context.Context, files ...string) (int, error)
...
+		0,
+		0,
+		now.Location(),
+	)
 
-	if midnight.Before(now) {
-		// since <= 11:59am rounds down.
-		midnight = midnight.Add(day)
+	// Ensure first cleanup is in the future.
+	for firstCleanupAt.Before(now) {
```
NyaaaWhatsUpDoc, line 149: why the for loop here?

internal/gotosocial/gotosocial.go
```diff
@@ -20,62 +20,47 @@ package gotosocial
...
-	}, nil
-}
-
-// gotosocial fulfils the gotosocial interface.
-type gotosocial struct {
-	db           db.DB
-	apiRouter    router.Router
-	federator    *federation.Federator
-	mediaManager *media.Manager
+	cleaner *cleaner.Cleaner,
```
NyaaaWhatsUpDoc, line 41: Do we need to embed this within the Server{} type? The long-running type that gets passed round everywhere and holds everything is the State{} type, which I think would be useful to hold the Cleaner{} in time (though depending cycling will need to be avoided :p). I'm not sure I understand the abstraction that this represents or why we're hiding the ScheduleJobs() part from the main server startup.

internal/cleaner/cleaner.go
tsmethurst: If someone sets their instance to clean every hour, but the first clean is four hours in the future (for example) then adding the interval to the firstCleanup time will not be enough to make it in the future. So the loop just ensures that it keeps adding the interval until it gets to a time that's actually in the future :)

internal/gotosocial/gotosocial.go
tsmethurst: Ah yes I suppose we could do that instead. Basically I did this because we need to be able to tell the scheduler to 'start' and return an error if something is misconfigured (instead of panicking deeply somewhere). Happy to move it to state if you prefer!

internal/cleaner/cleaner.go
NyaaaWhatsUpDoc: OHHHHHH i get it now. i was thinking of the cleanupFrom as being a "distance from now time" instead of being a 24hr clock time. so was wondering how on earth firstCleanupAt could *ever* **not** be in the future :woman_facepalming:

internal/gotosocial/gotosocial.go
NyaaaWhatsUpDoc: actually on second thoughts , remember i'm doing some wrapping of the scheduler in the polls PR and will need to do some fenangling (that's definitely a word) to get the rebase working correctly, maybe just leave this as-is for now and i'll deal with it during said fenangling :]

PR #2310, opened 2023-10-29
Remove account_suspended_at_idx to resolve slow query issues

### LINKED ISSUE #2309: [bug] Slow query warnings for CountInstanceUsers

As discussed in the Matrix chat, a user reports the following warnings associated with poor performance of their GtS instance (running SQLite):
```
timestamp="28/10/2023 09:18:40.891" func=bundb.queryHook.AfterQuery level=WARN duration="1m0.817367328s" query="SELECT count(*) FROM \"accounts\" AS \"account\" WHERE (\"account\".\"username\" != '[node.isonomia.net](http://node.isonomia.net/)') AND (\"account\".\"suspended_at\" IS NULL) AND (\"account\".\"domain\" IS NULL)" requestID=d4ks6xcb04001xv5kpcg msg="SLOW DATABASE QUERY"
```

This is a trivial query which shouldn't take minutes, and the user reports sometimes up to _hours_. Asking the query optimiser to explain produces:
```
sqlite> explain query plan SELECT count(*) FROM "accounts" AS "account" WHERE ("account"."username" != '[node.isonomia.net](http://node.isonomia.net/)') AND ("account"."suspended_at" IS NULL) AND ("account"."domain" IS NULL);
QUERY PLAN
`--SEARCH account USING INDEX accounts_suspended_at_idx (suspended_at=?)
```
For comparison, here's what my Postgres does:
```
gotosocial=# explain SELECT count(*) FROM "accounts" AS "account" WHERE ("account"."username" != 'social.sentynel.com') AND ("account"."suspended_at" IS NULL) AND ("account"."domain" IS NULL);
                                            QUERY PLAN
--------------------------------------------------------------------------------------------------
 Aggregate  (cost=4.76..4.77 rows=1 width=8)
   ->  Index Scan using accounts_domain_idx on accounts account  (cost=0.29..4.75 rows=2 width=0)
         Index Cond: (domain IS NULL)
         Filter: ((suspended_at IS NULL) AND ((username)::text <> 'social.sentynel.com'::text))
(4 rows)
```
Note that the problematic SQLite installation is using the `accounts_suspended_at_idx` index, whereas Postgres is using `accounts_domain_idx`. Using the domain index first will produce, on most installations, single digit result counts, and the filter for the other conditions is trivial. Conversely, the suspended_at index will hit on 50% or more of all accounts in the database, resulting in applying the other conditions to hundreds of thousands of accounts or more.

It's unclear to me why the SQLite optimiser is dropping the ball in this particular case. However, as far as I can tell `accounts_suspended_at_idx` should _never_ be used, so we can resolve this problem and save a little space and time on the accounts table by simply removing it. We're not currently making any queries which look for anything other than `NULL` on that index, which (as discussed above) will return _many_ results. I have not been able to come up with a scenario in which we might usefully want to query on this index.

v0.12.1

### GoToSocial Arch

_No response_

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

_No response_

Sentynel:
# Description

closes #2309

Resolve issue by removing index, as discussed in issue.

## Checklist

### DIFF
--- a/internal/db/bundb/migrations/20231029143819_remove_suspended_at_index.go
+++ b/internal/db/bundb/migrations/20231029143819_remove_suspended_at_index.go
@@ -1,0 +1,51 @@
+/*
+GoToSocial
+Copyright (C) 2021-2023 GoToSocial Authors admin@gotosocial.org
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+package migrations
+
+import (
+	"context"
+
+	"github.com/uptrace/bun"
+)
+
+func init() {
+	up := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			// Remove unhelpful index which can confuse the query optimiser
+			// for CountInstanceUsers()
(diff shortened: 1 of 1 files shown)


NyaaaWhatsUpDoc:
nice work!! :heart:

Issue #2144, opened 2023-08-21
[feature/chore] Relax `Mention` parsing, document expected format


tsmethurst:
Right now, GtS won't parse a mention properly unless it looks like the following:

```json
{
	"type": "Mention",
	"href": "https://gts.superseriousbusiness.org/users/gotosocial",
	"name": "@gotosocial@superseriousbusiness.org"
}
```

So, `href` should point to an ActivityPub URI/ID (not just a web URL), and `name` has to be a specifically formatted 'namestring'. This was adopted from the way Mastodon sends / receives Mentions, and we haven't really touched it in years.

HOWEVER! We don't really *need* all of the above things to deduce who the mention is aimed at. We could make do with just `name`, or just `href`. Or `href` could point to an account's web URL (assuming uniqueness -- https://github.com/superseriousbusiness/gotosocial/issues/2082), and that's still enough for us to figure out who the target is.

We might want to make our parsing a bit less fussy, then, so that other implementations don't have to worry too much about getting the format 100% to GtS's liking.

Also, we should document the Mention format alongside other incoming/outgoing federation types in our federating with gotosocial document.

Related: https://github.com/superseriousbusiness/gotosocial/issues/2135

tsmethurst:
Well, we could even use either! If only one is present, use that. If both are present, parse both and make sure they point to the same thing, something along those lines.

tsmethurst:
Just had my attention brought to a case where this actually causes issues: when trying to authenticate in an Owncast chat. Normally, Owncast will send a DM to the username you provide to it, containing a code to enter in the chat to prove that it's really you. Unfortunately GtS can't parse the mention owncast is providing:

```
Oct 29 21:03:02 goblin.technology gotosocial[1359664]: timestamp="29/10/2023 21:03:02.289" func=server.glob..func1.Logger.func13.1 level=INFO latency="428.044µs" userAgent=Go-http-client/1.1 method=GET statusCode=200 path=/.well-known/webfinger clientIP=65.21.145.62 requestID=25f0gzcb04001aj2pmf0 msg="OK: wrote 194B"
[...]

PR #2320, opened 2023-10-30
[bugfix] Relax `Mention` parsing, allowing either href or name

tsmethurst:
# Description

>

This pull request relaxes the format required for incoming Mention tags, by making the `name` property no longer a strict requirement. Instead, GtS will now try to use `href` first (either URI or URL, both should work), and fall back to `name` if that's not set. If neither are set, the mention will be discarded.

Also adds some documentation for this in our "Federating with GoToSocial" document :)

closes https://github.com/superseriousbusiness/gotosocial/issues/2144

## Checklist


docs/federation/federating_with_gotosocial.md
```diff
@@ -427,3 +427,58 @@ With just one tag, the `tag` property will be an object rather than an array, wh
 The `href` URL provided by GoToSocial in outgoing tags points to a web URL that serves `text/html`.
 
 GoToSocial makes no guarantees whatsoever about what the content of the given `text/html` will be, and remote servers should not interpret the URL as a canonical ActivityPub ID/URI property. The `href` URL is provided merely as an endpoint which *might* contain more information about the given hashtag.
+
+## Mentions
```
NyaaaWhatsUpDoc, line 431: these some good docs :woman_cook: :ok_hand:

tsmethurst: Thanks! :)

internal/federation/dereferencing/status.go
```diff
@@ -342,35 +343,104 @@ func (d *Dereferencer) enrichStatus(
...
+	// Mentions can be created using Name or Href.
+	// Prefer Href (TargetAccountURI), fall back to Name.
+	if mention.TargetAccountURI == "" {
+		// Href wasn't set. Find the target account using namestring.
+		username, domain, err := util.ExtractNamestringParts(mention.NameString)
+		if err != nil {
+			return nil, false, gtserror.Newf("failed to parse namestring %s: %w", mention.NameString, err)
+		}
+
+		mention.TargetAccount, _, err = d.getAccountByUsernameDomain(ctx, requestUser, username, domain)
```
NyaaaWhatsUpDoc, line 375: the "code smell" of performing this here followed by calling it again for an account we already have along with URI again below isn't great imo, i don't know how else to explain it other than that :p. (also even if using caches, these getAccountBy() functions are both pretty heavy code paths). i think the existing.GetMention() check is the only part of code they really share between those two paths. alternatively what if you did something like:

```
func deriveMention(...) (*gtsmodel.Mention, bool, error) {
	if mention.TargetURI == "" {
		// load mention account by username,domain

		existing, ok := existing.GetMentionByTargetURI(...)
		if ok && existing.ID != "" {
			return existing, true, nil
		}

		return mention, false, nil
	} else {
		existing, ok := existing.GetMentionByTargetURI(...)
		if ok && existing.ID != "" {
			return existing, true, nil
		}

		accountURI, err := url.Parse(...)
		if err != nil {
			return nil, false, err
		}

		// load mention account by uri
	}
}
```

PR #2329, opened 2023-11-02
[feature] support canceling scheduled tasks, some federation API performance improvements

NyaaaWhatsUpDoc:
# Description

- wraps the scheduler to support easy job cancelling
- copies over more of the required, (+ some niceties) logic changes from the ./internal/ap subpkg
- copies over more of the required, (+ some niceties) from the ./internal/{federation,processing/workers} subpkgs
    - create status handling now entirely within fediapi handler (removes need for double Statusable->Status parsing)
    - supports create of multiple status objects
    - vastly reduces federatingDB.NewID() complexity

## Checklist


internal/processing/workers/fromfediapi.go
```diff
@@ -526,23 +516,25 @@ func (p *fediAPI) UpdateStatus(ctx context.Context, fMsg messages.FromFediAPI) e
...
 		apStatus,
-		false,
+		true,
 	)
 	if err != nil {
 		return gtserror.Newf("error refreshing updated status: %w", err)
 	}
 
+	if statusable != nil {
+		// Status representation was refetched, uncache from timelines.
```
tsmethurst, line 534: Ahhh great catch here

internal/federation/federatingdb/update.go
```diff
@@ -117,15 +122,31 @@ func (f *federatingDB) updateStatusable(ctx context.Context, receivingAcct *gtsm
...
 	}
 
+	// Check if this is a forwarded object, i.e. did
+	// the account making the request also create this?
+	forwarded := !isSender(statusable, requestingAcct)
+
 	// Get the status we have on file for this URI string.
 	status, err := f.state.DB.GetStatusByURI(ctx, statusURIStr)
-	if err != nil {
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
```
tsmethurst, line 131: Nice, another good catch

internal/processing/workers/fromfediapi.go
```diff
@@ -211,31 +206,38 @@ func (p *fediAPI) CreateStatus(ctx context.Context, fMsg messages.FromFediAPI) e
...
-	// it (deref remote) if necessary.
-	var err error
-	status, _, err = p.federate.RefreshStatus(
+	if status == nil {
+		// No status was set, create a bare-bones
+		// model for the deferencer to flesh-out.
+		status = &gtsmodel.Status{
+
+			// if coming in here status will ALWAYS be remote.
+			Local: func() *bool { var false bool; return &false }(),
```
tsmethurst, line 225: you can use the util.ptr function here

internal/processing/workers/fromfediapi.go
```diff
@@ -358,15 +357,6 @@ func (p *fediAPI) CreateAnnounce(ctx context.Context, fMsg messages.FromFediAPI)
 		return gtserror.Newf("db error inserting status: %w", err)
 	}
 
-	// Ensure boosted status ancestors dereferenced. We need at least
-	// the immediate parent (if present) to ascertain timelineability.
-	if err := p.federate.DereferenceStatusAncestors(ctx,
```
tsmethurst, line 363: Is this check still performed somewhere? I remember this being quite important to avoid seeing unwanted statuses (replies to blocked domains etc) in the timeline

internal/federation/federatingdb/util.go
```diff
@@ -78,131 +101,31 @@ func (f *federatingDB) NewID(ctx context.Context, t vocab.Type) (idURL *url.URL,
...
-					if actorAccount, err := f.state.DB.GetAccountByURI(ctx, iter.GetIRI().String()); err == nil {
-						newID, err := id.NewRandomULID()
-						if err != nil {
-							return nil, err
-						}
-						return url.Parse(uris.GenerateURIForFollow(actorAccount.Username, newID))
-					}
+	// Most of our types set an ID already
+	// by this point, return this if found.
+	idProp := t.GetJSONLDId()
```
tsmethurst, line 106: Good tidying up here :) Thanks for getting out the broom

internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,131 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package scheduler
```
tsmethurst, line 18: Interesting. Feels like this is paving for the way using this to send out messages as well.

I've been thinking recently that it's quite annoying when you post a toot, delete it immediately because you notice a typo, and post the new one, how you end up with the old toot still being sent out even though it's no longer in the database anymore. If we used something like a cancellable scheduler, we could cancel the send job as soon as the old toot was deleted. A thought for the future!

tsmethurst:
Looking good. Just not sure about the panicking!

internal/processing/workers/fromfediapi.go
NyaaaWhatsUpDoc: :grin:

internal/federation/federatingdb/update.go
NyaaaWhatsUpDoc: :grin:

internal/processing/workers/fromfediapi.go
NyaaaWhatsUpDoc: and done :)

internal/federation/federatingdb/util.go
NyaaaWhatsUpDoc: :innocent:

internal/scheduler/scheduler.go
NyaaaWhatsUpDoc: yeah that's one i've thought quite a lot about too. i might try to address it with the http client queueing work i've mentioned on matrix before, will take some thonking to get it working nicely (e.g. dropping conflicting messages, moving repeat / expected failures to back of the queue, etc) while still being neat and not too hyper-specific for just AP messages, i.e. so we can use it for all our standard outgoing requests. :thinking:

internal/processing/workers/fromfediapi.go
NyaaaWhatsUpDoc: I'm not sure what you mean by check, as this is a dereference of status ancestors. Though it is something that will be handled by the `DereferenceAnnounce` function above, though I have just realized a quick tweak it might need now... I'll also add a note here explaining why we don't need to dereference ancestors, i.e. because it's handled within the `DereferenceAnnounce()` function.

internal/ap/properties.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+// Below lie a collection of ActivityStreams property getter
+// and setter functions. They are all without error returns
+// as they are fundamental property types, as such they will
+// panic on failed parse. This is because without these fundamental
+// property types / values being in a valid state, the ActivityPub
+// objects themselves will be unusable, so this panic behaviour
+// helps reduce the rather verbose error-checking code that the
+// go-fed/activity library tends to produce. Note that the panics
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: Ah yeah that is a good point :thinking:. I'll update these to use errors, but add in some occasional `Must___()` type functions as there are situations where we do know an ID (etc) will already have been parsed / set.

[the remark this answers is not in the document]
NyaaaWhatsUpDoc: I removed the panicking functions, which only required updating one thing. Instead it now returns nil in the cases where it isn't found, so it's up to the caller whether that check is needed or not (which depends where it's being called, e.g. in certain functions we already know it to be set to have reached that function in the first place).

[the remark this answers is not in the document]
NyaaaWhatsUpDoc: I also added (currently commented-out, until needed) `Must{Get,Set,Append}` generic functions which allow callers to wrap any of the properties functions to catch and panic on errors, e.g. in the case where things MUST be set at that point (i.e. it *is* an internal server if somehow a property has become unset during one of our functions :p)

internal/processing/workers/fromfediapi.go
tsmethurst: Right, sorry I wasn't very clear. What I meant was:

The below call to `p.surface.timelineAndNotifyStatus` will derive timelineability of the provided status, in part, from the visibility of the status's parent.

Previously, this call to `p.federate.DereferenceStatusAncestors` ensured that this status parent was set (if present) *before* the status gets passed into `p.surface.timelineAndNotifyStatus`, thus ensuring that it's not timelined if the parent is not visible for whatever reason.

By removing this synchronous call here to dereference the status's ancestors, you may be reintroducing a bug where statuses sometimes got timelined when they shouldn't (eg., the status is a boost of a reply to someone on a blocked domain).

internal/ap/properties.go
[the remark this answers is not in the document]
tsmethurst: Nice one, thanks for changing it :)

internal/processing/workers/fromfediapi.go
NyaaaWhatsUpDoc: Ahhhh gotcha. Makes sense. Will fix it :)

NyaaaWhatsUpDoc: I updated our `RefreshStatus()` and `GetStatusByURI()` fucntions to handle status parent dereferencing synchronously in the cases where they're dereferencing a _new_ status :)

Issue #1579, opened 2023-03-02
[bug] DereferenceStatusable: type name Question not supported as Statusable


Jofkos:
### Describe the bug with a clear and concise description of what the bug is.

Remote toot doesn't load and can't be searched for. I do not know why this toot is making problems, specifically, yet it might be due to it being a poll (pools have worked before, however).

[The toot in question.](https://social.treehouse.systems/@alyssa/109934257984594393)

The log prints the following error:
```
timestamp="26/02/2023 21:09:12.162" func=concurrency.(*WorkerPool).Queue.func1 level=ERROR type=worker.Worker[messages.FromFederator] error="populateStatusFields: error populating status repliedTo: populateStatusRepliedTo: couldn't get reply to status with uri https://social.treehouse.systems/users/alyssa/statuses/109934257984594393: wrong received type: GetRemoteStatus: error dereferencing statusable: wrong received type: DereferenceStatusable: type name Question not supported as Statusable" msg="message processing error"
```

### What's your GoToSocial Version?

v0.7.0 git-b599309

### GoToSocial Arch

amd64, binary installation

### What happened?

A toot from someone I am following did not show up in my feed, so I tried manually loading it by searching for it.

### What you expected to happen?

The toot to show up, as I am following that person. That not working, the toot loading after I sent an explicit search for it.

### How to reproduce it?

Search for https://social.treehouse.systems/@alyssa/109934257984594393.

tsmethurst:
We didn't implement polls yet, so that's why! It's planned for some time this year :)

tsmethurst:
Yeah agreed, it's a bit silly! But we'll be implementing it soon enough, so there might not be much point making a workaround yet.

tsmethurst:
closed by https://github.com/superseriousbusiness/gotosocial/pull/2330

PR #2237, opened 2023-09-29
[feature] add support for polls + receiving federated status edits

NyaaaWhatsUpDoc:
# Description

Adds support for the following features:
- creating, voting in polls via the API
- sending federating poll votes
- receiving federated poll updates
- receiving federated status updates
- notifying users of poll ends (both local and federated)

TODO:
- ~~docs on polls from an AP perspective~~ for another PR!
- ~~handling federation side effects on poll vote~~
- ~~inserting poll on client API status create~~
- ~~poll scheduler + required side-effects~~
- ~~add necessary database migration~~
- ~~finish writing docs for code comments~~
- ~~side-effects of _local_ vote in _local_ poll~~
- ~~side-effects of _local_ vote in _remote_ poll~~
- ~~side-effects of _remote_ vote in _local_ poll~~
- ~~final fixups (many i'm sure)~~
- poll database tests
- poll vote database  tests
- poll processor test
- poll vote processor test
- AS status w/ poll tests
- AS poll vote tests 

Notes:
- support for visible polls in the frontend will come in a later PR (soon)
- docs on polls from an AP perspective will come in a later PR (soon)
- support for edited status revision history will come in a later PR (probably much later)
- support for editing statuses by client API will come in a later PR (probably much later)

## Checklist
- [x] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [x] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [x] I/we have performed a self-review of added code.
- [x] I/we have written code that is legible and maintainable by others.
- [x] I/we have commented the added code, particularly in hard-to-understand areas.
- [ ] I/we have made any necessary changes to documentation.
- [ ] I/we have added tests that cover new code.
- [ ] I/we have run tests and they pass locally with the changes.
- [x] I/we have run `go fmt ./...` and `golangci-lint run`.

Issue #1990, opened 2023-07-16
[feature] Allow the editing of posts


Ash-Crow:
### Is your feature request related to a problem ?

"I'm always frustrated when": I see a spelling error or a formating issue on a post I just made, especially if it has already been liked or boosted. Other use cases include finding out that I used the wrong language tag (eg, left "English" selected while my message was in French) or the wrong visibility setting.

Most clients (eg Tusky or elk.zone) have an edit button that allow to edit a post (including replies), but as the feature is not yet implemented in gotosocial, using it results in an error message either when trying to edit the toot (Tusky) or when trying to post the edited toot (elk)

### Describe the solution you'd like.

I would like an edit feature that implements the "status.update" type used by Mastodon and possibly other ActivityPub implementations on the "write" side, complementing #1326 that implements it on the "receive" side.

### Describe alternatives you've considered.

The alternative is to delete and recreate the post/reply, but this has several downsides: faves and boosts are lost, along with the post time, and people that were mentioned in the original post are pinged twice.

### Additional context.

There was a previous issue on this topic (#1591)  but it was closed after the discussion became tense at some point. Per the previous discussion, it is not a simple task ("Implementing edit support requires backend changes, changes to federation logic, documentation, etc. The API endpoint for that doesn't exist yet in GtS.") and not a prioritary one (maybe a beta goal?)

Also, maybe the documentation at https://docs.gotosocial.org/en/latest/user_guide/posts/ should be updated to indicate that this feature is not yet available?

tsmethurst:
Thanks! This is something we'll probably look at implementing after the end of this year (ie., when we're out of alpha and into beta), since we have a bunch of other stuff we need to get to first.

(Also, do you mind if I reword the title of the issue a bit? I just wanna change 'edition' to 'editing' to make it easier for people to search in the issues.)

Ash-Crow:
No problem, you can reword the title and content if it isn't clear enough (English is not my first language... Which makes me think about a couple other usecases that I just added)

mirabilos:
For fixes right after posting, you can also use the “Delete and redraft” feature of most clients.

For later fixes… yeah. But first, GtS ought to implement actually notifying users if a post they interacted with (reply, favourite, boost, bookmark) was edited. Perhaps ideally have a way to review the old content, not just the new.

mirabilos:
[...]

PR #2330, opened 2023-11-02
[feature] add support for polls + receiving federated status edits

NyaaaWhatsUpDoc:
# Description

Adds support for the following features:
- creating, voting in polls via the API
- sending federating poll votes
- receiving federated poll updates
- receiving federated status updates
- notifying users of poll ends (both local and federated)

Closes:
https://github.com/superseriousbusiness/gotosocial/issues/1579
https://github.com/superseriousbusiness/gotosocial/issues/1990

Continuing where #2237 was closed, but with less horrific commit history. (I lost count of the number of rebases on that PR, it was very long-running). Is currently based upon #2329 so will rebase from that when it is merged.

TODO:
- ~~docs on polls from an AP perspective~~ for another PR!
- ~~handling federation side effects on poll vote~~
- ~~inserting poll on client API status create~~
- ~~poll scheduler + required side-effects~~
- ~~add necessary database migration~~
- ~~finish writing docs for code comments~~
- ~~side-effects of _local_ vote in _local_ poll~~
- ~~side-effects of _local_ vote in _remote_ poll~~
- ~~side-effects of _remote_ vote in _local_ poll~~
- ~~final fixups (many i'm sure)~~
- ~~poll database tests~~
- ~~poll vote database  tests~~
- ~~poll processor test~~
- ~~poll vote processor test~~
- ~~AS status w/ poll tests~~
- ~~AS poll vote tests~~

Notes:
- more robust AS tests will come in a later PR (soon)
- support for visible polls in the frontend will come in a later PR (soon)
- docs on polls from an AP perspective will come in a later PR (soon)
- support for edited status revision history will come in a later PR (probably much later)
- support for editing statuses by client API will come in a later PR (probably much later)

## Checklist


tsmethurst:
Holy crap! It's arrived! Will get my reviewing glasses on (imagine opera glasses held on a stick while I peer at your code discerningly from up on the balcony).

internal/processing/workers/fromfediapi.go
```diff
@@ -269,12 +296,10 @@ func (p *fediAPI) CreateFollowReq(ctx context.Context, fMsg messages.FromFediAPI
 	}
 
 	if *followRequest.TargetAccount.Locked {
-		// Account on our instance is locked:
-		// just notify the follow request.
+		// Account on our instance is locked: just notify the follow request.
 		if err := p.surface.notifyFollowRequest(ctx, followRequest); err != nil {
-			return gtserror.Newf("error notifying follow request: %w", err)
+			log.Errorf(ctx, "error notifying follow request: %v", err)
```
NyaaaWhatsUpDoc, line 301: all of these changes in the worker packages to more often log error, than directly return it, were part of a bug fix to get the poll votes working (since our federate step often return errors), and then widened in scoped to basically make our asynchronous processing more robust of errors. so unless a **blocking** stage like "db.Populate___()" fails which prevents further processing, we simply log and continue with the rest of the asynchronous processing.

internal/processing/workers/surfacenotify.go
```diff
@@ -228,17 +347,12 @@ func (s *surface) notifyAnnounce(
 func (s *surface) notify(
 	ctx context.Context,
 	notificationType gtsmodel.NotificationType,
-	targetAccountID string,
-	originAccountID string,
+	targetAccount *gtsmodel.Account,
```
NyaaaWhatsUpDoc, line 350: i updated this function signature to require the models themselves, instead of just the IDs, as most callers of this function already have populated models by this point. i added in extra "db.Populate___()" calls beforehand to make sure they are populated in every case, but fortunately those calls only bother doing database calls if the models aren't already set. so on average this *should* reduce the number of required database calls :)

internal/scheduler/scheduler.go
```diff
@@ -26,18 +26,16 @@ import (
 	"codeberg.org/gruf/go-sched"
 )
 
-// Scheduler wraps an underlying task scheduler
-// to provide concurrency safe tracking by 'id'
-// strings in order to provide easy cancellation.
+// Scheduler wraps an underlying scheduler to provide
```
NyaaaWhatsUpDoc, line 29: these code comments somehow ended up changed as part of some rebase weirdness :p they basically end up meaning the same thing so i'm just going to leave them. though obviously do let me know if you do prefer the "before" change and i can fix it :+1:

internal/typeutils/wrap.go
```diff
@@ -84,132 +85,86 @@ func (c *Converter) WrapPersonInUpdate(person vocab.ActivityStreamsPerson, origi
...
-	if objectIRIOnly {
-		// Only append the object IRI to objectProp.
-		objectProp.AppendIRI(status.GetJSONLDId().GetIRI())
-	} else {
-		// Our statusable's are always note types.
-		asNote := status.(vocab.ActivityStreamsNote)
-		objectProp.AppendActivityStreamsNote(asNote)
+func WrapPollOptionablesInCreate(options ...ap.PollOptionable) vocab.ActivityStreamsCreate {
+	if len(options) == 0 {
+		panic("no options")
```
NyaaaWhatsUpDoc, line 96: re: the last discussion we had about panics. this function is only ever called on our own statuses, so if something *isn't* set by this point, it really is an internal server error, at which point a panic with stack trace is very useful :)

testrig/db.go
```diff
@@ -330,7 +330,7 @@ func StandardDBSetup(db db.DB, accounts map[string]*gtsmodel.Account) {
 func StandardDBTeardown(db db.DB) {
 	ctx := context.Background()
 	if db == nil {
-		log.Panic(nil, "db was nil")
+		return
```
NyaaaWhatsUpDoc, line 333: if db is nil here and it panics, this usually ends up masking a panic stacktrace that caused the db to be nil to begin with. just returning here ensures the previous panic doesn't get masked :)

internal/processing/status/create.go
```diff
@@ -103,6 +131,15 @@ func (p *Processor) Create(ctx context.Context, requestingAccount *gtsmodel.Acco
 		OriginAccount:  requestingAccount,
 	})
 
+	if status.Poll != nil {
```
NyaaaWhatsUpDoc, line 134: the repeated "status.Poll != nil" checks in this same function aren't the best code smell, but i think this function could do with a bit of a rework in general as it's quite old, so it's probably worth saving any further thinking until then

internal/federation/dereferencing/status.go
```diff
@@ -681,6 +686,101 @@ func (d *Dereferencer) fetchStatusTags(ctx context.Context, existing, status *gt
 	return nil
 }
 
+func (d *Dereferencer) fetchStatusPoll(ctx context.Context, existing, status *gtsmodel.Status) error {
```
NyaaaWhatsUpDoc, line 689: this is the bastard function that caused the requirement of dereferencing locks :p. so now we don't handle "ErrAlreadyExists" separately, and just like any other error it will cause the dereferencing to fail and return early. at which point the "enrichStatusSafely()" function will catch it, (whether it be from "PutPoll()", "PutStatus()", etc), and appropriately check the database for an existing copy of that status by URI.

internal/db/bundb/poll.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+
+	return poll, nil
+}
+
+func (p *pollDB) GetOpenPolls(ctx context.Context) ([]*gtsmodel.Poll, error) {
+	var pollIDs []string
+
+	// Select all polls with unset `closed_at` time.
+	if err := p.db.NewSelect().
```
NyaaaWhatsUpDoc, line 103: This isn't the most efficient function, especially considering there's no caching of these IDs, but it is only ever called within "ScheduleAllPolls()", which itself is only ever called during cmd/server initialisation. Otherwise I would say we absolutely do need to cache this :p

internal/db/bundb/migrations/20231002153327_add_status_polls.go
```diff
@@ -0,0 +1,65 @@
...
+
+		// Add the new status `poll_id` column.
+		_, err := db.NewAddColumn().
+			Model(&gtsmodel.Status{}).
+			ColumnExpr("? CHAR(26)", bun.Ident("poll_id")).
+			Exec(ctx)
+		if err != nil && !(strings.Contains(err.Error(), "already exists") || strings.Contains(err.Error(), "duplicate column name") || strings.Contains(err.Error(), "SQLSTATE 42701")) {
+			return err
+		}
+
```
NyaaaWhatsUpDoc, line 52: in time it might be worth adding another migration to add some indices on the polls table (specifically for searching by status ID). but even on my heavily federated (+constrained) instance i haven't had any issues so far, so haven't felt the need to add it yet. worst case we end up having to add a simple index before v0.13.0

NyaaaWhatsUpDoc:
@tsmethurst ready to go :sunglasses:

internal/api/client/polls/polls_get.go
```diff
@@ -0,0 +1,103 @@
...
+		return
+	}
+
+	if _, err := apiutil.NegotiateAccept(c, apiutil.JSONAcceptHeaders...); err != nil {
+		errWithCode := gtserror.NewErrorNotAcceptable(err, err.Error())
+		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
+		return
+	}
+
+	pollID := c.Param(IDKey)
```
tsmethurst, line 84: Would you be up for using the `apiutil` functions for this instead? It doesn't offer much different functionality, it just helps to keep the query parameter logic in one place.

internal/api/client/polls/polls_vote.go
```diff
@@ -0,0 +1,115 @@
...
+		return
+	}
+
+	if _, err := apiutil.NegotiateAccept(c, apiutil.JSONAcceptHeaders...); err != nil {
+		errWithCode := gtserror.NewErrorNotAcceptable(err, err.Error())
+		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
+		return
+	}
+
+	pollID := c.Param(IDKey)
```
tsmethurst, line 87: same comment as above, you might use the apiutil functions here for getting/erroring on the id

internal/api/model/poll.go
```diff
@@ -63,11 +72,23 @@ type PollRequest struct {
...
+
 	// Hide vote counts until the poll ends.
 	HideTotals bool `form:"hide_totals" json:"hide_totals" xml:"hide_totals"`
 }
+
+// PollVoteRequest models a request to vote in a poll.
+//
+// swagger:parameters pollVote
+type PollVoteRequest struct {
+	// Choices contains poll vote choice indices. Note that XML
```
tsmethurst, line 91: You mean form right, not xml?

internal/db/bundb/poll.go
tsmethurst: Ack!

internal/federation/dereferencing/status.go
tsmethurst: Sneaky!

internal/federation/dereferencing/status.go
```diff
@@ -681,6 +686,101 @@ func (d *Dereferencer) fetchStatusTags(ctx context.Context, existing, status *gt
...
+	var (
+		// insertStatusPoll generates ID and inserts the poll attached to status into the database.
+		insertStatusPoll = func(ctx context.Context, status *gtsmodel.Status) error {
+			var err error
+
+			// Generate new ID for poll from the status CreatedAt.
+			// TODO: update this to use "edited_at" when we add
+			//       support for edited status revision history.
+			status.Poll.ID, err = id.NewULIDFromTime(status.CreatedAt)
```
tsmethurst, line 698: Oh that's a neat trick :) I'm glad we added those NewULIDFromTime functions!

internal/federation/dereferencing/status.go
```diff
@@ -681,6 +686,101 @@ func (d *Dereferencer) fetchStatusTags(ctx context.Context, existing, status *gt
...
+			if err != nil {
+				log.Errorf(ctx, "invalid created at date: %v", err)
```
tsmethurst, line 700: Probably worth adding 'falling back to Now()' here so that admins seeing this in their logs don't think the polls aren't working at all

internal/gtsmodel/poll.go
```diff
@@ -0,0 +1,74 @@
...
+
+package gtsmodel
+
+import (
+	"strings"
+	"time"
+)
+
+// Poll represents an attached (to) Status poll, i.e. a questionaire. Can be remote / local.
+type Poll struct {
```
tsmethurst, line 26: I don't think you need to provide most of these types, as they'll largely be inferred already. Better to leave them be, imo, unless you specifically need to override them to something different (as we do with ULID fields)

internal/gtsmodel/poll.go
```diff
@@ -0,0 +1,74 @@
...
+// Closed returns whether the Poll is closed (i.e. date is set and BEFORE now).
+func (p *Poll) Closed() bool {
+	return !p.ClosedAt.IsZero() &&
+		time.Now().After(p.ClosedAt)
+}
+
+// PollVote represents a single instance of vote(s) in a Poll by an account.
+// If the Poll is single-choice, len(.Choices) = 1, if multiple-choice then
+// len(.Choices) >= 1. Can be remote or local.
+type PollVote struct {
```
tsmethurst, line 66: Same comment as above re: types

internal/processing/status/create.go
tsmethurst: No worries!

internal/typeutils/astointernal.go
```diff
@@ -261,8 +261,10 @@ func (c *Converter) ASStatusToStatus(ctx context.Context, statusable ap.Statusab
 	// Attached poll information (the statusable will actually
 	// be a Pollable, as a Question is a subset of our Status).
 	if pollable, ok := ap.ToPollable(statusable); ok {
-		// TODO: handle decoding poll data
```
tsmethurst, line 264: Very satisfying to be able to delete a 'TODO' by doing the thing!

internal/api/client/polls/polls.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/superseriousbusiness/gotosocial/internal/api/util"
+	"github.com/superseriousbusiness/gotosocial/internal/processing"
+)
+
+const (
+	IDKey           = "id"                                 // IDKey is the key for poll IDs
+	BasePath        = "/:" + util.APIVersionKey + "/polls" // BasePath is the base API path for making poll requests through v1 or v2 of the api (for mastodon API compatibility)
+	PollWithID      = BasePath + "/:" + IDKey              //
```
tsmethurst, line 31: Some missing comments here.

internal/processing/polls/expiry.go
```diff
@@ -0,0 +1,126 @@
...
+		p.onExpiry(poll.ID),
+	)
+
+	if !ok {
+		// Failed to add the poll to the scheduler, either it was
+		// starting / stopping or there already exists a task for poll.
+		return gtserror.Newf("failed adding poll %s to scheduler", poll.ID)
+	}
+
+	atStr := poll.ExpiresAt.Local().Format("Jan _2 2006 15:04:05")
```
tsmethurst, line 70: What a weird format! What's this all about?

internal/processing/polls/expiry.go
```diff
@@ -0,0 +1,126 @@
...
+	log.Infof(ctx, "scheduled poll expiry for %s at '%s'", poll.ID, atStr)
+	return nil
+}
+
+// onExpiry returns a callback function to be used by the scheduler when the given poll expires.
+func (p *Processor) onExpiry(pollID string) func(context.Context, time.Time) {
```
tsmethurst, line 76: Love this!

internal/processing/workers/fromclientapi.go
```diff
@@ -228,27 +235,87 @@ func (p *clientAPI) CreateStatus(ctx context.Context, cMsg messages.FromClientAP
...
+		// These were poll votes in a local status,
+		// we only need to federate-out the updated
+		// status model with latest vote counts.
+		//
+		// Handle as a regular status update by
+		// updating the cMsg to contain the status.
+		cMsg.APActivityType = ap.ActivityUpdate
+		cMsg.APObjectType = ap.ObjectNote
+		cMsg.GTSModel = status
+		return p.UpdateStatus(ctx, cMsg)
```
tsmethurst, line 276: Ah interesting... so when a local user votes in a local poll, we also send out the poll to the followers of... the poll creator, i guess?

internal/processing/workers/fromfediapi.go
tsmethurst: Ack!

internal/processing/workers/surfacenotify.go
tsmethurst: Nice one :)

internal/scheduler/scheduler.go
tsmethurst: Ack! Let's keep it as is and change later if we need to.

tsmethurst (approved):
Brilliant work, well done! I have some small comments but no serious changes! Can't wait to try this out!

internal/processing/polls/expiry.go
tsmethurst: Ah nevermind, I can see it's a bit like the `time.Stamp` format but with the year too

internal/api/model/poll.go
NyaaaWhatsUpDoc: that's the one :woman_facepalming:

internal/config/helpers.gen.go
```diff
@@ -3548,3 +3623,4 @@ func GetRequestIDHeader() string { return global.GetRequestIDHeader() }
 
 // SetRequestIDHeader safely sets the value for global configuration 'RequestIDHeader' field
 func SetRequestIDHeader(v string) { global.SetRequestIDHeader(v) }
+
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: all done now :+1:

internal/gtsmodel/poll.go
tsmethurst: After thinking about this a bit more: sqlite and postgres end up with different types for arrays, iirc, so if you override array types it may not work as expected on Postgres.

internal/typeutils/astointernal.go
NyaaaWhatsUpDoc: right!!

internal/processing/polls/expiry.go
NyaaaWhatsUpDoc: :ok_hand:

internal/api/client/polls/polls.go
NyaaaWhatsUpDoc: even if they were empty i just left them so the existing ones all get aligned to the same height along the string const ends. i was gonna leave it, but if you really want i can remove?

internal/gtsmodel/poll.go
tsmethurst: Mmm, I get where you're coming from, the syntax is a bit weird, but you'll be overriding Postgres's storage of arrays as type `ARRAY` with the `VARCHAR` type used by SQLite by default, so I think it's just going to end up breaking things if you explicitly define the type (for slices, at least)

internal/api/client/polls/polls.go
tsmethurst: Ahh okay. Well, feel free to leave it as-is then! It's not a big deal either way

internal/api/client/polls/polls_vote.go
```diff
@@ -0,0 +1,115 @@
...
+	if pollID == "" {
+		const text = "no poll id specified"
+		errWithCode := gtserror.NewErrorBadRequest(errors.New(text), text)
+		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
+		return
+	}
+
+	var form apimodel.PollVoteRequest
```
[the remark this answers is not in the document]
tsmethurst: (We can also do this later in a chore PR btw, I realize this PR is getting a bit long in the tooth now and I don't wanna burden you with too many review changes when it's non-code stuff we can just fix before release)

internal/processing/workers/fromclientapi.go
NyaaaWhatsUpDoc: you have raised a good point i'm not sure i'm handling this the same for both client and fedi API processing. i'm gonna have a think and do a little reshuffle. should be a quick change either way :)

internal/api/client/polls/polls_vote.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: ah bollocks :thinking: will take a look!

internal/gtsmodel/poll.go
NyaaaWhatsUpDoc: Ah bollocks. *shakes fist at postgres*

NyaaaWhatsUpDoc: have updated that now here if you could double check it for me pls 1cb3be4f9

internal/processing/workers/fromclientapi.go
NyaaaWhatsUpDoc: Aaaaaaand there we go: 940fcdab3

internal/gtsmodel/poll.go
tsmethurst: Looks good!

internal/gtsmodel/poll.go
```diff
@@ -0,0 +1,74 @@
...
+	ID        string    `bun:"type:CHAR(26),pk,nullzero,notnull,unique"`                    // Unique identity string.
+	Choices   []int     `bun:"type:VARCHAR,nullzero,notnull"`                               // The Poll's option indices of which these are votes for.
```
tsmethurst: Ah, missed on here.

NyaaaWhatsUpDoc: that is actually fixed there, i'm not quite sure what happened here!

filipesmedeiros:
Hey, just to check, has this been released? Or still just on the dev (`main`) branch?

filipesmedeiros:
Can I please ask if this implementation will be able to get polls from older posts (from other instances), or only from the moment we upgrade?

filipesmedeiros:
@tsmethurst amazing! How does the team feel about using a non-release version, by building from source? Not recommended?

filipesmedeiros:
Hey! I'm confused about something: does this PR enable seeing edits on federated posts? It seems that at least image descriptions are not updating for me :)

Issue #1218, opened 2022-12-06
[feature] Prometheus metrics implementation


tsmethurst:
We should look into instrumenting GoToSocial with prometheus metrics gathering, so that admins can glean more insight into the performance and load of their instance.

We should start with just one or two useful metrics and play around with it, rather than worrying about implementing everything at once.

We should be wary of performance impact, and make this toggleable so that if it's off it doesn't cause any memory/cpu overhead.

Likely we should have this 'off' by default, since most people likely won't need it or want it.

Here's a little checklist to get us going:

- [ ] Decide which metrics to record.
- [x] Decide on implementation / library.
- [ ] Consider integrating metrics view into the admin panel - worth it? How to do it?
- [x] Minimize metrics overhead, allow disabling
- [x] Consider how this will interact with opentelemetry (if we end up doing that): https://github.com/superseriousbusiness/gotosocial/issues/1230

tsmethurst:
> Another checkmark worth adding (so it's not forgotten): minimize metrics overhead, allow disabling

Added :)

daenney:
One interesting thing in Prometheus that's currently under preview in 2.40.x is the new [native/sparse histograms](https://github.com/prometheus/prometheus/pull/11447). The current histograms have a slight usability issue that the buckets need to be defined upfront and you have to hope to get them right. In the case of GTS that might also be hard because instances of different sizes or running on smol devices might result in fairly different measurements for things like HTTP response latency. The native type doesn't suffer from this making them much more useful in practice.

LittleFox94:
Useful metrics I planned to add but just didn't get around to yet:
* the usual runtime metrics (goroutines, memory, ... standard Go collectors)
* number of users
* number of posts
* number of instances federated with
* the usual HTTP request metrics (count by status code, latency, request/response body size, ...)
  - split between server-to-server / server-to-client
  - s2s split between if requesting or replying
* database metrics, especially latency (as bun already warns about that)

LittleFox94:
Re the requested "Minimize metrics overhead, allow disabling" @NyaaaWhatsUpDoc: I think it's better to filter that on Prometheus level? Most applications I come across don't have something like that and for them you'd have to filter that in Prometheus anyway, making that a better source of truth - does this make sense, what do you think?

NyaaaWhatsUpDoc:
My point is to not have Prometheus metrics being gathered in the binary, and ideally not compiled into the binary, if not necessary. So no overhead if not enabled.

[...]

PR #2334, opened 2023-11-05
[feature] Initial metrics

Tsuribori:
# Description

Relating to #1218, this PR implements the following metrics:
* Go performance/runtime metrics
* Bun metrics
* Gin metrics

These are instrumented via OpenTelemetry.

Following settings are added:
```go
	MetricsEnabled:  false,
	MetricsExporter: "prometheus",
```

Turning metrics on exposes them on a public `/metrics` endpoint for Prometheus pull.

The code has largely been modeled after #1623

## Checklist


Tsuribori:
I don't really know if its possible to add tests for this

tsmethurst:
Thanks for this! Finishing off a couple things now, will try to have a look and review soon!

docs/advanced/metrics.md
```diff
@@ -0,0 +1,18 @@
...
+* Bun (database) metrics
+
+How to configure metrics is explained in the [Observability configuration reference][obs]. Quickstart: add the following to your GoToSocial configuration and restart your instance:
+
+```yaml
+metrics-enabled: true
+metrics-exporter: "prometheus"
+```
+
+This will expose the metrics under the **public** endpoint `/metrics`.
```
daenney, line 15: I think we need to change this. Some folks might run without a proxy or something in front of it, so they wouldn't have the ability to control exposure.

We can do one of two things:
* Have optional basic auth for that endpoint
* Have a second HTTP listener that people can configure to use a different ip:port combo and attach the metrics handler there

docs/advanced/metrics.md
```diff
@@ -0,0 +1,18 @@
...
+GoToSocial comes with [OpenTelemetry][otel] based metrics built-in with pull-style Prometheus exporter. Currently the following metrics are collected:
+* Go performance and runtime metrics
+* Gin (HTTP) metrics
+* Bun (database) metrics
+
+How to configure metrics is explained in the [Observability configuration reference][obs]. Quickstart: add the following to your GoToSocial configuration and restart your instance:
+
+```yaml
+metrics-enabled: true
+metrics-exporter: "prometheus"
```
daenney, line 12: I would suggest we don't add a configuration that can only ever be set to a single valid value. If and when we end up supporting multiple exposition formats, we can make that configurable.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+const (
+	serviceName = "GoToSocial"
+)
+
+func Initialize() error {
+	if !config.GetMetricsEnabled() {
+		return nil
+	}
+
+	fmt.Println("Initializing metrics")
```
daenney, line 46: No `fmt.Println`'s please. If it's necessary to log this, you can grab the logger. But I think we can do without this.

docs/advanced/metrics.md
daenney: I think basic auth is the easiest one, and it's fine if that's a single static secret in the config file. Secondary HTTP listener could also be nice for some other debug endpoints, but that's a lot more shenaniganery to wire up.

Tsuribori: So the Basic Auth username should be some constant (for example "metrics"?) and the password should be configurable through the config. Should there simply be a default value for it or do something it its empty/not set?

daenney: I would probably do something like `metrics-auth-enabled` as a boolean, and based on that ignore or use `metrics-auth-username` and `metrics-auth-password`.

I think in that case it's also fine to not have a default for the username, and require people explicitly configure both once they set `metrics-auth-enabled` to true.

Tsuribori: Should `metrics-auth-enabled` be false or true by default?

daenney: I would set it to `false` by default, since `metrics-enabled` is also `false` by default so there's no chance of accidental exposure.

Tsuribori: Added those three options

daenney (approved):
I think that looks good. This gives us a good starting point and we can start to add custom metrics for things later on. Thanks for the work!

tsmethurst (approved):
Great work! I'll just rebase this on main (if I can) to see if I can fix the merge conflicts...

PR #2344, opened 2023-11-09
[bugfix] actually decrement votes during poll vote delete ...

NyaaaWhatsUpDoc:
one last one today, i promise!!


tsmethurst:
More votes! Higher votes!

PR #2348, opened 2023-11-10
[bugfix] Don't try to update suspended accounts

tsmethurst:
# Description

>

This pull request fixes a bug where accounts that were suspended via the admin panel could become marked as unsuspended again on refresh of the account.

Bug only affected specifically-targeted accounts, not accounts that were blocked by a domain block action. All the account's relationships and posts and media and stuff were still deleted just fine by the side effects.

For good measure, add some early checks to bail fedi requests if a remote account was marked as suspended locally.

## Checklist

### DIFF
--- a/internal/federation/dereferencing/account.go
+++ b/internal/federation/dereferencing/account.go
@@ -42,6 +42,11 @@ import (
 // accountUpToDate returns whether the given account model is both updateable (i.e.
 // non-instance remote account) and whether it needs an update based on `fetched_at`.
 func accountUpToDate(account *gtsmodel.Account) bool {
+	if !account.SuspendedAt.IsZero() {
+		// Can't update suspended accounts.
+		return true
+	}
+
 	if account.IsLocal() {
 		// Can't update local accounts.
 		return true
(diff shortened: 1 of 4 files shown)


daenney (approved):

internal/federation/dereferencing/account.go
```diff
@@ -331,6 +336,11 @@ func (d *Dereferencer) enrichAccountSafely(
 	account *gtsmodel.Account,
 	apubAcc ap.Accountable,
 ) (*gtsmodel.Account, ap.Accountable, error) {
+	// Noop if account has been suspended.
+	if !account.SuspendedAt.IsZero() {
```
NyaaaWhatsUpDoc, line 340: you don't need to check this here as accountIsUpToDate() is always called before entering enrichSafely() :)

internal/federation/federatingprotocol.go
```diff
@@ -288,6 +288,13 @@ func (f *Federator) AuthenticatePostInbox(ctx context.Context, w http.ResponseWr
 		return nil, false, err
 	}
 
+	if !requestingAccount.SuspendedAt.IsZero() {
```
NyaaaWhatsUpDoc, line 291: i was going to say we should probably move this to AuthenticateFederatedRequest() to make sure we do this check in all applicable locations, but then saw that function doesn't guarantee to always have the account model by that point ... what if we updated AuthenticateFederatedRequest() to handle the instance model creation if necessary, and fetching updated account model? that way we can just move this check in there, and it won't need to be added anywhere else.

just this current way of doing it makes me worry we're going to miss out locations where it's needed as it needs to add them in all the locations.

tsmethurst: Will do a refactor of that in a subsequent PR :+1:

PR #2353, opened 2023-11-12
[feature/performance] Wrap incoming HTTP requests in timeout handler

tsmethurst:
# Description

>

This pull request wraps HTTP requests (API, web, etc) in a handler that will time out requests after 10 minutes (if the request hasn't completed already). This should prevent very slow client requests from hanging around for too long in the server, taking up PRECIOUS MEMORY and so on.

Also does a few little chores:

- Deinterface router.
- Tidy up router functions a bit.

## Checklist

### DIFF
--- a/internal/api/util/errorhandling.go
+++ b/internal/api/util/errorhandling.go
@@ -115,6 +115,11 @@ func ErrorHandler(
 		if ctxErr == router.ErrRequestDeadlineExpired {
 			// We timed out the request.
 			errWithCode = gtserror.NewErrorRequestTimeout(err)
+
+			// Be correct and write "close".
+			// See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection#close
+			// and: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408
+			c.Header("Connection", "close")
 		} else {
 			// Client timed out the request.
 			errWithCode = gtserror.NewErrorClientClosedRequest(err)
--- a/internal/api/util/errorhandling.go
+++ b/internal/api/util/errorhandling.go
@@ -27,6 +27,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/gtscontext"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
+	"github.com/superseriousbusiness/gotosocial/internal/router"
 )
 
 // TODO: add more templated html pages here for different error types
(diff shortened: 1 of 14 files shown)


internal/router/timeout.go
```diff
@@ -0,0 +1,65 @@
...
+		r.Context(),
+		requestTimeout,
+		ErrRequestTimeout,
+	)
+	defer cancelCtx()
+
+	// Replace request with new
+	// version, and serve it.
+	tor := r.WithContext(toCtx)
+	*r = *tor
```
NyaaaWhatsUpDoc, line 62: This may cause issues within the net/http package itself. The way to do it would just be pass in this request with separate context into the `.ServeHTTP()` function below.

internal/router/timeout.go
```diff
@@ -0,0 +1,65 @@
...
+		// Leave well enough alone.
+		th.Engine.ServeHTTP(w, r)
+		return
+	}
+
+	// Create timeout ctx.
+	toCtx, cancelCtx := context.WithTimeoutCause(
+		r.Context(),
+		requestTimeout,
+		ErrRequestTimeout,
```
NyaaaWhatsUpDoc, line 55: do we need to use a separate error here? as opposed to just `context.WithTimeout()` and checking for `context.DeadlineExceeded` since all we're doing is just checking for it above?

internal/router/router.go
```diff
@@ -167,78 +194,87 @@ func (r *router) Stop(ctx context.Context) error {
...
-	engine := gin.New()
-	engine.MaxMultipartMemory = maxMultipartMemory
-	engine.HandleMethodNotAllowed = true
-
-	// set up IP forwarding via x-forward-* headers.
-	trustedProxies := config.GetTrustedProxies()
-	if err := engine.SetTrustedProxies(trustedProxies); err != nil {
+// customTLS modifies the router's underlying
+// http server to use custom TLS cert/key pair.
+func (r *Router) customTLS(
```
NyaaaWhatsUpDoc, line 199: bit a subjective one so feel free to ignore, but it feels a bit unidiomatic to refer to the default way to set TLS certs as `customTLS()` as opposed to just like `withTLS(...)`

daenney: It is kinda "custom", since our default way is with the built-in Lets Encrypt support. So `withTLS` would be a bit odd to me since now there's two. `withManuallyProvisionedTLSCertificates` :wink:

NyaaaWhatsUpDoc: oh i forget about that... i just haven't used letsencrypt with gts once :p

internal/router/timeout.go
tsmethurst: Ack, thank you! https://github.com/superseriousbusiness/gotosocial/pull/2353/commits/411ee63ff8302c580a312968b89ce287dc3fd97d

internal/router/timeout.go
tsmethurst: Ah no indeed we don't have to, I just felt like being fancy with the new Cause stuff. I'll change it.

internal/api/util/errorhandling.go
```diff
@@ -90,19 +91,40 @@ func genericErrorHandler(c *gin.Context, instanceGet func(ctx context.Context) (
...
+) {
+	if ctxErr := c.Request.Context().Err(); ctxErr != nil {
+		// Context error means either client has left already,
+		// or server has timed out a very slow request.
+		//
+		// Rewrap the error with something less scary,
+		// and just abort the request gracelessly.
+		err := errWithCode.Unwrap()
+
+		if ctxErr == router.ErrRequestTimeout {
```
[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2353/commits/f86810edf8c387ce733237fa1d21459dc9deed79

internal/router/timeout.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2353/commits/f86810edf8c387ce733237fa1d21459dc9deed79

PR #2361, opened 2023-11-14
[bugfix] Update poll delete/update db queries

tsmethurst:
# Description

>

This pull request updates our db queries a little bit for polls:

- Don't bother checking for db.ErrNoEntries on delete, since this is never returned; use RowsAffected instead. You can never recover from an error inside a transaction, so checking for an ErrNoEntries and continuing would cause the next query to fail anyways.
- Fix choices slice return -- you can't do LIMIT on a Delete using bun, so bun doesn't know that only one row will be deleted, so you have to select into a slice of slices.
- Select fewer columns where possible (v. small performance tweak).

## Checklist

### DIFF
--- a/internal/db/bundb/migrations/20231110142330_small_poll_table_tweaks.go
+++ b/internal/db/bundb/migrations/20231110142330_small_poll_table_tweaks.go
@@ -44,7 +44,7 @@ func init() {
 				Table("polls").
 				Column("expires_at_new").
 				Set("? = ?", bun.Ident("expires_at_new"), bun.Ident("expires_at")).
-				Where("1"). // bun gets angry performing update over all rows
+				Where("TRUE"). // bun gets angry performing update over all rows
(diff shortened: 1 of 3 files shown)


internal/db/bundb/migrations/20231110142330_small_poll_table_tweaks.go
```diff
@@ -44,7 +44,7 @@ func init() {
 				Table("polls").
 				Column("expires_at_new").
 				Set("? = ?", bun.Ident("expires_at_new"), bun.Ident("expires_at")).
-				Where("1"). // bun gets angry performing update over all rows
+				Where("TRUE"). // bun gets angry performing update over all rows
```
tsmethurst, line 47: Fix migration for postgres: it wants a boolean and can't understand sqlite's 0 or 1 booleans. Sqlite doesn't mind either way.

PR #2364, opened 2023-11-16
[bugfix] self-referencing collection pages for status replies

NyaaaWhatsUpDoc:
# Description
- updates the paging package to use "slices" instead of the old experimental slices package
- updates StatusRepliesGet() processor function to use new collection and paging code
- updates GetStatusChildren() to be much simpler and return in case of reply fetching errors
- returns OrderedCollection instead of Collection from replies endpoint as our paging only works with an order ...
- updates the /replies endpoint to follow more AP compliant logic, this should be fine even with our compliance "guarantee" since this is an AP endpoint and not a client-side one

## Checklist


NyaaaWhatsUpDoc:
the new AP endpoint works as expected with other gotosocial instances, mastodon has issues though due to HTTP signatures and the fact the endpoint requires query parameters, so unrelated.

internal/api/activitypub/users/repliesget.go
```diff
@@ -120,46 +119,46 @@ func (m *Module) StatusRepliesGETHandler(c *gin.Context) {
...
+		page,
+		onlyOtherAccounts,
+	)
+	if errWithCode != nil {
+		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
 		return
 	}
 
-	c.Data(http.StatusOK, format, b)
+	c.JSON(http.StatusOK, resp)
```
tsmethurst, line 163: iirc the JSON function sets the returned content-type to application/json, where we want one of the ActivityPub content types to be returned.

internal/util/ptr.go
```diff
@@ -33,3 +33,11 @@ func EqualPtrs[T comparable](t1, t2 *T) bool {
 func Ptr[T any](t T) *T {
 	return &t
 }
+
+// PtrValueOr returns either value of ptr, or default.
+func PtrValueOr[T any](t *T, _default T) T {
```
tsmethurst, line 38: Ah handy :)

internal/api/activitypub/users/repliesget_test.go
```diff
@@ -76,21 +77,34 @@ func (suite *RepliesGetTestSuite) TestGetReplies() {
...
+
+	// Create JSON string of expected output.
+	expect := toJSON(map[string]any{
+		"@context":   "https://www.w3.org/ns/activitystreams",
+		"type":       "OrderedCollection",
+		"id":         targetStatus.URI + "/replies?only_other_accounts=false",
+		"first":      targetStatus.URI + "/replies?limit=20&only_other_accounts=false",
+		"totalItems": 1,
+	})
+	assert.Equal(suite.T(), expect, string(b))
```
tsmethurst, line 98: These `assert.Equal(suite.T(), ...)` functions within a suite can all be replace with just `suite.Equal(...)`. Definitely not urgent, and it's functionally the same, but if you feel like changing them then feel free, since it's a little more concise :P

internal/api/activitypub/users/repliesget.go
NyaaaWhatsUpDoc: ah gotcha will change that back :sunglasses:

internal/api/activitypub/users/repliesget.go
```diff
@@ -120,46 +119,46 @@ func (m *Module) StatusRepliesGETHandler(c *gin.Context) {
...
-		onlyOtherAccounts = i
-	}
+	var onlyOtherAccounts bool
 
-	minID := ""
-	minIDString := c.Query(MinIDKey)
-	if minIDString != "" {
-		minID = minIDString
+	// Look for supplied 'onlyOtherAccounts' query key.
+	if raw, ok := c.GetQuery("only_other_accounts"); ok {
```
tsmethurst, line 125: Could you put this in apiutil instead? There's quite a few Parse functions in there that more or less fit this signature. Probably not a bad idea to add a "Paging keys" section in the consts in `api/util/parsequery.go` and use that.

NyaaaWhatsUpDoc: ah yep i'll this only_other_accounts var there too. we don't need the paging keys in there as these are actually defined in the 'paging' package

internal/api/activitypub/users/repliesget_test.go
NyaaaWhatsUpDoc: ah yep, agreed! though may do this another time :p

internal/db/bundb/status_test.go
```diff
@@ -165,7 +165,7 @@ func (suite *StatusTestSuite) TestGetStatusTwice() {
 
 func (suite *StatusTestSuite) TestGetStatusChildren() {
 	targetStatus := suite.testStatuses["local_account_1_status_1"]
-	children, err := suite.db.GetStatusChildren(context.Background(), targetStatus, true, "")
+	children, err := suite.db.GetStatusReplies(context.Background(), targetStatus.ID)
```
tsmethurst, line 168: Hmm, how come this has changed to call GetStatusReplies?

internal/db/bundb/util.go
```diff
@@ -99,7 +100,7 @@ func loadPagedIDs(cache *cache.SliceCache[string], key string, page *paging.Page
 	// order. Depending on the paging requested
 	// this may be an unexpected order.
 	if page.GetOrder().Ascending() {
-		ids = paging.Reverse(ids)
+		slices.Reverse(ids)
```
tsmethurst, line 103: Love these new slices functions :)

internal/processing/fedi/common.go
```diff
@@ -20,67 +20,68 @@ package fedi
...
 	"github.com/superseriousbusiness/gotosocial/internal/gtscontext"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 )
 
-func (p *Processor) authenticate(ctx context.Context, requestedUsername string) (
-	*gtsmodel.Account, // requestedAccount
-	*gtsmodel.Account, // requestingAccount
+func (p *Processor) authenticate(ctx context.Context, requestedUser string) (
+	*gtsmodel.Account, // requester: i.e. user making the request
```
tsmethurst, line 31: Oh, you swapped the order of these right? Was that on purpose?

tsmethurst (commented):
Good stuff! Few comments but nothing breaking I believe except for the JSON type comment (should be activitystreams content-type)

internal/processing/fedi/common.go
NyaaaWhatsUpDoc: i didn't swap them intentionally, but i did end up updating all the calling pieces of code so it works :sweat_smile:

internal/api/activitypub/users/repliesget_test.go
tsmethurst: No problemo, we can do a chore PR at some point (*waves vaguely*) to replace them all everywhere at once, might be a bit easier.

internal/api/activitypub/users/repliesget.go
NyaaaWhatsUpDoc: and done :) 26d90e69a

internal/api/activitypub/users/repliesget.go
NyaaaWhatsUpDoc: done! 462e342f73dd28d31ddd7fc415caae63e23191d5

internal/db/bundb/status_test.go
NyaaaWhatsUpDoc: renamed it, and added a separate GetStatusChildren() test, which also expects the same results :p e37ca7c03

PR #2374, opened 2023-11-21
[performance] http response encoding / writing improvements

NyaaaWhatsUpDoc:
# Description

This adds new wrapper functions within `internal/api/util` that handle response writing with content type / length setting. Where possible these will stream responses, and an all cases except for pre-prepared byte slices has the addition of pooled memory buffers for response encoding (including by JSON and XML). And multiple cases this also enables the use of pre-prepared JSON responses to reduce unneeded encoding for simple but known cases.

## Checklist


internal/api/util/response.go
```diff
@@ -0,0 +1,281 @@
...
+// to ResponseWriter with given status code content-type.
+func WriteResponse(
+	rw http.ResponseWriter,
+	r *http.Request,
+	statusCode int,
+	contentType string,
+	data io.Reader,
+	length int64,
+) {
+	if length < 0 {
```
tsmethurst, line 79: Should this be `<=` instead of `<`? Do we always set length to -1 when we don't know it?

internal/api/util/response.go
```diff
@@ -0,0 +1,281 @@
...
+	r *http.Request,
+	statusCode int,
+	contentType string,
+	data any,
+) {
+	// Acquire buffer.
+	buf := getBuf()
+
+	// Wrap buffer in JSON encoder.
+	enc := json.NewEncoder(buf)
```
tsmethurst, line 125: We might want to do `SetEscapeHTML(false)` on JSON encoders, since we don't really need escaping of HTML characters within the JSON in this context (API).

tsmethurst (approved):
Looks good, very clean and tidy :)

daenney:
I'm not sure I understand the change. Gin's JSON, and associated, methods will do stuff like set the right content-type and I don't think I've ever observed the content-length missing either?

internal/api/util/response.go
NyaaaWhatsUpDoc: i mean a length of zero is a valid content-length, which might happen if say we just want to pass in an arbitrary reader + length (even if the reader contains zero data). -1 is a specifically invalid content-length so a useful flag to indicate non-standard behaviour

internal/api/util/response.go
NyaaaWhatsUpDoc: ooh i'm sure that should result in a small little performance boost too with less escape checking + handling to do :eyes:

NyaaaWhatsUpDoc: 50023f117

PR #2378, opened 2023-11-22
[docs] Add a warning about SQLite cache sizes

daenney:
# Description

People keep tweaking this and running into performance issues as a consequence.

## Checklist


daenney:
> Looks good, but I'm not sure about the 'consult with the developers bit'. Ideally we'd provide a tuning document in the rest of the documentation _at some point_ (during beta maybe?) which we can refer to instead, but for now I'd just leave it out I think and advise people not to touch it.

Fair yeah. I've tweaked it now to say:
* You shouldn't touch this
* If you've touched it, you need to tell us

PR #2384, opened 2023-11-24
[bugfix] Add Actor to outgoing poll vote Create; other fixes

tsmethurst:
# Description

>

Few small fixes to poll federation.

- Poll vote from one GtS instance was being rejected by another, because the Actor property wasn't populated on the outgoing Create. So make sure actor is set properly on Create.
- Use poll vote creation time as published time.
- Reorder poll vote ID creation.
- Add test for above, update docs with example Create.

## Checklist

### DIFF
--- a/internal/typeutils/wrap.go
+++ b/internal/typeutils/wrap.go
@@ -19,7 +19,6 @@ package typeutils
 
 import (
 	"net/url"
-	"time"
 
 	"github.com/superseriousbusiness/activity/pub"
 	"github.com/superseriousbusiness/activity/streams"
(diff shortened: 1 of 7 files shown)


internal/typeutils/wrap.go
```diff
@@ -91,7 +91,10 @@ func WrapStatusableInCreate(status ap.Statusable, iriOnly bool) vocab.ActivitySt
 	return create
 }
 
-func WrapPollOptionablesInCreate(options ...ap.PollOptionable) vocab.ActivityStreamsCreate {
+func WrapPollOptionablesInCreate(
+	publishedAt time.Time,
```
NyaaaWhatsUpDoc, line 95: the signature of this function feels a bit, odd. what if you passed in the vote object instead of just the publish time? or just perform `ap.SetPublished()` in the function calling this one?

NyaaaWhatsUpDoc:
aside from the 1 comment, looks good to me!

internal/typeutils/wrap.go
tsmethurst: Mmmm OK I took another look at this, and for now just removed the in-between step of converting it to a slice of Notes, and then into a Create. We didn't need that slice of Notes for anything *except* making it into a Create, so this feels like it makes sense (and fits with the signature of some of the other functions in the typeconverter). What do you think?

daenney (approved):

PR #2385, opened 2023-11-24
[docs] Add docs about memory requirements and swap

daenney:

### DIFF
--- a/docs/getting_started/index.md
+++ b/docs/getting_started/index.md
@@ -87,7 +87,10 @@ The BSD family of distributions don't document memory requirements as much, but 
 
 ### Swap
 
-Do not turn off swap. If you want to run without swap, ensure you understand the memory allocation patterns of the different workloads on your system and how the memory management subsystem of your kernel will behave under various conditions like memory pressure.
+!!! warning
+    Do not turn off swap.
+
+In order to safely run without swap, you'll need to tune the kernel, system and your workloads accordingly. This requires a good understanding of your kernel's memory management system as well as the memory usage patterns of the workloads you're running.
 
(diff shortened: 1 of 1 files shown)


docs/getting_started/index.md
```diff
@@ -65,6 +65,35 @@ If you decide to use a VPS instead, you can spin yourself up something cheap wit
...
+
+[archreq]: https://wiki.archlinux.org/title/installation_guide
+[debreq]: https://www.debian.org/releases/stable/amd64/ch02s05.en.html
+[ubireq]: https://ubuntu.com/server/docs/installation
+[rhelreq]: https://access.redhat.com/articles/rhel-limits#minimum-required-memory-3
+[fedorareq]: https://docs.fedoraproject.org/en-US/fedora/latest/release-notes/welcome/Hardware_Overview/#hardware_overview-specs
+
+### Swap
+
+!!! warning
```
tsmethurst, line 90: I would actually nuance this a little bit rather than just saying "don't turn off swap" -- if anything, I'd lean in the opposite direction and say "DO turn off swap", since GtS gets VERY slow when it starts swapping.

Perhaps we could update this to say that swap might be necessary for getting an operating system running on a very constrained environment, but GtS doesn't do well while swapping, so if you *can* turn it off while keeping your system running normally (and if GtS is the only thing running on that server) it's better to turn it off.

daenney: Fair yeah. I'll rewrite this a bit. I'll remove the outright warning, and just try to push a bit more towards the "unless you know what's up, don't ignore your distribution recommendations for swap etc."

daenney: There, I've rewritten the swap section a bit and added a section on implementing memory limits if you want to.

tsmethurst: nice one :)

PR #2386, opened 2023-11-25
[bugfix] Don't copy ptr fields in caches

tsmethurst:
# Description

>

This PR fixes the following scenario:

- Some of my statuses get cached after database calls.
- My account model is attached to those statuses.
- I update my profile picture.
- The statuses still show the old profile picture, on both the web view and via the API.

Essentially, what was happening in the above scenario is that the old account model was being kept on the cached statuses, and those statuses weren't being invalidated when the account model was updated.

There are two possible approaches to fixing this:

1. Invalidate all cache entries that contain a pointer to a given account (or whatever) when that account (or whatever) updates; OR --
2. Don't return cached struct pointers when copying a cached model and giving it to the caller.

This PR takes the second approach, on the basis that it will lead to fewer database calls than the first approach, and preclude super complicated cache invalidation logic.

## Checklist

### DIFF
--- a/internal/cache/gts.go
+++ b/internal/cache/gts.go
@@ -1073,17 +1073,25 @@ func (c *GTSCaches) initUser() {
 
 	log.Infof(nil, "cache size = %d", cap)
 
+	copyF := func(u1 *gtsmodel.User) *gtsmodel.User {
+		u2 := new(gtsmodel.User)
+		*u2 = *u1
+
+		// Don't include ptr fields that
+		// will be populated separately.
+		// See internal/db/bundb/user.go.
+		u2.Account = nil
+
+		return u2
+	}
+
 	c.user = result.New([]result.Lookup{
 		{Name: "ID"},
 		{Name: "AccountID"},
 		{Name: "Email"},
 		{Name: "ConfirmationToken"},
 		{Name: "ExternalID"},
-	}, func(u1 *gtsmodel.User) *gtsmodel.User {
-		u2 := new(gtsmodel.User)
-		*u2 = *u1
-		return u2
-	}, cap)
+	}, copyF, cap)
 
 	c.user.IgnoreErrors(ignoreErrors)
 }
--- a/internal/cache/gts.go
+++ b/internal/cache/gts.go
@@ -306,6 +306,20 @@ func (c *GTSCaches) initAccount() {
 
 	log.Infof(nil, "cache size = %d", cap)
 
+	copyF := func(a1 *gtsmodel.Account) *gtsmodel.Account {
+		a2 := new(gtsmodel.Account)
+		*a2 = *a1
+
+		// Don't include ptr fields that
+		// will be populated separately.
+		// See internal/db/bundb/account.go.
+		a2.AvatarMediaAttachment = nil
+		a2.HeaderMediaAttachment = nil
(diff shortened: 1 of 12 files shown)


internal/cache/gts.go
```diff
@@ -306,6 +306,20 @@ func (c *GTSCaches) initAccount() {
...
 	log.Infof(nil, "cache size = %d", cap)
 
+	copyF := func(a1 *gtsmodel.Account) *gtsmodel.Account {
+		a2 := new(gtsmodel.Account)
+		*a2 = *a1
+
+		// Don't include ptr fields that
+		// will be populated separately.
+		// See internal/db/bundb/account.go.
+		a2.AvatarMediaAttachment = nil
```
NyaaaWhatsUpDoc, line 316: i'm pretty sure these don't get copied anyways, because we specifically don't populate sub-models in the cache load functions :thinking:

NyaaaWhatsUpDoc: we call 'Populate()' after the cache load functions specifically so they don't populate sub-models in the cache loaders.

did these changes actually fix the issue?

NyaaaWhatsUpDoc: e.g.

```golang

func (a *accountDB) getAccount(ctx context.Context, lookup string, dbQuery func(*gtsmodel.Account) error, keyParts ...any) (*gtsmodel.Account, error) {
	// Fetch account from database cache with loader callback
	account, err := a.state.Caches.GTS.Account().Load(lookup, func() (*gtsmodel.Account, error) {
		var account gtsmodel.Account

		// Not cached! Perform database query
		if err := dbQuery(&account); err != nil {
			return nil, err
		}

		return &account, nil
	}, keyParts...)
	if err != nil {
		return nil, err
	}

	if gtscontext.Barebones(ctx) {
		// no need to fully populate.
		return account, nil
	}

	// Further populate the account fields where applicable.
	if err := a.PopulateAccount(ctx, account); err != nil {
		return nil, err
	}

	return account, nil
}
```

NyaaaWhatsUpDoc:
i think the issue you're likely seeing is probably in the timeline cache rather than the database cache, given that the timeline pre-prepares a frontend model with all the extra parts fully populated.

internal/cache/gts.go
tsmethurst: @NyaaaWhatsUpDoc since we talked about this together out of band yesterday (because I couldn't access github on my phone yesterday afternoon), I'll just post the summary here:

---

> the point in that PR is that we're putting stuff in the cache with pointers populated, so when it comes out the cache again, stuff is still populated, but with not necessarily up to date values... The test I wrote on the PR fails when you remove the changes I made in the cache package
> We could strip them on the way into the cache, but there's no function for that afaik, so stripping them on the way out in the copy function was the only way I saw to do it 
> This means we're still storing populated items with potentially out of date references on them, which is not ideal, but at least callers don't get them now 

So the problem is eg., a new status comes in from the Fedi API, we populate it by dereferencing everything, setting the appropriate Account on the status, etc, then we store the status in the DB (by calling insert) before passing it off somewhere else for further processing. That call to the DB functions puts the status in the cache using the callback. When someone else comes along later and calls Get on the DB, they get the cached version that was already populated, no matter whether they call Populate on it later.

Issue #1978, opened 2023-07-11
[bug] Improper handling of HTTP range headers with suffix-length > content-length


MaeIsBad:
GTS responds to range headers with their suffix length larger than content-length with 416-Unsatisfiable Range, it should instead attempt to return as much of the range as possible.

```
~ curl -r 0-9999999999 'https://gts.superseriousbusiness.org/fileserver/016VP4S3BP3QSJBACEQ6J1VFDX/attachment/original/01HVXG7ZB92RBAQ2GYRTK0T0Y9.png'
Unsatisfiable Range
```

According to RFC 9110:

> If the [last-pos](https://www.rfc-editor.org/rfc/rfc9110#rule.int-range) value is greater than or equal to the current length of the representation data, the byte range is interpreted as the remainder of the representation (i.e., the server replaces the value of [last-pos](https://www.rfc-editor.org/rfc/rfc9110#rule.int-range) with a value that is one less than the current length of the selected representation).

https://www.rfc-editor.org/rfc/rfc9110#section-14.1.2-6

This breaks akkoma's media proxy cache fetching of GTS posts, when using the recommended nginx configuration(https://akkoma.dev/AkkomaGang/akkoma/src/branch/develop/installation/nginx/akkoma.nginx#L86)

### What's your GoToSocial Version?

GoToSocial 0.10.0-rc1 git-2a99df0 (https://gts.superseriousbusiness.org/)
GoToSocial 0.8.1 git-4dbf9a2 (https://social.noyu.me/)

MaeIsBad:
If anyone wants to test it I'm pretty sure the fix is as simple as
```patch
diff --git a/fs.go b/fs.go
index 2b025ff..a664459 100644
--- a/fs.go
+++ b/fs.go
@@ -314,10 +314,7 @@ func (fs *FS) serveFileRange(rw http.ResponseWriter, r *http.Request, fsreq *FSR
                }

                if end > size {
-                       // This range exceeds length of the file, therefore unsatisfiable
-                       rw.Header().Set("Content-Range", "bytes *"+strconv.FormatInt(size, 10))
-                       http.Error(rw, "Unsatisfiable Range", http.StatusRequestedRangeNotSatisfiable)
-                       return
+                       end = size -1
                }
        } else {
                // No end supplied, implying file end
```

PR #2395, opened 2023-11-29
[bugfix] Correctly handle range > content-length

Jadeiin:
# Description

Hi I found this bug in Pleroma/Akkoma's MediaProxy. Currently GTS instance returns incorrect range headers when suffix length is larger than content-length:

```bash
$ url="https://goblin.technology/fileserver/01BPSX2MKCRVMD4YN4D71G9CP5/attachment/original/01E4BA3HFHWN7GDE2Y89F0J563.png"
$ curl -r 0-1048576 -X GET -I $url | grep Content-Range
Content-Range: bytes 0-1048576/11021
```

The correct range header should be `bytes 0-11020/11021`,  which only requires formatting correct `endRng` when range > content-length.

Related issue(closed but not actually fixed): #1978 

## Checklist

### DIFF
--- a/internal/api/fileserver/servefile.go
+++ b/internal/api/fileserver/servefile.go
@@ -206,10 +206,11 @@ func serveFileRange(rw http.ResponseWriter, r *http.Request, src io.Reader, rng 
 			return
 		}
 
-		if end > size {
+		if end >= size {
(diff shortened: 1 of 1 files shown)


tsmethurst:
Oh nice, thank you :)

PR #2399, opened 2023-11-29
[bugfix] return 400 Bad Request on more cases of malformed AS data

NyaaaWhatsUpDoc:
# Description

Adds a "malformed" error value flag which we now set on parse-type failures during AS->internal conversions. These are then caught after PostInbox() and returned as 400 Bad Request errors appropriately. Also has some small tweaks here and there, and updates the `go-errors` package with some tweaks (which we should eventually incorporate in our own `gtserror` package too).

## Checklist


NyaaaWhatsUpDoc:
linter is angry, and there's an error message in a test that needs updating. a job for later kim ... :]

internal/typeutils/astointernal.go
```diff
@@ -38,201 +37,204 @@ import (
...
+		err := gtserror.New("unusable iri property")
+		return nil, gtserror.SetMalformed(err)
 	}
-	uri := uriProp.GetIRI()
 
-	// we don't know the account, or we're being told to update it, so we need to generate it from the person -- at least we already have the URI!
-	acct := &gtsmodel.Account{}
-	acct.URI = uri.String()
+	// Stringify URI.
+	uri := _uri.String()
```
tsmethurst, line 50: I'm not really a big fan of this underscore syntax; does it have a special meaning in Go to prefix a variable name with an underscore? Reason I ask is because in JS/TS it signifies an unused variable, but I suppose that's not the case here. If possible, I'd prefer to avoid possibly-confusing syntax crossovers between the two languages in the codebase. Maybe you could use an inline function instead here to just extract the values we care about in this function?

For example:

```go
uri, host, err := func() (string, string, error) {
  // Extract uri and host, and stringify uri in here.
}()
```

What do you think? Perhaps there's a better way.

internal/typeutils/astointernal.go
```diff
@@ -38,201 +37,204 @@ import (
...
-	// SharedInboxURI:
-	// only trust shared inbox if it has at least two domains,
-	// from the right, in common with the domain of the account
+	// Extract the outbox IRI property.
+	outboxIRI := ap.GetOutbox(accountable)
+	if outboxIRI != nil {
+		acct.OutboxURI = outboxIRI.String()
+	}
+
+	// Extract a SharedInboxURI, but only trust it if a subdomain of account's domain.
```
tsmethurst, line 176: This comment isn't quite right, we also trust it if it's the *same* domain, not just a subdomain.

internal/typeutils/astointernal.go
```diff
@@ -38,201 +37,204 @@ import (
...
-
-	// FeaturedURI aka pinned collection:
-	// Only trust featured URI if it has at least two domains,
-	// from the right, in common with the domain of the account
-	if featured := accountable.GetTootFeatured(); featured != nil && featured.IsIRI() {
-		if featuredURI := featured.GetIRI(); // nocollapse
-		featuredURI != nil && dns.CompareDomainName(acct.Domain, featuredURI.Host) >= 2 {
-			acct.FeaturedCollectionURI = featuredURI.String()
-		}
+	// Extract a FeaturedURI, but only trust it if a subdomain of account's domain.
```
tsmethurst, line 195: Same comment as above.

internal/typeutils/astointernal.go
```diff
@@ -38,201 +37,204 @@ import (
...
-	status := new(gtsmodel.Status)
-
-	// status.URI
-	//
-	// ActivityPub ID/URI of this status.
-	idProp := statusable.GetJSONLDId()
-	if idProp == nil || !idProp.IsIRI() {
-		return nil, gtserror.New("no id property found, or id was not an iri")
+	// Extract URI from statusable
+	_uri := ap.GetJSONLDId(statusable)
```
tsmethurst, line 226: Same comment as above re: underscore syntax.

internal/typeutils/astointernal.go
```diff
@@ -329,20 +330,17 @@ func (c *Converter) ASStatusToStatus(ctx context.Context, statusable ap.Statusab
...
-	accountURI := attributedTo.String()
-
-	account, err := c.state.DB.GetAccountByURI(ctx, accountURI)
+	status.Account, err = c.getASAttributedToAccount(ctx,
+		status.URI,
+		statusable,
+	)
 	if err != nil {
-		err = gtserror.Newf("db error getting status author account %s: %w", accountURI, err)
 		return nil, err
```
tsmethurst, line 338: Is it worth returning malformed in this case? I'm not sure. What must have happened if we get to this point and we don't have the account?

internal/typeutils/astointernal.go
```diff
@@ -725,86 +680,164 @@ func (c *Converter) ASFlagToReport(ctx context.Context, flaggable ap.Flaggable)
...
+		err := gtserror.Newf("unusable object property iris for %s", uri)
+		return nil, gtserror.SetMalformed(err)
 	}
 
-	var targetAccountURI *url.URL
 	for _, object := range objects {
 		switch {
-		case object.Host != config.GetHost():
-			// object doesn't belong to us, just ignore it
+		case object.Host != host:
```
tsmethurst, line 691: Not something for this PR, but just noting it here: eventually we might also want to accept reports from trusted instances about other instances than our own (I believe Mastodon is introducing this feature, and it's quite cool). So this bit would have to change, if we ever implement that. Again, not something to do anything about here and now, it just came to mind.

internal/typeutils/astointernal.go
NyaaaWhatsUpDoc: yeah i agree it is a bit odd, i'll fix it :)

internal/typeutils/astointernal.go
NyaaaWhatsUpDoc: that would be cool to implement!

internal/typeutils/astointernal.go
NyaaaWhatsUpDoc: i'm not sure if it fits "malformed" though, malformed more indicates that the data doesn't represent a correct ActivityStreams object type that we're parsing. but if we don't have the account by this point it's more an internal server error due to likely one of our bugs :P

PR #2402, opened 2023-11-30
[bugfix] always go through status parent dereferencing on isNew, even on data-race

### LINKED ISSUE #2396: [bug] Parent status not always being dereferenced when a reply comes in
Think we made a cheeky little regression somewhere. When a status comes in from the fedi API, we don't seem to dereference the parent status properly when it first comes in, so it looks in the timeline (and the db) as though the status doesn't reply to anything, even though it does. This means that threads can appear to be severed, at least until you open the status again (at least) two hours later, and the thread gets dereferenced again.

I think the bug is somewhere in `(d *Dereferencer) RefreshStatus()`

NyaaaWhatsUpDoc:
# Description

As the title says really. This isn't ideal, but it's the only (easy) way of handling this right now, without performing a parent dereference within mutex lock and possibly risk lock-ups if there's any threads that end up self-referencing. This shouldn't happen *too* often so performance impact shouldn't be much, especially as all of those statuses will also just return fetched_at as too new (all of which will be cache hits).

Also removes the dereference account step _after_ dereferencing status in the fedi API worker, as to reach that point the account must already have been dereferenced / refreshed.

Closes #2396 

## Checklist

### DIFF
--- a/internal/federation/dereferencing/status.go
+++ b/internal/federation/dereferencing/status.go
@@ -297,8 +297,8 @@ func (d *Dereferencer) enrichStatusSafely(
 		// We leave 'isNew' set so that caller
 		// still dereferences parents, otherwise
 		// the version we pass back may not have
-		// these dereferenced yet (since those
-		// happen OUTSIDE the federator lock).
+		// these attached as inReplyTos yet (since
+		// those happen OUTSIDE federator lock).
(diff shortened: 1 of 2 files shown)


tsmethurst:
Nice one!

Issue #2394, opened 2023-11-28
[bug] boosted posts sometimes appear to be authored by the booster


illfygli:
### Describe the bug with a clear and concise description of what the bug is.

A boosted toot may change so that the booster appears as the author, the date of the toot becomes the boost date, and so on.

I've seen it change before my eyes, so it starts out correct, and then the booster usurps the original author. :)

Unfortunately, I found nothing even remotely interesting in the logs.

### What's your GoToSocial Version?

0.12.2

### GoToSocial Arch

x86_64 binary

### What happened?

A boost appeared on the timeline.

Then the boost metadata replaced that of the original post, making it look like an original post by the booster.

### What you expected to happen?

Post should keep the original author.

### How to reproduce it?

Stare at the timeline for extended periods of time. Sorry, I don't know. :(

### Anything else we need to know?

You're the best, keep up the good work.

tsmethurst:
Oh, I didn't mean to mark this as completed. I'll open it again until we can validate it's actually fixed.

illfygli:
Upgraded to 0.13.0-rc1 and will keep an eye out!

tsmethurst:
I haven't seen this again yet :eyes:

illfygli:
Me neither!

Seems like it was actually quite common, once I started looking for it, but now I haven't seen it on 0.13.0 or any of its RCs.

tsmethurst:
Great :) I'll close it for now and let's see if it pops up again in future.

PR #2405, opened 2023-12-01
[bugfix/chore] `Announce` reliability updates

tsmethurst:
# Description

>

- ~~Use per-URI deref locks for Announce dereferences to avoid races.~~
- Use new `BoostOfURI` field on statuses for doing Announce derefs (new field not stored in the db).
- Ensure faves and boosts are redirected properly to top-level status when the fave or boost was created targeting the ID of a boost wrapper.
- Don't serve boost wrappers via the web UI or the fedi API.
- Use `Unretrievable` to do noop nicely when a boost target is on a blocked domain.
- Fail to serialize boost via the client API if BoostOfID is set but BoostOf is not (for whatever reason).

This is rather a hodge-podge of fixes and updates. It MAY resolve https://github.com/superseriousbusiness/gotosocial/issues/2394 but we'll have to run it for a while to see, since I couldn't actually definitely identify the root cause of that bug.

## Checklist

### DIFF
--- a/internal/federation/dereferencing/announce.go
+++ b/internal/federation/dereferencing/announce.go
@@ -48,11 +48,11 @@ func (d *Dereferencer) EnrichAnnounce(
 	}
 
 	// Parse the boost target status URI.
-	targetURIObj, err := url.Parse(boost.BoostOfURI)
+	targetURIObj, err := url.Parse(targetURI)
 	if err != nil {
 		return nil, gtserror.Newf(
 			"couldn't parse boost target status URI %s: %w",
-			boost.BoostOfURI, err,
+			targetURI, err,
 		)
 	}
 
--- a/internal/federation/dereferencing/announce.go
+++ b/internal/federation/dereferencing/announce.go
@@ -56,14 +56,6 @@ func (d *Dereferencer) EnrichAnnounce(
 		)
 	}
 
-	// Ensure target status isn't from a blocked host.
-	if blocked, err := d.state.DB.IsDomainBlocked(ctx, targetURIObj.Host); err != nil {
-		return nil, gtserror.Newf("error checking blocked domain: %w", err)
-	} else if blocked {
-		err = gtserror.Newf("%s is blocked", targetURIObj.Host)
-		return nil, gtserror.SetUnretrievable(err)
-	}
-
 	// Fetch/deref status being boosted.
 	var target *gtsmodel.Status
 
@@ -72,6 +64,9 @@ func (d *Dereferencer) EnrichAnnounce(
 		target, err = d.state.DB.GetStatusByURI(ctx, targetURI)
 	} else {
 		// This is a remote status, we need to dereference it.
+		//
+		// d.GetStatusByURI will handle domain block checking for us,
+		// so we don't try to deref an announce target on a blocked host.
 		target, _, err = d.GetStatusByURI(ctx, requestUser, targetURIObj)
 	}
 
--- a/internal/federation/dereferencing/announce.go
+++ b/internal/federation/dereferencing/announce.go
@@ -29,28 +29,28 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/id"
 )
 
-// EnrichAnnounceSafely enriches the given boost wrapper status
+// EnrichAnnounce enriches the given boost wrapper status
 // by either fetching from the DB or dereferencing the target
 // status, populating the boost wrapper's fields based on the
 // target status, and then storing the wrapper in the database.
-// It uses per-URI locks to ensure the boost is only enriched once.
+// The wrapper is then returned to the caller.
 //
 // The provided boost wrapper status must have BoostOfURI set.
-func (d *Dereferencer) EnrichAnnounceSafely(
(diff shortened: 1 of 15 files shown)


internal/processing/common/status.go
```diff
@@ -119,6 +119,37 @@ func (p *Processor) GetVisibleTargetStatus(
 	return target, nil
 }
 
+// GetVisibleTargetStatusUnwrapped is like GetVisibleTargetStatus,
+// but if targetID points to a boost wrapper status, that status
+// will be 'unwrapped' to return the boosted status instead.
+func (p *Processor) GetVisibleTargetStatusUnwrapped(
```
NyaaaWhatsUpDoc, line 125: I think the function name of this would make a bit more sense if it was to say `GetVisibleTargetBoost()` :thinking: as `TargetBoost` still makes sense in terms of thinking about the status the boost is targeting. `Unwrapped()` sounds a little to vague imo

internal/federation/dereferencing/announce.go
```diff
@@ -20,66 +20,119 @@ package dereferencing
...
-		// we can't do anything unfortunately
-		return errors.New("DereferenceAnnounce: no URI to dereference")
+// EnrichAnnounceSafely enriches the given boost wrapper status
+// by either fetching from the DB or dereferencing the target
+// status, populating the boost wrapper's fields based on the
+// target status, and then storing the wrapper in the database.
+// It uses per-URI locks to ensure the boost is only enriched once.
+//
+// The provided boost wrapper status must have BoostOfURI set.
+func (d *Dereferencer) EnrichAnnounceSafely(
```
NyaaaWhatsUpDoc, line 39: I think it might fit the current exported package naming scheme a bit better if this was to be called `RefreshAnnounce()` or even just `EnrichAnnounce()`. the safely suffixes are largely for internal functions to indicate which are performing locks (which those outside the federator package shouldn't need to be aware of).

internal/processing/common/status.go
tsmethurst: It should be boost target then right, not target boost? But that's also a bit misleading because the passed-in status ID will not always correspond to a boost wrapper status; 99% of the time it will just point to an already-unwrapped status. The reason I chose the wrapped/unwrapped terminology here is because that's how it's referred to elsewhere in the codebase. I'm not sure if changing it to GetVisibleTargetBoost offers much more clarity. Is there something else that would work?

internal/federation/dereferencing/announce.go
```diff
@@ -20,66 +20,119 @@ package dereferencing
...
+		err = gtserror.Newf("%s is blocked", targetURIObj.Host)
+		return gtserror.SetUnretrievable(err)
 	}
 
-	var boostedStatus *gtsmodel.Status
+	// Acquire per-URI deref lock; safely defer in
+	// case of panic. Important: use the URI of the
+	// BOOST WRAPPER for this, not the target's URI;
+	// the latter will be locked by d.GetStatusByURI.
+	unlock := d.state.FedLocks.Lock(boost.URI)
```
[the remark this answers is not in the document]
tsmethurst: Ah got it, I'll change it now

internal/federation/dereferencing/announce.go
tsmethurst: Got it! Changed here: https://github.com/superseriousbusiness/gotosocial/pull/2405/commits/4cc95b3d343537160c9cf4bc6e4fe8823ac7d290

internal/federation/dereferencing/announce.go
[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2405/commits/4cc95b3d343537160c9cf4bc6e4fe8823ac7d290

internal/federation/dereferencing/announce.go
```diff
@@ -29,48 +29,41 @@ import (
...
 	targetURIObj, err := url.Parse(boost.BoostOfURI)
 	if err != nil {
-		return gtserror.Newf(
+		return nil, gtserror.Newf(
 			"couldn't parse boost target status URI %s: %w",
 			boost.BoostOfURI, err,
 		)
 	}
 
 	// Ensure target status isn't from a blocked host.
```
NyaaaWhatsUpDoc, line 59: can we get rid of this domain check now? since the GetStatusByURI() function handles that, or do we need this separately too?

tsmethurst: Oh yes we can! Nice.

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2405/commits/17cd7ffdc174869e563455c799f3cc9403101e70

internal/processing/common/status.go
tsmethurst: Yeah that's good :)

Did it here, but used UnwrapIfBoost as a name instead to hint that it's a noop if it's not a boost: https://github.com/superseriousbusiness/gotosocial/pull/2405/commits/dc7a4c7378566be80dba17200f653e95d2adc12f

PR #2409, opened 2023-12-03
[feature] request blocking by http headers

NyaaaWhatsUpDoc:
# Description

- adds new `HeaderFilter()` middleware to filter requests by their header content
- adds new database methods to get, store, delete and match against request header filters
- adds new cache types for both allow and block type header filters
- adds new headerfilter subpkg which handles the actual compiling of in-memory request header filters (header keys to value matching regular expressions) + stat collection
- adds new admin API endpoints (with matching swagger docs) to manage allow / block header filters
- adds new "advanced-header-filter-mode" configuration variable which matches the "instance-federation-mode" variable in how it functions re: allow vs. block modes
- add new `/docs/admin/header_filtering_modes.md` documentation laying out the logic of the http request filtering modes

## Checklist


cmd/gotosocial/action/server/server.go
```diff
@@ -321,14 +327,14 @@ var Start action.GTSAction = func(ctx context.Context) error {
 
 	// these should be routed in order;
 	// apply throttling *after* rate limiting
-	authModule.Route(router, clLimit, clThrottle, gzip)
-	clientModule.Route(router, clLimit, clThrottle, gzip)
-	metricsModule.Route(router, clLimit, clThrottle, gzip)
-	fileserverModule.Route(router, fsLimit, fsThrottle)
+	authModule.Route(router, clLimit, clThrottle, gzip, headerFilter)
```
NyaaaWhatsUpDoc, line 330: definitely worth double checking here whether these are the route groups we want to filter.

tsmethurst: Tbh I think it would make sense to filter *everything*. If someone adds some search engine user agent to their blocked headers, they won't want it to be able to hit well known, node info, or the web endpoints either, I would imagine.

docs/admin/request_filtering_modes.md
```diff
@@ -0,0 +1,30 @@
...
+
+    Management via settings panel is TBA. Until then you will need to manage these directly via API endpoints.
+
+## Disabled header filtering mode (default)
+
+When `request-header-filtering-mode` is set to `""`, i.e. an empty string, all request header filtering will be disabled.
+
+## Block filtering mode
+
+When `request-header-filtering-mode` is set to `"block"`, your instance will openly accept HTTP requests, without restriction, with the exception of matching block header filters you have explicitly created via the settings panel.
```
tsmethurst, line 18: I would change the wording a bit here to reassure users that other ACLs will still be applied. So instead of 

> your instance will openly accept HTTP requests, without restriction

I would say something like

> your instance will allow HTTP requests through as normal (pending API token checks, signature checks, etc)

docs/admin/request_filtering_modes.md
```diff
@@ -0,0 +1,30 @@
...
+
+A request in block mode will be accepted if it is EXPLICITLY ALLOWED OR NOT EXPLICITLY BLOCKED.
+
+## Allow filtering mode
+
+When `request-header-filtering-mode` is set to `"allow"`, your instance will only accept HTTP requests for which a matching allow header filter has been explicitly created via the settings panel. All other requests will be refused.
+
+In allow mode, a block header filter can be used to override an existing allow filter, providing an extra level of granularity.
+
+A request in allow mode will only be accepted if it is EXPLICITLY ALLOWED AND NOT EXPLICITLY BLOCKED.
```
tsmethurst, line 30: You might consider adding a warning here that this is a very restrictive mode. Something like:

```
!!! danger
    Allow filtering mode is an extremely restrictive mode that will almost certainly prevent many (legitimate) clients from being able to access your instance, including yourself. You should only enable this mode if you know exactly what you're trying to achieve.
```

example/config.yaml
```diff
@@ -361,6 +361,24 @@ instance-deliver-to-shared-inboxes: true
 # Default: false
 instance-inject-mastodon-version: false
 
+# String. HTTP request header filtering mode to use for this instance.
```
tsmethurst, line 364: Perhaps it's worth renaming this to `advanced-http-request-header-filtering-mode` and putting it in the `advanced` section instead? I know it's a bit more verbose, but it's also a potentially very breaking setting, so it would make sense to put it down there imo.

internal/api/client/admin/headerfilter.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	if _, err := apiutil.NegotiateAccept(c, apiutil.JSONAcceptHeaders...); err != nil {
+		errWithCode := gtserror.NewErrorNotAcceptable(err, err.Error())
+		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
+		return
+	}
+
+	filterID := c.Param("id")
+	if filterID == "" {
```
tsmethurst, line 56: This check (and other checks like it that have been added) could probably use the code in `apiutil` instead.

internal/cache/cache.go
```diff
@@ -18,21 +18,26 @@
...
 )
 
 type Caches struct {
-	// GTS provides access to the collection of gtsmodel object caches.
-	// (used by the database).
+	// GTS provides access to the collection of
+	// gtsmodel object caches. (used by the database).
 	GTS GTSCaches
 
-	// AP provides access to the collection of ActivityPub object caches.
```
tsmethurst, line 30: RIP AP caches

internal/cache/domain/domain.go
```diff
@@ -36,18 +35,15 @@ import (
 //
 // The .Clear() function can be used to invalidate the cache,
 // e.g. when an entry is added / deleted from the database.
-type Cache struct {
-	// atomically updated ptr value to the
-	// current domain cache radix trie.
-	rootptr unsafe.Pointer
-}
+type Cache struct{ rootptr atomic.Pointer[root] }
```
tsmethurst, line 38: Oh, I liked the comment you removed, I thought it was sort of helpful. Was it no longer accurate with these changes?

internal/cache/domain/domain.go
```diff
@@ -60,35 +56,32 @@ func (c *Cache) Matches(domain string, load func() ([]string, error)) (bool, err
 
 		// Allocate new radix trie
 		// node to store matches.
-		root := new(root)
+		ptr = new(root)
```
tsmethurst, line 59: Is this functionally different than what was happening before, or is it just a tidier way of doing the same thing? Not a rhetorical question, I just don't know.

internal/config/config.go
```diff
@@ -85,6 +85,7 @@ type Configuration struct {
 	InstanceDeliverToSharedInboxes bool               `name:"instance-deliver-to-shared-inboxes" usage:"Deliver federated messages to shared inboxes, if they're available."`
 	InstanceInjectMastodonVersion  bool               `name:"instance-inject-mastodon-version" usage:"This injects a Mastodon compatible version in /api/v1/instance to help Mastodon clients that use that version for feature detection"`
 	InstanceLanguages              language.Languages `name:"instance-languages" usage:"BCP47 language tags for the instance. Used to indicate the preferred languages of instance residents (in order from most-preferred to least-preferred)."`
+	RequestHeaderFilterMode        string             `name:"request-header-filter-mode" usage:"Set incoming request header filtering mode."`
```
tsmethurst, line 88: Don't forget to add this as a flag too, else people will wonder why it's not working.

internal/db/bundb/basic.go
```diff
@@ -120,39 +119,6 @@ func (b *basicDB) CreateTable(ctx context.Context, i interface{}) error {
 	return err
 }
 
-func (b *basicDB) CreateAllTables(ctx context.Context) error {
```
tsmethurst, line 123: Ah good, thanks for removing this cruft.

internal/db/bundb/headerfilter.go
```diff
@@ -0,0 +1,207 @@
...
+		Table("header_filter_blocks").
+		Where("? = ?", bun.Ident("id"), id).
+		Exec(ctx); err != nil {
+		return err
+	}
+	h.state.Caches.BlockHeaderFilters.Clear()
+	return nil
+}
+
+// NOTE:
```
tsmethurst, line 176: Blimey, that's a bit cheeky. This is faster than casting it another way I guess?

internal/db/bundb/headerfilter_test.go
```diff
@@ -0,0 +1,108 @@
+package bundb_test
```
tsmethurst, line 1: license header

internal/db/bundb/migrations/20231212144715_add_header_filters.go
```diff
@@ -0,0 +1,55 @@
...
+	up := func(ctx context.Context, db *bun.DB) error {
+		for _, model := range []any{
+			&gtsmodel.HeaderFilterAllow{},
+			&gtsmodel.HeaderFilterBlock{},
+		} {
+			_, err := db.NewCreateTable().
+				IfNotExists().
+				Model(model).
+				Exec(ctx)
+			if err != nil && !(strings.Contains(err.Error(), "already exists") || strings.Contains(err.Error(), "duplicate column name") || strings.Contains(err.Error(), "SQLSTATE 42701")) {
```
tsmethurst, line 38: I think you don't need the specific error checking here, since this is table creation (not column creation) and you already have IfNotExists

internal/db/headerfilter.go
```diff
@@ -0,0 +1,73 @@
...
+package db
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+)
+
+type HeaderFilter interface {
```
tsmethurst, line 27: This looks good but man, we should probably deinterface the DB at some point huh?

internal/gtsmodel/headerfilter.go
```diff
@@ -0,0 +1,54 @@
...
+
+// HeaderFilterBlock represents a block HTTP header filter in the database.
+type HeaderFilterBlock struct{ HeaderFilter }
+
+// HeaderFilter represents an HTTP request filter in
+// the database, with a header to match against, value
+// matching regex, and details about its creation.
+type HeaderFilter struct {
+	ID        string    `bun:"type:CHAR(26),pk,nullzero,notnull,unique"`                    // ID of this item in the database
+	Header    string    `bun:",nullzero,notnull"`                                           // Request header this filter pertains to
```
tsmethurst, line 48: You could probably put a unique constraint on `header,regex` here; you'll not want to be able to store multiple identical regexes for the same header, I'd say.

internal/headerfilter/filter.go
```diff
@@ -0,0 +1,161 @@
...
+	// match count.
+	n atomic.Uint64
+}
+
+// Append will add new header filter expression under given header key.
+func (fs *Filters) Append(key string, expr string) error {
+	var filter *headerfilter
+
+	// Get key in canonical mime header format.
+	key = textproto.CanonicalMIMEHeaderKey(key)
```
tsmethurst, line 66: Aha, I was about to ask about this very thing, and here it is :)

internal/headerfilter/filter.go
```diff
@@ -0,0 +1,161 @@
...
+		if (*fs)[i].key == key {
+			filter = &((*fs)[i])
+			break
+		}
+	}
+
+	if filter == nil {
+		// No existing filter found, create new.
+		(*fs) = append((*fs), headerfilter{})
+		filter = &((*fs)[len((*fs))-1])
```
tsmethurst, line 80: This is a bit dense. Probably worth breaking it across multiple lines for readability and/or commenting what's happening.

internal/headerfilter/filter.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// regexpr wraps a regular expression
+// with an atomically updated uint64 in
+// order to count no. positive matches.
+type regexpr struct {
+	*regexp.Regexp
+
+	// match count.
+	n atomic.Uint64
```
tsmethurst, line 58: This can probably be dropped, if we're going to use prometheus/otel metrics for this then we can use a [counter](https://opentelemetry.io/docs/instrumentation/go/manual/#using-counters) for this instead when we get around to it.

internal/headerfilter/filter.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// RegularMatch returns whether any values in http header
+// matches any of the receiving filter regular expressions.
+func (fs Filters) RegularMatch(h http.Header) (bool, error) {
+	for _, filter := range fs {
+		for _, value := range h[filter.key] {
+			// Don't perform match on large values
+			// to mitigate denial of service attacks.
+			if len(value) > MaxHeaderValue {
```
tsmethurst, line 105: Nice one :)

internal/headerfilter/filter.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+	}
+	return false, nil
+}
+
+// Stats compiles each of the filters and their match counts into
+// a readable set of HeaderKey:{ValueExpr: MatchCount} statistics.
+// TODO: may be worth updating this to be more prometheus readable
+func (fs Filters) Stats() map[string]map[string]uint64 {
```
tsmethurst, line 147: Can probably also be replaced with some suitable otel type :)

internal/middleware/headerfilter.go
```diff
@@ -0,0 +1,148 @@
...
+
+	case config.RequestHeaderFilterModeBlock:
+		return headerFilterBlockMode(state)
+
+	default:
+		panic("unrecognized filter mode: " + mode)
+	}
+}
+
+func headerFilterAllowMode(state *state.State) func(c *gin.Context) {
```
tsmethurst, line 46: very tidy and readable!

internal/middleware/headerfilter_test.go
```diff
@@ -0,0 +1,299 @@
...
+				{"User-Agent", ".*Firefox.*"},
+			},
+			block: []filter{},
+			input: http.Header{
+				"User-Agent": []string{"Chromium v169.42; Extra Tracking Info"},
+			},
+			expect: false,
+		},
+		{
+			// Allow mode with too long header value expecting 403 Forbidden.
```
tsmethurst, line 75: Is it worth returning [Header Too Large](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/431) in these cases?

internal/processing/admin/headerfilter.go
```diff
@@ -0,0 +1,214 @@
...
+	if errWithCode := validateHeaderFilter(
+		request.Header,
+		request.Regex,
+	); errWithCode != nil {
+		return nil, errWithCode
+	}
+
+	// Create new database model with ID.
+	var filter gtsmodel.HeaderFilter
+	filter.CreatedAt = time.Now()
```
tsmethurst, line 156: We don't need to set created_at manually here since it's set by the db anyway. Unless there's some other reason for setting it, in which case ignore this :)

internal/processing/admin/headerfilter.go
```diff
@@ -0,0 +1,214 @@
...
+		Regex:     filter.Regex,
+		CreatedBy: filter.AuthorID,
+		CreatedAt: util.FormatISO8601(filter.CreatedAt),
+	}
+}
+
+// validateHeaderFilter validates incoming filter's header key, and regular expression.
+func validateHeaderFilter(header, regex string) gtserror.WithCode {
+	// Canonicalize the mime header key and check validity.
+	header = textproto.CanonicalMIMEHeaderKey(header)
```
tsmethurst, line 201: Aha great! Disregard my comment above!

tsmethurst (commented):
It's looking great! Coming along nicely!

example/config.yaml
NyaaaWhatsUpDoc: Yeah agreed, good idea 👍. the name of it will look a bit more sensible when we do the config refactor that allows us to have more sub sections in the config anywho :p

internal/cache/cache.go
NyaaaWhatsUpDoc: RIP AP caches that never even got to see the light of day

internal/db/bundb/headerfilter.go
NyaaaWhatsUpDoc: there isn't another way of casting it! since it's using structs. and it's extremely marginally faster, interfaces usually require extra allocations. it wasn't done out of speed, just I preferred being able to use solid struct types here 😊

internal/db/headerfilter.go
NyaaaWhatsUpDoc: yeah there are a lot of them ....

internal/gtsmodel/headerfilter.go
NyaaaWhatsUpDoc: good idea!

internal/headerfilter/filter.go
NyaaaWhatsUpDoc: oh I didn't canonicalize them on the way in so that the returned value shown in the settings API eventually is the original typed value, but when it comes to usage we'll use the canonical format

NyaaaWhatsUpDoc: or is that maybe not so important?I'm not fussed either way, it's a very simple change.

internal/headerfilter/filter.go
NyaaaWhatsUpDoc: oh I was just going to return this as a counter value, as I'm pretty sure that's what it uses under the hood is atomic ints etc. the plan is to make that stats function export Prometheus compatible data at some point. also this way I have some stat collection enabled required for one of the milestone bulletpoints

internal/headerfilter/filter.go
NyaaaWhatsUpDoc: see: above

internal/middleware/headerfilter_test.go
NyaaaWhatsUpDoc: honestly the only case there'd be a header this large is likely some request fuckery. this way we reveal less internal stuff and they have no idea why they're blocked :p

internal/headerfilter/filter.go
tsmethurst: The way counters work in OTel is I think to just call Add(i) on the counter when a value must be incremented, and let OTel handle the rest; you don't actually need to keep track in the application of the counter's current value afaik. That's one thing that makes OTel so nice :)

internal/middleware/headerfilter_test.go
tsmethurst: Ah that's a fair point. Alrighty!

internal/cache/domain/domain.go
NyaaaWhatsUpDoc: oh no! just wasn't thinking. will add it back :)

internal/cache/domain/domain.go
NyaaaWhatsUpDoc: it's just using the new atomic.Pointer generic type now that it's available :) so (arguably) somewhat tidier

internal/headerfilter/filter.go
NyaaaWhatsUpDoc: gotcha. maybe in the tracing / metric package i'll wrap it so it works with our enabling / disabling mechanism (otherwise this will make it a hard dependency in all cases)

docs/admin/request_filtering_modes.md
NyaaaWhatsUpDoc: Have added it here: 87393309c

docs/admin/request_filtering_modes.md
NyaaaWhatsUpDoc: Have added it here: e78369e42

internal/cache/domain/domain.go
NyaaaWhatsUpDoc: Have added here: 12f66c169

internal/db/bundb/migrations/20231212144715_add_header_filters.go
NyaaaWhatsUpDoc: Have added here: 036c6ccd2

internal/gtsmodel/headerfilter.go
NyaaaWhatsUpDoc: Have added here: 78bfd0842

cmd/gotosocial/action/server/server.go
NyaaaWhatsUpDoc: Have changed this here: a5a80ae6f

NyaaaWhatsUpDoc: Have changed this here: a5a80ae6f

internal/processing/admin/headerfilter.go
NyaaaWhatsUpDoc: Have added here: 057b68a40

internal/config/config.go
NyaaaWhatsUpDoc: have added that here: fe86527afb5ffdbd6d6362f810e91395c528eacd

example/config.yaml
NyaaaWhatsUpDoc: Have renamed it here: d03efa043013935d80564ad57c6bc5b334b9c210

internal/config/config.go
NyaaaWhatsUpDoc: have added that here: fe86527afb5ffdbd6d6362f810e91395c528eacd

internal/headerfilter/filter.go
NyaaaWhatsUpDoc: so for now i've kept-in an initial statistics tracking implementation, but i have rejigged the header matching so the matched header key and regular expression are returned which should make it easier to instrument with otel in the future. see: eaa29f9f4

PR #2418, opened 2023-12-05
[feature] Push status edit messages into open streams

### LINKED ISSUE #2417: [feature] Push status edit messages to streams

If you have a client using the streaming API connected, and a status in your timeline is edited, you don't currently get a streamed message telling you this, so the status doesn't get updated.

Add support for the "status.update" message type on the streaming API.

n/a

### Additional context.

_No response_

Sentynel:
# Description

When a status update message comes in (i.e. a status is edited), push the appropriate status.update message to any connected streams that would have received the original status.

Slightly WIP still but does work.

closes #2417 

## Checklist


tsmethurst:
Nice one; I think we will save this for after 0.13.0 so we can try it out on main and tweak it as necessary :)

Sentynel:
Yep, wasn't expecting this to make it into this release - I've tested that it sends messages that look like the Mastodon ones, but I haven't tested it with different clients or anything.

Outstanding questions I think are:
1) I've basically duplicated a chunk of the new status -> timeline processing code in surfacetimeline.go for the update route instead, but maybe it's better to add an "edit" parameter to the existing code and alter the flow as appropriate instead? It's basically the same code but without the actual inserts and the notification generation, which could be stuck behind an "is_new_status" guard instead. Slight reduction of copy-and-paste code, slightly harder to follow the logic.
2) Speaking of notifications, I'm not notifying on edited statuses. This feels like the right move in general, but does mean that if somebody edits a mention into a post we don't notify. I think that's a reasonable trade-off? I can't think of a clean way of working out if a mention is new in the edited version in order to generate a new notification (you'd have to add a bunch of accounting to keep track of the notifications you've sent for each status, which sucks, or diff the edit history, which sucks less but still sounds annoying).

NyaaaWhatsUpDoc:
> Yep, wasn't expecting this to make it into this release - I've tested that it sends messages that look like the Mastodon ones, but I haven't tested it with different clients or anything.
> 
> Outstanding questions I think are:
> 
>     1. I've basically duplicated a chunk of the new status -> timeline processing code in surfacetimeline.go for the update route instead, but maybe it's better to add an "edit" parameter to the existing code and alter the flow as appropriate instead? It's basically the same code but without the actual inserts and the notification generation, which could be stuck behind an "is_new_status" guard instead. Slight reduction of copy-and-paste code, slightly harder to follow the logic.
> 
>     2. Speaking of notifications, I'm not notifying on edited statuses. This feels like the right move in general, but does mean that if somebody edits a mention into a post we don't notify. I think that's a reasonable trade-off? I can't think of a clean way of working out if a mention is new in the edited version in order to generate a new notification (you'd have to add a bunch of accounting to keep track of the notifications you've sent for each status, which sucks, or diff the edit history, which sucks less but still sounds annoying).

great work here! nice one :). regarding 1) honestly i quite like how it is currently, as you say making a singular function share logic for both makes it a bit more complicated now, and in the future if there's further diversion it could start to get a bit hairy :P. and for 2), as you say that's a bit of an awkward situation adding in a mention to an edited post, i think that's something we can safely leave for now as it's a bit of a niche situation. also, if anyone further replies after that the person will still get notified for any new mentions of themselves. we can probably safe that functionality until we actually support full edited statuses logic including revision history.

internal/processing/stream/statusupdate_test.go
```diff
@@ -0,0 +1,139 @@
...
+	StreamTestSuite
+}
+
+func (suite *StatusUpdateTestSuite) TestStreamNotification() {
+	account := suite.testAccounts["local_account_1"]
+
+	openStream, errWithCode := suite.streamProcessor.Open(context.Background(), account, "user")
+	suite.NoError(errWithCode)
+
+	//followAccount := suite.testAccounts["remote_account_1"]
```
NyaaaWhatsUpDoc, line 41: did you mean to remove these commented-out lines? :p

NyaaaWhatsUpDoc:
only one comment on the added test, otherwise looks all good to me @Sentynel :ok_hand:

NyaaaWhatsUpDoc (approved):

tsmethurst:
Thank you!

Issue #2425, opened 2023-12-08
[feature] Ensure we analyze SQLite tables regularly


daenney:
### Describe the bug with a clear and concise description of what the bug is.

We currently don't seem to have code that runs `PRAGMA optimize` for SQlite on a regular basis. Without it, SQLite internal statistics aren't updated and indexes might go unused.

> Applications with long-lived databases that use complex queries should consider running the following commands just prior to closing each database connection:
> 
> ```
> PRAGMA analysis_limit=400;
> PRAGMA optimize;
> ```
> 
> The [optimize pragma](https://sqlite.org/pragma.html#pragma_optimize) is usually a no-op but it will occasionally run ANALYZE if it seems like doing so will be useful to the query planner. The [analysis_limit pragma](https://sqlite.org/pragma.html#pragma_analysis_limit) limits the scope of any ANALYZE command that the [optimize pragma](https://sqlite.org/pragma.html#pragma_optimize) runs so that it does not consume too many CPU cycles. The constant "400" can be adjusted as needed. Values between 100 and 1000 work well for most applications.
> 
> https://sqlite.org/lang_analyze.html

daenney:
The recommendation from SQLite devs is to run `PRAGMA optimize` regularly on the connection, since that lets SQLite use information from the queries that were run on a connection to decide if an `ANALYZE` should run and for which tables. This is what ends up making `PRAGMA optimize` a no-op in many cases, avoiding unnecessary work scanning tables.

However, there's not a very easy way to get to the underlying connections, we generally interface through the connection pool. The alternative would be for us to run `ANALYZE` on a schedule instead, and perhaps limit that to the few tables we know experience a lot of churn (hello there `statuses`). That'll probably run `ANALYZE` more often than necessary, but since `ANALYZE` is the thing that takes `analysis_limit` into account, it's still possible to do so without risking too much impact on the DB.

NyaaaWhatsUpDoc:
I'm handling this by (hopefully) adding an upstream connection close hook: https://gitlab.com/cznic/sqlite/-/merge_requests/64

PR #2427, opened 2023-12-08
[chore] Run `ANALYZE` for SQLite after latest migrations

tsmethurst:
See https://github.com/superseriousbusiness/gotosocial/issues/2425 for more details.

We may want to run VACUUM as well, but since this can take quite a long time (10-15 minutes-ish), it might be better not to, but instead tell folks to run it manually. I'm not entirely sure yet.

### DIFF
--- a/internal/db/bundb/migrations/20231208152242_sqlite_analyze.go
+++ b/internal/db/bundb/migrations/20231208152242_sqlite_analyze.go
@@ -1,0 +1,51 @@
+// GoToSocial
+// Copyright (C) GoToSocial Authors admin@gotosocial.org
+// SPDX-License-Identifier: AGPL-3.0-or-later
+//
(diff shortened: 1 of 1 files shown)


daenney:
I'm wondering if instead of doing this as a migration itself, we can see if migrations happened during startup and then trigger an analyze? That way we don't have to remember to do this after each migration.

daenney:
But for the case of the 0.13 cycle I think this is fine. We can just merge this.

tsmethurst:
Ack, yeah that makes sense. Let's do it properly (and perhaps include autovacuum as well) once 0.13.0 is out.

daenney:
We can hook it here I think: https://github.com/superseriousbusiness/gotosocial/blob/455064fec743cc42b25097a489904893ad92f0e0/internal/db/bundb/bundb.go#L106

PR #2428, opened 2023-12-08
[feature] Run ANALYZE after migrations on SQLite

daenney:
# Description

This ensures that at the end of migrations, we run ANALYZE if we're using SQLite. This should be relatively quick and guarantees that the table and index statistics have been updated. This helps to ensure the query planner makes better choices when it comes to picking which indexes are used when running queries.

## Checklist

### DIFF
--- a/internal/db/bundb/bundb.go
+++ b/internal/db/bundb/bundb.go
@@ -44,6 +44,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/state"
 	"github.com/superseriousbusiness/gotosocial/internal/tracing"
 	"github.com/uptrace/bun"
+	"github.com/uptrace/bun/dialect"
 	"github.com/uptrace/bun/dialect/pgdialect"
 	"github.com/uptrace/bun/dialect/sqlitedialect"
 	"github.com/uptrace/bun/migrate"
@@ -113,6 +114,14 @@ func doMigration(ctx context.Context, db *bun.DB) error {
 	}
 
 	log.Infof(ctx, "MIGRATED DATABASE TO %s", group)
+
+	if db.Dialect().Name() == dialect.SQLite {
+		log.Info(ctx, "running ANALYZE to update table and index statistics")
+		_, err := db.Exec("ANALYZE")
+		if err != nil {
+			log.Warnf(ctx, "ANALYZE failed, query planner may make poor life choices: %s", err)
+		}
+	}
 	return nil
 }
 
--- a/internal/db/bundb/bundb.go
+++ b/internal/db/bundb/bundb.go
@@ -117,7 +117,7 @@ func doMigration(ctx context.Context, db *bun.DB) error {
 
 	if db.Dialect().Name() == dialect.SQLite {
 		log.Info(ctx, "running ANALYZE to update table and index statistics")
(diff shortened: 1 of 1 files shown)


internal/db/bundb/bundb.go
```diff
@@ -113,6 +114,14 @@ func doMigration(ctx context.Context, db *bun.DB) error {
 	}
 
 	log.Infof(ctx, "MIGRATED DATABASE TO %s", group)
+
```
daenney, line 117: This was the most logical place I could think of hooking it in. Happy to move it elsewhere.

internal/db/bundb/bundb.go
```diff
@@ -113,6 +114,14 @@ func doMigration(ctx context.Context, db *bun.DB) error {
...
+	if db.Dialect().Name() == dialect.SQLite {
+		log.Info(ctx, "running ANALYZE to update table and index statistics")
+		_, err := db.Exec("ANALYZE")
+		if err != nil {
+			log.Warnf(ctx, "ANALYZE failed, query planner may make poor life choices: %s", err)
```
daenney, line 122: I'm not returning the error here since I don't want this to fail just because `ANALYZE` might've had an issue. We'll still work, just perhaps less optimally.

internal/db/bundb/bundb.go
```diff
@@ -113,6 +114,14 @@ func doMigration(ctx context.Context, db *bun.DB) error {
 	}
 
 	log.Infof(ctx, "MIGRATED DATABASE TO %s", group)
+
+	if db.Dialect().Name() == dialect.SQLite {
```
daenney, line 118: I think right now this means it always runs, even if there were no migrations. But in that case it should be very cheap either way. We can guard this with an additional `err != nil && !strings.Contains(err.Error(), "no migrations")` like we did after `migrator.Migrate()`.

daenney: Actually no, since we have the check for if migrations were run earlier and we return out of the function there. So this only triggers if a migration happened.

internal/db/bundb/bundb.go
```diff
@@ -113,6 +114,14 @@ func doMigration(ctx context.Context, db *bun.DB) error {
...
+		log.Info(ctx, "running ANALYZE to update table and index statistics")
+		_, err := db.ExecContext(ctx, "ANALYZE")
```
daenney, line 120: I'm not calling `PRAGMA analysis_limit=$num` here since this runs after migrations, and not during normal runtime. When we add it as a timer functionality we should make sure to set the analysis limit to limit the impact.

internal/db/bundb/bundb.go
daenney: This is something an admin can always correct themselves by opening the DB with the sqlite3 CLI and calling `ANALYZE` themselves.

NyaaaWhatsUpDoc: i enjoy the choice of log text here :grin:

NyaaaWhatsUpDoc (approved):

internal/db/bundb/bundb.go
tsmethurst: I think this is fine :)

PR #2432, opened 2023-12-09
Use the issue config.yaml file

filipesmedeiros:
# Description

Use the [config.yaml](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser) file that Github has, instead of having a separate custom issue template.

It will look like this:
Screenshot 2023-12-09 at 12 28 09

Obviously, this is a non-urgent change :) I just noticed it when opening issues!

## Checklist


.github/ISSUE_TEMPLATE/config.yaml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: true
```
daenney, line 1: This seems like it's `true` by default, so I don't think we need to have it set. I can see it already if I try to create a new issue now.

filipesmedeiros: Ahaa so the whole file is unnecessary in our case? Do you wanna redirect people to a Matrix room or something?

daenney: We already mention in the README that you should use Matrix for questions etc. I think the reason the template was added is because it gives you a much more obvious big green button in the same list, instead of the much de-emphasised "looks like a normal link" that you get without it.

filipesmedeiros: What do you think of just at least changing the description of the custom issue?

tsmethurst: Is there much point? I think we can probably just close this PR. Thanks for poking at it though: it's good to know that config.yml is an option if we ever need it.

daenney: Actually, I think that would make sense. Right now, it renders like this:
Screenshot from 2023-12-10 15-47-40

That's a bit odd. It's probably nice to change that to just be "Other type of issue" and set a very simple description.

daenney: Also that one's in Markdown format whereas the others are `.yaml`, and it would be nice to have those all be the same I think.

filipesmedeiros: I guess this is an admin issue mostly, so maybe you can take over the PR as you see fit? With a description you like etc. I can also close it if you want

daenney: Sure, I'll close this for now then. Thanks for the effort!

PR #2433, opened 2023-12-09
[bugfix] Fix web media not showing as sensitive

tsmethurst:
# Description

>

Web statuses weren't having their media marked as sensitive correctly, because I (it's my bug) was trying to modify a non-pointer attachment inside a slice. This PR fixes the bug by changing status.MediaAttachments to a slice of pointers.

Probably could have fixed this by just replacing items in the array with different versions, but that seems more expensive than just setting a field value on a ptr.

## Checklist


NyaaaWhatsUpDoc:
> Web statuses weren't having their media marked as sensitive correctly, because I (it's my bug) was trying to modify a non-pointer attachment inside a slice. This PR fixes the bug by changing status.MediaAttachments to a slice of pointers.
> 
> Probably could have fixed this by just replacing items in the array with different versions, but that seems more expensive than just setting a field value on a ptr.

in the future you can do any of the following to modify slice values:
```golang
for i := range slice {
    slice[i].Member = newMemberValue
}

for i := range slice {
    value := &(slice[i])
    value.Member = newMemberValue
}
```

tsmethurst:
Ahh thank you :)

Issue #2411, opened 2023-12-04
[bug] `@` not handled in autocompletion queries


mirabilos:
### Describe the bug with a clear and concise description of what the bug is.

When beginning to type `@meow` for autocompletion in Semaphore, I get this in the log:

```
func=bundb.queryHook.AfterQuery level=WARN duration="1m19.636948962s" query="SELECT \"account\".\"id\" FROM \"accounts\" AS \"account\" WHERE ((\"account\".\"domain\" IS NULL) OR (\"account\".\"domain\" != \"account\".\"username\")) AND (\"account\".\"id\" < 'ZZZZZZZZZZZZZZZZZZZZZZZZZZ') AND ((SELECT LOWER(CONCAT(\"account\".\"username\", COALESCE(\"account\".\"display_name\", ''))) AS \"account_text\") LIKE '%@meow%' ESCAPE '\\') ORDER BY \"account\".\"id\" DESC LIMIT 4" requestID=8gmewcmc0400166f0vg0 msg="SLOW DATABASE QUERY"
```

However, the string searched for (the `LOWER(…)`) does not even begin with a `@`.

Furthermore, this makes it good as impossible to search for “common” names as the domain is not contained in the string *and* the `LIMIT` is very low.

May I at least suggest doing:
- strip the leading `@` (but _only_ __one__ leading one!) from the search text
  - so we get `LIKE '%meow%'` but people can still search for `@` explicitly by doubling it or putting one later
- change the `LOWER(…)` sub-clause to something like `LOWER(CONCAT("account"."username", '@', COALESCE("account"."domain", 'gts.superseriousbusiness.example.org'), COALESCE("account"."display_name", '')))`
  - where `gts.superseriousbusiness.example.org` is the __account-domain__ from the configuration, of course (since GtS knows that *and* it will never change, it can put this as hardcoded string into the query, no extra hardparse penalty)
  - so we get 'lowerusername@lowerdomainlowerdisplayname' as text compared with

Additionally, perhaps raise the `LIMIT` a bit?

Thanks in advance!

### What's your GoToSocial Version?

0.13.0-rc1 git-0e2c342

### GoToSocial Arch

amd64 binary

tsmethurst:
GtS not stripping the leading '@' looks indeed like a regression, I believe we used to do that. I'll check it out.

[...]

PR #2435, opened 2023-12-10
[bugfix] Narrow search scope for accounts starting with '@'; don't `LOWER` SQLite text searches

tsmethurst:
# Description

>

This pull request tightens up our search logic a little:

- Don't use LOWER for SQLite `LIKE` searches. SQLite LIKE search is case-insensitive anyway, and LOWER is expensive. Postgres still uses LOWER.
- Passing a query string starting with `@` into the accounts search DB function will make it only look for accounts that *start with* the text after the `@`. This is a shitload faster than doing a full wildcard search, and is more suitable for an autocomplete-style search anyway. The leading `@` is stripped properly when doing a search like this.
- Searches that start with `@` will return early after looking for accounts, and not proceed to full text search (unless the search query parameters are used to indicate explicitly that the search is not looking for accounts).

closes https://github.com/superseriousbusiness/gotosocial/issues/2411

## Checklist

### DIFF
--- a/internal/db/bundb/search.go
+++ b/internal/db/bundb/search.go
@@ -67,7 +67,7 @@ type searchDB struct {
 //	WHERE (("account"."domain" IS NULL) OR ("account"."domain" != "account"."username"))
 //	AND ("account"."id" < 'ZZZZZZZZZZZZZZZZZZZZZZZZZZ')
 //	AND ("account"."id" IN (SELECT "target_account_id" FROM "follows" WHERE ("account_id" = '016T5Q3SQKBT337DAKVSKNXXW1')))
-//	AND ((SELECT LOWER("account"."username" || COALESCE("account"."display_name", '') || COALESCE("account"."note", '')) AS "account_text") LIKE '%turtle%' ESCAPE '\')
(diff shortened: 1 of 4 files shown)


internal/db/bundb/search.go
```diff
@@ -128,12 +128,20 @@ func (s *searchDB) SearchForAccounts(
...
-	accountTextSubq := s.accountText(following)
-
-	// Search using LIKE for matches of query
-	// string within accountText subquery.
-	q = whereLike(q, accountTextSubq, query)
+	if strings.HasPrefix(query, "@") {
+		// Query looks a bit like a username.
+		// Normalize it and just look for
+		// usernames that start with query.
+		query = query[1:]
```
mirabilos, line 135: this is an improvement alright, but it *still* won’t let me search for `@user@domain` (when the `@user` search returns the four wrong results because many people are called `user`); should I open up a separate feature request for this?

tsmethurst: Heya! Indeed that's right. You can autocomplete for `@user[......]` but you can't currently autocomplete for `@user@domain[.......]`. A feature request issue is a good idea :)

mirabilos: Filed as #2449

PR #2444, opened 2023-12-11
[bugfix] poll vote count fixes

NyaaaWhatsUpDoc:
# Description

- only sets the `voters_count` frontend field when the poll is multiple choice (changed to ptr to accomodate this)
- slightly more robust ActivityStreams poll data extraction

## Checklist

### DIFF
--- a/internal/api/model/poll.go
+++ b/internal/api/model/poll.go
@@ -28,7 +28,7 @@ type Poll struct {
 	ID string `json:"id"`
 
 	// When the poll ends. (ISO 8601 Datetime).
-	ExpiresAt *string `json:"expires_at,omitempty"`
+	ExpiresAt *string `json:"expires_at"`
 
 	// Is the poll currently expired?
 	Expired bool `json:"expired"`
@@ -40,7 +40,7 @@ type Poll struct {
 	VotesCount int `json:"votes_count"`
 
 	// How many unique accounts have voted on a multiple-choice poll.
-	VotersCount *int `json:"voters_count,omitempty"`
+	VotersCount *int `json:"voters_count"`
 
 	// When called with a user token, has the authorized user voted?
 	//
@@ -68,7 +68,7 @@ type PollOption struct {
 	Title string `json:"title"`
 
 	// The number of received votes for this option.
-	VotesCount *int `json:"votes_count,omitempty"`
+	VotesCount *int `json:"votes_count"`
 }
 
 // PollRequest models a request to create a poll.
--- a/internal/api/model/poll.go
+++ b/internal/api/model/poll.go
@@ -40,7 +40,7 @@ type Poll struct {
 	VotesCount int `json:"votes_count"`
 
 	// How many unique accounts have voted on a multiple-choice poll.
-	VotersCount int `json:"voters_count"`
+	VotersCount int `json:"voters_count,omitempty"`
 
 	// When called with a user token, has the authorized user voted?
 	//
--- a/internal/api/model/poll.go
+++ b/internal/api/model/poll.go
@@ -28,7 +28,7 @@ type Poll struct {
 	ID string `json:"id"`
 
 	// When the poll ends. (ISO 8601 Datetime).
-	ExpiresAt string `json:"expires_at"`
+	ExpiresAt *string `json:"expires_at,omitempty"`
 
 	// Is the poll currently expired?
 	Expired bool `json:"expired"`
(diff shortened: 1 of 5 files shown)


internal/typeutils/internaltofrontend.go
```diff
@@ -1498,21 +1497,28 @@ func (c *Converter) PollToAPIPoll(ctx context.Context, requester *gtsmodel.Accou
...
-		emojis, err = c.convertEmojisToAPIEmojis(ctx, pStatus.Emojis, pStatus.EmojiIDs)
-		if err != nil {
-			// Fall back to empty slice.
-			log.Errorf(ctx, "error converting emojis from parent status: %v", err)
-			emojis = make([]apimodel.Emoji, 0)
-		}
+	var err error
+
+	// Try to inherit emojis from parent status.
+	emojis, err = c.convertEmojisToAPIEmojis(ctx,
```
NyaaaWhatsUpDoc, line 1515: considering our call to populate earlier, and that the status should always be set, i just removed this lil check

internal/api/model/poll.go
```diff
@@ -40,7 +40,7 @@ type Poll struct {
 	VotesCount int `json:"votes_count"`
 
 	// How many unique accounts have voted on a multiple-choice poll.
-	VotersCount int `json:"voters_count"`
+	VotersCount *int `json:"voters_count,omitempty"`
```
tsmethurst, line 43: Just to check, does this line up with the Masto API? Does that also omit voters_count when it's nil?

tsmethurst: Right, I checked and it shouldn't have `omitEmpty`. voters_count should be null for single-choice polls, and set for multiple-choice polls: https://docs.joinmastodon.org/entities/Poll/#voters_count

tsmethurst: Also, when results of a poll are not yet published, it looks like `votes_count` on each option (PollOption below) should be null: https://docs.joinmastodon.org/entities/Poll/#option-votes_count

Might as well fix that in this PR too :)

tsmethurst: The templates will need to be changed slightly if you do that, so that 0 values aren't understood as being nil. Here's a patch!

```patch
diff --git a/internal/router/template.go b/internal/router/template.go
index 804f532b..a3516a36 100644
--- a/internal/router/template.go
+++ b/internal/router/template.go
@@ -22,6 +22,7 @@ import (
        "html/template"
        "os"
        "path/filepath"
+       "reflect"
        "strings"
        "time"
 
@@ -185,5 +186,18 @@ func LoadTemplateFunctions(engine *gin.Engine) {
                "emojify":          emojify,
                "acctInstance":     acctInstance,
                "increment":        increment,
+               "isNil":            isNil,
        })
 }
+
+// isNil returns true if given
+// value is a nil pointer.
+func isNil(value any) bool {
+       vOf := reflect.ValueOf(value)
+
+       if vOf.Kind() != reflect.Pointer {
+               return false
+       }
+
+       return vOf.IsNil()
+}
diff --git a/web/template/status_poll.tmpl b/web/template/status_poll.tmpl
index f6acecfa..d2604628 100644
--- a/web/template/status_poll.tmpl
+++ b/web/template/status_poll.tmpl
@@ -47,6 +47,9 @@
                                <meter aria-hidden="true" id="poll-{{- $pollOption.PollID -}}-option-{{- increment $index -}}" min="0" max="100" value="{{- $pollOption.VoteShare -}}">{{- $pollOption.VoteShare -}}&#37;</meter>
                                <div class="sr-only">Option {{ increment $index }}:&nbsp;<span lang="{{ .LanguageTag.TagStr }}">{{ emojify .Emojis (noescape $pollOption.Title) -}}</span></div>
                                <div class="poll-vote-summary">
+                                       {{- if isNil $pollOption.VotesCount }}
+                                       Results not yet published.
+                                       {{- else -}}
                                        <span class="poll-vote-share">{{- $pollOption.VoteShareStr -}}&#37;</span>
                                        <span class="poll-vote-count">
                                                {{- if eq $pollOption.VotesCount 1 -}}
@@ -55,6 +58,7 @@
                                                        {{- $pollOption.VotesCount }} votes
                                                {{- end -}}
                                        </span>
+                                       {{- end -}}
                                </div>
                        </li>
                {{- end }}

```

NyaaaWhatsUpDoc: thank you for this :D

NyaaaWhatsUpDoc: [b304084](https://github.com/superseriousbusiness/gotosocial/pull/2444/commits/b304084df82d8ce5c3aa78f590b8e173bd64111a)

tsmethurst: Looks good, but you shouldn't have the `omitempty` tags, since we actually want to serialize null.

NyaaaWhatsUpDoc: *shakes fist* mastodonnnnn!!

PR #2453, opened 2023-12-14
[bugfix] use a much shorter refresh limit for statuses with polls

NyaaaWhatsUpDoc:
# Description

Adds an upToDate flag to the common processor target status fetching functions, which specifies whether to force a refresh of the given status. We use this to then force a refresh within the poll handler.

The dereferencer has also been updated to enforce a 5 minute cooldown on the force flag, so spamming the refresh button can't accidentally DOS another server, (or us acciddentally passing a force refresh flag elsewhere...)

This also updates the common processor functions to refresh the status synchronously instead of async when the flag is provided, which should help with issues of content "pop-in".

## Checklist

### DIFF
--- a/internal/db/bundb/migrations/20231215115920_add_status_poll_index.go
+++ b/internal/db/bundb/migrations/20231215115920_add_status_poll_index.go
@@ -33,7 +33,6 @@ func init() {
 			}
 
 			for _, spec := range []spec{
-				// Rename the admin actions indexes.
 				{
 					index:   "statuses_poll_id_idx",
 					table:   "statuses",
--- a/internal/db/bundb/migrations/20231215115920_add_status_poll_index.go
+++ b/internal/db/bundb/migrations/20231215115920_add_status_poll_index.go
@@ -1,0 +1,67 @@
+// GoToSocial
+// Copyright (C) GoToSocial Authors admin@gotosocial.org
+// SPDX-License-Identifier: AGPL-3.0-or-later
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package migrations
+
+import (
+	"context"
+
+	"github.com/uptrace/bun"
+)
+
+func init() {
+	up := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			type spec struct {
+				index   string
+				table   string
+				columns []string
+			}
+
+			for _, spec := range []spec{
+				// Rename the admin actions indexes.
+				{
+					index:   "statuses_poll_id_idx",
+					table:   "statuses",
+					columns: []string{"poll_id"},
+				},
+			} {
+				if _, err := tx.
+					NewCreateIndex().
+					Table(spec.table).
+					Index(spec.index).
+					Column(spec.columns...).
+					IfNotExists().
+					Exec(ctx); err != nil {
+					return err
+				}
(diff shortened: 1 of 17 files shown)


internal/db/bundb/status.go
```diff
@@ -87,6 +87,17 @@ func (s *statusDB) GetStatusByURL(ctx context.Context, url string) (*gtsmodel.St
 	)
 }
 
+func (s *statusDB) GetStatusByPollID(ctx context.Context, pollID string) (*gtsmodel.Status, error) {
+	return s.getStatus(
+		ctx,
+		"PollID",
+		func(status *gtsmodel.Status) error {
+			return s.db.NewSelect().Model(status).Where("? = ?", bun.Ident("status.poll_id"), pollID).Scan(ctx)
```
NyaaaWhatsUpDoc, line 95: is it worth adding an index on the poll_id column now do we think? this is only used in one place, and the cache should help back it up given we added a cache index for it

daenney: Probably. It's a high cardinality field so it ought to help. Cache is good but it's nice to keep the queries filling the cache fast too.

daenney: Unless there's a unique on that field, in which case the database already maintains an index.

internal/federation/dereferencing/status.go
```diff
@@ -40,14 +40,25 @@ import (
...
-	if next := status.FetchedAt.Add(2 * time.Hour); time.Now().Before(next) {
+	// Default limit we allow
+	// statuses to be refreshed.
+	limit := 2 * time.Hour
+
+	if force {
+		// We specifically allow the force flag
+		// to force an early refresh (on a much
+		// smaller cooldown period).
+		limit = 5 * time.Minute
```
tsmethurst, line 57: Nice one, this is smart.

internal/db/bundb/status.go
tsmethurst: Index is a good idea I think.

We could update the table to set that column as unique (since more than one status should never share a poll), but since it's a nullable field it's a [bit of a pain for Postgres versions < 15](https://stackoverflow.com/questions/8289100/create-unique-constraint-with-null-columns) (though no issue for SQLite). So a simple index seems OK for now, and perhaps later we can rejig this table (and specify a minimum postgres version requirement).

internal/processing/common/status.go
```diff
@@ -30,10 +30,12 @@ import (
 
 // GetTargetStatusBy fetches the target status with db load function, given the authorized (or, nil) requester's
 // account. This returns an approprate gtserror.WithCode accounting for not found and visibility to requester.
+// The upToDate argument allows specifying whether the returned copy MUST be up-to-date.
 func (p *Processor) GetTargetStatusBy(
 	ctx context.Context,
 	requester *gtsmodel.Account,
 	getTargetFromDB func() (*gtsmodel.Status, error),
+	upToDate bool,
```
tsmethurst, line 38: Not sure about this parameter name. I think `refresh` would make more sense and fit in a bit better with what we already use elsewhere. What do you think?

tsmethurst (commented):
Looks good :) I have a small quibble about parameter names, and an index on poll_id would be good, but I reckon that's it :)

internal/processing/common/status.go
NyaaaWhatsUpDoc: 'thar we go 7cbc68065

internal/db/bundb/status.go
NyaaaWhatsUpDoc: i actually thought i already had set the `poll_id`column as unique, but i guess past kim ran into this same issue testing it. shall add the index :ok_hand:

NyaaaWhatsUpDoc: annnnnnnd there we go 7a3453f00

internal/db/bundb/migrations/20231215115920_add_status_poll_index.go
```diff
@@ -0,0 +1,67 @@
...
+	up := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			type spec struct {
+				index   string
+				table   string
+				columns []string
+			}
+
+			for _, spec := range []spec{
+				// Rename the admin actions indexes.
```
tsmethurst, line 36: This comment looks like a copy paste remnant, could you change/remove it?

NyaaaWhatsUpDoc: :innocent:

PR #2465, opened 2023-12-16
[docs] Fix typo in help text regarding environment variable for instance-languages

bacardi55:
Fix typo in help text about instance-languages configuration using docker environment variable

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -42,7 +42,7 @@ If you're in doubt about any of the names of these environment variables, just c
     
     If you need to use an environment variable to set a configuration option that accepts an array, provide each value in a comma-separated list.
     
-    For example, `instance-languages` may be set in the config.yaml file as an array like so: `["nl", "de", "fr", "en"]`. To set the same values as an environment variable, use: `GTS_INSTANCE_LANGUAGES="nl,de,fr,en"`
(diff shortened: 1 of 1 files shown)


docs/configuration/index.md
```diff
@@ -42,7 +42,7 @@ If you're in doubt about any of the names of these environment variables, just c
     
     If you need to use an environment variable to set a configuration option that accepts an array, provide each value in a comma-separated list.
     
-    For example, `instance-languages` may be set in the config.yaml file as an array like so: `["nl", "de", "fr", "en"]`. To set the same values as an environment variable, use: `GTS_INSTANCE_LANGUAGES="nl,de,fr,en"`
+    For example, `instance-languages` may be set in the config.yaml file as an array like so: `["nl", "de", "fr", "en"]`. To set the same values as an environment variable, use: `GTS_INSTANCE_LANGUAGES: "nl,de,fr,en"`
```
daenney, line 45: I'm afraid this is incorrect. Environment variables are assigned using a `KEY=value` approach. It's only in places like the `docker-compose.yml`, because they're a YAML file, that you'll see `KEY: value`.

daenney: However, even in `docker-compose, you can use the `KEY=value` approach, as you can see in the examples here: https://docs.docker.com/compose/compose-file/compose-file-v3/#environment.

bacardi55: Sorry, I was wrong indeed (even though docker compose didn't like using the `=` and I had to change it to `:`. But this PR is wrong so I'm closing it, sorry for the noise

daenney: No worries Thanks for taking the time!

Issue #2474, opened 2023-12-18
[bug] Friendica accounts sometimes stored in the database with `null` domain


tsmethurst:
A few people lately have reported seeing "ghost" accounts in the "users" count of their single user instance. Where they expect to see 1 user, they see 2 or 3.

This seems to be caused by Friendica service accounts being stored occasionally with `null` for their `domain` value, which the count query then interprets as them being local accounts.

We should fix this by:

1. Modifying our logic to ensure we don't store `null` value accounts in the database when they're not definitely local accounts AND
2. Adding a migration to correct the existing null domain Friendica accounts.

PR #2475, opened 2023-12-18
[bugfix] misc dereferencer fixes

NyaaaWhatsUpDoc:
# Description

- fixes issue of new statusUpToDate() checks skipping status edits made in < 5mins after creation (as noted + tracked down by @Sentynel in matrix, thank you!). now we only perform the up-to-date check if no incoming data is provided, i.e. if we need to make an outgoing dereference request.
- updates account dereferencing up-to-date logic to match that of statuses
- adds a check after all webfingering + dereferencing in the account dereferencer for an empty domain #2474 

the migration itself i'm thinking we write after christmas, as it will involve dropping accounts with NULL domain entries, which is VERY SCARY IN CASE WE DROP LOCAL USER ACCOUNTS. i'm thinking we perform a double check by parsing the account uris and checking the host field maybe??

## Checklist

### DIFF
--- a/internal/federation/dereferencing/account.go
+++ b/internal/federation/dereferencing/account.go
@@ -41,7 +41,7 @@ import (
 
 // accountUpToDate returns whether the given account model is both updateable (i.e.
 // non-instance remote account) and whether it needs an update based on `fetched_at`.
-func accountUpToDate(account *gtsmodel.Account) bool {
+func accountUpToDate(account *gtsmodel.Account, force bool) bool {
 	if !account.SuspendedAt.IsZero() {
 		// Can't update suspended accounts.
 		return true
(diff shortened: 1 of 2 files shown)


internal/federation/dereferencing/account.go
```diff
@@ -57,8 +57,19 @@ func accountUpToDate(account *gtsmodel.Account) bool {
 		return true
 	}
 
-	// If this account was updated recently (last interval), we return as-is.
-	if next := account.FetchedAt.Add(6 * time.Hour); time.Now().Before(next) {
+	// Default limit we allow
+	// statuses to be refreshed.
```
tsmethurst, line 61: This should say accounts, not statuses, right?

tsmethurst:
Looks good, just the one comment on a comment, but feel free to squerge if you like, it's very very very minor

Issue #1311, opened 2023-01-06
[feature] Allow unsetting contact account


tsmethurst:
Right now one can set a contact account via the admin panel + the api.

We should update the API to also allow *unsetting* the contact user by submitting an empty string.

This requires some pretty small changes here: https://github.com/superseriousbusiness/gotosocial/blob/67919204af970803ba707cb40537dfd9ad811c27/internal/processing/instance.go#L137-L170

vaguely relates to https://github.com/superseriousbusiness/gotosocial/issues/499

Issue #707, opened 2022-07-12
[feature] Footer links for imprint and privacy policy


boerge30:
### Is your feature request related to a problem ?

No footer links for the imprint and privacy policy (DSGVO / GDPR Stuff).

### Describe the solution you'd like.

I am looking for a simple way to use footer links for the imprint and the privacy policy (DSGVO / GDPR). Best via the admin interface, as it already works with the e-mail and the username.

### Describe alternatives you've considered.

None

tsmethurst:
Thanks for raising this! In future we'll have a 'terms of service' or 'terms and conditions' thing linked on the front page of the instance that users can read. You could put GDPR info there. Would that fit the bill?

boerge30:
Definitely! Thanks :)

tsmethurst:
Whoops

Issue #143, opened 2021-08-17
[feature]  Instance terms of service \ code of conduct pages


jk-na:
Lots of responsible fediverse users and admin sensibly don't follow users or allow instances to use their relay services without a comprehensive terms of service or code of conduct.  It would be good to have this and also great if this was easily editable via the admin interface.

PR #2481, opened 2023-12-28
[feature] Parse instance descriptors as markdown, show T&C on /about

tsmethurst:
# Description

>

This pull request updates our settings panel to allow parsing instance descriptor fields as markdown (short description, description, terms and conditions), presenting the original "unparsed" version of the field to the user for editing, similar to what we do for status text and account note/bio.

Also updates the /about page to add a table of contents (since it's getting long), and expose terms and conditions.

Few little other nice changes:

- Allow code formatting + syntax highlighting outside of just statuses.
- Allow contact account to be unset (closes https://github.com/superseriousbusiness/gotosocial/issues/1311)
- Show character limits + do validation for descriptor fields.

Closes https://github.com/superseriousbusiness/gotosocial/issues/143
Closes https://github.com/superseriousbusiness/gotosocial/issues/707

## Checklist


internal/gtsmodel/instance.go
```diff
@@ -31,8 +31,11 @@ type Instance struct {
 	DomainBlockID          string       `bun:"type:CHAR(26),nullzero"`                                      // ID of any existing domain block for this instance in the database
 	DomainBlock            *DomainBlock `bun:"rel:belongs-to"`                                              // Domain block corresponding to domainBlockID
 	ShortDescription       string       `bun:""`                                                            // Short description of this instance
-	Description            string       `bun:""`                                                            // Longer description of this instance
-	Terms                  string       `bun:""`                                                            // Terms and conditions of this instance
+	ShortDescriptionText   string       `bun:",nullzero"`                                                   // Raw text version of short description (before parsing).
```
NyaaaWhatsUpDoc, line 34: should we use "Raw" as the suffix here for these instead of "Text"? just to match how we handle account notes.

also is there a reason that these use nullzero but the non-raw versions don't? would they not have the same null value semantics regarding their existence or not?

NyaaaWhatsUpDoc:
only a single comment, otherwise looks good :)

internal/gtsmodel/instance.go
tsmethurst: It actually was `raw` in the original version of this PR, but I changed it to `text` to match what we do for statuses. I actually don't know why we use one term in accounts and one term in statuses, some Mastodon API stuff I suppose.

Re: null, yeah we could just set an empty string I suppose, it doesn't really matter too much imo. I can change it if you like.

tsmethurst: Ack, I'll change the bun tags :)

PR #2502, opened 2024-01-06
Update s3 storage configuration with ACL set to public

martijndeb:
# Description

This PR changes the s3cmd sync command to include the --acl-public flag. This is required because otherwise your files can't be accessed from the web. It also adds a setacl command to indicate what you need to do if you've omitted the public acl before.

## Checklist

### DIFF
--- a/docs/configuration/storage.md
+++ b/docs/configuration/storage.md
@@ -121,7 +121,13 @@ This step is only needed if you have a running instance. Ignore this if you are 
 We have provided [s3cmd](https://github.com/s3tools/s3cmd) command for the copy operation.
 
 ```bash
-s3cmd sync --add-header="Cache-Control:public, max-age=315576000, immutable" ./ s3://<bucket name>
+s3cmd sync --acl-public --add-header="Cache-Control:public, max-age=315576000, immutable" ./ s3://<bucket name>
+```
+
+If you've omitted the public permission, you can set this permission afterwards.
+
+```bash
+s3cmd setacl s3://<bucket name> --acl-public --recursive
 ```
 
 


tsmethurst:
This public permission shouldn't be necessary. When using S3, GoToSocial either proxies media, or forwards media requests using signed URLs, so your bucket/media should not need to be public. For example, try opening this in your browser (the profile picture for the GoToSocial account): https://gts.superseriousbusiness.org/fileserver/016VP4S3BP3QSJBACEQ6J1VFDX/attachment/original/01N7F4KRTAWQP5NTQTHQ56XHD7.png. You'll be forwarded to the GtS S3 bucket using a signed URL that allows you to retrieve the media.

martijndeb:
Hm. I had the files being rejected before setting them public. But I just created the bucket before that, perhaps that's a timing thing of my host (Vultr). I just restored the ACL to private and indeed it seems to work now.

What is still missing from the docs I think is that you most likely need to configure the new advanced-csp-extra-uris: [] param (which is documented under advanced to include the bucket endpoints.

tsmethurst:
Ah, most of the time you shouldn't need to touch `advanced-csp-extra-uris` either. On startup in S3 mode your instance will probe the S3 bucket to derive the host part of the path and add that to your content-security-policy headers. Did you need to change those settings when you migrated over?

martijndeb:
No, I just set them because the config file mentioned the s3 buckets there. So I've been running with them configured there since moving to S3. I just assumed that was a small oversight here.

So I think we can then conclude this PR can be closed and docs are up to date and my initial issue was some timing thing because the bucket was created literally a few minutes before

tsmethurst:
Alright! Thanks for the PR anyways, it's appreciated :)

PR #2513, opened 2024-01-09
[bugfix] fix higher-level explicit domain rules causing issues with lower-level domain blocking

NyaaaWhatsUpDoc:
# Description

Turns out the children in the radix trie were being sorted in the wrong direction :woman_facepalming: 

## Checklist

### DIFF
--- a/internal/cache/domain/domain.go
+++ b/internal/cache/domain/domain.go
@@ -195,13 +195,11 @@ func (n *node) Match(parts []string) bool {
 
 		if nn == nil {
 			// No match :(
-			fmt.Println("no match")
 			return false
 		}
 
 		if len(nn.child) == 0 {
 			// It's a match!
-			fmt.Println("match!")
 			return true
 		}
 
--- a/internal/cache/domain/domain.go
+++ b/internal/cache/domain/domain.go
@@ -19,10 +19,9 @@ package domain
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 	"sync/atomic"
-
-	"golang.org/x/exp/slices"
 )
 
 // Cache provides a means of caching domains in memory to reduce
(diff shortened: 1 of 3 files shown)


internal/cache/domain/domain.go
```diff
@@ -57,6 +56,24 @@ func (c *Cache) Matches(domain string, load func() ([]string, error)) (bool, err
 			return false, fmt.Errorf("error reloading cache: %w", err)
 		}
 
+		// Ensure the domains being inserted into the cache
+		// are sorted by number of domain parts. i.e. those
+		// with less parts are inserted last, else this can
+		// allow domains to fall through the matching code!
+		slices.SortFunc(domains, func(a, b string) int {
+			const k = +1
```
NyaaaWhatsUpDoc, line 64: i'm using a constant "k" here because then it makes it easier to toggle the sort order just by modifying this value

internal/cache/domain/domain.go
```diff
@@ -178,11 +195,13 @@ func (n *node) match(parts []string) bool {
 
 		if nn == nil {
 			// No match :(
+			fmt.Println("no match")
```
tsmethurst, line 198: :eyes:

tsmethurst:
Oh my goodness :D Well fixed! Could you remove the `fmt.Print` calls? Then I think we can squerge this and make a bugfix release :)

internal/cache/domain/domain.go
NyaaaWhatsUpDoc: nooo! my debugging secrets have been revealed!!

NyaaaWhatsUpDoc: and done: [ed38960](https://github.com/superseriousbusiness/gotosocial/pull/2513/commits/ed389609fc8d3259548d20873481c12c3928a09d)

Issue #130, opened 2021-08-03
[feature] Account `move` activity for instance migration


tsmethurst:
Mastodon allows an account to migrate from one instance to another, as documented here:

https://docs.joinmastodon.org/user/moving/#migration

Under the hood, this uses a combination of the activitypub `move` activity, and the `alsoKnownAs` field on profiles, as documented here:

https://docs.joinmastodon.org/spec/activitypub/#profile

It would be very cool if GoToSocial was able to support this as well, to allow accounts to move freely across instances with different AP implementations and features.

In order to do this, we need to first add `alsoKnownAs` to the library that GoToSocial uses for federation: https://github.com/go-fed/activity -- using astool: https://github.com/go-fed/activity/tree/master/astool. Then, either we make a PR against that repo (preferred), or we fork the repo and use our own fork (much less preferred, and silly).

Once that's done, we need to implement some functionality in GtS for setting `alsoKnownAs` -- probably through a PATCH to the `/api/v1/accounts/update_credentials` path (https://github.com/superseriousbusiness/gotosocial/blob/main/internal/api/client/account/accountupdate.go). This PATCH would update this field in the account model: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/gtsmodel/account.go#L112

Finally, we'd need to implement the `move` activity in the federating database -- https://github.com/superseriousbusiness/gotosocial/tree/main/internal/federation/federatingdb -- and process side effects in the processor somewhere -- https://github.com/superseriousbusiness/gotosocial/blob/main/internal/processing/fromfederator.go

GitpayPop:
This issue was added to Gitpay, check the task [https://gitpay.me/#/task/612](https://gitpay.me/#/task/612) 💝

[About Gitpay platform](https://gitpay.me).

For portuguese, you can [access the portuguese version](https://github.com/worknenjoy/gitpay-github-app/blob/master/first-comment.pt-br.md)

### 📋 Step by Step

#### If you are the task owner who created this issue
- [x] 🔔 ***Issue open*** an issue was created
- [x] 👌 **Issue added on Gitpay** this issue was added on gitpay [https://gitpay.me/#/task/612](https://gitpay.me/#/task/612)
- [ ] 📝 **Add bounties** A bounty is added to get things done
- [ ] 📝 **Set a deadline** A deadline to finish the task (optional)
- [ ] 💾 **Choose a contributor interested** You can choose someone interested that will be able to solve your issue

[...]

PR #2518, opened 2024-01-13
[feature] Account alias / move API + db models

tsmethurst:
# Description

>

This pull request adds client API support and documentation for setting account aliases ("alsoKnownAs") and triggering the account Move activity.

- add POST to /api/v1/account/alias to set/unset aliases.
- add POST to /api/v1/account/move to trigger a move to target account.
- update database models and support functions to set/unset/retrieve aka and move URIs properly
- update internal-to-frontend conversion logic to include moved and aka URIs information

Currently while it is possible to send a `Move` activity to the clientAPI processor via the newly-added endpoints, assuming all preconditions are met, nothing is yet processed. This will come in a subsequent PR.

relates to #130 

## Checklist


internal/db/bundb/migrations/20230328203024_migration_fix/account.go
```diff
@@ -0,0 +1,79 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package gtsmodel
+
+import (
+	"crypto/rsa"
+	"time"
+)
+
+// Account represents either a local or a remote fediverse account, gotosocial or otherwise (mastodon, pleroma, etc).
```
tsmethurst, line 25: Needed to store this version of the account model to ensure that migrations didn't break. Previously `internal/db/bundb/migrations/20230328203024_migration_fix.go` was using whatever latest version of the account model was stored in gtsmodel. Now it uses this version, which reflects what the account model looked like at the time the migration was written.

internal/gtsmodel/account.go
```diff
@@ -52,8 +52,10 @@ type Account struct {
 	Note                    string           `bun:""`                               // A note that this account has on their profile (ie., the account's bio/description of themselves)
 	NoteRaw                 string           `bun:""`                               // The raw contents of .Note without conversion to HTML, only available when requester = target
 	Memorial                *bool            `bun:",default:false"`                 // Is this a memorial account, ie., has the user passed away?
-	AlsoKnownAs             string           `bun:"type:CHAR(26),nullzero"`         // This account is associated with x account id (TODO: migrate to be AlsoKnownAsID)
-	MovedToAccountID        string           `bun:"type:CHAR(26),nullzero"`         // This account has moved this account id in the database
+	AlsoKnownAsURIs         []string         `bun:"also_known_as_uris,nullzero"`    // This account is associated with these account URIs.
```
tsmethurst, line 55: I changed this from IDs to URIs here because we might also want to store URIs for accounts that we don't necessarily want to keep ID'd in our database (for example, accounts from suspended instances).

NyaaaWhatsUpDoc: why providing the name in the bun tag here? how was bun formatting it otherwise?

internal/processing/account/move.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/log"
+	"github.com/superseriousbusiness/gotosocial/internal/messages"
+	"github.com/superseriousbusiness/gotosocial/internal/oauth"
+	"golang.org/x/crypto/bcrypt"
+)
+
+const (
+	moveTargetSuspended  = "target account %s is suspended from this instance; you will not be able to Move to that account; if you believe this is an error, please contact your instance admin"
```
NyaaaWhatsUpDoc, line 36: i think it's probably worth moving these into the locations where they're actually being used below, since they're only used once and it makes the code itself harder to read at a glance needing to "dereference" the constants to the top of the file. if they're feeling a bit long for the code, it probably indicates the error string could do with being reworded / shortened imo

internal/processing/account/move.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+func (p *Processor) MoveSelf(
+	ctx context.Context,
+	authed *oauth.Auth,
+	form *apimodel.AccountMoveRequest,
+) gtserror.WithCode {
+	// Ensure valid MovedToURI.
+	if form.MovedToURI == "" {
+		err := gtserror.New("no moved_to_uri provided in account Move request")
```
NyaaaWhatsUpDoc, line 48: for all these returned error messages intended to largely just be user facing, i think it's worth just using `errors.New()` otherwise we're leaking internal function names which aren't providing extra useful context to users or administrators

internal/processing/account/alias.go
```diff
@@ -0,0 +1,137 @@
...
+	"errors"
+	"net/url"
+
+	apimodel "github.com/superseriousbusiness/gotosocial/internal/api/model"
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+)
+
+const (
+	aliasTargetSuspended = "target account %s is suspended from this instance; you will not be able to set alsoKnownAs to that account; if you believe this is an error, please contact your instance admin"
```
NyaaaWhatsUpDoc, line 31: same comments here re: constant error strings + use of `gtserror.New_()` in user-facing error strings as in `move.go`

internal/processing/account/move.go
```diff
@@ -0,0 +1,145 @@
...
+		return gtserror.NewErrorBadRequest(err, err.Error())
+	}
+
+	movedToURI, err := url.Parse(form.MovedToURI)
+	if err != nil {
+		err := gtserror.Newf("invalid moved_to_uri provided in account Move request: %w", err)
+		return gtserror.NewErrorBadRequest(err, err.Error())
+	}
+
+	if movedToURI == nil || (movedToURI.Scheme != "https" && movedToURI.Scheme != "http") {
```
NyaaaWhatsUpDoc, line 58: i don't think we need to do the nill check here

internal/processing/account/alias.go
```diff
@@ -0,0 +1,137 @@
...
+		// get and check the target account, and set.
+		for _, rawURI := range newAKAURIs {
+			alsoKnownAsURI, err := url.Parse(rawURI)
+			if err != nil {
+				err := gtserror.Newf(aliasBadURI, rawURI, err)
+				return nil, gtserror.NewErrorBadRequest(err, err.Error())
+			}
+
+			// We only deref http or https, so check this.
+			if alsoKnownAsURI == nil || (alsoKnownAsURI.Scheme != "https" && alsoKnownAsURI.Scheme != "http") {
```
NyaaaWhatsUpDoc, line 79: i don't think we need to do a nil check here

internal/processing/account/alias.go
```diff
@@ -0,0 +1,137 @@
...
+		update = true
+
+	default:
+		// Prepare to update AlsoKnownAs entries on the account.
+		account.AlsoKnownAs = make([]*gtsmodel.Account, 0, newLen)
+		account.AlsoKnownAsURIs = make([]string, 0, newLen)
+		update = true
+
+		// Use a map to deduplicate desired URIs.
+		newAKAsMap := make(map[string]struct{}, len(newAKAURIs))
```
NyaaaWhatsUpDoc, line 67: tbh we end up deduplicating string slices quite often, might be useful to separate out that part of this into the utils subpkg as a `utils.Deduplicate[T comparable](in []T) []T` that does a quick deduplication on input, which would then also make the code here a little simpler

NyaaaWhatsUpDoc:
have left a few comments here and there but otherwise looks good to me! :)

internal/gtsmodel/account.go
tsmethurst: It was formatting it as, get this, `also_known_as_uri_s` :)

internal/processing/account/move.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/2518/commits/097585ef44a2d0dac5bac2c6b4b829d06ee33315

internal/processing/account/move.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/2518/commits/097585ef44a2d0dac5bac2c6b4b829d06ee33315

internal/processing/account/alias.go
tsmethurst: Also done here! https://github.com/superseriousbusiness/gotosocial/pull/2518/commits/097585ef44a2d0dac5bac2c6b4b829d06ee33315

internal/processing/account/move.go
tsmethurst: Right, I thought url parsing could sometimes return nil but I misremembered it, it sometimes returns a url with an empty scheme. But since we're checking for scheme here, it's all good. Changed that here: https://github.com/superseriousbusiness/gotosocial/pull/2518/commits/097585ef44a2d0dac5bac2c6b4b829d06ee33315

internal/processing/account/alias.go
tsmethurst: Changed in the same commit as above: https://github.com/superseriousbusiness/gotosocial/pull/2518/commits/097585ef44a2d0dac5bac2c6b4b829d06ee33315

internal/processing/account/alias.go
tsmethurst: Done! Good shout! https://github.com/superseriousbusiness/gotosocial/pull/2518/commits/8021dfe97536dc0851a5c66d995a46e4b4814d70

PR #2521, opened 2024-01-13
[docs] Update roadmap for 2024 onwards

tsmethurst:
As the title says :) This updates our roadmap to shuffle the things we didn't finish in 2023 into early 2024, and also add some more items for the BETA period of development.


mirabilos:
Mute accounts and mute/filter words/hashtags still not on there? :'(

Whom must I bribe, and with how much of what? ;-)

PR #2525, opened 2024-01-14
[bugfix] Replace named unique constraint on header filter header with generic unique directive

tsmethurst:
# Description

>

This pull request fixes an issue where postgres migration was failing, because the migration was trying to create two identically-named unique constraints (which postgres doesn't like).

Instead, uniqueness is now NOT enforced for canonical header name + regex on HeaderFilter,.

It adds a migration to drop any partially created tables caused by running the previous migration on Postgres, and to tidy up any constraints that might still be kicking around in the database. 

This is what a previous failed migration looks like in Postgres (note only allow table was created because the second table creation failed on the unique constraint).

Screenshot from 2024-01-15 16-28-28

Screenshot from 2024-01-15 16-28-36

## Checklist


internal/gtsmodel/headerfilter.go
```diff
@@ -45,8 +45,8 @@ type HeaderFilterBlock struct{ HeaderFilter }
 // matching regex, and details about its creation.
 type HeaderFilter struct {
 	ID        string    `bun:"type:CHAR(26),pk,nullzero,notnull,unique"`                    // ID of this item in the database
-	Header    string    `bun:",nullzero,notnull,unique:header_regex"`                       // Request header this filter pertains to
```
NyaaaWhatsUpDoc, line 48: honestly if this causes such a serious issue, it's probably just worth skipping out on the unique constraints. this is an advanced setting, so any admin will realize that having multiple copies of the same matching regex for the same header key is wasted performance. that's all the unique constraint is adding, a mild performance safeguard. the code itself has no issues with having multiple copies of the same matching mechanism

tsmethurst: Ah fair, I'm happy to drop it altogether! I'll check if that means we need to create an index on `header`; ie., do we ever select using it.

Issue #2522, opened 2024-01-13
[bug] Postgres username search has become case-sensitive


mirabilos:
### Describe the bug with a clear and concise description of what the bug is.

#2435 changed search to not lowercase the search term; ostensibly only for SQLite.

However, currently, my searches (by typing `@foo` or `@Foo` in Semaphore, as well as in the search form) are case-sensitive *with PostgreSQL* :(

### What's your GoToSocial Version?

0.13.1 git-ccecf5a

### GoToSocial Arch

amd64 binary

PR #2526, opened 2024-01-14
[bugfix] Better Postgres search case insensitivity

tsmethurst:
# Description

>

This pull request provides slightly more rigorous case-insensitivity on Postgres when searching for accounts + statuses.

closes https://github.com/superseriousbusiness/gotosocial/issues/2522

## Checklist


internal/db/bundb/util.go
```diff
@@ -37,6 +37,24 @@ var likeEscaper = strings.NewReplacer(
...
+// prepareSearchString prepares a query
+// string for use in a LIKE search.
+func prepareSearchString(search string) string {
+	// Lowercase the query string. This
+	// doesn't make a difference for the
+	// case-insensitive SQLite, but for
+	// Postgres it matches the LOWER()
+	// function calls that should be used
+	// in the LIKE subquery.
+	search = strings.ToLower(search)
```
mirabilos, line 49: lowercasing should not make a difference for LIKE-escaping, so you could just let the DB do the lowercasing instead.

The query optimiser will ensure it’ll be done only once.

That would have the benefit of using the same lowercasing rules for both sides of the comparision; otherwise you could get into trouble if e.g. the database uses C.UTF-8 but GtS is run in tr_TR.UTF-8 where it’s not I→i but I→ı and i←İ

mirabilos: … actually, PostgreSQL has `ILIKE`, so unless you’re comparing with `=` or `<>` you could just use that instead

mirabilos: https://www.grouparoo.com/blog/sql-dialect-differences#case-insensitive-sting-comparisons

tsmethurst: Oh thanks, I remember reading about ILIKE and not choosing it for some reason, though I now can't remember what the reason was. I'll play around with it and see if it's better for this case.

mirabilos:
tobi dixit:

>Oh thanks, I remember reading about ILIKE and not choosing it for some
>reason, though I now can't remember what the reason was.

>From the article I linked, SQLite does not support ILIKE. Maybe that?

>I'll play around with it and see if it's better for this case.

Thanks,
//mirabilos

internal/db/bundb/util.go
NyaaaWhatsUpDoc: if this PR still needs some playing with do you want to move it to draft? just so i don't end up accidentally merging it early :p

tsmethurst: Will do yeah.

tsmethurst:
Alright, this seems to work well. Thanks @mirabilos.

@NyaaaWhatsUpDoc this can be reviewed now :)

mirabilos:
Thank you!

daenney (approved):
That seems fine!

PR #2535, opened 2024-01-15
[performance] overhaul struct (+ result) caching library for simplicity, performance and multiple-result lookups

NyaaaWhatsUpDoc:
# Description

- updates gotosocial to use `codeberg.org/gruf/go-structr` instead of `codeberg.org/gruf/go-cache/v3/result` for result / struct caching which includes the following changes:
  - rewrite caching library for reduced complexity
  - improved caching library code commenting, design
  - now uses just a doubly linked list for tracking entries, instead a linked list + hashmap unnecessarily
  - uses linked lists is more places where advantageous over slices
  - increases the number of internal types that can be pooled / reused
  - adds much more robust tests
  - adds support for loading + caching multiple values at a time
- updates gotosocial to use the new multi value cache loader functions to reduce database calls
- removes the unnecessary step of making the sub-caches package private and only accessible by helper methods
- updates regular cache sweep to keep caches within "60" threshold instead of "80" to give us more headroom during load spikes

## Checklist


NyaaaWhatsUpDoc:
this diff is a lot more intimidating than it looks in terms of actual gotosocial code, because it includes dropping the old result library from vendor/, and adding the new one.

the separate library itself can be found at https://codeberg.org/gruf/go-structr

cmd/gotosocial/action/server/server.go
```diff
@@ -133,7 +133,7 @@ var Start action.GTSAction = func(ctx context.Context) error {
 		time.Time{},   // start
 		time.Minute,   // freq
 		func(context.Context, time.Time) {
-			state.Caches.Sweep(80)
+			state.Caches.Sweep(60)
```
NyaaaWhatsUpDoc, line 136: this helps reduce the chance we'll always be pushing against the cache maximum capacity, gives us a solid 40% of extra cache space to play with during peak load without worrying about causing an eviction on every new insert

internal/cache/cache.go
```diff
@@ -49,12 +48,41 @@ type Caches struct {
 func (c *Caches) Init() {
 	log.Infof(nil, "init: %p", c)
 
-	c.GTS.Init()
-	c.Visibility.Init()
-
-	// Setup cache invalidate hooks.
-	// !! READ THE METHOD COMMENT
-	c.setuphooks()
+	c.initAccount()
```
NyaaaWhatsUpDoc, line 51: i moved the init functions to be a part of the top-level caches model, instead of the sub-types (e.g. GTS), because then during cache init functions we have full access to all the other caches for cache invalidation hooks etc. it wasn't necessary previously due to slight API changes between the old and new struct cache libraries

internal/cache/cache.go
```diff
@@ -73,207 +101,37 @@ func (c *Caches) Stop() {
 	c.Visibility.Stop()
 }
 
-// setuphooks sets necessary cache invalidation hooks between caches,
```
NyaaaWhatsUpDoc, line 76: these hooks are now set during cache init, so i separated them out into separate functions in `invalidate.go` which we pass to the caches during init, but also leaves the possibility of calling these hooks in other locations in the code which could be useful

internal/cache/db.go
```diff
@@ -0,0 +1,1083 @@
...
+
+		return a2
+	}
+
+	c.GTS.Account.Init(structr.Config[*gtsmodel.Account]{
+		Indices: []structr.IndexConfig{
+			{Fields: "ID"},
+			{Fields: "URI"},
+			{Fields: "URL"},
+			{Fields: "Username,Domain", AllowZero: true},
```
NyaaaWhatsUpDoc, line 193: notice the field naming mechanism now uses commas to separate fields now, this is because periods are reserved instead for nested fields e.g. on a status could do something like `Account.URI`

internal/db/bundb/emoji.go
```diff
@@ -568,28 +568,69 @@ func (e *emojiDB) PopulateEmoji(ctx context.Context, emoji *gtsmodel.Emoji) erro
...
-	for _, id := range emojiIDs {
-		emoji, err := e.GetEmojiByID(ctx, id)
-		if err != nil {
-			log.Errorf(ctx, "emojisFromIDs: error getting emoji %q: %v", id, err)
-			continue
-		}
+	// Load all emoji IDs via cache loader callbacks.
+	emojis, err := e.state.Caches.GTS.Emoji.Load("ID",
+
+		// Load cached + check for uncached.
```
NyaaaWhatsUpDoc, line 582: the initial passed function is used to load the wanted keys from the cache, it doesn't return them here but notice we specifically add load failures to the `uncached` slice to indicate we want to load those from the database in the later callback function below it

internal/db/bundb/emoji.go
```diff
@@ -568,28 +568,69 @@ func (e *emojiDB) PopulateEmoji(ctx context.Context, emoji *gtsmodel.Emoji) erro
...
+			}
+
+			return emojis, nil
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// Reorder the emojis by their
```
NyaaaWhatsUpDoc, line 612: unfortunately we need to re-order them because internally the cache gathers all the values it does have cached, then appends the newly fetched models from the database to the end of those, as it's only dealing with generic parameters it can't actually deal with any ordering itself without it getting very hairy. instead i just wrote this nice function that orders an input slice according to a slice of given IDs (without needing to do any slice copying or anything, it just moves things around in the provided slice)

NyaaaWhatsUpDoc:
the cache size estimator i'm going to do in a separate PR, this PR is already large enough, and enough of the internals are similar that the size estimations won't actually be *that* far off

cmd/gotosocial/action/server/server.go
tsmethurst: Sensible. Perhaps in future this ought to be tuneable under the `advanced` config section? Just a thought, not urgent.

internal/cache/cache.go
tsmethurst: Got it :+1:

internal/cache/db.go
```diff
@@ -0,0 +1,1083 @@
...
+
+	// User provides access to the gtsmodel User database cache.
+	User structr.Cache[*gtsmodel.User]
+
+	// Webfinger provides access to the webfinger URL cache.
+	// TODO: move out of GTS caches since unrelated to DB.
+	Webfinger *ttl.Cache[string, string] // TTL=24hr, sweep=5min
+}
+
+// Start will attempt to start all of the gtsmodel caches, or panic.
```
tsmethurst, line 145: So only TTL caches need to be started here, right?

internal/cache/db.go
tsmethurst: Ack, got it :+1: That's a sensible change. This syntax makes more sense anyway!

internal/cache/db.go
```diff
@@ -0,0 +1,1083 @@
...
+
+// NOTE:
+// all of the below init functions
+// are receivers to the main cache
+// struct type, not the database cache
+// struct type, in order to get access
+// to the full suite of caches for
+// our invalidate function hooks.
+
+func (c *Caches) initAccount() {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: that is indeed correct, straight-copy paste and just modifed to support the new structr.Cache init function signatures :)

internal/cache/db.go
NyaaaWhatsUpDoc: indeed ! worth changing the comment?

tsmethurst: I think it's probably worth mentioning yeah, if you wouldn't mind :) Just helps thembos like me to understand what's happening.

internal/db/bundb/emoji.go
tsmethurst: Ohh very interesting!

internal/db/bundb/emoji.go
tsmethurst: How does that compare to how slices.Sort() does it? Does that allocate / copy? I know slices.Sort doesn't return a slice, it just operates on the slice you give it, but perhaps it allocates while sorting, I'm not sure.

internal/db/bundb/emoji.go
```diff
@@ -601,36 +642,51 @@ func (e *emojiDB) getEmojiCategory(ctx context.Context, lookup string, dbQuery f
...
+			for _, id := range ids {
+				if !load(id) {
+					uncached = append(uncached, id)
+				}
+			}
+		},
 
-	return emojiCategories, nil
-}
+		// Uncached follow req loader function.
```
tsmethurst, line 665: Copy paste artifact? :P

internal/db/bundb/emoji.go
NyaaaWhatsUpDoc: the thing is we don't actually know what the sort of the incoming slice of IDs is, it may be descending / ascending / unrelated to the objects themselves, that's why we just order by the provided ID slice

internal/db/bundb/emoji.go
NyaaaWhatsUpDoc: :eyes:

testrig/testmodels.go
```diff
@@ -2842,20 +2842,20 @@ func NewTestFediPeople() map[string]vocab.ActivityStreamsPerson {
...
 		),
-		"https://example.org/users/Some_User": newAPPerson(
-			URLMustParse("https://example.org/users/Some_User"),
-			URLMustParse("https://example.org/users/Some_User/following"),
-			URLMustParse("https://example.org/users/Some_User/followers"),
-			URLMustParse("https://example.org/users/Some_User/inbox"),
-			URLMustParse("https://example.org/sharedInbox"),
-			URLMustParse("https://example.org/users/Some_User/outbox"),
-			URLMustParse("https://example.org/users/Some_User/collections/featured"),
+		"http://example.org/users/Some_User": newAPPerson(
```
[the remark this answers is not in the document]
tsmethurst: Weird!

internal/db/bundb/util.go
```diff
@@ -158,3 +158,88 @@ func parseWhere(w db.Where) (query string, args []interface{}) {
 	args = []interface{}{bun.Ident(w.Key), w.Value}
 	return
 }
+
+// dropIDs drops given ID string from IDs slice.
+func dropID(ids []string, id string) []string {
```
tsmethurst, line 163: I think you could probably put these functions over in `internal/util` instead somewhere. Feels like we'll end up wanting to use them elsewhere at some point. Maybe in `internal/util/slices.go` or something, since they're all slice functions? And you could put those deduplicate functions I wrote into that file as well. What do you think?

internal/db/bundb/mention.go
```diff
@@ -63,21 +63,61 @@ func (m *mentionDB) GetMention(ctx context.Context, id string) (*gtsmodel.Mentio
...
+	orderByIDs(mentions, ids, getID)
 
-		// Append mention
-		mentions = append(mentions, mention)
+	if gtscontext.Barebones(ctx) {
+		// no need to fully populate.
+		return mentions, nil
+	}
+
+	// Populate all loaded blocks.
```
tsmethurst, line 112: Another c+p artifact. Maybe it's worth just changing these comments to say "items" instead of blocks or mentions or whatever. It's still gonna be fairly clear to the reader what the comment is referring to.

internal/db/bundb/relationship_follow.go
```diff
@@ -316,22 +354,30 @@ func (r *relationshipDB) DeleteAccountFollows(ctx context.Context, accountID str
...
+	_, err := r.GetAccountFollows(ctx, accountID, nil)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return err
+	}
 
-		// Delete each follow from DB.
-		if err := r.deleteFollow(ctx, follow.ID); err != nil &&
-			!errors.Is(err, db.ErrNoEntries) {
+	// Delete all follows from DB.
+	_, err = r.db.NewDelete().
```
tsmethurst, line 370: Now that all follows are being deleted in this way, instead of calling DeleteFollow one by one, the invalidation hooks from DeleteFollow are no longer being called. Is this OK?

internal/db/bundb/status.go
```diff
@@ -48,18 +48,57 @@ func (s *statusDB) GetStatusByID(ctx context.Context, id string) (*gtsmodel.Stat
...
+	orderByIDs(statuses, ids, getID)
 
-		// Append status to return slice.
-		statuses = append(statuses, status)
+	if gtscontext.Barebones(ctx) {
+		// no need to fully populate.
+		return statuses, nil
+	}
+
+	// Populate all loaded statuses.
```
tsmethurst, line 97: One logical change I notice in this and other functions is that previously if a `Populate` call failed, then `GetStatusByID` would return an error, and the status would be skipped from being inserted into the returned slice of Statuses. Now, a `Populate` call might fail but the status will still be returned. This means this `GetStatusesByIDs` function no longer guarantees that each status will be populated, and so calls to pointers of structs attached to the status might now panic.

This isn't necessarily undesirable behavior, but we'll have to be on the lookout for `panic`s in the logs while we run on this main, to make sure we haven't missed any nil checks in any of the functions calling this one (and others). It might be worth having a quick look-through before we merge it to see which functions are calling these `Get[Whatever]sByIDs` functions, and make sure they do appropriate nil checking.

internal/cache/db.go
NyaaaWhatsUpDoc: done! c907370c5

internal/db/bundb/emoji.go
NyaaaWhatsUpDoc: done! 8a6304e96

internal/db/bundb/mention.go
NyaaaWhatsUpDoc: done! 462697d47

though i just fixed the issues, i didn't have the energy to change them all to "items". though going forwards i'll keep that in mind to use more generic language if it's gonna end up being copy-pasted :D

internal/db/bundb/util.go
NyaaaWhatsUpDoc: done! see:
dee652415142cf2323bd0ecda34ba968074ee515
2b258c715a3c9ac1bf3da4cd3cd1ac028b7e2501
a002417d2842d8c661331049e4a52a775a3882c8

internal/db/bundb/util.go
```diff
@@ -158,3 +158,88 @@ func parseWhere(w db.Where) (query string, args []interface{}) {
...
+			ts = append(ts, t)
+			tm[t] = struct{}{}
+		}
+	}
+
+	return ts
+}
+
+// orderByIDs orders a slice of database models by the provided slice of IDs.
+func orderByIDs[T any](models []T, ids []string, getID func(T) string) {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: done! see:
dee652415142cf2323bd0ecda34ba968074ee515
2b258c715a3c9ac1bf3da4cd3cd1ac028b7e2501
a002417d2842d8c661331049e4a52a775a3882c8

internal/db/bundb/status.go
NyaaaWhatsUpDoc: that is a good point... i think i might fix this to maintain existing behaviour then, because while it's not a clear-cut case of which is the preferred outcome, it certainly is simpler not having to nil check all over the codebase :p

internal/db/bundb/util.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: resolving this one so we can keep comments in one place since they're 2 (kinda) duplicate threads

internal/db/bundb/status.go
NyaaaWhatsUpDoc: and done: a16a68fa8

NyaaaWhatsUpDoc: switched to using `slices.DeleteFunc()` as per matrix discussion 8242e5aa3

tsmethurst (approved):
Good job kimbe :+1:

PR #2539, opened 2024-01-17
[chore] update viper version

NyaaaWhatsUpDoc:
Updates the viper version and updates areas of the library relying on the old exp/slices function signature to use the new stdlib slices package


tsmethurst (approved):
Feel free to squerge when this passes :) Thanks for picking it up.

PR #2540, opened 2024-01-17
[chore] chore rationalise http return codes for activitypub handlers

NyaaaWhatsUpDoc:
We do this by dropping transient activities (those with missing ID property) instead of returning 400 bad request, since they are *technically* valid activitypub data.

Also some other misc fixes / cleanup i noticed while in the code


daenney (approved):
Nice! Ignoring transient ones is a good catch.

Issue #2491, opened 2024-01-05
[bug] Failing to fetch data from threads.net


inetol:
### Describe the bug with a clear and concise description of what the bug is.

As you may already know, Meta is carrying out some controversial tests on Threads with some users to make their posts (?) readable through ActivityPub. When trying to search any federated Threads user account via GoToSocial, the instance (GTS) responds with an HTTP 500 error. (detailed issue details appears in the instance logs, see below...)

### What's your GoToSocial Version?

v0.13.0 git-f4fcffc

### GoToSocial Arch

x86_64 Docker (podman)

### What you expected to happen?

The user "should" be accessible and the request would return the user information.

*(You can try searching for it on [mastodon.online](https://mastodon.online) or any Mastodon instances where threads.net is not blacklisted)*

### How to reproduce it?

- Use a web client or mobile app to interact with the GoToSocial instance
- Look for some Threads user accessible through ActivityPub (e.g. `@mosseri@threads.net`)
- Request will fail with this error on the instance (GTS) logs: 
```bash
errors="Error #01: Get: error searching by namestring: accountsByUsernameDomain: error looking up @mosseri@threads.net as account: enrichAccount: error resolving accountable from data for account https://www.threads.net/ap/users/mosseri/: ResolveAccountable: error unmarshalling bytes into json: invalid character '<' looking for beginning of value\n" requestID=ngwz1pcc0XXXXXXXXXXX msg="Internal Server Error: wrote 54B"
```

### Anything else we need to know?

I know this issue involves a third party with proprietary software and the pain in the ass that goes with it, but I hope it will serve as a reference for future users who experience the same problem.

tsmethurst:
This is a threads bug afaik. They're returning HTML in response to valid ActivityPub json requests, for whatever reason. Nothing we can do about it from our side.

The 500 error on the other hand, we should do something about, because that's a bit weird. But I'd rather have a separate issue for that.

inetol:
Once the Threads federation is GA *and* if the issue persists I would like to look at it further, for the moment this is a bit up in the air.

tsmethurst:
[...]

PR #2550, opened 2024-01-21
[bugfix] Don't return Internal Server Error when searching for URIs that don't return AP JSON

tsmethurst:
# Description

>

This PR fixes an issue where searching for a URL that doesn't return valid AP JSON would return an Internal Server Error to the caller. Instead, failures to parse AP JSON from a dereference now return WrongType instead, which is checked for in the search processor.

Relates to https://github.com/superseriousbusiness/gotosocial/issues/2491, since that was triggering the bug, as threads.net doesn't return valid AP in response to our requests.

This could also be triggered just by putting something like "https://example.org" in the search bar. 

## Checklist

### DIFF
--- a/internal/ap/resolve.go
+++ b/internal/ap/resolve.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/superseriousbusiness/activity/pub"
 	"github.com/superseriousbusiness/activity/streams"
+	"github.com/superseriousbusiness/activity/streams/vocab"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 )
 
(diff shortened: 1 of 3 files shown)


internal/ap/resolve.go
```diff
@@ -56,6 +57,35 @@ func putMap(m map[string]any) {
...
+// as a JSON ActivityPub type, failing if the input
+// bytes are not parseable as JSON, or do not parse
+// to an ActivityPub that we can understand.
+//
+// The given map pointer will also be populated with
+// the parsed JSON, to allow further processing.
+func bytesToType(
+	ctx context.Context,
+	b []byte,
+	raw *map[string]any,
```
daenney, line 70: You don't need a pointer here. Maps are passed by value, but the value that gets passed is the pointer to them: https://go.dev/play/

tsmethurst: Ah thanks! I'm constantly unsure about that :')

internal/ap/resolve.go
```diff
@@ -56,6 +57,35 @@ func putMap(m map[string]any) {
...
+) (vocab.Type, error) {
```
daenney, line 71: I think this can probably be sped up by checking if the first non-space thing we run into is `{`. If it's not, then we can avoid the unmarshalling etc. entirely.

internal/ap/resolve.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2550/commits/e2df9f79fcdcefca89beecd48e8849bfb92e79e9

internal/ap/resolve.go
daenney: Oh ya, sorry, I was just musing about it. Wasn't meant as a "we should implement" this. I should've been clearer about that.

daenney (approved):

internal/ap/resolve.go
NyaaaWhatsUpDoc: @daenney i mean if the JSON unmarshaler is clever that's what it'll be doing internally, just with some extra type checking on the passed "interface{}" type

Issue #2552, opened 2024-01-21
[bug] 500 error when setting alias to mastodon


CDN18:
### Describe the bug with a clear and concise description of what the bug is.

I encountered this while testing #2519 and trying to set my gts account as an alias of my mastodon account. 

### What's your GoToSocial Version?

v0.13.1-SNAPSHOT git-b2cacd6

### GoToSocial Arch

arm64 Docker

### What happened?

I entered `https://mastodon.social/@<my_username>` and clicked save, then got an 500 error.

### What you expected to happen?

The alias was set successfully.

### How to reproduce it?

1. Enter mastodon user url (e.g.  `https://mastodon.social/@<my_username>` ) in migration panel
2. See error

## Logs

```
gotosocial | timestamp="21/01/2024 14:43:43.728" func=server.glob..func1.Logger.func13.1 level=ERROR latency="4.70556ms" userAgent="***" method=POST statusCode=500 path=/api/v1/accounts/alias clientIP=223.97.49.77 errors="Error #01: Alias: db error updating also_known_as_uri: ERROR: malformed array literal: \"[\"https://mastodon.social/users/icarus22\"]\" (SQLSTATE 22P02)\n" requestID=xcvqmb4d04001g4zrrzg msg="Internal Server Error: wrote 54B"
```

tsmethurst:
Thanks, looks like something specific to Postgres, hmm.

CDN18:
Thanks for the quick fix! Can confirm it's working now on postgres.

tsmethurst:
Great. Just be aware that the setting doesn't actually *do* anything yet; it's not plumbed in to anything, so don't expect it to work yet :P

PR #2553, opened 2024-01-21
[bugfix] fix array type for also_known_as_uris

tsmethurst:
Closes https://github.com/superseriousbusiness/gotosocial/issues/2552

### DIFF
--- a/internal/gtsmodel/account.go
+++ b/internal/gtsmodel/account.go
@@ -52,7 +52,7 @@ type Account struct {
 	Note                    string           `bun:""`                               // A note that this account has on their profile (ie., the account's bio/description of themselves)
 	NoteRaw                 string           `bun:""`                               // The raw contents of .Note without conversion to HTML, only available when requester = target
 	Memorial                *bool            `bun:",default:false"`                 // Is this a memorial account, ie., has the user passed away?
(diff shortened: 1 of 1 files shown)


internal/gtsmodel/account.go
```diff
@@ -52,7 +52,7 @@ type Account struct {
 	Note                    string           `bun:""`                               // A note that this account has on their profile (ie., the account's bio/description of themselves)
 	NoteRaw                 string           `bun:""`                               // The raw contents of .Note without conversion to HTML, only available when requester = target
 	Memorial                *bool            `bun:",default:false"`                 // Is this a memorial account, ie., has the user passed away?
-	AlsoKnownAsURIs         []string         `bun:"also_known_as_uris,nullzero"`    // This account is associated with these account URIs.
+	AlsoKnownAsURIs         []string         `bun:"also_known_as_uris,array"`       // This account is associated with these account URIs.
```
daenney, line 55: This seems sensible. Do you know if Bun has a reference of their different tags and modifiers somewhere? Would be useful to be able to consult that but I can't find it in the documentation.

daenney (approved):

internal/gtsmodel/account.go
tsmethurst: They do! https://bun.uptrace.dev/guide/models.html#struct-tags

PR #2556, opened 2024-01-21
[docs] Enhance log-timestamp-format documentation

### LINKED ISSUE #2455: [question] Ambiguos log-timestamp-format in config.yml

Since there is ambiguity between the US-American way of describing dates and basically the rest of the world, the example has to be sensible enough to not confuse day with month.

v0.13.0-rc2

### GoToSocial Arch

AMD64

### What happened?

I installed GtS on my VPS bare metal as per the documentation.

When editing the config.yml I was unsure how the example and default are meant to be interpreted:

https://github.com/superseriousbusiness/gotosocial/blob/v0.13.0-rc2/example/config.yaml#L38-L47

The example says `[true, false]` which doesn't look like a string.
The default says `"02/01/2006 15:04:05.000"` which is a string but the [Go documentation on time](https://pkg.go.dev/time#pkg-constants) gives `01/02 03:04:05PM '06 -0700` as layout example.

Personally I would prefer an RFC3339 format: 2006-01-02T15:04:05Z.

As European I'm not sure, whether `02/01` is referring to 1st of February or 2nd of January.

I would either expect another format or use 13 etc. as day part of the example.
The example confused me.

### How to reproduce it?

Copy the `example/config.yml` and open it in a text editor.

Go To Social is awesome!

mirabilos:
# Description

- use ISO 8601 timestamp and empty string (no timestamp logged) as examples, not [true, false] which isn’t of the right type
- make it clearer that the timestamp template format is imposed by an external standard library, not a GtS invention
    
Having an ISO 8601 example should also make it easier to understand.

Also cleans up whitespace at end of line/file for the touched files.

closes #2455 (hopefully)

## Checklist

### DIFF
--- a/docs/configuration/general.md
+++ b/docs/configuration/general.md
@@ -32,10 +32,11 @@ log-client-ip: true
 # If set to the empty string, the timestamp will be
 # ommitted from the logs entirely.
 #
-# The format must be compatible with Go's time.Layout, as
-# documented on https://pkg.go.dev/time#pkg-constants.
+# This uses the timestamp template format of the golang
+# time.Layout standard library, which is documented at:
+# https://pkg.go.dev/time#pkg-constants
 #
-# Examples: [true, false]
+# Examples: ["2006-01-02T15:04:05.000Z07:00", ""]
 # Default: "02/01/2006 15:04:05.000"
 log-timestamp-format: "02/01/2006 15:04:05.000"
 
--- a/example/config.yaml
+++ b/example/config.yaml
@@ -297,7 +297,7 @@ instance-languages: []
 
 # String. Federation mode to use for this instance.
 #
-# "blocklist" -- open federation by default. Only instances that are explicitly 
+# "blocklist" -- open federation by default. Only instances that are explicitly
 #                blocked will be denied (unless they are also explicitly allowed).
 #
 # "allowlist" -- closed federation by default. Only instances that are explicitly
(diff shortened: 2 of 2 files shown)


docs/configuration/general.md
```diff
@@ -32,10 +32,11 @@ log-client-ip: true
 # If set to the empty string, the timestamp will be
 # ommitted from the logs entirely.
 #
-# The format must be compatible with Go's time.Layout, as
-# documented on https://pkg.go.dev/time#pkg-constants.
+# This uses the timestamp template format of the golang
```
daenney, line 35:
```suggestion
# This uses the timestamp template format of the Go
```

example/config.yaml
```diff
@@ -40,10 +40,11 @@ log-client-ip: true
 # If set to the empty string, the timestamp will be
 # ommitted from the logs entirely.
 #
-# The format must be compatible with Go's time.Layout, as
-# documented on https://pkg.go.dev/time#pkg-constants.
+# This uses the timestamp template format of the golang
```
daenney, line 43:
```suggestion
# This uses the timestamp template format of the Go
```

docs/configuration/general.md
daenney: I don't really find this rewording any clearer, and it's the Go programming language, not "golang". I would suggest reverting that change. Idem for the other one.

tsmethurst:
Agree with Daenney, changing `Go` to `golang` doesn't offer any additional clarity in this context, especially when we already link to the relevant document right there so folks don't need to search for it. The timestamp examples are good though :)

daenney:
We picked up the example change in #2572, so I'm going to close this one. Thanks again!

Issue #2527, opened 2024-01-14
[bug] [federation] honk (tedu)


mirabilos:
### Describe the bug with a clear and concise description of what the bug is.

After benjojo, I now have federation issues with tedu’s honk instance.

Searching for `https://honk.tedunangst.com/u/tedu/h/189775FqjZ7tN7qz6F` gives “The data
couldn’t be read because it isn’t in the correct format.” in FediText, and is probably connected to the following log line:

```
timestamp="2024-01-14T19:37:47.528Z" func=server.glob..func1.Logger.func13.1 level=ERROR stacktrace="()\n\t/usr/local/go/src/runtime/panic.go:261\nruntime.sigpanic()\n\t/usr/local/go/src/runtime/signal_unix.go:861\ntypeutils.(*Converter).AccountToAPIAccountPublic()\n\t/drone/src/internal/typeutils/internaltofrontend.go:218\nsearch.(*Processor).packageAccounts()\n\t/drone/src/internal/processing/search/util.go:79\nsearch.(*Processor).packageSearchResult()\n\t/drone/src/internal/processing/search/util.go:184\nsearch.(*Processor).Get()\n\t/drone/src/internal/processing/search/get.go:238\nsearch.(*Module).SearchGETHandler()\n\t/drone/src/internal/api/client/search/searchget.go:232\n()\n\t/drone/src/vendor/github.com/gin-gonic/gin/context.go:174\ngzip.(*gzipHandler).Handle()\n\t/drone/src/vendor/github.com/gin-contrib/gzip/handler.go:60\n" requestID=0vpqm2cd04001f77cg60 msg="recovered panic: runtime error: invalid memory address or nil pointer dereference"
```

Searching for `https://honk.tedunangst.com/u/tedu` gives me an… empty profile with one toot? Screenshot follows, it better describes this issue.

![Screenshot_20240114_204252](https://github.com/superseriousbusiness/gotosocial/assets/861078/a8b13e06-8c05-41b5-bc47-6f5386ea52c1)

Of course, I cannot follow him, etc. from here.

### What's your GoToSocial Version?

0.13.1 git-ccecf5a

### GoToSocial Arch

amd64 binary

### What you expected to happen?

I was under the impression that it was possible to interact with those honks as Fedi instances, and I searched the bugtracker and found no open bug mentioning honk, so I thought it would be possible.

If it is out of scope, that’s okay as well.

tsmethurst:
[...]

Issue #2514, opened 2024-01-10
[bug] Panic in `AccountToAPIAccountPublic` when looking up account from Catodon


CDN18:
### Describe the bug with a clear and concise description of what the bug is.

Catodon is a new fork of firefish/misskey: https://catodon.social/about-catodon
However, while gts can interact with misskey instances without problem, it get panicked while federating with catodon.

### What's your GoToSocial Version?

v0.13.1

### GoToSocial Arch

arm64 Docker

### What happened?

Gotosocial can't federate with catodon instances.

### What you expected to happen?

Being able to federate with catodon without problem.

### How to reproduce it?

1. search for an account on catodon.social
2. see error

### Anything else we need to know?

Log:

```
timestamp="10/01/2024 16:26:41.359" func=server.glob..func1.Logger.func13.1 level=INFO latency="2.67464ms" userAgent=MoshidonAndroid method=GET statusCode=200 path=/api/v2/search clientIP=0.0.0.0 requestID=9p3k5x4c04000y0pzhxg msg="OK: wrote 59B"

timestamp="10/01/2024 16:26:43.482" func=server.glob..func1.Logger.func13.1 level=ERROR stacktrace="()\n\t/usr/local/go/src/runtime/panic.go:261\nruntime.sigpanic()\n\t/usr/local/go/src/runtime/signal_unix.go:861\ntypeutils.(*Converter).AccountToAPIAccountPublic()\n\t/drone/src/internal/typeutils/internaltofrontend.go:218\nsearch.(*Processor).packageAccounts()\n\t/drone/src/internal/processing/search/util.go:79\nsearch.(*Processor).packageSearchResult()\n\t/drone/src/internal/processing/search/util.go:184\nsearch.(*Processor).Get()\n\t/drone/src/internal/processing/search/get.go:199\nsearch.(*Module).SearchGETHandler()\n\t/drone/src/internal/api/client/search/searchget.go:232\n()\n\t/drone/src/vendor/github.com/gin-gonic/gin/context.go:174\ngzip.(*gzipHandler).Handle()\n\t/drone/src/vendor/github.com/gin-contrib/gzip/handler.go:60\n" requestID=k27k5x4c04001xzaafj0 msg="recovered panic: runtime error: invalid memory address or nil pointer dereference"

[...]

PR #2563, opened 2024-01-22
[bugfix] Don't return Account or Status if new and dereferencing failed, other small fixes

tsmethurst:
# Description

>

This pull request fixes up some of our account and status dereferencing logic in the following ways:

- Don't return a near-zero Account or Status when it's new and a call to the remote instance returned a 4xx status code. Log + return an error instead. Previously, returning a partial account or status which hadn't yet been inserted properly in the database was causing all sorts of problems, since it would be put in the cache with zero values all over it. See https://github.com/superseriousbusiness/gotosocial/issues/2527 and https://github.com/superseriousbusiness/gotosocial/issues/2514.
- Derive published time for accounts using the `published` property. This should allow the actual remote creation time of the account to be stored, instead of just the time that the GtS instance first saw the account, which was misleading in that it was making some accounts look way newer than they actually are.
- Refactor some of the webfinger response parsing logic a bit to reuse existing regular expressions, add a new regular expression for the user web path (`/@username`).

## Checklist


internal/federation/dereferencing/account.go
```diff
@@ -581,12 +659,29 @@ func (d *Dereferencer) enrichAccount(
...
+	published := ap.GetPublished(apubAcc)
+
+	if account.IsNew() {
+		// Derive published/created time from
+		// apubAcc, fall back to FetchedAt value.
+		if published.IsZero() {
+			latestAcc.CreatedAt = published
+		} else {
+			latestAcc.CreatedAt = latestAcc.FetchedAt
+		}
```
daenney, line 682: Shouldn't this be the other way around? If `published.IsZero()`, then we don't want to set it to published but the fetch time?

daenney: Idem in the other block where we do this?

tsmethurst: Whoops, thanks!

internal/federation/dereferencing/finger.go
```diff
@@ -21,53 +21,113 @@ import (
...
-		return
+		err = gtserror.Newf("error webfingering %s: %w", target, err)
+		return "", nil, err
 	}
 
-	resp := &apimodel.WellKnownResponse{}
-	if err = json.Unmarshal(b, resp); err != nil {
-		err = fmt.Errorf("fingerRemoteAccount: could not unmarshal server response as WebfingerAccountResponse while dereferencing @%s@%s: %s", targetUsername, targetHost, err)
-		return
+	var resp = new(apimodel.WellKnownResponse)
```
daenney, line 62: I would one of `var resp *apimodel.WellKnownResponse`, `resp := new(apimodel.WellKnownResponse)` or the previous initialisation? But `var name = new()` is a bit odd and not something we do anywhere else I believe.

internal/federation/dereferencing/finger.go
```diff
@@ -21,53 +21,113 @@ import (
...
-	// look through the links for the first one that matches what we need
-	for _, l := range resp.Links {
-		if l.Rel == "self" && (strings.EqualFold(l.Type, "application/activity+json") || strings.EqualFold(l.Type, "application/ld+json; profile=\"https://www.w3.org/ns/activitystreams\"")) {
-			if uri, thiserr := url.Parse(l.Href); thiserr == nil && (uri.Scheme == "http" || uri.Scheme == "https") {
-				// found it!
-				accountURI = uri
-				return
-			}
+	if accountDomain == "" {
+
```
daenney, line 85: This is an empty branch

tsmethurst: Fixed! Silly. https://github.com/superseriousbusiness/gotosocial/pull/2563/commits/66f5ecad7c0b631efc286b55f3f83177afbf553e

internal/federation/dereferencing/finger.go
tsmethurst: Ah you're right, I'll change it.

internal/federation/dereferencing/finger.go
```diff
@@ -21,53 +21,109 @@ import (
...
 
 	if resp.Subject == "" {
-		err = fmt.Errorf("fingerRemoteAccount: no subject found in webfinger response %s", string(b))
-		return
+		err = gtserror.Newf("no subject found in response for %s", target)
+		return "", nil, err
 	}
 
-	_, accountDomain, err = util.ExtractWebfingerParts(resp.Subject)
+	_, accountDomain, err := util.ExtractWebfingerParts(resp.Subject)
```
daenney, line 78: While we're here, maybe change this to `if _, accountDomain, err := ... ; err != nil`? We do a lot of "accidental" shadowing/overwriting of `err` and it makes things a bit hard to read at times.

daenney: Actually nvm, we need to keep `accountDomain` around.

internal/federation/dereferencing/finger.go
tsmethurst: There, made it just not use a pointer since we don't really need one outside of parsing the JSON. https://github.com/superseriousbusiness/gotosocial/pull/2563/commits/ac6054267dd694ee47abefdd0444768a0e874883

internal/federation/dereferencing/finger.go
```diff
@@ -21,53 +21,109 @@ import (
...
+			)
+
+			// Funky URL, ignore.
+			continue
+		}
+
+		if uri.Scheme != "http" && uri.Scheme != "https" {
+			log.Warnf(ctx,
+				"invalid href for ActivityPub self link %s for %s: %v",
+				link.Href, target, errors.New("schema must be http or https"),
```
NyaaaWhatsUpDoc, line 116: there's no need for this to be an error, just include it in the log string :p

internal/federation/dereferencing/status.go
```diff
@@ -294,10 +295,28 @@ func (d *Dereferencer) enrichStatusSafely(
 		apubStatus,
 	)
 
-	if gtserror.StatusCode(err) >= 400 {
-		// Update fetch-at to slow re-attempts.
+	if code := gtserror.StatusCode(err); code >= 400 {
+		err = fmt.Errorf("status enrichment failed with status code %d: %w", code, err)
```
NyaaaWhatsUpDoc, line 299: do we need to further wrap the error here? i *think* the error returned should still be a status code error just now we'll end up repeating it in the same error string and making it longer. (same goes for here and the similar handling of accounts in the file above)

daenney: I'm not sure about returning a status code error here. We got a 4xx from the other side, but I don't think we want that status code to bubble up further.

internal/federation/dereferencing/account.go
```diff
@@ -394,10 +396,28 @@ func (d *Dereferencer) enrichAccountSafely(
...
+
+		// We had this account stored already
+		// before this enrichment attempt.
+		//
+		// Update fetched_at to slow re-attempts
+		// but don't return early. We can still
+		// return the model we had stored already.
 		account.FetchedAt = time.Now()
-		_ = d.state.DB.UpdateAccount(ctx, account, "fetched_at")
+		if updateErr := d.state.DB.UpdateAccount(ctx, account, "fetched_at"); updateErr != nil {
```
NyaaaWhatsUpDoc, line 418: this is in a separate clause here so you can overshadow the error here and do `if err := d.state.DB.UpdateAccount()` without touching the outer clause's error, the separate named error values makes it seem like we're going to wrapping / combining the errors imo

internal/federation/dereferencing/account.go
```diff
@@ -394,10 +396,28 @@ func (d *Dereferencer) enrichAccountSafely(
 		accountable,
 	)
 
-	if gtserror.StatusCode(err) >= 400 {
-		// Update fetch-at to slow re-attempts.
+	if code := gtserror.StatusCode(err); code >= 400 {
+		err = fmt.Errorf("account enrichment failed with status code %d: %w", code, err)
+		log.Info(ctx, err)
+
+		if account.IsNew() {
```
NyaaaWhatsUpDoc, line 403: honestly it probably makes more sense here to instead change the check to be `if gtserror.StatusCode(err) >= 400 && !account.IsNew()` and skip doing this extra logic in here. this will likely result in double logging happening as the error w/ code contained within this clause still gets returned and logged by either GetAccountByUR() or the callers of that function. Though the logging of the UpdateAccount() error below definitely is useful

NyaaaWhatsUpDoc: same goes re: the relevant enrichStatus function below

internal/federation/dereferencing/status.go
NyaaaWhatsUpDoc: (this comment is only worth addressing once you've read the comments on the enrichAccount function for this similar section btw)

internal/federation/dereferencing/account.go
```diff
@@ -440,32 +460,44 @@ func (d *Dereferencer) enrichAccount(
...
 		// A username was provided so we can attempt a webfinger, this ensures up-to-date accountdomain info.
 		accDomain, accURI, err := d.fingerRemoteAccount(ctx, tsport, account.Username, account.Domain)
-		if err != nil {
-			if account.URI == "" {
-				// this is a new account (to us) with username@domain
-				// but failed webfinger, nothing more we can do.
-				err := gtserror.Newf("error webfingering account: %w", err)
-				return nil, nil, gtserror.SetUnretrievable(err)
-			}
+		switch {
```
NyaaaWhatsUpDoc, line 463: is there any logic change happening here? i'm finding this rewritten switch case a bit harder to read and i genuinely can't tell

internal/federation/dereferencing/status.go
NyaaaWhatsUpDoc: > I'm not sure about returning a status code error here. We got a 4xx from the other side, but I don't think we want that status code to bubble up further.

the status code will probably be at the bottom layer of the error chain, so it won't directly be an http error at this level, and given the callers of these functions it won't ever be received by the user

internal/federation/dereferencing/account.go
tsmethurst: No not really! I just found the switch statement easier to read, myself. I can change it back if you prefer though :) It's not a strong preference on my part.

internal/federation/dereferencing/finger.go
tsmethurst: :facepalm:

internal/federation/dereferencing/account.go
```diff
@@ -581,12 +659,29 @@ func (d *Dereferencer) enrichAccount(
...
+	//
+	// This allows us to set CreatedAt properly for
+	// a remote account, rather than assuming it was
+	// created at the moment we first saw it.
+	//
+	// DON'T set CreatedAt on the latestAcc outside
+	// of the subsequent IsNew() check, since that
+	// uses CreatedAt to check if we already had an
+	// account in our database.
+	published := ap.GetPublished(apubAcc)
```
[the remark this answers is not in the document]
tsmethurst: Ah sure, then it's more consistent with what we do with statuses. I'll change it.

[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2563/commits/67a75cbd3d01558dede8207a4eed45ca40186688

internal/federation/dereferencing/finger.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2563/commits/99ae3c3a4e7a8ae145228c725e7e5b596e7dc3ef

internal/federation/dereferencing/account.go
tsmethurst: The issue is that we don't currently log it in the search functions (which is why this bug was so hard to catch), we just do this:

```go
	// Domain and username were both set.
	// Caller is likely trying to search for an exact
	// match, from either a remote instance or local.
	foundAccount, err := p.accountByUsernameDomain(
		ctx,
		requestingAccount,
		username,
		domain,
		resolve,
	)
	if err != nil {
		// Check for semi-expected error types.
		// On one of these, we can continue.
		if !gtserror.IsUnretrievable(err) && !gtserror.IsWrongType(err) {
			err = gtserror.Newf("error looking up @%s@%s as account: %w", username, domain, err)
			return gtserror.NewErrorInternalError(err)
		}
	} else {
		appendAccount(foundAccount)
	}
```

So either we log it here, or we update the search functions to always log the error and then maybe return it too if it's not one of the expected error types. Personally I'd rather log it here since it means less code duplication in the calling functions, if that makes sense.

internal/federation/dereferencing/account.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2563/commits/bf3dbc86ea2bd92295c88e1fd3fbf60ada38d6ad

internal/federation/dereferencing/status.go
tsmethurst: The error here will look something like:

`status enrichment failed with status code 4xx: error deferencing https://example.org/users/someone/some_status_id: GET request to https://example.org/users/someone/some_status_id failed: status="4xx" body="aaaaaa"`

So indeed we probably could drop the "with status code 4xx" part.

tsmethurst: There, truncated the wrapping a bit: https://github.com/superseriousbusiness/gotosocial/pull/2563/commits/3c66c30ffc161943b9ea31a69caf545bdd01fa0e

The URL will still get repeated in case the code is 4xx, but it won't necessarily be repeated if dereferencing goes wrong for some other reason, so I think it's fine.

daenney (approved):

internal/federation/dereferencing/status.go
NyaaaWhatsUpDoc: do we need to wrap it again though is my point? it's not telling us anything new since we know from the dereferencing message it's during enrichment and that also solves the problem of repeated url :p

tsmethurst: Ah right fair point yeah. I'll drop the wrapping and just log the error as-is then :)

PR #2576, opened 2024-01-27
Fix EmptyJSONObject/EmptyJSONArray

VyrCossont:
These are meant to be the bytes representing an empty object and array in JSON: `{}` and `[]`. They are actually the strings `"{}"` and `"[]"`. This causes clients expecting an object or array to not be able to parse the response.

## Checklist

### DIFF
--- a/internal/api/util/response.go
+++ b/internal/api/util/response.go
@@ -51,8 +51,8 @@ var (
 	ErrorRateLimited = mustJSON(map[string]string{
 		"error": "rate limit reached",
 	})
-	EmptyJSONObject = []byte{'{', '}'}
-	EmptyJSONArray  = []byte{'[', ']'}
+	EmptyJSONObject = json.RawMessage(`{}`)
+	EmptyJSONArray  = json.RawMessage(`[]`)
 
 	// write buffer pool.
 	bufPool sync.Pool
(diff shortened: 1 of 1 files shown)


internal/api/util/response.go
```diff
@@ -51,8 +51,8 @@ var (
 	ErrorRateLimited = mustJSON(map[string]string{
 		"error": "rate limit reached",
 	})
-	EmptyJSONObject = mustJSON("{}")
-	EmptyJSONArray  = mustJSON("[]")
+	EmptyJSONObject = []byte{'{', '}'}
+	EmptyJSONArray  = []byte{'[', ']'}
```
daenney, line 55: So the reason this went wrong is because `mustJSON` simply calls `json.Marshall`. Calling `json.Marshall` on a string simply results in the string being returned, so `"{}"`, not the intended empty JSON object `{}`.

This fix works, but I can easily see this regressing since it's not entirely obvious from the code why we're building a byte slice like that. Since we're already importing `econding/json` in there anyway, I would suggest to do this instead:

```go
EmptyJSONObject = json.RawMessage(`{}`)
EmptyJSONArray  = json.RawMessage(`(]`)
```

That's a bit more obvious

daenney: The alternative is something like `EmptyJSONObject = mustJSON(map[byte]struct{}{})` but that's hideous and imho much less comprehensible than `json.RawMessage()`.

VyrCossont: That hideousness is why I went with the byte array in the first place, but I agree that using the `json.RawMessage` type conveys the intent clearly. Let's use that.

PR #2579, opened 2024-01-28
Improve context descendant sorting

### LINKED ISSUE #2479: [feature] Better conversation thread sorting
Currently when we return thread context (aka conversation) from the API or on the web frontend, our sorting is a little strange because it interleaves replies from other accounts into the "main" thread of the conversation, interrupting the flow:

- Post by account A (start of thread) 1/4
- Self reply by account A 2/4
- Reply from account B to 2/4
- Self reply by account A 3/4
- Reply from account C to 3/4
- etc etc
- Self reply by account A (end of thread) 4/4

We should try to sort conversation threads a little nicer so that posts are more readably arranged / divided into main and comments sections:

- Post by account A (start of thread) 1/4
- Self reply by account A 2/4
- Self reply by account A 3/4
- Self reply by account A (end of thread) 4/4
- **Comments section**
- Reply from account B to 2/4
- Reply from account C to 3/4
- etc etc

VyrCossont:
# Description

Topologically sort replies, moving self-replies to top of list. The topo sort is decent at producing runs of reply chains in the output, although it's probably possible to improve on that.

Fixes #2479

## Checklist

- [x] I/we have run `go fmt ./...`
- [x] and `golangci-lint run`. *(for some reason this is incredibly slow on my Mac, like 10 minutes at 800% CPU and it hasn't finished)*

### DIFF
--- a/internal/processing/status/get.go
+++ b/internal/processing/status/get.go
@@ -109,8 +109,7 @@ func (p *Processor) contextGet(
 		}
 	}
 
-	// Topologically sort descendants.
-	topoSort(descendants, targetStatus.AccountID)
+	TopoSort(descendants, targetStatus.AccountID)
 
 	context := &apimodel.Context{
 		Ancestors:   make([]apimodel.Status, 0, len(ancestors)),
@@ -126,10 +125,10 @@ func (p *Processor) contextGet(
 	return context, nil
 }
 
-// Sort statuses topologically. Break ties by sorting by ID.
+// TopoSort sorts statuses topologically, by self-reply, and by ID.
 // Can handle cycles but the output order will be arbitrary.
 // (But if there are cycles, something went wrong upstream.)
-func topoSort(apiStatuses []*apimodel.Status, targetAccountID string) {
+func TopoSort(apiStatuses []*apimodel.Status, targetAccountID string) {
 	if len(apiStatuses) == 0 {
 		return
 	}
--- a/internal/processing/status/get.go
+++ b/internal/processing/status/get.go
@@ -111,6 +111,7 @@ func (p *Processor) contextGet(
 
 	TopoSort(descendants, targetStatus.AccountID)
 
+	//goland:noinspection GoImportUsedAsName
 	context := &apimodel.Context{
 		Ancestors:   make([]apimodel.Status, 0, len(ancestors)),
 		Descendants: make([]apimodel.Status, 0, len(descendants)),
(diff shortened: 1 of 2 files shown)


internal/processing/status/get.go
```diff
@@ -74,46 +75,139 @@ func (p *Processor) contextGet(
...
 			if err == nil {
-				context.Descendants = append(context.Descendants, *apiStatus)
+				descendants = append(descendants, apiStatus)
 			}
 		}
 	}
 
+	TopoSort(descendants, targetStatus.AccountID)
+
+	context := &apimodel.Context{
```
tsmethurst, line 114: You can make this a little neater here (and avoid reslicing ancestors + descendants):

```suggestion
	context := &apimodel.Context{
		Ancestors:   make([]apimodel.Status, len(ancestors)),
		Descendants: make([]apimodel.Status, len(descendants)),
	}
	for i, ancestor := range ancestors {
		context.Ancestors[i] = *ancestor
	}
	for i, descendant := range descendants {
		context.Descendants[i] = *descendant
	}
```

internal/processing/status/get.go
```diff
@@ -74,46 +75,139 @@ func (p *Processor) contextGet(
...
+		Ancestors:   make([]apimodel.Status, 0, len(ancestors)),
+		Descendants: make([]apimodel.Status, 0, len(descendants)),
+	}
+	for _, ancestor := range ancestors {
+		context.Ancestors = append(context.Ancestors, *ancestor)
+	}
+	for _, descendant := range descendants {
+		context.Descendants = append(context.Descendants, *descendant)
+	}
```
tsmethurst, line 123: You can make this a bit neater imo, since you know in advance how long ancestor + descendant slices should be, and you're not filtering out any entries:

```suggestion
	context := &apimodel.Context{
		Ancestors:   make([]apimodel.Status, len(ancestors)),
		Descendants: make([]apimodel.Status, len(descendants)),
	}
	for i, ancestor := range ancestors {
		context.Ancestors[i] = *ancestor
	}
	for i, descendant := range descendants {
		context.Descendants[i] = *descendant
	}
```

internal/processing/status/get.go
```diff
@@ -74,46 +75,139 @@ func (p *Processor) contextGet(
...
+// TopoSort sorts statuses topologically, by self-reply, and by ID.
+// Can handle cycles but the output order will be arbitrary.
+// (But if there are cycles, something went wrong upstream.)
+func TopoSort(apiStatuses []*apimodel.Status, targetAccountID string) {
+	if len(apiStatuses) == 0 {
+		return
+	}
+
+	// Map of status IDs to statuses.
+	lookup := map[string]*apimodel.Status{}
```
tsmethurst, line 137: Since you know the size of this map already you can do:

```suggestion
	lookup := make(map[string]*apimodel.Status, len(apiStatuses))
```

tsmethurst:
Looks neat to me :) @NyaaaWhatsUpDoc do you have any comments on the sorting?

internal/processing/status/get.go
NyaaaWhatsUpDoc: i'd also do this for the tree map below, using a worst-case scenario value of len(apiStatuses). it's better to overestimate an allocation once by a little than have the runtime have to repeatedly resize a map from zero

internal/processing/status/get.go
tsmethurst: Ah well fair enough, feel free to disregard this comment then.

PR #2590, opened 2024-01-30
[bugfix] fix possible infinite loops in media / emoji cleanup

NyaaaWhatsUpDoc:

### DIFF
--- a/internal/cleaner/emoji.go
+++ b/internal/cleaner/emoji.go
@@ -27,6 +27,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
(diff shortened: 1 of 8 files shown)


internal/cleaner/emoji.go
```diff
@@ -224,23 +240,30 @@ func (e *Emoji) PruneUnused(ctx context.Context) (int, error) {
 func (e *Emoji) FixCacheStates(ctx context.Context) (int, error) {
```
daenney, line 240: A good part of this function looks like a dupe of the `PruneUnused` body. Would it make sense to extract that and have both functions call into it instead?

daenney: The same seems to happen between `PruneUnused` and `FixCacheStates` on `Media`.

daenney (approved):
I had one comment about what used like a bunch of code duplication between a pair of functions. But that was already there before this PR, so we don't have to address it here either.

It's nice to see the page stuff make it through in more places and solve bugs.

Issue #894, opened 2022-10-06
[investigation] HTTP signature query params issues


tsmethurst:
While GtS's implementation of http signatures is technically correct (https://github.com/superseriousbusiness/gotosocial/issues/107#issuecomment-1188257189), Mastodon and many other implementations do not include query parameters when generating http signatures. This leads to some incompatibilities with GoToSocial when it comes to fetching status replies.

We should investigate how to alleviate this incompatibility from GtS's side, while still adhering correctly to the rfc.

My initial idea about this would be to maintain current behavior for first attempts, BUT: if an incoming request with query parameters fails http sig authentication, retry it without including the query params in the reconstructed signature, and if an outgoing request with query params gets a 401 returned, retry it without including query params in the outgoing signature.

benjojo:
I am using https://humungus.tedunangst.com/r/honk and it appears to suffer from this issue as well.

Might I suggest (since the rest of the ecosystem appears to do the Mastodon(?) way) that you validate "both" methods (the technically correct sig, and the Mastodon style one)?

I am unable to follow some accounts due to this and it appears the only instances that have this issue are GTS ones, and it's quite annoying to miss out on some accounts

tsmethurst:
Related: https://github.com/mastodon/mastodon/issues/18638

NyaaaWhatsUpDoc:
Just noting here, i will be handling this task once we've finalized funding with NLNet for this proposed task :)

tsmethurst:
Oh great, I forgot we'd included that :)

milas:
Hi! I am a happy GoToSocial user, and I decided to give fixing this a shot a bit back, and have been running a fork on my instance with the changes for a bit.

I followed the approach outlined in the original issue:
> My initial idea about this would be to maintain current behavior for first attempts, BUT: if an incoming request with query parameters fails http sig authentication, retry it without including the query params in the reconstructed signature, and if an outgoing request with query params gets a 401 returned, retry it without including query params in the outgoing signature.

The changes on the GtS side are fairly small, the `go-fed/httpsig` changes were a little trickier when keeping semver compatibility:
 * go-fed/httpsig#31

Apologies for not posting here sooner, and I won't be at all offended if you opt to go with a different implementation!

For reference, here's more or less what the GtS changes look like when using my `go-fed/httpsig` fork:

```diff
commit 3b265efb1985e09e9a9f80eb09f7a41a1e5f30cb
[...]

PR #2591, opened 2024-01-31
[feature] Try HTTP signature validation with and without query params for incoming requests

tsmethurst:
# Description

>

This pull request addresses the first part of https://github.com/superseriousbusiness/gotosocial/issues/894 by implementing code to fall back to excluding query params when validating incoming HTTP signatures. The "default" first-attempt behavior (including query params when reassembling signature strings) remains the same.

## Checklist


daenney (approved):

daenney:
The one thing maybe worth considering here is caching which of the two verify schemes worked for a domain, so we don't always have to try both for a particular domain if it happens to want the second scheme.

NyaaaWhatsUpDoc:
> The one thing maybe worth considering here is caching which of the two verify schemes worked for a domain, so we don't always have to try both for a particular domain if it happens to want the second scheme.

have squerged for now as i think compared to the other side of attempting both times of httpsigs this is less of a priority, at least re: tobi being able to get on with the other Move work. though definitely a low-hanging fruit performance gain imo

PR #2594, opened 2024-02-01
[feature] Filters v1

### LINKED ISSUE #1472: [feature] filter / mute words and hashtags

When I'm trying to add filters on hashtags or words in Tusky, I'm getting an error. I want to block posts about certain topics, this isn't possible so far (unless this problem is tusky exclusive).
I see this as an important feature, and I guess that this isn't implemented so far as I don't see it on the roadmap or in any other closed or open issues for "filter"

I am able to filter posts.

None 

### Additional context.

_No response_

VyrCossont:
# Description

This pull request implements a client-side filter API compatible with the Mastodon filter v1 API.

It does not implement the Mastodon filter v2 API, although it does add the DB and cache plumbing to trivially implement that API later. It also doesn't implement the server-side filtering part of the v1 API: the `irreversible` filter option will return an error message if set to `true`.

Fixes #1472

## Checklist


VyrCossont:
There are two cache-related TODOs: one is for the multi-invalidate cache method that doesn't exist yet, and one is for cache memory ratios, which I don't understand.

The Swagger tool GtS uses seems to have a lot of bugs and omissions, so I'm not 100% on the Swagger parts.

VyrCossont:
Using the CI for `golangci-lint run` because it's 100% broken on my dev box. Sorry!

internal/gtsmodel/filter.go
```diff
@@ -0,0 +1,71 @@
...
+	ContextAccount       bool             `bun:",notnull"`                                                    // Apply filter when viewing an account profile.
+}
+
+// FilterEntry is the fields common across FilterKeyword and FilterStatus.
+type FilterEntry struct {
+	ID        string    `bun:"type:CHAR(26),pk,nullzero,notnull,unique"`                                     // id of this item in the database
+	CreatedAt time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"`                  // when was item created
+	UpdatedAt time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"`                  // when was item last updated
+	AccountID string    `bun:"type:CHAR(26),notnull,nullzero"`                                               // ID of the local account that created the filter keyword.
+	FilterID  string    `bun:"type:CHAR(26),notnull,nullzero,unique:filter_keywords_filter_id_keyword_uniq"` // ID of the filter that this keyword belongs to.
```
VyrCossont, line 46: After factoring out `FilterEntry` from `FilterKeyword` and `FilterStatus`, these uniqueness constraints won't work. Can we add them in the migration?

internal/api/model/filterv1.go
```diff
@@ -0,0 +1,100 @@
...
+	Irreversible bool `json:"irreversible"`
+	// When the filter should no longer be applied. Not present if the filter does not expire.
+	//
+	// Example: 2024-02-01T02:57:49Z
+	ExpiresAt string `json:"expires_at,omitempty"`
+}
+
+// FilterCreateUpdateRequestV1 captures params for creating or updating a v1 filter.
+//
+// swagger:model filterCreateUpdateRequestV1
```
VyrCossont, line 70: This should use https://goswagger.io/use/spec/params.html instead.

JJGadgets:
Please do let me know if I'm being a nuisance, but I really want this feature badly: is there any chance of this making it into 0.14?

Unfollowing accounts just because one of their views don't align or because they're reporting on news I'm not interested in, isn't a solution for me.

Also, more of a curiosity question, why only implement v1 (client-side) initially and not go straight to implementing v2 (server-side)?

VyrCossont:
Client-side filtering doesn't need any server work past filter storage, so it's low risk, and many Mastodon API clients already support it (including GtS favorites Semaphore and Feditext), so it's already useful.

There are some newer clients that only support server-side filtering, such as Phanpy, so I designed this patch's storage layer to make it easy to add in a followup.

VyrCossont:
@JJGadgets aside from the missing index, this should work fine on 0.14 if you want to apply it yourself. I've been running the 0.13 prototype on my own instance for weeks.

JJGadgets:
> @JJGadgets aside from the missing index, this should work fine on 0.14 if you want to apply it yourself. I've been running the 0.13 prototype on my own instance for weeks.

@VyrCossont should be safe to just build directly from your branch... right? (Those failed checks kept me on edge haha)

I'm already building my own distroless container images for GoToSocial and other apps anyway, so it shouldn't be too hard.

(I build my own images mainly for the fun of it, and because my environment is Kubernetes on Talos Linux which I use because of its minimal and secure distro design for running Kubernetes and nothing more on the host level, but to be clear, nothing's inherently wrong with the upstream image, distroless isn't for everyone.)

JJGadgets:
Also, do you know if v1 filters are compatible with Tusky and Elk?

I mainly use Mona (which isn't compatible with either kinds of filters), Tusky and Phanpy depending on which OS I'm on, but I'm okay with switching back from Phanpy to Elk if it means I can get filters.

VyrCossont:
@JJGadgets not sure about Tusky or Elk, sorry. When you find out, please let me know. And Mona doesn't have filter support at *all*? Yikes. Try Feditext.

VyrCossont:
Oops, that close was entirely accidental.

VyrCossont:
I take back my "this should work fine on 0.14"; there are some DB changes I need to rebase on top of first.

internal/api/model/filterv1.go
VyrCossont: Swagger/OpenAPI 2 seems like it *really* wasn't built to handle the sloppy mess that is Rails-compatible "you can send this as form params or JSON" REST APIs; maybe OpenAPI 3 does better, but for now we're just going to do what the rest of the GtS codebase does: the currently prevailing approach is to use `swagger:ignore` on the objects representing the parsed versions of form parameters, and document the form parameters.

VyrCossont:
The unused code warnings are a false positive, possibly due to dominikh/go-tools#1440, and I've fixed the issue with GoLand custom folding region comments in go-critic/go-critic#1408. For now, I've added very specific exclusions to our linter config.

JJGadgets:
> @JJGadgets not sure about Tusky or Elk, sorry. When you find out, please let me know. And Mona doesn't have filter support at *all*? Yikes. Try Feditext.

Mona has filters but they don't talk to the instance at all, they're synced over iCloud instead and thus only work on Mona. Which I currently use because GtS doesn't support filters (before this PR), but that means my filtered timeline only shows on my Apple devices.

I'm in the Feditext beta, yeah. I generally prefer Mona though.

Will find out about Tusky and Elk and report back when I do.

tsmethurst:
> The unused code warnings are a false positive, possibly due to [dominikh/go-tools#1440](https://github.com/dominikh/go-tools/issues/1440), and I've fixed the issue with GoLand custom folding region comments in [go-critic/go-critic#1408](https://github.com/go-critic/go-critic/pull/1408). For now, I've added very specific exclusions to our linter config.
> 
> I have no idea why the docs build failed, but I haven't changed anything in the docs.

I'll check it out :)

test/envparsing.sh
```diff
@@ -31,6 +31,9 @@ EXPECT=$(cat << "EOF"
         "boost-of-ids-mem-ratio": 3,
         "emoji-category-mem-ratio": 0.1,
         "emoji-mem-ratio": 3,
+        "filter-keyword-mem-ratio": 1,
```
tsmethurst, line 34: These ratios are probably a little higher than they need to be, I'd say 0.5 or something for each of them would probably be OK. Happy to defer to @NyaaaWhatsUpDoc though since she knows more about how these values end up being used.

internal/api/client/filters/v1/filterdelete.go
```diff
@@ -0,0 +1,92 @@
...
+//	-
+//		name: id
+//		type: string
+//		description: ID of the list
+//		in: path
+//		required: true
+//
+//	security:
+//	- OAuth2 Bearer:
+//		- write:lists
```
tsmethurst, line 51: Think you mean write:filters here.

internal/api/client/filters/v1/filterdelete.go
```diff
@@ -0,0 +1,92 @@
...
+		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	if _, err := apiutil.NegotiateAccept(c, apiutil.JSONAcceptHeaders...); err != nil {
+		apiutil.ErrorHandler(c, gtserror.NewErrorNotAcceptable(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	id := c.Param(IDKey)
```
tsmethurst, line 78: For parsing IDs and stuff like that, you can use the utility functions in apiutil like this:

```go
	id, errWithCode := apiutil.ParseID(c.Param(apiutil.IDKey))
	if errWithCode != nil {
		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
		return
	}
```

It doesn't really lead to fewer lines of code per se, but it does mean the errors we return are a bit more consistently worded.

Just note that `apiutil.IDKey` is "id" and not "ID".

internal/api/client/filters/v1/filterput.go
```diff
@@ -0,0 +1,160 @@
...
+//		name: whole_word
+//		in: formData
+//		description: Should the filter consider word boundaries?
+//		type: boolean
+//		default: false
+//		example: true
+//
+//	security:
+//	- OAuth2 Bearer:
+//		- read:filters
```
tsmethurst, line 103: Should be "write" again.

internal/api/model/filterv1.go
```diff
@@ -0,0 +1,105 @@
...
+	WholeWord bool `json:"whole_word"`
+	// Should matching entities be removed from the user's timelines/views, instead of hidden?
+	//
+	// Required: true
+	// Example: false
+	Irreversible bool `json:"irreversible"`
+	// When the filter should no longer be applied. Not present if the filter does not expire.
+	//
+	// Example: 2024-02-01T02:57:49Z
+	ExpiresAt string `json:"expires_at,omitempty"`
```
tsmethurst, line 70: I think this needs to a *string and be null if it doesn't expire, instead of omitted, see: https://docs.joinmastodon.org/entities/Filter/#expires_at

internal/api/model/filterv1.go
```diff
@@ -0,0 +1,105 @@
...
+//
+// swagger:model filterV1
+//
+// ---
+// tags:
+// - filters
+type FilterV1 struct {
+	// The ID of the filter in the database.
+	//
+	// Required: true
```
tsmethurst, line 37: Unless a filter is used as the body of a request somewhere, I don't think we need the `Required` tags on this model.

internal/db/filter.go
```diff
@@ -0,0 +1,105 @@
...
+package db
+
+import (
+	"context"
+
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+)
+
+type Filter interface {
+	//region Filter methods
```
tsmethurst, line 27: What do these region comments do? Or are they just for clarity while reading?

internal/gtsmodel/filter.go
```diff
@@ -0,0 +1,71 @@
...
+import "time"
+
+// Filter stores a filter created by a local account.
+type Filter struct {
+	ID                   string           `bun:"type:CHAR(26),pk,nullzero,notnull,unique"`                    // id of this item in the database
+	CreatedAt            time.Time        `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"` // when was item created
+	UpdatedAt            time.Time        `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"` // when was item last updated
+	ExpiresAt            time.Time        `bun:"type:timestamptz,nullzero"`                                   // Time filter should expire. If null, should not expire.
+	AccountID            string           `bun:"type:CHAR(26),notnull,nullzero"`                              // ID of the local account that created the filter.
+	Title                string           `bun:",nullzero,notnull,unique:filter_title_uniq"`                  // The name of the filter.
```
tsmethurst, line 29: If you're just making this one field unique on its own, without combining it with another field, you don't need to name the unique constraint; you can just tag it as `,unique` and let the db implementation name it for you.

internal/gtsmodel/filter.go
```diff
@@ -0,0 +1,71 @@
...
+	AccountID string    `bun:"type:CHAR(26),notnull,nullzero"`                              // ID of the local account that created the filter keyword.
+	FilterID  string    `bun:"type:CHAR(26),notnull,nullzero"`                              // ID of the filter that this keyword belongs to.
+	Filter    *Filter   `bun:"-"`                                                           // Filter corresponding to FilterID
+}
+
+// FilterKeyword stores a single keyword to filter statuses against.
+type FilterKeyword struct {
+	FilterEntry
+	Keyword   string `bun:",nullzero,notnull"` // The keyword or phrase to filter against.
+	WholeWord bool   `bun:",notnull"`          // Should the filter consider word boundaries?
```
tsmethurst, line 54: Probably good to set a default value in the db for this, eg:

``go
WholeWord *bool `bun:",nullzero,notnull,default:false"`
``

internal/gtsmodel/filter.go
```diff
@@ -0,0 +1,71 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package gtsmodel
+
+import "time"
+
+// Filter stores a filter created by a local account.
+type Filter struct {
```
tsmethurst, line 23: We've had issues with booleans in bun before, so the sort of best practice we arrived at was to be very explicit about bools in tags. As such, I'd recommend making bools in these filter models into ptrs, and adding default values for each of them (false in every case I think?):

``
*bool `bun:",nullzero,notnull,default:false"`
``

internal/db/bundb/filter.go
```diff
@@ -0,0 +1,655 @@
...
+}
+
+func (f *filterDB) GetFiltersForAccountID(ctx context.Context, accountID string) ([]*gtsmodel.Filter, error) {
+	// Fetch IDs of all filters owned by this account.
+	var filterIDs []string
+	if err := f.db.
+		NewSelect().
+		Model((*gtsmodel.Filter)(nil)).
+		Column("id").
+		Where("account_id = ?", accountID).
```
tsmethurst, line 72: Generally speaking in all these functions instead of doing `Where("whatever = ?", whatever)`, it's better to do `Where("? = ?", bun.Ident("whatever"), whatever)`, so that it can be properly escaped by bun depending on which database implementation is being used.

internal/db/bundb/filter.go
```diff
@@ -0,0 +1,655 @@
...
+	return nil
+}
+
+//endregion
+
+//region Filter entries
+
+// entry wraps a pointer to a model extending [gtsmodel.FilterEntry]
+// and provides operations common across those models.
+type entry[Model any, Self any] interface {
```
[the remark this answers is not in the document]
tsmethurst: By this I mean, I can see you're cooking up these generic types to avoid code repetition here, but it ends up being quite complex and difficult to reason about (for me, at least, I struggle to understand what's happening here). In my view it might be better to just put up with the repetition instead of trying to do these interface workarounds.

internal/db/bundb/migrations/20240126064004_add_filters.go
```diff
@@ -0,0 +1,105 @@
...
+				unique  bool
+				columns []string
+			}{
+				"filters": {
+					"filters_account_id_idx": {false, []string{"account_id"}},
+				},
+				"filter_keywords": {
+					"filter_keywords_account_id_idx":         {false, []string{"account_id"}},
+					"filter_keywords_filter_id_idx":          {false, []string{"filter_id"}},
+					"filter_keywords_filter_id_keyword_uniq": {true, []string{"filter_id", "keyword"}},
```
tsmethurst, line 68: For these unique indexes, it might just be better to set `unique` constraints on the appropriate fields of the models in `gtsmodel` and let the database create the constraints (and accompanying indexes) instead.

tsmethurst:
Not sure why the docs build failed, but I'll kick it :)

internal/api/model/filterv1.go
VyrCossont: These are only used in responses so we can drop that.

internal/db/filter.go
VyrCossont: They're semi-standard markers for custom code folding. So they're for clarity in reading, but also, in editors that are aware of region markers (I use IntelliJ/GoLand), you can do this:

this file with two regions folded up to a single line with their name, and one unfolded and showing code

internal/db/bundb/filter.go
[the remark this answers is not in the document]
VyrCossont: No problem. We should have adequate test coverage to de-generify it without breaking anything if that's the consensus. I personally find shorter code with less repetition easier to read and maintain, but this generic type has only two concrete instantiations so I'm fine doing it either way.

test/envparsing.sh
VyrCossont: Happy to defer as well.

internal/validate/formvalidation.go
```diff
@@ -44,6 +44,7 @@ const (
 	maximumProfileFieldLength     = 255
 	maximumProfileFields          = 6
 	maximumListTitleLength        = 200
+	maximumFilterKeywordLength    = 40
```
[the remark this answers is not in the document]
VyrCossont: Actually I think we can remove this entirely.

Most implementations (both client-side filters and probably our future server-side filters) are going to weld all your keywords together as alternate branches of a compiled regex for fast matching, so the real limit is how much the regex engine can take, which is as much down to how many filters you have as how long they are.

internal/api/model/filterv1.go
VyrCossont: You're correct. Sample output from Mastodon:

```json
[
  {
    "id": "145781",
    "phrase": "test",
    "context": [
      "home",
      "public"
    ],
    "whole_word": true,
    "expires_at": null,
    "irreversible": false
  }
]
```

internal/gtsmodel/filter.go
VyrCossont: That's super weird! I'll go make those changes.

internal/db/bundb/filter.go
VyrCossont: Have we had a ton of identifier collisions with builtins, or identifiers that have to be quoted for other reasons? I'll still make the changes but I'm curious because I've never seen this pattern before for *known* IDs. (Arbitrary IDs, totally understandable.)

daenney: I think it's nice because it avoids treating some identifiers are special or trusted. If you always do it this way, then you never have to think about it and you don't risk a situation in which you forget and boom SQL injection.

tsmethurst: > Have we had a ton of identifier collisions with builtins, or identifiers that have to be quoted for other reasons?

Not collisions really, but we have had cases where generated SQL queries fail for one db driver and work for another, due to different escaping methods, which has lead to lots of confusion when testing. Can't remember any specific examples off the top of my head, but it's easier to just avoid potential issues entirely by hinting to bun that they'd idents.

internal/db/bundb/migrations/20240126064004_add_filters.go
tsmethurst: Mmm, right, I see what you mean. The common nested struct is a little neater, but I think I'd rather we be more explicit and just inline those models and use the DB tags. Then it's also clearer by just looking at the models which fields have uniqueness constraints on them, without having to look at the db structure or look through the migrations. Would you mind making the change?

internal/db/bundb/filter.go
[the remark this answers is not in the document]
tsmethurst: Personally I always lean towards writing more verbose but more explicit code, and I think this is sort of facilitated by Go as well, given how difficult the language tends to make it to do the clever generics stuff you can do in other languages (as you write in a comment here, you had to do a workaround to get this working). So my preference would be to de-generify it, but I'll wait to see if kim has any opinions on the matter. Thanks for your patience with our curmudgeonly, deliberate style :P

tsmethurst:
This is looking almost ready to go :eyes: Just a few discussions left to resolve. Thanks for your continued hard work @VyrCossont !

internal/db/bundb/filter.go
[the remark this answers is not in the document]
VyrCossont: I've committed a de-generified version for comparison. If we decide we want to keep the generic version, we can revert a few specific commits, but now we can see both at once: commits up to f162d22 for generic and up to c69b420 for non-generic.

internal/api/client/filters/v1/validate.go
```diff
@@ -0,0 +1,68 @@
...
+
+	// Normalize filter expiry if necessary.
+	// If we parsed this as JSON, expires_in
+	// may be either a float64 or a string.
+	if ei := form.ExpiresInI; ei != nil {
+		switch e := ei.(type) {
+		case float64:
+			form.ExpiresIn = util.Ptr(int(e))
+
+		case string:
```
NyaaaWhatsUpDoc, line 54: should we be expecting this as a string? this feels wrong. or is this mastodon normalized behaviour we're stuck with :p

internal/db/bundb/filter.go
```diff
@@ -0,0 +1,345 @@
...
+	if gtscontext.Barebones(ctx) {
+		return filters, nil
+	}
+
+	// Populate the filters. Remove any that we can't populate from the return slice.
+	errs := gtserror.NewMultiError(len(filters))
+	filters = slices.DeleteFunc(filters, func(filter *gtsmodel.Filter) bool {
+		if err := f.populateFilter(ctx, filter); err != nil {
+			// %w is allowed here.
+			//goland:noinspection GoPrintFunctions
```
NyaaaWhatsUpDoc, line 120: is therre any way around needing to add this to every single `errs.Appendf()` with %w operator?

internal/api/client/filters/v1/validate.go
VyrCossont: This is the same set of types accepted by the poll expiration time mechanism, so I assume we've seen that happen at least once 😞

internal/db/bundb/filter.go
VyrCossont: There may be, but I haven't found it yet.

test/envparsing.sh
VyrCossont: Dropping these to 0.5 after discussion in Matrix.

nikclayton:
Hi - now this is merged is 0.15.0 the first release it will appear in? Or if you released a hypothetical 0.14.1 patch release would it appear there?

(Broadly, do you make patch branches from HEAD, or from the tag of the minor release you're patching?)

I'm asking so I can add it to https://github.com/pachli/pachli-android/blob/32a6d3465b19e90898e59cd21fa29f12941341d7/core/network/src/main/kotlin/app/pachli/core/network/Server.kt#L277

tsmethurst:
0.15.0

nikclayton:
Thanks @tsmethurst , https://github.com/pachli/pachli-android/pull/523 adds the support in Pachli. And thanks @VyrCossont for the implementation work (and keeping the Swagger definitions up to date, that's incredibly useful).

PR #2596, opened 2024-02-01
[feature/performance] sqlite pragma optimize on close

NyaaaWhatsUpDoc:
# Description

This replaces our database connection wrapping to be more concise, instead only wrapping the database driver implementations instead of the many bun types. In doing so we separate the retryBusy functionality to *only* exist for SQLite (as it should), and on SQLite connection close we attempt to run `PRAGMA analysis_limit=1000; PRAGMA optimize;`, though we purposely don't check error return as we don't know whether this will actually be a live connection or not.

In time maybe we can make the analysis limit configurable, but SQLite's analysis limit handling is already fairly intelligent so we may not even need to.

closes https://github.com/superseriousbusiness/gotosocial/issues/2425

TODO: test with postgres (currently my personal postgres testing script is broken)

## Checklist


internal/db/bundb/drivers.go
```diff
@@ -0,0 +1,267 @@
...
+		err = processSQLiteError(err)
+		return err
+	})
+	return
+}
+
+func (c *SQLiteConn) Close() error {
+	// see: https://www.sqlite.org/pragma.html#pragma_optimize
+	const onClose = "PRAGMA analysis_limit=1000; PRAGMA optimize;"
+	_, _ = c.conn.ExecContext(context.Background(), onClose, nil)
```
tsmethurst, line 189: Cheeky. Is it worth logging any error here?

Also I'd probably use a context with timeout value of whatever we use when shutting down GtS, normally (I think it's 1 min before it gets forcibly killed?). We probably don't wanna hang around because a connection is still open + optimizing.

NyaaaWhatsUpDoc: the analysis limit will ensure it doesn't run for too long. the underlying sqlite library doesn't pay a huge amount of attention to contexts other than an initial check (since blocking a Go channel against transpiled C code is understandably very difficult).

tsmethurst: Ahh okay, fair enough! How about the error? Fine to just leave it if you don't think it's worth logging something.

tsmethurst (approved):
Seems to work OK for Postgres too, feel free to squerge this once you've considered the "should we log errors on close" comment and decided what to do about it either way :)

internal/db/bundb/drivers.go
NyaaaWhatsUpDoc: i think we can probably leave it honestly. errors are likely to be database closing type errors and the like which might just raise false flags that people end up reporting. i'll leave it for now but in the future may consider logging it :)

PR #2600, opened 2024-02-02
[feature] serdes for moved/also_known_as

tsmethurst:
# Description

>

This pull request implements serdes for incoming and outgoing movedTo and alsoKnownAs properties via the activitypub API. By extension, this means movedTo and alsoKnownAs URIs will be set in the database for remote AP accounts, though we don't yet do anything with these properties aside from store them.

It also renames the `getIRIs` utility function to `extractIRIs` to reflect that it can extract IRIs from a slice that contains either/both IRIs and Types. It adds a specific `getIRIs` function back in to get IRIs from a slice that the caller knows must contain only IRIs and not Types.

~~Draft while I write up some documentation for this.~~ https://gotosocial--2600.org.readthedocs.build/en/2600/federation/federating_with_gotosocial/#actor-migration-aliasing

## Checklist

### DIFF
--- a/internal/db/bundb/account.go
+++ b/internal/db/bundb/account.go
@@ -279,7 +279,12 @@ func (a *accountDB) PopulateAccount(ctx context.Context, account *gtsmodel.Accou
 		}
 	}
 
-	if !account.AlsoKnownAsPopulated() {
+	// Only try to populate AlsoKnownAs for local accounts,
+	// since those are the only accounts to which it's relevant.
+	//
+	// AKA from remotes might have loads of random-ass values
+	// set here, and we don't want to do lots of failing DB calls.
+	if account.IsLocal() && !account.AlsoKnownAsPopulated() {
 		// Account alsoKnownAs accounts are
 		// out-of-date with URIs, repopulate.
 		alsoKnownAs := make([]*gtsmodel.Account, 0)
--- a/internal/typeutils/astointernal.go
+++ b/internal/typeutils/astointernal.go
@@ -206,7 +206,15 @@ func (c *Converter) ASRepresentationToAccount(ctx context.Context, accountable a
 	}
 
 	alsoKnownAsURIs := ap.GetAlsoKnownAs(accountable)
-	for _, uri := range alsoKnownAsURIs {
+	for i, uri := range alsoKnownAsURIs {
+		// Don't store more than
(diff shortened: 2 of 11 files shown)


internal/db/bundb/account.go
```diff
@@ -279,7 +279,12 @@ func (a *accountDB) PopulateAccount(ctx context.Context, account *gtsmodel.Accou
 		}
 	}
 
-	if !account.AlsoKnownAsPopulated() {
+	// Only try to populate AlsoKnownAs for local accounts,
+	// since those are the only accounts to which it's relevant.
+	//
+	// AKA from remotes might have loads of random-ass values
+	// set here, and we don't want to do lots of failing DB calls.
+	if account.IsLocal() && !account.AlsoKnownAsPopulated() {
```
tsmethurst, line 287: It's possible for remotes to set basically whatever they want in `alsoKnownAs` and we don't try dereferencing these URIs when receiving them, or validate them when inserting them into the DB.

As such, if we try to populate remote AKAs here we might end up trying to select things that aren't accounts we have stored (deleted accounts, things that aren't actually accounts, etc). This will cause calls to PopulateAccount to fail, which can have annoying knock-on effects elsewhere. Easier to only try populating this for local accounts where we do actually validate and deref the alsoKnownAs URI before storing it :)

internal/typeutils/astointernal.go
```diff
@@ -198,7 +198,25 @@ func (c *Converter) ASRepresentationToAccount(ctx context.Context, accountable a
...
+		acct.MovedToURI = movedToURI.String()
+	}
+
+	alsoKnownAsURIs := ap.GetAlsoKnownAs(accountable)
+	for i, uri := range alsoKnownAsURIs {
+		// Don't store more than
+		// 20 AKA URIs for remotes,
+		// to prevent people playing
+		// silly buggers.
+		if i >= 20 {
```
tsmethurst, line 214: This isn't in response to any particular vulnerability, but it seems smart to limit this to some high but still feasible value. We don't want someone trying to store 100s of URIs here for some weird purpose.

internal/db/bundb/account.go
NyaaaWhatsUpDoc: absolutely makes sense, good idea!

PR #2611, opened 2024-02-06
Make GtS reported version SemVer-compatible

### LINKED ISSUE #1953: [feature] SemVer version in /api/vX/instance and /nodeinfo/2.0

VyrCossont:
# Description

This pull request makes GtS's version string compatible with SemVer by concatenating the VCS metadata with the version using a plus sign (SemVer's build tag separator) instead of a space. It relies on `main.Version` being set to a SemVer-compatible version, which should be the case for release and most development builds.

Fixes #1953

## Checklist

### DIFF
--- a/cmd/gotosocial/main.go
+++ b/cmd/gotosocial/main.go
@@ -77,6 +77,7 @@ func main() {
 }
 
 // version will build a version string from binary's stored build information.
+// It is SemVer-compatible so long as Version is SemVer-compatible.
 func version() string {
 	// Read build information from binary
 	build, ok := godebug.ReadBuildInfo()
@@ -115,5 +116,5 @@ func version() string {
 		}
 	}
 
-	return strings.Join(info, " ")
+	return strings.Join(info, "+")
 }


tsmethurst:
One character change, nice! Looks fine to me, and I can't think of any negatives for doing things this way. @NyaaaWhatsUpDoc any thots?

PR #2614, opened 2024-02-07
[chore] Allow (internal) callers to set their own freshness window for Accounts + Statuses

tsmethurst:
# Description

>

This pull request implements super-duper force refresh of accounts and statuses, by allowing callers to the dereferencing functions to specify what they consider to be "fresh enough" for their purposes, instead of a simple bool. 

This will be required for `Move`, which necessitates that we have a very up to date version of the target and origin accounts of a Move. I don't wanna lower the account/status force deref cooldown of 5 mins globally, so passing in a custom duration to the refresh functions (but providing consts for callers to use) seemed like a good way of doing things.

## Checklist

### DIFF
--- a/internal/federation/dereferencing/dereferencer.go
+++ b/internal/federation/dereferencing/dereferencer.go
@@ -20,11 +20,49 @@ package dereferencing
 import (
 	"net/url"
 	"sync"
+	"time"
 
 	"github.com/superseriousbusiness/gotosocial/internal/media"
 	"github.com/superseriousbusiness/gotosocial/internal/state"
 	"github.com/superseriousbusiness/gotosocial/internal/transport"
 	"github.com/superseriousbusiness/gotosocial/internal/typeutils"
+	"github.com/superseriousbusiness/gotosocial/internal/util"
+)
+
+// FreshnessWindow represents a duration in which a
+// Status or Account is still considered to be "fresh"
+// (ie., not in need of a refresh from remote), if its
+// last FetchedAt value falls within the window.
+//
+// For example, if an Account was FetchedAt 09:00, and it
+// is now 12:00, then it would be considered "fresh"
+// according to DefaultAccountFreshness, but not according
+// to Fresh, which would indicate that the Account requires
+// refreshing from remote.
+type FreshnessWindow time.Duration
+
+var (
+	// 6 hours.
+	//
+	// Default window for doing a
+	// fresh dereference of an Account.
+	DefaultAccountFreshness = util.Ptr(FreshnessWindow(6 * time.Hour))
+
+	// 2 hours.
+	//
+	// Default window for doing a
+	// fresh dereference of a Status.
+	DefaultStatusFreshness = util.Ptr(FreshnessWindow(2 * time.Hour))
+
+	// 5 minutes.
+	//
+	// Fresh is useful when you're wanting
+	// a more up-to-date model of something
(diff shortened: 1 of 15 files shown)


internal/federation/dereferencing/dereferencer.go
```diff
@@ -20,13 +20,44 @@ package dereferencing
...
+// Status or Account is still considered to be "fresh"
+// (ie., not in need of a refresh from remote), if its
+// last FetchedAt value falls within the window.
+//
+// For example, if an Account was FetchedAt 09:00, and it
+// is now 12:00, then it would be considered "fresh"
+// according to DefaultAccountFreshness, but not according
+// to Fresh, which would indicate that the Account requires
+// refreshing from remote.
+type FreshnessWindow time.Duration
```
NyaaaWhatsUpDoc, line 41: i don't think we need a separate type for this, it honestly confused me for a bit looking at the func signatures thinking there was more abstractions than there were. Just using time.Duration should be fine

tsmethurst: Well, I considered it, but a separate type does have advantages since:

1. It prevents new devs on the codebase from thinking "window, what's that? I don't need that" and passing in `0 * time.Second` or something, since they'll have to go and look at the type, which has a nice explanatory comment on it.
2. Instead of just `window time.Duration` in function signatures you see `window FreshnessWindow` which immediately indicates that it pertains to freshness. This lets you hint to callers what's happening without having the variable name `freshnessWindow`.

So, it would be fine to use time.Duration, but I'd rather keep the type for the above reasons. Unless there's some noticeable performance disadvantage to using the type of course!

internal/processing/common/status.go
```diff
@@ -68,13 +69,14 @@ func (p *Processor) GetTargetStatusBy(
 		// a requester (i.e. request is authorized)
 		// to prevent a possible DOS vector.
 
-		if refresh {
-			// Refresh required, forcibly do synchronously.
+		if window < dereferencing.DefaultStatusFreshness {
```
NyaaaWhatsUpDoc, line 72: I think the semantics of this might make a bit more sense if you were to make the window a ptr value and replace this check with a `!= nil`. What do you think?

tsmethurst: Mm, I'm not sure, because then someone could provide a pointer to DefaultStatusFreshness, we'd have to then check the pointer and also check anyway if the window is tighter than the default. Tbh I think it's OK like this. In a previous version of this code the `window` variable was called `freshness` and then having a less than comparison definitely didn't make sense :P

NyaaaWhatsUpDoc: I mean I was thinking we have the pointer semantic *instead* of the tightness check. This also comes with the bonuses that callers wanting defaults don't need to pass in a value at all, and those passing in a value can still require synchronous dereferencing on any timeframe they want vs. at the moment which requires all synchronous fetching be done with a very fast freshness window

tsmethurst: Ahhh okay I see, that makes sense. So passing `nil` into the function call just uses the default. That could work. I'll fiddle with it some more.

tsmethurst: Alright, I did it!

NyaaaWhatsUpDoc:
looks good to me! feel free to squerge when it's done :sunglasses:

Issue #1887, opened 2023-06-12
[bug/performance] Slow Postgres `COUNT(*)` queries


mirabilos:
My instance is now up for over a year, but some things have become really slow. For example, the Pinafore frontend times out 3–5 times trying to load my followers (~160) or followees (~200). These classes of slowness seem to be database-related.

I notice, from the GtS log, that you use many `COUNT(*)` calls, which, on PostgreSQL, [must be slow](https://wiki.postgresql.org/wiki/Count_estimate) due to its data consistency guarantees.

Looking at the slow queries, many of them are…

    SELECT count(*) FROM \"statuses\" AS \"status\" WHERE (\"status\".\"account_id\" = '…')

… which seems to list the number of posts done by each account.

Which Pinafore doesn’t even display.

Which GtS cannot even get because their old toots from before I followed them aren’t backfilled.

So please just use one of the estimate alternatives for that. The number will be slightly off, but not by much, and given the above I doubt anyone would complain overly seriously.

(My `statuses` table is 375k rows long, and indicēs only help somewhat far but not for this.)

darrinsmart:
I would have thought PostgreSQL would still be quite fast for that query and row count. 

My instance uses sqlite, but as a test I copied the database to PostgreSQL to try the same type of query. It was quick - less than 1ms.

```
darrin=# explain analyze select count(*) from statuses as status where status.account_id='012345';
                                                                        QUERY PLAN                                                                         
-----------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=75.12..75.13 rows=1 width=8) (actual time=0.634..0.634 rows=1 loops=1)
   ->  Index Only Scan using statuses_account_id_idx on statuses status  (cost=0.42..68.29 rows=2735 width=0) (actual time=0.036..0.376 rows=2855 loops=1)
         Index Cond: (account_id = '012345'::bpchar)
         Heap Fetches: 0
 Planning Time: 0.132 ms
 Execution Time: 0.654 ms
(6 rows)
```

(I've obfuscated the actual account_id.. but I picked one with a lot of statuses. My "statuses" table has 512k rows)

Are you able to connect to your database with psql and run the query with "explain analyze"?

tsmethurst:
> Which Pinafore doesn’t even display.

[...]

PR #2620, opened 2024-02-09
[performance] temporarily cache account status counts to reduce no. account counts

NyaaaWhatsUpDoc:
should temporarily help with #1887 until my account status IDs PR is ready (will be post v0.14.0 release)

whilst i was here also ended up changing:
- fixing our database driver wrapping to include stmts (previously they were forgotten, though i'm not sure we actually use any)
- outlining `retryOnBusy()` to have a slow and path set of functions, allowing the fast path (case of just 1 try) to be inlined
- fix return of wrapped tx types on error

### DIFF
--- a/internal/db/bundb/drivers.go
+++ b/internal/db/bundb/drivers.go
@@ -141,6 +141,9 @@ func (c *SQLiteConn) BeginTx(ctx context.Context, opts driver.TxOptions) (tx dri
 		err = processSQLiteError(err)
 		return err
 	})
+	if err != nil {
+		return nil, err
+	}
 	return &SQLiteTx{Context: ctx, Tx: tx}, nil
 }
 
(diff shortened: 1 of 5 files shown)


NyaaaWhatsUpDoc:
i'm really just, having a time today, huh

internal/db/bundb/drivers.go
```diff
@@ -213,6 +239,29 @@ func (tx *SQLiteTx) Rollback() (err error) {
 	return
 }
 
+type SQLiteStmt struct {
+	context.Context
```
daenney, line 243: Why are we storing the context here but not in the Pg one?

daenney: From what I can see this shouldn't be necessary, since `ctx` can be passed directly to both `retryOnBusy` and `(Exec|Query)Context`

NyaaaWhatsUpDoc: ah good point!!

tsmethurst:
If this looks good to both of you @daenney and @NyaaaWhatsUpDoc feel free to squerge when everything is passing

PR #2633, opened 2024-02-12
[bugfix] fix possible mutex lockup during streaming code

NyaaaWhatsUpDoc:
# Description

This fixes an issue I have noticed a few times on my own instance. It rewrites Stream{} to use much less mutex locking, and updates related code calling it.

## Checklist

### DIFF
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -354,10 +354,13 @@ func (m *Module) writeToWSConn(
 
 	for {
 		// Wrap stream context with timeout to send a ping.
-		ctx, _ := context.WithTimeout(stream.Context(), ping)
+		ctx, cncl := context.WithTimeout(stream.Context(), ping)
 
 		// Block on receipt of next message.
 		msg, ok, open := stream.Recv(ctx)
+
+		// Stop timer.
+		cncl()
 
 		if !open {
 			// Stream was
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -368,19 +368,20 @@ loop:
 		// Stop timer.
 		cncl()
 
-		switch {
-		case pingctx.Err() != nil:
-			l.Trace("writing websocket ping")
+		if !ok {
+			if pingctx.Err() != nil {
+				// The ping context timed out!
+				l.Trace("writing websocket ping")
 
-			// Wrapped context time-out, send a keep-alive "ping".
-			if err := wsConn.WriteControl(websocket.PingMessage, nil, time.Time{}); err != nil {
-				l.Debugf("error writing websocket ping: %v", err)
-				break
+				// Wrapped context time-out, send a keep-alive "ping".
+				if err := wsConn.WriteControl(websocket.PingMessage, nil, time.Time{}); err != nil {
+					l.Debugf("error writing websocket ping: %v", err)
+					break
+				}
+
+				continue loop
 			}
 
-			continue loop
-
-		case !ok:
 			// Stream was
 			// closed.
 			return
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -266,6 +266,10 @@ func (m *Module) handleWSConn(l *log.Entry, wsConn *websocket.Conn, stream *stre
 	// to be closed.
 	<-ctx.Done()
 
+	// Close stream
+	// straightaway.
+	stream.Close()
+
 	// Tidy up underlying websocket connection.
 	if err := wsConn.Close(); err != nil {
 		l.Errorf("error closing websocket connection: %v", err)
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -22,10 +22,10 @@ import (
 	"slices"
 	"time"
 
-	"codeberg.org/gruf/go-kv"
 	apiutil "github.com/superseriousbusiness/gotosocial/internal/api/util"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/id"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
 	"github.com/superseriousbusiness/gotosocial/internal/oauth"
 	streampkg "github.com/superseriousbusiness/gotosocial/internal/stream"
@@ -213,10 +213,8 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
 
 	l := log.
 		WithContext(c.Request.Context()).
-		WithFields(kv.Fields{
-			{"username", account.Username},
-			{"streamID", stream.ID},
-		}...)
+		WithField("streamID", id.NewULID()).
+		WithField("username", account.Username)
 
 	// Upgrade the incoming HTTP request. This hijacks the
 	// underlying connection and reuses it for the websocket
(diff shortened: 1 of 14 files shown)


internal/api/client/streaming/stream.go
```diff
@@ -202,7 +202,7 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
 	// functions pass messages into a channel, which we can
 	// then read from and put into a websockets connection.
 	stream, errWithCode := m.processor.Stream().Open(
-		c.Request.Context(),
+		context.Background(), // use new ctx for async Stream{}
```
daenney, line 205: I think this means we break traces now, as the tracing context is lost this way.

internal/api/client/streaming/stream.go
```diff
@@ -238,56 +236,37 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
...
-		WithFields(kv.Fields{
-			{"username", username},
-			{"streamID", stream.ID},
-		}...)
-
-	// Create ticker to send keepalive pings
-	pinger := time.NewTicker(m.dTicker)
-
-	// Read messages coming from the Websocket client connection into the server.
+func (m *Module) handleWSConn(l *log.Entry, wsConn *websocket.Conn, stream *streampkg.Stream) {
```
daenney, line 247: I'm a bit surprised we pass the `log.Entry` down here. I don't think we do that in other parts of the codebase?

internal/api/client/streaming/stream.go
```diff
@@ -298,86 +277,61 @@ func (m *Module) handleWSConn(username string, wsConn *websocket.Conn, stream *s
 // This is a blocking function; will return only on read error or
 // if the given context is canceled.
```
daenney, line 278: I think this comment needs updating. There's no context anymore that gets passed in that could cause cancellation.

daenney (commented):
The storing a context on the Stream feels wrong. It's something the context package warns against doing and the Go team wrote a whole blog post about. We might have a good reason for it, but I don't know what the motivation for it is and why it has to be done this way.

internal/api/client/streaming/stream.go
NyaaaWhatsUpDoc: we already weren't doing anything with this context in terms of actually keeping it around for trace values, though i see your point i'll have a think about what can be done in terms of future possible logging

internal/api/client/streaming/stream.go
NyaaaWhatsUpDoc: we basically were passing in a number of variables each time purely for creating a log entry with them, so instead of recreating the log entry on every function call i just figured why not pass round the log entry

NyaaaWhatsUpDoc:
> The storing a context on the Stream feels wrong. It's something the context package warns against doing and the Go team wrote a whole blog post about. We might have a good reason for it, but I don't know what the motivation for it is and why it has to be done this way.

ah i wasn't aware of this. will have a think about rearchitecting it, and mark it draft for now

internal/api/client/streaming/stream.go
NyaaaWhatsUpDoc: the context is introduced again, the comment is relevant again :D

NyaaaWhatsUpDoc:
just wanna make one last change, changing to draft until then even though all tests are passing

internal/api/client/streaming/stream.go
```diff
@@ -299,89 +286,64 @@ func (m *Module) handleWSConn(username string, wsConn *websocket.Conn, stream *s
...
-
-				// The connection is gone; no
-				// further streaming possible.
-				break readLoop
+		// Read JSON objects from the client and act on them.
+		if err := wsConn.ReadJSON(&msg); err != nil {
+			// Only log an error if something weird happened.
+			// See: https://www.rfc-editor.org/rfc/rfc6455.html#section-11.7
+			if websocket.IsUnexpectedCloseError(err, []int{
+				websocket.CloseNormalClosure,
```
daenney, line 306: `websocket.CloseNormalClosure` feels out of place given the comment above it

> Only log an error if something weird happened

`CloseNormalClosure` doesn't sound that weird. Should we be passing that into `IsUnexpectedCloseError`?

internal/processing/stream/notification.go
```diff
@@ -18,20 +18,27 @@
...
 )
 
 // Notify streams the given notification to any open, appropriate streams belonging to the given account.
-func (p *Processor) Notify(n *apimodel.Notification, account *gtsmodel.Account) error {
-	bytes, err := json.Marshal(n)
+func (p *Processor) Notify(ctx context.Context, account *gtsmodel.Account, notif *apimodel.Notification) {
+	b, err := json.Marshal(notif)
 	if err != nil {
-		return fmt.Errorf("error marshalling notification to json: %s", err)
+		panic(err) // this should never happen
```
daenney, line 34: Should we debug log and return instead? I'm not super fond of functions calling `panic` and potentially crashing the whole lot. The worst that could happen is a notify doesn't get sent out?

internal/processing/stream/statusupdate.go
```diff
@@ -18,21 +18,24 @@
...
-// StatusUpdate streams the given edited status to any open, appropriate
-// streams belonging to the given account.
-func (p *Processor) StatusUpdate(s *apimodel.Status, account *gtsmodel.Account, streamTypes []string) error {
-	bytes, err := json.Marshal(s)
+// StatusUpdate streams the given edited status to any open, appropriate streams belonging to the given account.
+func (p *Processor) StatusUpdate(ctx context.Context, account *gtsmodel.Account, status *apimodel.Status, streamType string) {
+	b, err := json.Marshal(status)
 	if err != nil {
-		return fmt.Errorf("error marshalling status to json: %s", err)
+		panic(err) // this should never happen
```
daenney, line 34: Same here as for the other panic?

internal/processing/stream/update.go
```diff
@@ -18,20 +18,24 @@
...
 )
 
 // Update streams the given update to any open, appropriate streams belonging to the given account.
-func (p *Processor) Update(s *apimodel.Status, account *gtsmodel.Account, streamTypes []string) error {
-	bytes, err := json.Marshal(s)
+func (p *Processor) Update(ctx context.Context, account *gtsmodel.Account, status *apimodel.Status, streamType string) {
+	b, err := json.Marshal(status)
 	if err != nil {
-		return fmt.Errorf("error marshalling status to json: %s", err)
+		panic(err) // this should never happen
```
daenney, line 34: Idem here.

internal/api/client/streaming/stream.go
NyaaaWhatsUpDoc: i'll take a look. this was the previous logic but definitely worth looking into

internal/processing/stream/notification.go
NyaaaWhatsUpDoc: good point :+1:

NyaaaWhatsUpDoc: i opted for error log but just returns early, no panic: ac28a47b4

internal/processing/stream/statusupdate.go
NyaaaWhatsUpDoc: see: ac28a47b4

internal/processing/stream/update.go
NyaaaWhatsUpDoc: see: ac28a47b4

internal/api/client/streaming/stream.go
NyaaaWhatsUpDoc: IsUnexpectedCloseError does an inverse check to ensure it's **not** one of the provided codes, but this error check still then drops any non websocket close error. I updated this slightly confusing (and error dropping in some situations) logic here: 98179114e

daenney: Ohh. Right. That works then!

daenney (approved):

PR #2637, opened 2024-02-14
[docs] Inform new contributors to `git fetch`

NoraCodes:
# Description

Adds two lines describing how to avoid an error `fatal: No names found, cannot describe anything.` that I faced when initially setting up the repo on my machine.

This error happens because, when using the given instructions, `git` doesn't know about any tags.

## Checklist


tsmethurst:
Ah nice one, thanks!

PR #2642, opened 2024-02-15
[feature/oidc] Add support for very basic RBAC

9p4:
# Description

This pull request implements checking roles for all OpenID users logging in against a list of allowed roles. If the list of roles is empty, then the check is skipped. Only one role has to match to pass.

Closes https://github.com/superseriousbusiness/gotosocial/issues/1915

Future: revalidate "admin" claim on login (the role may have changed); if user is part of admin group, bypass check.

## Checklist

### DIFF
--- a/internal/api/auth/callback.go
+++ b/internal/api/auth/callback.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"slices"
 	"strings"
 
 	"github.com/gin-contrib/sessions"
@@ -156,6 +157,14 @@ func (m *Module) CallbackGETHandler(c *gin.Context) {
 		apiutil.TemplateWebPage(c, page)
 		return
 	}
+
+	// Check user permissions on login
+	if !allowedGroup(claims.Groups) {
+		err := fmt.Errorf("User groups %+v do not include an allowed group", claims.Groups)
+		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
 	s.Set(sessionUserID, user.ID)
 	if err := s.Save(); err != nil {
 		m.clearSession(s)
@@ -297,6 +306,11 @@ func (m *Module) createUserFromOIDC(ctx context.Context, claims *oidc.Claims, ex
 		return nil, gtserror.NewErrorConflict(err, help)
 	}
 
+	if !allowedGroup(claims.Groups) {
+		err := fmt.Errorf("User groups %+v do not include an allowed group", claims.Groups)
+		return nil, gtserror.NewErrorUnauthorized(err, err.Error())
+	}
+
 	// We still need to set something as a password, even
 	// if it's not a password the user will end up using.
 	//
@@ -356,17 +370,37 @@ func (m *Module) createUserFromOIDC(ctx context.Context, claims *oidc.Claims, ex
 // adminGroup returns true if one of the given OIDC
 // groups is equal to at least one admin OIDC group.
 func adminGroup(groups []string) bool {
-	for _, ag := range config.GetOIDCAdminGroups() {
-		for _, g := range groups {
-			if strings.EqualFold(ag, g) {
-				// This is an admin group,
-				// ∴ user is an admin.
-				return true
-			}
+	adminGroups := config.GetOIDCAdminGroups()
+	for _, claimedGroup := range groups {
+		if slices.ContainsFunc(adminGroups, func(allowedGroup string) bool {
+			return strings.EqualFold(claimedGroup, allowedGroup)
+		}) {
+			return true
 		}
 	}
 
 	// User is in no admin groups,
 	// ∴ user is not an admin.
+	return false
+}
+
+// allowedGroup returns true if one of the given OIDC
+// groups is equal to at least one allowed OIDC group.
+func allowedGroup(groups []string) bool {
+	allowedGroups := config.GetOIDCAllowedGroups()
+	if len(allowedGroups) == 0 {
+		// If no groups are configured, allow access (for backwards compatibility)
+		return true
+	}
+	for _, claimedGroup := range groups {
+		if slices.ContainsFunc(allowedGroups, func(allowedGroup string) bool {
+			return strings.EqualFold(claimedGroup, allowedGroup)
+		}) {
+			return true
+		}
+	}
+
+	// User is in no allowed groups,
+	// ∴ user is not allowed to log in
 	return false
 }
--- a/internal/api/auth/callback_test.go
+++ b/internal/api/auth/callback_test.go
@@ -1,0 +1,59 @@
+package auth
+
+import (
+	"testing"
+
+	"github.com/superseriousbusiness/gotosocial/testrig"
+)
+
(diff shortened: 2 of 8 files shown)


internal/api/auth/callback.go
```diff
@@ -370,3 +383,25 @@ func adminGroup(groups []string) bool {
...
+				// This is a group,
+				// ∴ user is allowed to log in
+				return true
+			}
+		}
+	}
+
+	// User is in no allowed groups,
+	// ∴ user is not allowed to log in
+	return false
```
tsmethurst, line 406: You can do this a bit nicer I think, using one of the new `slices` functions:

```suggestion
	allowedGroups := config.GetOIDCGroups()
	if len(allowedGroups) == 0 {
		// If no groups are configured as allowed,
		// allow access (for backwards compatibility).
		return true
	}

	// OK if one of claimed groups == an allowed group.
	for _, claimedGroup := range groups {
		if slices.ContainsFunc(allowedGroups, func(allowedGroup string) bool {
			return strings.EqualFold(claimedGroup, allowedGroup)
		})
	}

	// User is in no allowed groups,
	// ∴ user is not allowed to log in
	return false
```

internal/config/config.go
```diff
@@ -133,6 +133,7 @@ type Configuration struct {
 	OIDCClientSecret     string   `name:"oidc-client-secret" usage:"ClientSecret of GoToSocial, as registered with the OIDC provider."`
 	OIDCScopes           []string `name:"oidc-scopes" usage:"OIDC scopes."`
 	OIDCLinkExisting     bool     `name:"oidc-link-existing" usage:"link existing user accounts to OIDC logins based on the stored email value"`
+	OIDCGroups           []string `name:"oidc-groups" usage:"Membership of one of the listed groups allows access to GtS. If this is empty, all groups are allowed."`
```
tsmethurst, line 136: I think the name of this could be a bit more descriptive. How about `OIDCAllowedGroups`?

internal/api/auth/callback.go
9p4: Should I also apply that change to the same logic in `adminGroup`?

internal/api/auth/callback.go
```diff
@@ -156,6 +156,14 @@ func (m *Module) CallbackGETHandler(c *gin.Context) {
 		apiutil.TemplateWebPage(c, page)
 		return
 	}
+
+	// Check user permissions on login
+	if !allowedGroup(claims.Groups) {
+		err := gtserror.New("The user is not allowed access to the application")
```
tsmethurst, line 162: I think the error message could be a bit more generous by letting the caller know which groups the user is in; this might make things a bit easier for debugging. For example:

```suggestion
		err := fmt.Errorf("User groups %+v do not include an allowed group", claims.Groups)
```

Or something along those lines. What do you think?

internal/api/auth/callback.go
tsmethurst: Sure!

internal/api/auth/callback.go
9p4: I agree.

9p4: Do you think that the error message should also dump the list of groups the user is currently in?

tsmethurst: Nah probably not, that's something that could be debugged on the oauth side, besides which it's slightly more sensitive information that might allow someone to get enough info to eg., escalate an attack. Think it's fine to just return what groups GtS considers "allowed".

internal/api/auth/callback.go
```diff
@@ -156,6 +157,14 @@ func (m *Module) CallbackGETHandler(c *gin.Context) {
 		apiutil.TemplateWebPage(c, page)
 		return
 	}
+
+	// Check user permissions on login
+	if !allowedGroup(claims.Groups) {
+		err := gtserror.Newf("User groups %+v do not include an allowed group", claims.Groups)
```
tsmethurst, line 163: I'd drop the `gtserror.Newf` part from this and the other error below, and just use `fmt.Errorf` instead. `gtserror.Newf` is mostly useful for helping us debug internal errors because it includes the calling function as a prefix, but the caller of an API function like this doesn't really need to know the name of the calling function.

tsmethurst:
Alright, looks good to me :) I think @daenney might want to have a look over this since it pertains to other OIDC stuff that he's worked on.

daenney:
At first glance, it looks right, but I'll take a closer look.

I'd want some tests for the `allowedGroups` function though before we merge this, covering the base cases of empty list, case-insensitive match etc. It's relatively easy for bugs to sneak in there during a refactoring and that potentially creates a security issue.

internal/api/auth/callback.go
```diff
@@ -356,17 +370,41 @@ func (m *Module) createUserFromOIDC(ctx context.Context, claims *oidc.Claims, ex
 // adminGroup returns true if one of the given OIDC
 // groups is equal to at least one admin OIDC group.
 func adminGroup(groups []string) bool {
```
daenney, line 372: The behaviour of this function has changed. The docs say "if the given OIDC groups is equal to at least one admin OIDC group", but we now skip this check if the admin groups is 0. Previously, it had to match, and it could only return true if there was a match.

I think now everyone becomes an admin if oidc-admin-groups is empty.

daenney: If oidc-admin-groups is empty, it should return `false`, not `true`. Admin should only ever be granted on an explicit match.

9p4: Whoops, that should not have gotten through. Apologies. I must've copied the other function when switching to the new slices API and did not remove the check.

tsmethurst: Good catch! :facepalm:

tsmethurst:
If you could add some small tests for the `allowedGroup` and `adminGroup` functions, this will be ready to squerge I think :) Do you need any help writing the tests?

9p4:
I will be busy with schoolwork for the next week, but I should be able to do it sometime next weekend :)

daenney:
Unless there's a merge conflict or you're actually changing something, there's no need to constantly push rebases of your changes. This creates a lot of notification noise for us, despite the fact that there's nothing new to review.

If you're modifying existing code, please push additional commits on top of your current changes. That's easier to review, as we can see what the new code has changed compared to the previous state. We'll take care of squashing it into a single commit when we merge it.

internal/api/auth/callback_test.go
```diff
@@ -0,0 +1,59 @@
+package auth
+
+import (
+	"testing"
+
+	"github.com/superseriousbusiness/gotosocial/testrig"
+)
+
```
daenney, line 8: These tests can be simplified a bit by changing them to table driven tests.

```go
func TestAdminGroup(t *testing.T) {
	testrig.InitTestConfig()
	for _, test := range []struct {
		name     string
		groups   []string
		expected bool
	}{} {
		test := test // loopvar capture
		t.Run(test.name, func(t *testing.T) {
			if got := adminGroup(test.groups); got != test.expected {
				t.Fatalf("got: %t, wanted: %t", got, test.expected)
			}
		})
	}
}
```

That means we now only need 2 top-level test functions and we can have multiple tests per test function. That also makes it easier to add more cases later on.

tsmethurst:
Looks good! :) Thanks!

PR #2645, opened 2024-02-16
[chore] Simplify the User-Agent string

daenney:
# Description

[RFC 9110][1] includes a definition for the format of a user-agent header:

```
User-Agent = product *( RWS ( product / comment ) )
             product         = token ["/" product-version]
             product-version = token
	     comment        = "(" *( ctext / quoted-pair / comment ) ")"
             ctext          = HTAB / SP / %x21-27 / %x2A-5B / %x5D-7E / obs-text
```

An example given in the RFC: `User-Agent: CERN-LineMode/2.15 libwww/2.17b3`

The idea is typically start with the most important `product/version`, add a `(comment)` if necessary and then include any auxiliary products. However, the RFC warns against including too many auxiliary products as those can be unnecessarily revealing.

For automated systems (i.e not a browser), the common and recommended format is `<product></version> (+uri-for-contact)`, followed with any additional `<product>/<version>` pairs that are relevant.

This changes our UA to match that convention more closely. This makes it easier for administrators who do user-agent parsing for statistics or other purposes to correctly identify the version of GoToSocial. Currently tools tend to get confused by the lack of a `/<version>` on the start of our string.

[1]: https://www.rfc-editor.org/rfc/rfc9110.html#name-user-agents

## Checklist

### DIFF
--- a/internal/transport/controller.go
+++ b/internal/transport/controller.go
@@ -77,7 +77,7 @@ func NewController(state *state.State, federatingDB federatingdb.DB, clock pub.C
 		clock:     clock,
 		client:    client,
 		trspCache: cache.NewTTL[string, *transport](0, 100, 0),
-		userAgent: fmt.Sprintf("%s (+%s://%s) gotosocial/%s", applicationName, proto, host, version),
+		userAgent: fmt.Sprintf("%s/%s (+%s://%s)", applicationName, version, proto, host),
 		senders:   senders,
 	}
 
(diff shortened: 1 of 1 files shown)


daenney:
* `Misskey/2023.10.2 (https://instance/)`
* `http.rb/5.1.1 (Mastodon/4.2.0; +https://instance/)`
* `Firefish/1.0.4-beta31 (https://instance)`
* `Pleroma 2.5.5; https://instance <admin@instance>`

The Masto format is basically correct, though they should swap the `http.rb/5.1.1` and `Mastodon/4.2.0` around as the "most important" one should come first. Browsers have a different format where that's not the cause due to historic browser wars.

The Forkeys generally follow the expected format, though they don't start the URL with a `+` which is the prevalent convention.

The Omas one is a little out there and doesn't resemble any the IETF recommendation or the format MDN documents that's typically used for browsers.

tsmethurst:
Nice one, looks good to me!

PR #2646, opened 2024-02-16
[chore] refactor extractFromCtx a bit

tsmethurst:
A little chore PR that comes out of the Move stuff I've been doing. PRing it separately here to avoid the Move PR becoming too big.

Just refactors extractFromCtx -> getActivityContext and simplifies the return type. No actual logic changes, but there might be in the Move PR.

### DIFF
--- a/internal/federation/federatingdb/accept.go
+++ b/internal/federation/federatingdb/accept.go
@@ -41,10 +41,13 @@ func (f *federatingDB) Accept(ctx context.Context, accept vocab.ActivityStreamsA
 		l.Debug("entering Accept")
 	}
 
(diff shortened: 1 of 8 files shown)


daenney (approved):
This looks good to me, it's just moving the data we got returned by the function into a struct.

One thing maybe for the future, if you change the `requestingAcct` and `receivingAcct` to be `requestingAccount` and `receivingAccount` to retain the same names as we originally had, the diff becomes a lot smaller making this kind of change a bit simpler to review.

tsmethurst:
Ah yes, I actually did that in a previous version of this, but I ended up preferring the brevity of requestingAcct compared to requestingAccount etc. Bit easier on the eyes and we don't use `acct` to mean anything else.

PR #2647, opened 2024-02-16
[feature/chore] Add Move database functions + cache

tsmethurst:
# Description

>

This pull request adds database functions, cache, and helpers for storing and retrieving account Move activities (or indeed any kind of Move activity).

Will be needed for the upcoming Move PR; PR'd separately here to avoid that one becoming massive.

Part of https://github.com/superseriousbusiness/gotosocial/issues/130

## Checklist


internal/db/move.go
```diff
@@ -0,0 +1,56 @@
...
+	// GetMoveByURI gets one Move with the given AP URI.
+	GetMoveByURI(ctx context.Context, uri string) (*gtsmodel.Move, error)
+
+	// GetMoveByOriginTarget gets one move with the given originURI and targetURI.
+	GetMoveByOriginTarget(ctx context.Context, originURI string, targetURI string) (*gtsmodel.Move, error)
+
+	// GetLatestMoveSuccessInvolvingURIs gets the time of
+	// the latest successfully-processed Move that includes
+	// either uri1 or uri2 in target or origin positions.
+	GetLatestMoveSuccessInvolvingURIs(ctx context.Context, uri1 string, uri2 string) (time.Time, error)
```
NyaaaWhatsUpDoc, line 40: so just to be clear this doesn't have to involve *both* uri1 and uri2, it's just the last attempt that includes either one of those 2? and how often will this be called because i definitely think this is cacheable in the struct cache :thinking: 

(same goes for the similar function below)

NyaaaWhatsUpDoc:
one or two questions but otherwise looking good :) as discussed in matrix will be saving squerge of this (once questions resolved) util after v0.14.0 final release :+1:

internal/db/move.go
tsmethurst: Exactly, it's just the last attempt that includes either of the two. It'll only be called when checking if a Move should be processed, so not v. often at all. It's proooobably not worth caching at this point, imo.

NyaaaWhatsUpDoc (approved):

Issue #2460, opened 2023-12-16
[bug] Parent status / replied-to status not dereferenced sometimes


tsmethurst:
This is sorta related to a bug I thought we'd fixed already. Sometimes when a status from someone I follow gets inserted into the timeline, after being federated in via a Create activity, the parent status (`in_reply_to_uri`) doesn't get dereferenced, so the reply appears to be orphaned, since it's serialized via the API as having `in_reply_to = null`.

Checking in the db, I see that the status does have an `in_reply_to_uri` set, and it's the URI of a public / unlisted status on an instance I don't block, by a user I don't block, so it should have been derefed before the child status was timelined.

We should check again through our dereferencing logic to ensure that this situation doesn't happen, since it's rather jarring and may also lead to statuses being put in the timeline when they oughtn't.

daenney:
I got this chain recently too. Initially:

```
sqlite> select in_reply_to_id, in_reply_to_uri, in_reply_to_account_id  from statuses where statuses.url = 'https://goblin.technology/@tobi/statuses/XXXX';
in_reply_to_id  in_reply_to_uri                                               in_reply_to_account_id
--------------  ------------------------------------------------------------  ----------------------
NULL            https://k.iim.gay/users/kim/statuses/YYYY  NULL                                                                                           
```

I tried to refresh that status a few times, but the thread kept being stuck in detached head state.

This morning, after refreshing it:

```
sqlite> select in_reply_to_id, in_reply_to_uri, in_reply_to_account_id  from statuses where statuses.url = 'https://goblin.technology/@tobi/statuses/XXXX';
in_reply_to_id              in_reply_to_uri                                               in_reply_to_account_id    
--------------------------  ------------------------------------------------------------  --------------------------
ZZZZ  https://k.iim.gay/users/kim/statuses/YYYY  AAAA
```

benjaminbhollon:
I've been seeing this a number of times on my instance but hadn't had time to investigate yet. At least once, my replies to the chain weren't being federated to the other instance—though that may be a completely unrelated issue or a one-time error.

mirabilos:
This does not seem to be fixed in 0.13.1-SNAPSHOT git-d59cba4 🦥

[...]

PR #2652, opened 2024-02-17
[bugfix] Fix dereferencing ancestors on new status create

tsmethurst:
# Description

>

Likely fix for https://github.com/superseriousbusiness/gotosocial/issues/2460.

Looks like on new status creates, we were passing the "barebones" status into the dereferenceThread function, which did not include the correct inReplyToURI values. This should fix it by passing in the properly dereferenced and populated status instead.

As a result of this fix, more followers-only posts were showing up in the timeline, even when the post they reply to is not visible. This PR fixes that too, by not assuming that 404 always means deleted when trying to deref a status. Many implementations (including Mastodon) just return 404 when we don't have permission to view a status because of its visibility settings, so we need to take account of that by leaving the in_reply_to_uri in place in such cases for visibility filtering purposes.

The PR also tones down the "status not deref'd" warnings to debug level instead, since they represent a legitimate state that GtS can and will end up in when checking visibility.

## Checklist

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -50,10 +50,6 @@ Take a look at the [list of open bugs](https://github.com/superseriousbusiness/g
 
 These warnings are caused by Mastodon forgetting to take query parameters into account when signing HTTP requests. Fixes on both Mastodon and GoToSocial sides are in the works, and other instance software is expected to follow.
 
-## Warnings “status not yet deref'd” in the logs
-
-The warning is mostly to say that the derived visibility of the given status may not be entirely accurate due to a parent not being dereferenced yet, so do not worry about it.
-
 ## Will you support tables in Markdown?
 
 Not at the moment, as most clients handle them terribly.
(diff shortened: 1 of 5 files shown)


daenney:
I've plonked it on my instance too. I'll keep an eye out. There's a few people who seem to manage to provoke broken threads somewhat regularly as they like to chain-post a lot.

NyaaaWhatsUpDoc:
this definitely won't fix *all* situations. some broken threads occur due to failed fetched at which point it should be picked up by the visibility filter, these i'm fixing in a separate patch i'm working on as previously mentioned in the matrix. but this fix here should absolutely fix a bunch of situations given we're not passing in the latest dereferenced / stored status :woman_facepalming:

tsmethurst:
I have found another issue with this now, which is that lots of followers-only replies by people you do follow are showing up in the timeline, but they respond to followers-only posts by people you don't follow. I've found where the problem is and I'll fix it a tomorrow or Monday.

tsmethurst:
But yes, it already seems better :)

Issue #1135, opened 2022-11-24
[bug] Missing documentation of API authentication


utrenkner:
### Describe the bug with a clear and concise description of what the bug is.

The individual [API calls are excellently documented](https://docs.gotosocial.org/en/latest/api/swagger/). However, I am missing an important piece: How do I supply user or other needed information to the API.

Most API calls require authentication. I was able to create an auth token using the [auth_flow.sh](https://github.com/superseriousbusiness/gotosocial/blob/main/scripts/auth_flow.sh) script. But still I am not able supply this token together with user(name) in a way the API understands.

Could you possibly add to the documentation one or several complete curl calls to the API so that I can pick up from that? Or even a step-by-step intro to the API authentication?

### What's your GoToSocial Version?

0.5.2

### GoToSocial Arch

amd64 (on FreeBSD 13.1-RELEASE)

### What happened?

The API documentation says that a call like ``curl https://<myhost>/api/v1/accounts/verify_credentials`` should work without any parameters. But it answers:
```
Unauthorized: token not supplied
```
So, I add the token to the http header: ``curl -H 'Authorization: Bearer <mytoken>' https://<myhost>/api/v1/accounts/verify_credentials``. But the API answers:
```
Unauthorized: user not supplied or not authorized
```
And this is where I am lost. I tried adding the things like ``-H 'User: <myuser>'`` and similar. But without success.

### What you expected to happen?

Above given error messages.

anantshri:
@utrenkner https://github.com/anantshri/gotosocial_bulk_follow/blob/main/README.md#note-about-auth-token might help. i had a bad time with that too. ultimately i used toolcli token in itself.

drym3r:
It happens the same to me. Your script seems to do the same @anantshri as published, it uses the bearer token (if I understood it correctly).

What do you mean when you said `ultimately i used toolcli token in itself`?

I've also tried posting with [toot](https://toot.readthedocs.io/en/latest/index.html) and the error was even weirder:

```
...
        <section class="error">
          <span>❌</span> <pre>Bad Request: json: cannot unmarshal string into Go struct field AdvancedStatusCreateForm.sensitive of type bool</pre>
        </section>
...
```

dennyabrain:
Hey all, I faced a similar problem and after struggling with it for a while, I was able to make authentication work. Should I add this documentation somewhere and make a PR for this?

NyaaaWhatsUpDoc:
[...]

Issue #2421, opened 2023-12-06
[bug] `scripts/auth-flow.sh` does not work as described


9ssi7:
### Describe the bug with a clear and concise description of what the bug is.

In its simplest definition;
- I clone the project
- I go to the examples folder and run the `docker compose up -d` command
- Then I run the `scripts/auth-flow.sh` file from the root directory.
- Then an error like the following appears:

```json
{
  "error": "Unauthorized: application not supplied"
}
```

I tried this in different ways. However, there seems to be an error in the token generation or verification part. I got an error like above in a few different ways and I can't take action.

### What's your GoToSocial Version?

v0.12.2 / latest

### GoToSocial Arch

arm64 Docker

tsmethurst:
Oh, I remember discussing this a while back as well, let me scroll back up through the chat and see if I can find what I thought was causing it...

9ssi7:
Any progress? If the work to be done is clear, I can contribute. Or if you could explain if something different needs to be done to fix it, I would appreciate it.

tsmethurst:
No progress, but it's pretty low priority since it's just an example script. The way to fix it at least in the example is to use the admin CLI to 'confirm' the account after it's been created, then restart the instance.

We'll probably rejig this entire part of the code when we properly implement self-registration of new accounts. This current endpoint that we have now is essentially just a stub for what will come then.

But it will likely never be the case that people can self-register via the API and then immediately start using a token to post on the instance etc, without at least having to confirm their email address first. I don't think there's any fedi software that does that, and for good reason :P

NyaaaWhatsUpDoc:
i'm going to remove this from the v0.14.0 milestone as the new authentication workflow doc better handles what this script handled initially re: teaching a user about how to authenticate an application with an account.

though i'll leave this issue open as i think this script should either be removed / improved when working on the registration flow as @tsmethurst previously mentioned. though i don't think there should be any logic changes done to how we always require email confirmation (though useful to note somewhere, if not already, this can be worked around via admin CLI commands), again as @tsmethurst mentions there's good reason for this.

tsmethurst:
I think we can just remove that script, and close the issue. We now have the auth flow documented anyways, so the script doesn't add that much -- https://docs.gotosocial.org/en/latest/api/authentication/

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/2421.

PR #2655, opened 2024-02-18
[docs] Add oauth token authentication workflow

fyrfli:
# Add the api authentication workflow to the documentation

>

This pull request adds the api authentication workflow to the documents and addresses #1135.

Closes: #1135

## Checklist

### DIFF
--- a/docs/api/authentication.md
+++ b/docs/api/authentication.md
@@ -27,8 +27,8 @@ Note that `scopes` can be one of:
 - admin
 
 !!! warning
- Be **very** careful what scopes you grant to your application. 
- The default is `read` if `scopes` is not set.
+    Be **very** careful what scopes you grant to your application.
+    The default is `read` if `scopes` is not set.
 
 A successful call returns a response with a client_id and client_secret that we are going need to use in the rest of the process. It looks something like this: 
 ```
@@ -42,7 +42,7 @@ A successful call returns a response with a client_id and client_secret that we 
 ```
 
 !!! important
- Ensure you save the client_id and client_secret somewhere so you can refer to it as we go.
+    Ensure you save the client_id and client_secret somewhere so you can refer to it as we go.
 
 ## authorizing your application
 
@@ -55,19 +55,19 @@ https://your.instance.url/oauth/authorize?client_id=your_new_client_id-id&redire
 
 You'll get a login form to your instance and be prompted to login if you aren't already logged in. Once logged in, you will get a screen that says something like this:
 !!! note
- Hi `your_username`!
+    Hi `your_username`!
  
- Application `your_app_name` would like to perform actions on your behalf, with scope *`read`*.
- The application will redirect to urn:ietf:wg:oauth:2.0:oob to continue.
+    Application `your_app_name` would like to perform actions on your behalf, with scope *`read`*.
+    The application will redirect to urn:ietf:wg:oauth:2.0:oob to continue.
  
- <span style="padding:12px;background-color:#66befe;color:black;">Allow</span>
+    <span style="padding:12px;background-color:#66befe;color:black;">Allow</span>
  
 
 Once you click `Allow`, you will get a window that looks something like this:
 
 !!! note
- Here's your out-of-band token with scope "read", use it wisely:
-  WBANQAXXHDN9KJQZXGWNQANA4V9EJWMUDHVCUUM3JHYAB3DP
+    Here's your out-of-band token with scope "read", use it wisely:
+     WBANQAXXHDN9KJQZXGWNQANA4V9EJWMUDHVCUUM3JHYAB3DP
 
 
 ## Getting your token
@@ -82,8 +82,8 @@ curl -X POST 'https://your.instance.url/oauth/token' -H 'Content-Type:applicatio
 }' 
 ```
 !!! warning
- Please note that the characters used above are just a random selection of characters and cannot be used.
-Make sure you replace it with the code *you* get from your instance.
+    Please note that the characters used above are just a random selection of characters and cannot be used.
+    ake sure you replace it with the code *you* get from your instance.
 
 You'll get a response that includes your access token and looks something like this:
 ```
--- a/docs/api/authentication.md
+++ b/docs/api/authentication.md
@@ -22,12 +22,14 @@ Note that `scopes` can be any space-separated combination of:
 
 - read
 - write
-- follow
-- push
 - admin
 
 !!! warning
-    Be **very** careful what scopes you grant to your application. While GoToSocial does not yet support properly scoped tokens, it is always good practice to grant your application the lowest tier permissions it needs to do its job. e.g. If your application won't be making posts, use scope=read. Also note that "read" is the default if none is set.
+   GoToSocial does not currently support scoped authorization tokens, so any token you obtain in this process will be able to perform all actions on your behalf, including admin actions if your account has admin permissions. Nevertheless, it is always good practice to grant your application the lowest tier permissions it needs to do its job. e.g. If your application won't be making posts, use scope=read.
+   
+   In this spirit, "read" is used in the example above, which means that in the future when scoped tokens are supported, the application will be restricted to only being able to do "read" actions.
+   
+   You can read more about additional planned OAuth security features [right here](https://github.com/superseriousbusiness/gotosocial/issues/2232).
 
 A successful call returns a response with a `client_id` and `client_secret` that we are going need to use in the rest of the process. It looks something like this: 
 ```json
--- a/docs/api/authentication.md
+++ b/docs/api/authentication.md
@@ -1,0 +1,113 @@
+# Authentication with the API
+
+GoToSocial implements an API that is compatible with the Mastodon API which means that most everything to do with the API follows many of the conventions and techniques that work for the Mastodon API. That includes authentication.
+
+Find below the steps used to authenticate with the GoToSocial API.
+
+---
+
+## Create a new application
+
+We need to get a new key:pair for our application. This is done by making a `POST` request to the `/api/v1/apps` endpoint. Replace `your_app_name` in the code below:
+
+```
+curl -X POST 'https://your.instance.url/api/v1/apps' -H 'Content-Type:application/json' -d '{
+  "client_name": your_app_name,
+  "redirect_uris": "urn:ietf:wg:oauth:2.0:oob",
+  "scopes": "read"
+}'
+
+```
+
+Note that `scopes` can be one of:
+- read
+- write
+- follow
+- push
+- admin
+
+> [!WARNING]
+> Be **very** careful what scopes you grant to your application. 
+> The default is `read` if `scopes` is not set.
+
+A successful call returns a response with a client_id and client_secret that we are going need to use in the rest of the process. It looks something like this: 
+```
+{
+  "id": randomised_id,
+  "name": your_app_name,
+  "redirect_uri": "urn:ietf:wg:oauth:2.0:oob",
+  "client_id": your_new_client_id,
+  "client_secret": your_new_client_secret
+}
+```
+
+> [!IMPORTANT]
+> Ensure you save the client_id and client_secret somewhere so you can refer to it as we go.
+
+## authorizing your application
+
+We've registered a new application with GoToSocial, but it isn't connected to your account just yet. Now we need to tell GoToSocial that that new application is actually going to interact as you. To do this, we need to authenticate with your instance via a browser to process the login and permission granting process.
+
+Using the `client_id` from the previous step, create a URL with a query string like so:
+```
+https://your.instance.url/oauth/authorize?client_id=your_new_client_id-id&redirect_uri=urn:ietf:wg:oauth:2.0:oob&response_type=code
+```
+
+You'll get a login form to your instance and be prompted to login if you aren't already logged in. Once logged in, you will get a screen that says something like this:
+> [!NOTE]
+> Hi `your_username`!
+> 
+> Application `your_app_name` would like to perform actions on your behalf, with scope *`read`*.
+> The application will redirect to urn:ietf:wg:oauth:2.0:oob to continue.
+> 
+> <span style="padding:8px;background-color:#66befe;color:black;">Allow</span>
+> 
+
+Once you click `Allow`, you will get a window that looks something like this:
+
+> [!NOTE]
+> Here's your out-of-band token with scope "read", use it wisely:
+>  WBANQAXXHDN9KJQZXGWNQANA4V9EJWMUDHVCUUM3JHYAB3DP
+
+
+## Getting your token
+The next step is to send a `POST` request to the oauth/token endpoint to get an access token that you will use to authenticate your future requests. 
+```
+curl -X POST 'https://your.instance.url/oauth/token' -H 'Content-Type:application/json' -d '{
+  "redirect_uri": "urn:ietf:wg:oauth:2.0:oob",
+  "client_id": "your_new_client_id",
+  "client_secret": "your_new_client_secret",
+  "grant_type": "authorization_code",
+  "code": "WBANQAXXHDN9KJQZXGWNQANA4V9EJWMUDHVCUUM3JHYAB3DP"
+}' 
+```
+> [!warning] 
+> Please note that the characters used above are just a random selection of characters and cannot be used.
+> Make sure you replace it with the code *you* get from your instance.
+
+You'll get a response that includes your access token and looks something like this:
+```
+{
+  "access_token": "your_brand_new_access_token",
+  "created_at": unixtimestamp,
+  "scope": "read",
+  "token_type": "Bearer"
+}
+```
+## Verifying it all
+To make sure everything went through successfully, target the `/api/v1/verify_credentials` endpoint, adding your new access token to the Header as `Authentication:Bearer your_brand_new_token`.
+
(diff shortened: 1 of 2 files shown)


docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
+# Authentication with the API
+
+GoToSocial implements an API that is compatible with the Mastodon API which means that most everything to do with the API follows many of the conventions and techniques that work for the Mastodon API. That includes authentication.
```
daenney, line 3: I would suggest moving this to either the `swagger.md`, as that's the introduction page for the API section, or dropping it. It's generally understood that alternative implementations provide a Mastodon-compatible Client API.

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+
+Find below the steps used to authenticate with the GoToSocial API.
+
+---
+
+## Create a new application
+
+We need to get a new key:pair for our application. This is done by making a `POST` request to the `/api/v1/apps` endpoint. Replace `your_app_name` in the code below:
```
daenney, line 11:
```suggestion
We need to get a new key pair for our application. This is done by making a `POST` request to the `/api/v1/apps` endpoint. Replace `your_app_name` in the code below:
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+
+```
+curl -X POST 'https://your.instance.url/api/v1/apps' -H 'Content-Type:application/json' -d '{
+  "client_name": your_app_name,
+  "redirect_uris": "urn:ietf:wg:oauth:2.0:oob",
```
daenney, line 16: I would add an explanation after this code sample for why `urn:ietf:wg:oauth:2.0:oob` specifically.

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+- write
+- follow
+- push
+- admin
+
+!!! warning
+    Be **very** careful what scopes you grant to your application.
+    The default is `read` if `scopes` is not set.
+
+A successful call returns a response with a client_id and client_secret that we are going need to use in the rest of the process. It looks something like this:
```
daenney, line 33:
```suggestion
A successful call returns a response with a `client_id` and `client_secret` that we are going need to use in the rest of the process. It looks something like this: 
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+- follow
+- push
+- admin
+
+!!! warning
+    Be **very** careful what scopes you grant to your application.
+    The default is `read` if `scopes` is not set.
+
+A successful call returns a response with a client_id and client_secret that we are going need to use in the rest of the process. It looks something like this: 
+```
```
daenney, line 34:
```suggestion
```json
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+  "redirect_uri": "urn:ietf:wg:oauth:2.0:oob",
+  "client_id": your_new_client_id,
+  "client_secret": your_new_client_secret
+}
+```
+
+!!! important
+    Ensure you save the client_id and client_secret somewhere so you can refer to it as we go.
+
+## authorizing your application
```
daenney, line 47:
```suggestion
## Authorizing your application
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+  "id": randomised_id,
+  "name": your_app_name,
+  "redirect_uri": "urn:ietf:wg:oauth:2.0:oob",
+  "client_id": your_new_client_id,
+  "client_secret": your_new_client_secret
+}
+```
+
+!!! important
+    Ensure you save the client_id and client_secret somewhere so you can refer to it as we go.
```
daenney, line 45:
```suggestion
    Ensure you save the `client_id` and `client_secret` somewhere so you can refer to it as we go.
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+
+## authorizing your application
+
+We've registered a new application with GoToSocial, but it isn't connected to your account just yet. Now we need to tell GoToSocial that that new application is actually going to interact as you. To do this, we need to authenticate with your instance via a browser to process the login and permission granting process.
```
daenney, line 49:
```suggestion
We've registered a new application with GoToSocial, but it isn't connected to your account just yet. Now we need to tell GoToSocial that that new application is actually going to act on your behalf. To do this, we need to authenticate with your instance via a browser to initiate the login and permission granting process.
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+{
+  "id": randomised_id,
+  "name": your_app_name,
+  "redirect_uri": "urn:ietf:wg:oauth:2.0:oob",
+  "client_id": your_new_client_id,
+  "client_secret": your_new_client_secret
+}
+```
+
+!!! important
```
daenney, line 44:
```suggestion
!!! tip
```

`important` isn't a recognised admonition, so it renders just like Note. I think `tip` is suitable here instead.

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+
+We've registered a new application with GoToSocial, but it isn't connected to your account just yet. Now we need to tell GoToSocial that that new application is actually going to interact as you. To do this, we need to authenticate with your instance via a browser to process the login and permission granting process.
+
+Using the `client_id` from the previous step, create a URL with a query string like so:
+```
+https://your.instance.url/oauth/authorize?client_id=your_new_client_id-id&redirect_uri=urn:ietf:wg:oauth:2.0:oob&response_type=code
+```
+
+You'll get a login form to your instance and be prompted to login if you aren't already logged in. Once logged in, you will get a screen that says something like this:
+!!! note
```
daenney, line 57: I would change this to a code block. Admonitions are for callouts or separate calls to action. I wouldn't use them to display the response of and endpoint or result of an action.

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+ 
+    Application `your_app_name` would like to perform actions on your behalf, with scope *`read`*.
+    The application will redirect to urn:ietf:wg:oauth:2.0:oob to continue.
+ 
+    <span style="padding:12px;background-color:#66befe;color:black;">Allow</span>
+ 
+
+Once you click `Allow`, you will get a window that looks something like this:
+
+!!! note
```
daenney, line 68: Idem here, this should be a code block.

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+  "grant_type": "authorization_code",
+  "code": "WBANQAXXHDN9KJQZXGWNQANA4V9EJWMUDHVCUUM3JHYAB3DP"
+}' 
+```
+!!! warning
+    Please note that the characters used above are just a random selection of characters and cannot be used.
+    ake sure you replace it with the code *you* get from your instance.
+
+You'll get a response that includes your access token and looks something like this:
+```
```
daenney, line 89:
```suggestion
```json
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
...
+{
+  "access_token": "your_brand_new_access_token",
+  "created_at": unixtimestamp,
+  "scope": "read",
+  "token_type": "Bearer"
+}
+```
+## Verifying it all
+To make sure everything went through successfully, target the `/api/v1/verify_credentials` endpoint, adding your new access token to the Header as `Authentication:Bearer your_brand_new_token`.
```
daenney, line 98:
```suggestion
To make sure everything went through successfully, query the `/api/v1/verify_credentials` endpoint, adding your new access token to the Header as `Authorization: Bearer your_brand_new_token`.
```

docs/api/authentication.md
```diff
@@ -0,0 +1,113 @@
+# Authentication with the API
+
+GoToSocial implements an API that is compatible with the Mastodon API which means that most everything to do with the API follows many of the conventions and techniques that work for the Mastodon API. That includes authentication.
+
+Find below the steps used to authenticate with the GoToSocial API.
+
+---
```
daenney, line 7:
```suggestion
```

The headers already create separate sections, we don't really need additional horizontal lines here.

daenney (changes requested):

daenney:
Overall this looks pretty great. I have some small stylistic/markup changes and once those get resolved I think this should be good to go!

docs/api/authentication.md
daenney: If you want to keep an intro paragraph, I would go with something like "Using the client API requires authentication. This page documents the general flow for retrieving an authentication token with examples for doing this on the CLI".

fyrfli:
sweet ... thanks for these. i will try and have them done today

daenney:
No worries, it's the weekend! We'll be here all of next week :wink:.

docs/api/authentication.md
fyrfli: maybe use a quote block instead?

docs/api/authentication.md
fyrfli: question: would it be helpful to use ```bash for the curl lines?

docs/api/authentication.md
fyrfli: done. i hope it's ok to use your suggestion as is for this. i did want some kind of intro paragraph, but i wrote it independently in obsidian and copied it directly into a new file, which is how i got that somewhat redundant paragraph.

docs/api/authentication.md
fyrfli: to be honest, i had to go searching for what it means because *i* didn't know what it meant.

docs/api/authentication.md
```diff
@@ -0,0 +1,116 @@
...
+```json
+{
+  "access_token": "your_brand_new_access_token",
+  "created_at": unixtimestamp,
+  "scope": "read",
+  "token_type": "Bearer"
+}
+```
+## Verifying it all
+To make sure everything went through successfully, query the `/api/v1/verify_credentials` endpoint, adding your new access token to the Header as `Authentication: Bearer your_brand_new_token`.
```
daenney, line 99:
```suggestion
To make sure everything went through successfully, query the `/api/v1/verify_credentials` endpoint, adding your new access token to the Header as `Authorization: Bearer your_brand_new_token`.
```

daenney (commented):
Thanks for resolving the comments! There's one with a wrong header name and after that I think we're good to go.

fyrfli:
done. surprised i missed that especially since i caught a few other typos earlier.

fyrfli:
my pleasure. i've been looking for the opportunity to contribute for a while.

daenney:
I'll leave it open for a day or two so the rest of the team has the opportunity to go through it too. If it doesn't get merged by anyone else I'll click the button on Wednesday.

tsmethurst:
This is looking good :) I'll have a proper look through it this afternoon. Thanks for doing this!

Linking this issue because it's relevant: https://github.com/superseriousbusiness/gotosocial/issues/2421

docs/api/authentication.md
```diff
@@ -0,0 +1,116 @@
+# Authentication with the API
+
+Using the client API requires authentication. This page documents the general flow for retrieving an authentication token with examples for doing this on the CLI
+
+## Create a new application
+
+We need to get a new key pair for our application. This is done by making a `POST` request to the `/api/v1/apps` endpoint. Replace `your_app_name` in the code below:
```
tsmethurst, line 7: Key pair is a bit confusing here because it suggests that cryptography is involved somehow. I'd just say something like:

```suggestion
We need to register a new application, which we can then use to request an OAuth token. This is done by making a `POST` request to the `/api/v1/apps` endpoint. Replace `your_app_name` in the code below:
```

docs/api/authentication.md
```diff
@@ -0,0 +1,116 @@
...
+
+```bash
+curl -X POST 'https://your.instance.url/api/v1/apps' \ 
+  -H 'Content-Type:application/json' \
+  -d '{
+      "client_name": your_app_name,
```
tsmethurst, line 13: I'd wrap this as a string as well so that callers don't try to provide it as a bare value.

```suggestion
      "client_name": "your_app_name",
```

docs/api/authentication.md
```diff
@@ -0,0 +1,116 @@
...
+Note that `scopes` can be any space-separated combination of:
+
+- read
+- write
+- follow
+- push
+- admin
+
+!!! warning
+    Be **very** careful what scopes you grant to your application.
```
tsmethurst, line 30: To avoid lulling people into a false sense of security, I think it would be a good idea to add a note here that GoToSocial does not currently support proper scopes on tokens. This is something we still need to add during the beta phase (this year).

fyrfli: So, something along the lines of:

"While GoToSocial does not currently support properly scoped tokens, it is still good practice to grant your application the lowest level permissions it needs to do its job. e.g. if your application won't be making posts, limit it to scope=read."

docs/api/authentication.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+The string "urn:ietf:wg:oauth:2.0:oob" is an indication of what is known as out-of-band authentication - a technique used in multi-factor authentication to reduce the number of ways that a bad actor can intrude on the authentication process. In this instance, it allows us to view and manually copy the tokens created to use further in this process.
+
+Note that `scopes` can be any space-separated combination of:
+
+- read
+- write
+- follow
+- push
```
mirabilos, line 26: I can sort of guess what the scopes read, write, follow and admin do, but push not so much. Maybe add a description of the scopes as relating to GtS or (maybe better) link to one?

docs/api/authentication.md
```diff
@@ -0,0 +1,115 @@
...
+If all goes well, you should get your user profile as a response response.
+
+## Final notes
+Thereafter, whenever you make an api call, say to query your notifications, add a Header to your request `Authorization:Bearer your_brand_new_token` like this:
+```bash
+curl --request GET \
+  --url https://your.instance.url/api/v1/notifications \
+  --header 'Authorization: Bearer your_brand_new_token' \
+```
+
```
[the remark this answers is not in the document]
mirabilos: OK, good point.

Is there an API to remove apps yet?

docs/api/authentication.md
fyrfli: Funnily, I copied that directly from the mastodon docs assuming the API is implemented in whole. My mistake. To your specific point: I have no idea how one would implement "push" permission. From the mastodon docs:

> Grants access to [Web Push API subscriptions.](https://docs.joinmastodon.org/methods/push/) Added in Mastodon 2.4.0.

Do we even have that feature/facility/functionality in Gts?

Re-reading the docs, it also seems as if follow is depreacted in favour of "child permissions" instead (read:follows, write:follows, etc.). So I will remove that on next commit as well.

docs/api/authentication.md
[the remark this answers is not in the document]
fyrfli: I have used tinysubversions before for my akkoma install but I agree that that should be an individual decision to use so that we don't expose ourselves as a project.

docs/api/authentication.md
mirabilos: Probably best to align here with what GtS supports, and possibly indicate what was added in which GtS version starting from when it becomes beta or so.

tsmethurst: Description of what scopes do is not in scope for this PR, which is just about how to go through the authorization flow to get a token. And token scopes are not implemented yet anyway, as mentioned in the warning box below. Once they are, we can revisit this bit of the docs and add appropriate links. So I think this is fine for now :)

fyrfli: Which makes me think I should remove "follow and "push" since they are outdated in home mastodon anyway? And leave the rest as is?

Thoughts?

tsmethurst: > I should remove "follow and "push"

Mmm. I thought about it a bit more; we probably could just replace all this stuff about scopes with something like:

```
!!! warning
   GoToSocial does not currently support scoped authorization tokens, so any token you obtain in this process will be able to perform all actions on your behalf. Nevertheless, it is always good practice to grant your application the lowest tier permissions it needs to do its job. e.g. If your application won't be making posts, use scope=read.
   
   In this spirit, "read" is used in the example above, which means that in the future when scoped tokens are supported, the application will be restricted to only being able to do "read" actions.
   
   You can read more about additional planned OAuth security features [right here](https://github.com/superseriousbusiness/gotosocial/issues/2232).
```

Then we're really not giving any false impressions that we support scoped tokens yet. What do you think?

mirabilos: That, but perhaps mention explicitly that that includes administrative access as well if the token is for an admin account. (I thought a read/write token would explicitly not do that but…)

tsmethurst: Edited the above to mention that ^

fyrfli: Alrighty. Apologies for only getting to this now. For some reason notifications stopped coming through either in email or in the app. Applied that change to the warning immediately following the scopes discussion and removed push/follow from the list.

tsmethurst: > Apologies for only getting to this now.

No problem at all, nothing to apologize for :)

tsmethurst:
Merged it! :) Thank you very much for this!

fyrfli:
It was my pleasure! I love this project!

PR #2659, opened 2024-02-18
feat: add Mastodon-compatible HTTP signature fallback

milas:
# Description
On outgoing `GET` requests that are signed (e.g. authorized fetch), if the initial request fails with `401`, try again, but _without_ the query parameters included in the HTTP signature.

This is primarily useful for compatibility with Mastodon; though hopefully this can be removed in the not-too-distant future, as they've started changing their behavior here.

See #894.

## Checklist

### DIFF
--- a/internal/gtscontext/context.go
+++ b/internal/gtscontext/context.go
@@ -136,7 +136,7 @@ func HTTPSignatureVerifier(ctx context.Context) httpsig.VerifierWithOptions {
 
 // SetHTTPSignatureVerifier stores the given http signature verifier and returns the
 // wrapped context. See HTTPSignatureVerifier() for further information on the verifier value.
-func SetHTTPSignatureVerifier(ctx context.Context, verifier httpsig.Verifier) context.Context {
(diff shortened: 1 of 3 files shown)


daenney:
Right, that ought to do it. We've been considering caching which signer is used for a domain so we don't end up computing them twice all the time. But with Mastodon having updated I think the first one including the query string should be quickly becoming the most common path so maybe we don't need to bother with the caching at all.

NyaaaWhatsUpDoc:
code looks good great to me :) nice work!

tsmethurst:
Thank you very much for this! :)

milas:
Woo! That was fast :)

FYI I've been running this on my single-user instance for a while now (changes were originally off 0.12) and threads have looked better ever since; I use Elk on Web and Tusky on mobile. My original changeset also had copious verbose logging and I analyzed it _very_ scientifically (aka `kubectl logs -f` and 🙏🏻) and didn't see any weirdness.

Issue #2273, opened 2023-10-17
[bug] Empty timeline pages can't be properly handled by some clients


benjaminbhollon:
### Describe the bug with a clear and concise description of what the bug is.

While the recent release candidate fixes some inconsistent timeline paging by providing a `next` link in the header even if the current page is empty, some clients aren't handling this correctly. Specifically, `tut` and `Semaphore` so far are not allowing you to scroll past an empty timeline page.

In contrast, someone using Tusky *is* able to scroll indefinitely.

I'm expecting this is due to clients ignoring the `link` header and using the ID of the most recent successfully fetched post as the `maxid` field, leading to repeatedly requesting the same empty page instead of continuing past it.

### What's your GoToSocial Version?

v0.12.0-rc1

### GoToSocial Arch

arm64 binary

### What happened?

On scrolling through the local timeline on `alpha.polymaths.social` with some clients, the timeline ends unexpectedly.

### What you expected to happen?

Being able to continue to scroll through the local timeline.

### How to reproduce it?

Use Semaphore and try to scroll through the local timeline; if necessary, I'm happy to provide an account on my instance since I'm aware most GoToSocial instances are single-user and don't really use the local timeline.

### Anything else we need to know?

I'm happy to provide any diagnostics you need. Hopefully we can get this fixed before the full release of Spaghetti Sloth comes out? Would really be frustrating to have to wait another whole release cycle. I get that development pace may or may not fit that, though. If necessary, we can find clients that work.

benjaminbhollon:
Tested some more clients:

- phanpy.social and elk.zone let you keep scrolling, but feed you federated posts, not local posts? Prett odd behavior imo, not sure what's causing it.
- Ice Cubes stops at the blank page, like tut and semaphore do.
- the toot CLI works perfectlye letting you scroll as far as you like and only showing local posts.

tsmethurst:
This feels to me like a client issue, since we are returning the link headers now per the mastodon API. We could try to resolve this in GtS by not returning until we've paged down a few times more to see if there's more posts ahead, but this is always going to be imprecise since we'll never know how far we need to page to find posts that aren't filtered out.

That said, since semaphore and pinafore and the like aren't really maintained anymore, perhaps we ought to see if we can do something from GtS's side..

>  feed you federated posts, not local posts

What on earth :')

tsmethurst:
Oh, the phanpy issue is this: https://github.com/superseriousbusiness/gotosocial/issues/2253

[...]

PR #2670, opened 2024-02-19
[bugfix] Account timeline: exclude self-replies that mention other accounts

VyrCossont:
# Description

This pull request excludes self-replies that mention other accounts when fetching an account's timeline with replies excluded. Normally, self-replies should be included, but it's possible to reply to one of your own replies to somebody else, and thus create a reply that mentions someone else but is included anyway, which looks weird. (I thought this was a Feditext client bug when I first noticed it.)

It doesn't specially handle the case where an account explicitly mentions themselves. (Most clients only include mentions for *other* accounts when replying to an account's own post.) Such posts will now be excluded from an account's timeline when replies are excluded. I could see a case for doing that either way (we could check that `mentions` contains a single entry that is the timeline account's ID), but this is the simpler option.

## Impact

Cosmetic. The bug doesn't expose any statuses that wouldn't be visible anyway. But it is widespread, potentially affecting any account timeline viewed through a GtS instance.

## Checklist

### DIFF
--- a/internal/db/bundb/account.go
+++ b/internal/db/bundb/account.go
@@ -604,13 +604,16 @@ func (a *accountDB) GetAccountStatuses(ctx context.Context, accountID string, li
 		Where("? = ?", bun.Ident("status.account_id"), accountID)
 
 	if excludeReplies {
-		q = q.WhereGroup(" AND ", func(*bun.SelectQuery) *bun.SelectQuery {
+		q = q.WhereGroup(" AND ", func(q *bun.SelectQuery) *bun.SelectQuery {
 			return q.
 				// Do include self replies (threads), but
 				// don't include replies to other people.
 				Where("? = ?", bun.Ident("status.in_reply_to_account_id"), accountID).
 				WhereOr("? IS NULL", bun.Ident("status.in_reply_to_uri"))
 		})
+		// Don't include replies that mention other people:
+		// for example, an account's reply to its own reply to someone else.
+		q = whereArrayIsNullOrEmpty(q, bun.Ident("status.mentions"))
 	}
 
 	if excludeReblogs {
--- a/internal/db/bundb/account_test.go
+++ b/internal/db/bundb/account_test.go
@@ -22,6 +22,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"errors"
+	"fmt"
 	"reflect"
 	"strings"
 	"testing"
@@ -32,6 +33,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/db"
 	"github.com/superseriousbusiness/gotosocial/internal/db/bundb"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/util"
 	"github.com/uptrace/bun"
 )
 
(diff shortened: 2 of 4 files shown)


internal/db/bundb/account.go
```diff
@@ -604,13 +604,16 @@ func (a *accountDB) GetAccountStatuses(ctx context.Context, accountID string, li
...
+		q = q.WhereGroup(" AND ", func(q *bun.SelectQuery) *bun.SelectQuery {
 			return q.
 				// Do include self replies (threads), but
 				// don't include replies to other people.
 				Where("? = ?", bun.Ident("status.in_reply_to_account_id"), accountID).
 				WhereOr("? IS NULL", bun.Ident("status.in_reply_to_uri"))
 		})
+		// Don't include replies that mention other people:
+		// for example, an account's reply to its own reply to someone else.
+		q = whereArrayIsNullOrEmpty(q, bun.Ident("status.mentions"))
```
tsmethurst, line 616: Since we don't have an index on `status.mentions`, do you know if this would add a lot of extra latency to these requests? It might be good to have some query analysis comparisons here from before + after this change, just so we can have a peek at what sqlite and postgres each do with their query planning.

VyrCossont: I'll pull some query plans. If an index is needed, we can probably use an expression index for these queries.

VyrCossont: PostgreSQL:

```sql
SELECT "status"."id"
FROM "statuses" AS "status"
WHERE ("status"."account_id" = '01E89AMV7M37APZWDKEW4Q1AFP')
    AND (("status"."in_reply_to_account_id" = '01E89AMV7M37APZWDKEW4Q1AFP') OR ("status"."in_reply_to_uri" IS NULL))
    AND (("status"."mentions" IS NULL) OR (CARDINALITY("status"."mentions") = 0))
    AND ("status"."id" < 'ZZZZZZZZZZZZZZZZZZZZZZZZZZ')
ORDER BY "status"."id"
DESC LIMIT 20
```

```
Limit  (cost=6481.15..6481.16 rows=4 width=27)
  ->  Sort  (cost=6481.15..6481.16 rows=4 width=27)
        Sort Key: id DESC
        ->  Bitmap Heap Scan on statuses status  (cost=29.94..6481.11 rows=4 width=27)
              Recheck Cond: (account_id = '01E89AMV7M37APZWDKEW4Q1AFP'::bpchar)
              Filter: (((in_reply_to_account_id = '01E89AMV7M37APZWDKEW4Q1AFP'::bpchar) OR (in_reply_to_uri IS NULL)) AND (id < 'ZZZZZZZZZZZZZZZZZZZZZZZZZZ'::bpchar) AND ((mentions IS NULL) OR (cardinality(mentions) = 0)))
              ->  Bitmap Index Scan on statuses_account_id_idx  (cost=0.00..29.94 rows=1802 width=0)
                    Index Cond: (account_id = '01E89AMV7M37APZWDKEW4Q1AFP'::bpchar)
```

VyrCossont: If I wanted this to go Really Fast for common timeline queries, I'd might try denormalizing: add boolean columns corresponding to timeline API filter params, and build indexes including those. The check for whether something is a top-level post or a self-reply with no mentions can thus be done when the status is saved or updated, getting us from four conditions, one of which depends on the requested account ID, down to just one.

NyaaaWhatsUpDoc: how does this look on the SQLite side of things? 🤔

VyrCossont: Still working on that. Test suite is locking up so I think I did something it didn't like.

VyrCossont: I get the same query plan in SQLite with or without the patch:

```
QUERY PLAN
`--SEARCH status USING INDEX sqlite_autoindex_statuses_1 (id<?)
```

VyrCossont:
The index works on PG and speeds up queries like the sample to ~5× slower than the original (0.7 ms to 4 ms). Unindexed, it's another order of magnitude slower, 30 ms. That's on production data for a small instance.

For SQLite, I don't have a small instance's worth of test data to play with, so I've only tested using the test rig. The index is created successfully, but with test rig data, SQLite ignores the new index and uses an existing index for both pre-patch and post-patch queries, and there's no measurable difference in performance. I'm not sure if this is an artifact of the small data size or if SQLite's query planner doesn't recognize that it can use that index.

If someone wants to try it on a copy of their production SQLite database, the command to create the index would be:

```sql
CREATE INDEX IF NOT EXISTS "statuses_mentions_is_null_or_empty_idx" ON "statuses" (
    account_id,
    in_reply_to_account_id,
    in_reply_to_uri,
    (mentions IS NULL OR json_array_length(mentions) = 0),
    id DESC
);
```

internal/db/bundb/account.go
daenney: I think this looks sensible. I realise the old query is much faster in that it completes in 0.1ms, but even 5ms is unlikely to be noticeable by a user. Most of the process of fetching a timeline updated will be dominated by other factors. Unless there's a fear that for a large account this might get expensive, I'm not sure maintaining the extra index is worth it just for the 2ms difference.

daenney:
I wonder if this ends up being faster if we do the filtering of self-replies in process, through something like `slices.DeleteFunc()`, rather than solving this in SQL?

tsmethurst:
> I wonder if this ends up being faster if we do the filtering of self-replies in process, through something like slices.DeleteFunc(), rather than solving this in SQL?

Mmm I was thinking the same. Unfortunately that also increases the chance of hitting bugs like https://github.com/superseriousbusiness/gotosocial/issues/2273, where we end up filtering out everything because there's nothing suitable, set paging links appropriately, but some clients cannot follow/understand empty pages.

tsmethurst:
Thanks for doing the detailed query analysis @VyrCossont :) Given the apparently small difference in performance, and the benefits of changing the query, I'd say it might be worth merging this as-is, with the proviso that we can always add the suggested index in later if we need to.

By that I mean, it's hard to test this kind of a change on a variety of machines and setups, so it's difficult to anticipate all the cases where this kind of query change might mess bugger up performance for some specific setup. But at least if someone later says "these particular queries are really slow on my deployment, what gives?" then we know already what we can probably do to speed it up.

Think it's also worth bearing in mind the context in which a db query takes place. If we were discussing a performance hit for a query selecting a single status, then 5ms difference would be quite serious because it adds up quickly. But for selecting a whole page of IDs, 5ms difference or even more is not really very significant. @daenney and @NyaaaWhatsUpDoc what do you think?

VyrCossont:
@tsmethurst to clarify: the second commit 534dcdf01e265c07b3ad92e43d29d21ff33fe78d adds the index on both supported DBs, and speeds up the modified query on PG significantly: 4 ms vs 30 ms. We should definitely include the index for PG deployments. I just can't figure out if there are circumstances under which the SQLite query planner makes use of it.

daenney:
> Think it's also worth bearing in mind the context in which a db query takes place. If we were discussing a performance hit for a query selecting a single status, then 5ms difference would be quite serious because it adds up quickly. But for selecting a whole page of IDs, 5ms difference or even more is not really very significant. @daenney and @NyaaaWhatsUpDoc what do you think?

Completely agree.

tsmethurst:
After discussion with @VyrCossont in Matrix I pushed a commit to drop a previous index which should be more or less covered by the new one :sunglasses: Gonna squerge this now so we can try it out!

Issue #2678, opened 2024-02-22
[feature/admin] Allow enabling very basic anti-spam heuristics to drop messages coming in via the fedi API


tsmethurst:
It would be good if there was a toggle somewhere on the settings panel (or in the config.yaml) to enable an extremely basic spam filter that would drop messages from remotes if several of the following conditions are met:

1. The message is from someone that nobody on the instance follows.
2. The message is from a brand-new account that the instance hasn't seen before.
3. The message mentions more than one person.
4. The message contains a link and/or the message contains an image.

Messages meeting those conditions could be dropped. This might result in legit messages getting caught in the crossfire, so this setting should probably be off by default, or maybe admins could customize how many conditions should be met before the message is dropped.

NyaaaWhatsUpDoc:
it would be useful if we had a frontend for the built-in header-filtering we currently have to go alongside this

benjaminbhollon:
This would be really handy, if only to turn on temporarily during a spam wave.

PR #2685, opened 2024-02-24
[feature] Add experimental `instance-federation-spam-filter` option

tsmethurst:
# Description

>

This pull request adds some very basic spam filter heuristics that admins can enable to drop messages that look spammy.

It's marked as v. much experimental, and is disabled by default, but may be useful in fighting spam when an instance is really getting smashed by spammers. We'll definitely want to refine this. For example, we might want to check if we've already seen some statuses from an account before (ie., is it brand new).

The PR is split into two main commits, the first one (https://github.com/superseriousbusiness/gotosocial/pull/2685/commits/15df68214dcf2708ddf0d0219b4943b2bfc5329c) just moves the `visibility` stuff into a new `filter` package where we can keep stuff like this, and the second one (https://github.com/superseriousbusiness/gotosocial/commit/14c250859f25313b525481084e68453b690371c6) is where the spam stuff is actually added. Likely easier when doing a review to just review the second commit, since the first one doesn't contain any logic changes.

~~Marking this as a draft until I can add some tests to it.~~

Closes https://github.com/superseriousbusiness/gotosocial/issues/2678

## Checklist


internal/federation/federatingdb/create.go
```diff
@@ -321,25 +318,24 @@ func (f *federatingDB) createStatusable(
 	statusable ap.Statusable,
 	forwarded bool,
 ) error {
-
-	// Check whether we should accept this new status,
-	// we do this BEFORE even handling forwards to us.
-	accept, err := f.shouldAcceptStatusable(ctx,
+	// Check whether this status looks like spam.
+	ok, err := f.spamFilter.StatusableOK(ctx,
```
NyaaaWhatsUpDoc, line 322: how do you feel about moving the status acceptability check back into a separate function again? So have something like `f.spamFilter.BasicCheckStatus(ctx, statusable)`, and then `f.spamFilter.DeepCheckStatus()`, something like that at least (the wording of those examples i gave doesn't feelt great :p). it means we can continue only trace logging on relevancy check fails, but info log on spam check fails.

it also means for the "deeper check" function we can start right off the bat with the config value check and return early which is quite nice.

docs/admin/spam.md
```diff
@@ -0,0 +1,23 @@
...
+
+Messages that are considered to be spam will not be stored on your instance, and will not generate notifications.
+
+!!! warning
+    Spam filters are necessarily imperfect tools, since they will likely catch at least a few legitimate messages in the filter, or indeed fail to catch some messages that *are* spam.
+    
+    Enabling `instance-federation-spam-filter` should be viewed as a "battening down the hatches" option for when the fediverse is facing a spam wave. Under normal circumstances, you will likely want to leave it turned off to avoid filtering out legitimate messages by accident.
+
+!!! tip
+    If you want to check what's being caught by the spam filter (if anything), check your logs for the phrase "looked like spam".
```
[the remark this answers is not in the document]
tsmethurst: Right! Thanks, didn't spot that I'd written one thing and done another. Changed it now to include 'looked like spam' in the command. Reason for this is to avoid catching just the word 'spam', which -- who knows -- could be in an account name or instance name or whatever. Grepping for 'looked like spam' is a lot more specific and unlikely to pick up as many false positives.

internal/federation/federatingdb/create.go
tsmethurst: Ah sure, I'll see if I can fiddle with it a bit!

tsmethurst: Ah... right... I remember now why I did it this way. It's a bit awkward because both of those functions need the prepared mentions. If you call them in order then either you need to extract + prepare the mentions twice (once for each function), or you have the caller prepare the mentions, which is real messy.

An alternative would be to keep it in one function, but have the function return just an error (not a bool and an error), with custom error types like NotRelevant and Spam, and check for those errors when deciding what + how to log. So something like `if gtserror.IsNotRelevant(err) { ... } else if gtserror.IsSpam(err) { ... } else { // real error }`. What do you think?

tsmethurst: I've gone ahead and done the above because it seems OK :)

tsmethurst:
Sorry for all the force pushes, trying to keep the branch clean.

internal/filter/spam/statusable.go
```diff
@@ -0,0 +1,472 @@
...
+	// HEURISTIC 5: Four or fewer people are mentioned,
+	// do we follow (request) at least one of them?
+	// If so, we're probably interested in the message.
+	knowsOne := f.knowsOneMentioned(ctx, receiver, mentions)
+	if knowsOne {
+		return nil
+	}
+
+	// HEURISTIC 6: Are there any media attachments?
+	attachments, _ := ap.ExtractAttachments(statusable)
```
NyaaaWhatsUpDoc, line 145: i think for the moment this is really good, also import to get this out there as some level of protection. but in the future i would like to be able to make these a bit more controllable e.g. by the methods mentioned original in the thread about spam control (more fine-grained toggles / dials for these kinda things).

NyaaaWhatsUpDoc:
really good work! :rocket:

Issue #2307, opened 2023-10-29
[chore] Ensure export/import is still working


tsmethurst:
We haven't tested export + import for a while, and we don't really automate testing for this. We should test it and, if it doesn't work, fix anything we've broken.

novakeith:
Assuming you are talking about the same thing... if you are referring to `gotosocial admin export --path file.csv` then I was having trouble with that the other day, at least on my deployment which is containerized. The import function fails because gotosocial immediately creates an entry for the host (IIRC its a hidden account in the database). As soon I tried importing data I had done with the export command, the import command failed because there was already that hidden entry for the host. And if I tried manually deleting that entry, it was recreated immediately.

 I ended up having to change my host name in the docker-compose.yaml to my.host.com-TEMP, let the server spin up, run the import, then go in to psql and delete the TEMP entry. Then relaunch the gotosocial container with the correct host listed. .

This might be way off topic for the chore you are describing, if so, I am happy to delete and re-open it as a separate issue.

tsmethurst:
Yeah, I think it was seeing your travails on the `#GoToSocial` tag that made me open this issue, wanted to see if I could reproduce and fix it!

darcyamurphy:
In case this is relevant to this issue: when you import a blocklist using the bulk import/export tool, it's possible to get to the step of uploading a file without selecting the radio button on the previous page to say if it's a bulk allow or bulk block. If you do this it fails with a 404 message which is a bit confusing.

tsmethurst:
This issue is about a totally separate thing -- export and import of vital instance data using the CLI.

PR #2693, opened 2024-02-27
[feature] add script to test import / export cycle of a gotosocial instance

NyaaaWhatsUpDoc:
closes https://github.com/superseriousbusiness/gotosocial/issues/2307

### DIFF
--- a/test/import-export.sh
+++ b/test/import-export.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+set -ex
+
+HOST='localhost:8080'
+ADDR='127.0.0.1'
+PORT='8080'
+DB=$(mktemp)
+STORAGE=$(mktemp -d)
+EXPORT="$(mktemp).json"
+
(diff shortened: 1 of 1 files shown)


NyaaaWhatsUpDoc:
not my neatest scripting work but at this point i'm just antsy to get v0.14.0 out :D

test/import-export.sh
```diff
@@ -0,0 +1,85 @@
...
+    # Confirm each of the accounts
+    gotosocial admin account confirm \
+        --username ${username}
+done
+
+# Ensure server runs once
+gotosocial server start &
+
+# Kill server
+sleep 5
```
NyaaaWhatsUpDoc, line 59: these long sleeps just let the server gracefully startup + then shutdown, since we don't know what machine speed we'll be running on and we're dealing with it running async

tsmethurst:
I'll just test this on postgres as well

PR #2698, opened 2024-02-28
[bugfix] Fix Swagger spec and add test script

### LINKED ISSUE #2196: [bug] Errors while generating client from swagger definition

Firstly with validation:

> swagger generate client -f ./docs/api/swagger.yaml -A go-gts
> 2023/09/11 23:47:34 validating spec .../docs/api/swagger.yaml
> The swagger spec at ".../docs/api/swagger.yaml" is invalid against swagger specification 2.0. see errors :
> - "paths./api/v1/reports.post.parameters" must validate one and only one schema (oneOf). Found none valid
> - paths./api/v1/reports.post.parameters.example in body is a forbidden property
> - paths./api/v1/reports.post.parameters.in in body should be one of [header]
> - "paths./api/v1/list.post.parameters" must validate one and only one schema (oneOf). Found none valid
> - paths./api/v1/list.post.parameters.example in body is a forbidden property
> - paths./api/v1/list.post.parameters.in in body should be one of [header]
> - "paths./api/v1/list.put.parameters" must validate one and only one schema (oneOf). Found none valid
> - paths./api/v1/list.put.parameters.example in body is a forbidden property
> - paths./api/v1/list.put.parameters.in in body should be one of [header]
> - "paths./api/v1/accounts/update_credentials.patch.parameters" must validate one and only one schema (oneOf). Found none valid
> - paths./api/v1/accounts/update_credentials.patch.parameters.items.type in body should be one of [string number integer boolean array]
> - paths./api/v1/accounts/update_credentials.patch.parameters.in in body should be one of [header]
> - "paths./api/v1/admin/reports/{id}/resolve.post.parameters" must validate one and only one schema (oneOf). Found none valid
> - paths./api/v1/admin/reports/{id}/resolve.post.parameters.example in body is a forbidden property
> - paths./api/v1/admin/reports/{id}/resolve.post.parameters.in in body should be one of [header]

Then without validation, running into this showstopper:

> $ swagger generate client --skip-validation -f docs/api/swagger.yaml -A go-gts
> 2023/09/11 23:49:55 preprocessing spec with option:  minimal flattening
> 2023/09/11 23:49:56 building a plan for generation
> 2023/09/11 23:49:56 generation target ./
> 2023/09/11 23:49:56 planning definitions (found: 69)
> 2023/09/11 23:49:57 planning operations (found: 89)
> object has no key "adminEmojiCategory"

I can try fix up stuff if someone shares some pointers!

v0.11.x

### GoToSocial Arch

x86_64

### What happened?

Client was not generated.

Client is generated.

### How to reproduce it?

`swagger generate client -f ./docs/api/swagger.yaml -A go-gts`

_No response_

VyrCossont:
# Description

- Fixes all errors and warnings related to validating the Swagger/OpenAPI 2 spec
  - `example` was used in several places where it's not valid
    - these uses have been moved to `Sample:` text in the `description`
  - `createStatus.poll` was an object, but Swagger `formData` parameters can't take an object
    - wrote equivalent operation param docs and removed the Swagger annotations from the API model
    - tests have been updated, and an incompatibility with the Mastodon API fixed
  - numeric range validation params were used to describe strings
    - string length params are now used instead
  - various paths, names, example values were obviously incorrect or missing
  - converted unused models to operation params or removed them where they conflict with existing operation param docs
- Adds a script for checking whether the Swagger spec is up to date and valid

Fixes #2196

## Checklist


VyrCossont:
I'd like to add the new `test/swagger.sh` as a CI check but am not familiar with Drone CI, so any help there would be appreciated.

VyrCossont:
And as a proof of concept, I've implemented https://github.com/VyrCossont/slurp on top of the updated Swagger spec.

internal/api/client/admin/emojicategoriesget.go
```diff
@@ -52,7 +52,7 @@ import (
 //			schema:
 //				type: array
 //				items:
-//					"$ref": "#/definitions/adminEmojiCategory"
+//					"$ref": "#/definitions/emojiCategory"
```
daenney, line 55: Whoops :smile:

internal/api/client/admin/domainkeysexpire.go
```diff
@@ -61,8 +61,7 @@ import (
 //	-
 //		name: domain
 //		in: formData
-//		description: Domain to expire keys for.
-//		example: example.org
+//		description: 'Domain to expire keys for. Sample: example.org'
```
daenney, line 64: Can we change this one to match the other description fields, so no wrapping single quotes and `Sample: ` on a new line?

internal/api/client/lists/listupdate.go
```diff
@@ -57,9 +57,8 @@ import (
 //	-
 //		name: title
 //		type: string
-//		description: Title of this list.
+//		description: 'Title of this list. Sample: Cool People'
```
daenney, line 60: Same here, with putting sample on its own line.

daenney (commented):
Did a quick first pass, looks awesome. Thanks for dealing with this!

For CI, we run the `envparsing` script here: https://github.com/superseriousbusiness/gotosocial/blob/504c4f227f82a44e7851dafb1614af7a944080e7/.drone.yml#L40. I would suggest adding the Swagger one there too.

internal/api/client/admin/emojicategoriesget.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/oauth"
 )
 
-// EmojiCategoriesGETHandler swagger:operation GET /api/v1/admin/custom_emojis/categories emojiCategoriesGet
+// EmojiCategoriesGETHandler swagger:operation GET /api/v1/admin/custom_emojis/categories/{id} emojiCategoriesGet
```
tsmethurst, line 30: I don't think this `/{id}` should be here, unless I'm missing something.

internal/api/client/accounts/accountupdate.go
```diff
@@ -120,12 +120,17 @@ import (
...
 //	-
-//		name: fields_attributes
+//		name: fields_attributes[0][name]
 //		in: formData
-//		description: Profile fields to be added to this account's profile
-//		type: array
-//		items:
-//			type: object
+//		description: Name of profile field to be added to this account's profile.
+//			(The index may be any string; add more indexes to send more fields.)
```
tsmethurst, line 126: Man I forgot how bonkers this part of the API is :')

internal/api/model/status.go
```diff
@@ -179,8 +179,26 @@ type StatusCreateRequest struct {
 	// in: formData
 	MediaIDs []string `form:"media_ids[]" json:"media_ids" xml:"media_ids"`
 	// Poll to include with this status.
-	// in: formData
+	// The SwaggerPoll… parameters should only be used to document this complex parameter,
+	// and should not be used otherwise.
+	// swagger:ignore
```
tsmethurst, line 184: Damn, is there really no way to get swagger to handle this properly? I'm not super keen on adding struct fields in this way purely for documentation purposes, if there's a way to avoid it.

internal/api/model/poll.go
```diff
@@ -78,21 +78,21 @@ type PollRequest struct {
 	// Array of possible answers.
 	// If provided, media_ids cannot be used, and poll[expires_in] must be provided.
 	// name: poll[options]
-	Options []string `form:"options" json:"options" xml:"options"`
+	Options []string `form:"poll[options][]" json:"options" xml:"options"`
```
tsmethurst, line 81: This is an actual API change, right, and not just a documentation change. Does this still work with clients that use form data? Or was is the case that this was broken before and this now fixes it?

VyrCossont: Broken before and this fixes it. I noticed that I wasn't able to construct a form data create status request with a poll that actually worked.

internal/api/model/status.go
VyrCossont: I hate this as much as you do, but Swagger does not make it easy to handle the Rails style where form data and JSON are alternate serializations of the same structure. It's not possible to have two definitions of an HTTP operation with the same path and method but different accepted MIME types. I'm not sure if OpenAPI 3 fixes this.

I actually think it'd be better to make the JSON representation the blessed one used for client code generation and the form representation be the secret one that you have to know about, as opposed to the other way around, which is what we have now. But that's a large change.

VyrCossont: Another alternative is we codify both into the Swagger spec: include an optional body param with the JSON definition, and make all form data parameters optional, noting which ones are *actually* required if you're not using a JSON body.

internal/api/client/admin/emojicategoriesget.go
VyrCossont: You're 100% right, that doesn't make sense. The `id` path parameter below is in error, not the URL itself.

internal/api/model/status.go
tsmethurst: Hmmm. We could also consider just manually write the swagger definition for parameters for POSTing to this endpoint, instead of relying on the parameters being generated by the apimodel in this case. That is, just add the fields here: https://github.com/superseriousbusiness/gotosocial/blob/fcecd0c952d6aa286ab326c4fecd06a42de06977/internal/api/client/statuses/statuscreate.go#L35-L74

In that way, we could write whatever we like there without having to worry about adding unused structs to the apimodel (and just add `swagger:ignore` to this model). We've done this in a few other places in the code too, where models were a bit too complex to be described just by swagger annotations.

What do you think?

VyrCossont: Oh! Yeah, we can absolutely do it by hand, or rather, by running the generator to capture the current state, pasting the parameters into the operation definition, and then removing the dummy fields and adding a `swagger:ignore` annotation.

internal/api/client/statuses/statuscreate.go
```diff
@@ -48,6 +48,143 @@ import (
...
+//			- unlisted
+//			- private
+//			- mutuals_only
+//			- direct
+//		in: formData
+//	-
+//		name: scheduled_at
+//		x-go-name: ScheduledAt
+//		description: |-
+//			ISO 8601 Datetime at which to schedule a status.
```
tsmethurst, line 143: Probs worth adding a mention here that this isn't implemented yet.

tools/tools.go
```diff
@@ -0,0 +1,28 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// tools exists to pull in command-line tools that we need to go get,
+// and is behind a build tag that is otherwise unused and thus only visible
+// to dependency management commands. See https://stackoverflow.com/a/54028731.
```
tsmethurst, line 22: Ah interesting, I wasn't aware of this pattern. So we can vendor our tools but not add them to the build, right?

VyrCossont: Yep! Unless you explicitly build with the `tools` build tag, these shouldn't be included in the build.

NyaaaWhatsUpDoc:
had a look through myself, looks all good to me :) nice one!

NyaaaWhatsUpDoc (approved):

tsmethurst:
Ah, I merged main into this after merging https://github.com/superseriousbusiness/gotosocial/pull/2594 and now the swagger test fails :D Do you fancy fixing that @VyrCossont or should I?

VyrCossont:
@tsmethurst from the test output, I *think* they're mostly simple fixes. If you want to fix them, go ahead, else I can take a look when I'm formally awake in a few hours.

VyrCossont:
Poking at it now.

VyrCossont:
Good to go.

decentral1se:
tysm @VyrCossont @tsmethurst! are there plans to maintain a generated Go client lib? Am interested to help!

PR #2700, opened 2024-02-28
[bugfix] Disable the test for long syslog messages over Unix datagram sockets on macOS

VyrCossont:
# Description

This pull request disables the test for long syslog messages over Unix datagram sockets on macOS, where it's known to hang the test runner. I suspect this is an implementation limit for macOS sockets. Given #493, which seems to cap the message size, I'm not sure this is an issue anything's going to hit in production. Also, we don't support macOS for production deployments anyway.

## Checklist

### DIFF
--- a/internal/log/syslog_test.go
+++ b/internal/log/syslog_test.go
@@ -19,12 +19,9 @@ package log_test
 
 import (
 	"fmt"
-	"os"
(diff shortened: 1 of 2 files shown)


tsmethurst:
Thanks :)

Issue #2701, opened 2024-02-28
[bug] It's possible to reply directly to a boost


benjaminbhollon:
### Describe the bug with a clear and concise description of what the bug is.

Somehow, and I have no idea how this happened, someone on my instance replied directly to my boost of a post and not to the post that I was boosting.

### What's your GoToSocial Version?

v0.13.3

### GoToSocial Arch

amd64 Binary

### What you expected to happen?

I'd have expected a reply to a boost to end up as a reply to the post being boosted.

### How to reproduce it?

I genuinely have no idea.Their client was IceCubesApp.

### Anything else we need to know?

This is the thread with the reply to the boost: https://alpha.polymaths.social/@spaceraser/statuses/01HQRFHNSKJCRMPQPK9SK5C6S8

This is the thread with the post that was boosted: https://alpha.polymaths.social/@ivan/statuses/01HQRFCSSB6G1R50W4PSB57MZB

PR #2702, opened 2024-02-29
[bugfix] unwrap boosts when checking in-reply-to status

NyaaaWhatsUpDoc:
# Description

closes https://github.com/superseriousbusiness/gotosocial/issues/2701

## Checklist

### DIFF
--- a/internal/processing/status/create.go
+++ b/internal/processing/status/create.go
@@ -30,6 +30,7 @@ import (
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/id"
+	"github.com/superseriousbusiness/gotosocial/internal/log"
 	"github.com/superseriousbusiness/gotosocial/internal/messages"
 	"github.com/superseriousbusiness/gotosocial/internal/text"
 	"github.com/superseriousbusiness/gotosocial/internal/typeutils"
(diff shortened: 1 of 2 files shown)


daenney:
Took a look at the tests, it's mad about:

```
--- FAIL: TestStatusCreateTestSuite (13.14s)

    --- FAIL: TestStatusCreateTestSuite/TestReplyToNonexistentStatus (1.57s)

        statuscreate_test.go:287: 

            	Error Trace:	/drone/src/internal/api/client/statuses/statuscreate_test.go:287

            	Error:      	Not equal: 

            	            	expected: 400

            	            	actual  : 404

            	Test:       	TestStatusCreateTestSuite/TestReplyToNonexistentStatus

        statuscreate_test.go:293: 

            	Error Trace:	/drone/src/internal/api/client/statuses/statuscreate_test.go:293

            	Error:      	Not equal: 

            	            	expected: "{\"error\":\"Bad Request: cannot reply to status that does not exist\"}"

            	            	actual  : "{\"error\":\"Not Found: target status not found\"}"

            	            	

            	            	Diff:

            	            	--- Expected

            	            	+++ Actual

            	            	@@ -1 +1 @@

            	            	-{"error":"Bad Request: cannot reply to status that does not exist"}

            	            	+{"error":"Not Found: target status not found"}

            	Test:       	TestStatusCreateTestSuite/TestReplyToNonexistentStatus
```

daenney:
So, for HTTP 400, the spec says:

> The 400 (Bad Request) status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

I think the 404 is probably more semantically correct in this case?

internal/processing/status/create.go
```diff
@@ -128,58 +142,49 @@ func (p *Processor) Create(ctx context.Context, requestingAccount *gtsmodel.Acco
...
+		OriginAccount:  requester,
 	})
 
 	if status.Poll != nil {
 		// Now that the status is inserted, and side effects queued,
 		// attempt to schedule an expiry handler for the status poll.
 		if err := p.polls.ScheduleExpiry(ctx, status.Poll); err != nil {
-			err := gtserror.Newf("error scheduling poll expiry: %w", err)
-			return nil, gtserror.NewErrorInternalError(err)
+			log.Errorf(ctx, "error scheduling poll expiry: %v", err)
```
tsmethurst, line 152: Hmm, should this not still be an internal error?

NyaaaWhatsUpDoc: basically the status has already been inserted, the poll has already been inserted, this is only a background expiry worker. it might look a little confusing if we return internal server error here, but the status actually got inserted. doing it like this treats it more like a side effect :)

tsmethurst: Aha got it, OK!

PR #2708, opened 2024-03-02
[chore] expose statuses table row count to prometheus

machiel:
# Description

>

I have exposed a new Prometheus Gauge, that tracks the amount of statuses in the statuses table.

## Checklist

I have not discussed these changes before implementing them, but since the work is already done it might just be easier to discuss whether or not this change is desirable in this MR itself. Hope that's okay.

### DIFF
--- a/internal/db/bundb/status.go
+++ b/internal/db/bundb/status.go
@@ -699,3 +699,15 @@ func (s *statusDB) IsStatusBookmarkedBy(ctx context.Context, status *gtsmodel.St
 		Where("? = ?", bun.Ident("status_bookmark.account_id"), accountID)
 	return exists(ctx, q)
 }
+
+func (s *statusDB) CountStatuses(ctx context.Context) (int, error) {
+	q := s.db.
+		NewSelect().
+		TableExpr("? AS ?", bun.Ident("statuses"), bun.Ident("status"))
+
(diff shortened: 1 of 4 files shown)


machiel:
Perhaps, instead of exposing this gauge, I could rely on something like https://github.com/albertodonato/query-exporter. Adding a guage like this will possibly lead to a rabbit hole of more and more metrics requiring exposure, and thus more code and maintenance.

tsmethurst:
Mm. I'll hold off on this one until after we've released 0.14.0, and then we can maybe have a discussion about what metrics it's sensible to expose, and come up with some kind of schema for doing so. I'm worried if folks keep adding metrics here and there without some kind of overarching plan, it might become a bit messy.

daenney:
The statuses table itself will grow over time, as you federate with things. In and of itself, it's not a metric you can act on. Is it good if it's 0, small, or huge?

We do report the number of local statuses as a metric, as that can tell an operator something about how the instance is used, peak posting times for local users etc. This can be useful to schedule maintenance activity around, help identify a performance issue during active times etc.

I don't think we should be exposing "number of records in a table" metrics, or generic database metrics, through GoToSocial. The query-exporter is a good one if that's of interest to someone.

machiel:
Thanks, that makes sense - closing PR.

Issue #1186, opened 2022-11-30
[investigation] How can we fix the awkward public key situation?


tsmethurst:
As I explained here -- https://github.com/superseriousbusiness/gotosocial/issues/1181#issuecomment-1331877047 -- we have an awkward situation right now as a result of a dodgy early design decision I made to serve a stub of an Actor model at the `/main-key` endpoint.

We should investigate what is required to fix this, and whether it's possible to do it without breaking federation for existing GtS instances.

- Should we serve public keys at a fragment path instead like other implementations do?
- Should we serve just the `publicKey` at the `/main-key` endpoint instead of an Actor stub, a bit like Misskey used to do before they switched to a fragment?
- Can we do either of these things without breaking existing GtS instances?
- Can we do it without carrying remnants of the bad design decision forwards forever?

tsmethurst:
For the record, my preference would be to serve just the `publicKey` at the `/main-key` endpoint, since this is both compliant with AP, and preserves the motivation of the original decision to serve only the minimum necessary data when a public key is requested.

edent:
I've recently received a mention from a GoToSocial 0.13.2 git-f5314c0 server which used `hs2019` as the algorithm.

`keyId="https://example.social/users/x/main-key",algorithm="hs2019",headers="(request-target) host date digest" ...`

My understanding is that the use of hs2019 is deprecated. Looking at [your documentation](https://docs.gotosocial.org/en/latest/federation/federating_with_gotosocial/#http-signatures) no mention is made of it.  Should it explicitly say the algorithm? Or should I try each of sha256, 512, ED25519 until I get a match?

tsmethurst:
You can safely just try sha256. AFAIK all implementations that specify hs2019 just use sha256.

In terms of deprecation, the entire cavage http signature specification is actually deprecated :P It's just used everywhere on the fedi for historical reasons, and because updating it to httpbis across multiple fedi server implementations, without breaking all sorts of things, is going to be very difficult.

cmw:
@tsmethurst would you consider it a way forward to include a configuration option that would enable fragment-based public key sharing? I assume that would allow for federation to work in most cases while you can roll out changes to the `/main-key` endpoint to serve a `publicKey`.

I would also prefer that latter solution to work properly, but in the meantime would also like to be able to federate my newly deployed GTS with my WriteFreely and BookWorm instances.

tsmethurst:
[...]

PR #2710, opened 2024-03-02
[chore] Try to parse public key as both Actor + bare key

tsmethurst:
# Description

>

This pull request is the first part of a fix for https://github.com/superseriousbusiness/gotosocial/issues/1186.

At this point, it *just* adds some logic to try to parse bytes received from a dereference to a public key endpoint as both an Actor, and as a bare Public Key. There are no real logic changes aside from that.

As far as I know, nobody actually serves just the key at the main-key endpoint yet, but it's not a bad idea to get this merged early in the GtS codebase for forwards compatibility reasons. Ie., if we include this change in 0.15.0, and then in 0.16.0 or later change our logic to serve only the key at `/main-key`, then GtS 0.15.0 will be compatible with the new way of doing things introduced in 0.16.0.

~~Draft so we don't squerge this by accident before we release 0.14.0.~~

## Checklist

### DIFF
--- a/internal/ap/extract.go
+++ b/internal/ap/extract.go
@@ -515,9 +515,9 @@ func ExtractURL(i WithURL) (*url.URL, error) {
 	return nil, gtserror.New("no valid URL property found")
 }
 
-// ExtractPublicKey extracts the public key, public key ID, and public
+// ExtractPubKeyFromActor extracts the public key, public key ID, and public
 // key owner ID from an interface, or an error if something goes wrong.
-func ExtractPublicKey(i WithPublicKey) (
+func ExtractPubKeyFromActor(i WithPublicKey) (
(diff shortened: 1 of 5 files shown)


daenney:
I think this looks sensible, though I'm wondering about also handling the `RSA PUBLIC KEY`. Is that a thing anyone uses?

tsmethurst:
> I think this looks sensible, though I'm wondering about also handling the `RSA PUBLIC KEY`. Is that a thing anyone uses?

Ah sorry I forgot to reply to this! I'm honestly not sure. I wrote the original public key parsing code ages ago so I can't remember now why that's in there, but I guess since PUBLIC KEY is first in the list, having RSA PUBLIC KEY second doesn't really do any harm :thinking:

Issue #2503, opened 2024-01-06
[bug] OpenTelemetry Tracing resource merge error


Tsuribori:
### Describe the bug with a clear and concise description of what the bug is.

There is an uncaught error that occurs with tracing in `internal/tracing/tracing.go`:

```go
	r, _ := resource.Merge(
		resource.Default(),
		resource.NewWithAttributes(
			semconv.SchemaURL,
			semconv.ServiceName("GoToSocial"),
		),
	)
```

The error that occurs is "cannot merge resource due to conflicting Schema URL", and is caused by the mismatch of otel library versions being used is `v1.21.0` but the semconv schema version in use is `v1.20.0`.

```go
	semconv "go.opentelemetry.io/otel/semconv/v1.20.0"
	httpconv "go.opentelemetry.io/otel/semconv/v1.20.0/httpconv"
```

Traces are still produced, but Grafana Tempo won't show them at all I guess due to missing attributes and they are shown funky in Jaeger too (they are shown as being produced by `OTLPResourceNoServiceName` instead of `GoToSocial`)

Bumping the semconv schema version used to `v1.21.0` makes it work again but the problem is that the httpconv package was [moved](https://github.com/open-telemetry/opentelemetry-go/issues/4081) to being an *internal* package of `opentelemetry-go-contrib` starting from `v1.21.0`, so the version for it can't be bumped. 

### What's your GoToSocial Version?

This occurs in all versions above v.11.0

### What happened?

OpenTelemetry traces don't show correctly in services that consume them

### What you expected to happen?

That they would be shown correctly like they did in GoToSocial version v0.11.0

### How to reproduce it?

Enable tracing grpc export to Grafana Tempo or Jaeger

tsmethurst:
Ah, thanks!

Tsuribori:
Would using
```go
	httpconv "go.opentelemetry.io/otel/semconv/v1.20.0/httpconv"
	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
```
mismatch be an acceptable quick fix? I guess it technically makes the whole thing schema non-compliant and is not a future-proof solution but it makes traces show up correctly. I haven't found anything to replace `httpconv`, though I'm not really an OTel expert

tsmethurst:
Mmm I'm not sure, possibly... I'll try to have a fiddle with this before we do our next minor release.

daenney:
[...]

PR #2712, opened 2024-03-03
[bugfix/tracing] fix broken tracing due to conflicting schema url

milas:
# Description
The OpenTelemetry SDK is very strict about the schema version when the `Resource` is initialized.

Specifically, different schema versions _CANNOT_ be mixed, and since the default SDK resource (which is merged with the user-defined one) defines a schema URL, the `semconv` imports are really prone to being out-of-sync.

The best way to avoid this is to merge a _schemaless_ resource. This is fine...there's plenty of other ways to get `semconv` out of sync, and the core service attributes (e.g. `service.name`) should not ever change.

Additionally, any errors here are now propagated so that they'll be visible instead of silently swallowed.

## Checklist

### DIFF
--- a/internal/tracing/tracing.go
+++ b/internal/tracing/tracing.go
@@ -34,8 +34,8 @@ import (
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	"go.opentelemetry.io/otel/sdk/trace"
-	semconv "go.opentelemetry.io/otel/semconv/v1.20.0"
-	httpconv "go.opentelemetry.io/otel/semconv/v1.20.0/httpconv"
+	"go.opentelemetry.io/otel/semconv/v1.20.0/httpconv"
+	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
 	oteltrace "go.opentelemetry.io/otel/trace"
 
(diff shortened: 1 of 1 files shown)


tsmethurst:
Ah thanks for this :)

daenney:
Ah yeah, that seems better. Then we should be able to update the otel dependencies too going forwards without problems.

milas:
> So this won't cause a regression to the issue described here https://github.com/superseriousbusiness/gotosocial/issues/2503 ?

Nope! Using schemaless should insulate us from it breaking again.

Longer explanation follows with more than you want to know because the `semconv` situation is...weird, and I've broken & fixed it across several Docker[^1] repos now 🙈 

---

`semconv` are well-known/standardized fields, and while there have been breaking changes, the `service.name` key is pretty fundamental and changing it would break EVERYTHING OTel-related, so I don't think it's a risk.

Another possibility is to add a test for constructing the `Resource`, which would catch the breakage when updating `go.mod` in CI, but that's a bit of a nuisance since it means looking at the OTel SDK source to find what `semconv` version it uses.

[^1]: disclaimer: I work at Docker Inc but my contributions to GoToSocial are my own work and not sponsored or endorsed by them.

PR #2724, opened 2024-03-06
[feature] Process incoming `Move` activity

tsmethurst:
# Description

>

This pull request implements processing of incoming Move activities (ie., remote accounts moving from one account to another).

Part of https://github.com/superseriousbusiness/gotosocial/issues/130

~~Still a draft while I write tests.~~

## Checklist

### DIFF
--- a/docs/federation/federating_with_gotosocial.md
+++ b/docs/federation/federating_with_gotosocial.md
@@ -846,4 +846,44 @@ GoToSocial will only set `movedTo` on outgoing Actors when an account `Move` has
 
 ### `Move` Activity
 
-TODO: document how `Move` works!
+To actually trigger account migrations, GoToSocial uses the `Move` Activity with Actor URI as Object and Target, for example:
+
+```json
+{
+  "@context": "https://www.w3.org/ns/activitystreams",
+  "id": "https://example.org/users/1happyturtle/moves/01HR9FDFCAGM7JYPMWNTFRDQE9",
+  "actor": "https://example.org/users/1happyturtle",
+  "type": "Move",
+  "object": "https://example.org/users/1happyturtle",
+  "target": "https://another-server.com/users/my_new_account_hurray",
+  "to": "https://example.org/users/1happyturtle/followers"
+}
+```
+
+In the above `Move`, Actor `https://example.org/users/1happyturtle` indicates that their account is moving to the URI `https://another-server.com/users/my_new_account_hurray`.
+
+#### Incoming
+
+On receiving a `Move` activity in an Actor's Inbox, GoToSocial will first validate the `Move` by making the following checks:
+
+1. Request was signed by `actor`.
+2. `actor` and `object` fields are the same (you can't `Move` someone else's account).
+3. `actor` has not already moved somewhere else.
+4. `target` is a valid Actor URI: retrievable, not suspended, not already moved, and on a domain that's not defederated by the GoToSocial instance that received the `Move`.
+5. `target` has `alsoKnownAs` set to the `actor` that sent the `Move`. In this example, `https://another-server.com/users/my_new_account_hurray` must have an `alsoKnownAs` value that includes `https://example.org/users/1happyturtle`.
+
+If checks pass, then GoToSocial will process the `Move` by redirecting followers to the new account:
+
+1. Select all followers on this GtS instance of the `actor` doing the `Move`.
+2. For each local follower selected in this way, send a follow request from that follower to the `target` of the `Move`.
+3. Remove all follows targeting the "old" `actor`.
+
+The end result of this is that all followers of `https://example.org/users/1happyturtle` on the receiving instance will now be following `https://another-server.com/users/my_new_account_hurray` instead.
+
+GoToSocial will also remove all follow and pending follow requests owned by the `actor` doing the `Move`; it's up to the `target` account to send follow requests out again.
+
+To prevent potential DoS vectors, GoToSocial enforces a 7-day cooldown on `Move`s. Once an account has successfully moved, GoToSocial will not process further moves from the new account until 7 days after the previous move.
+
+#### Outgoing
+
+Outgoing account migrations use the `Move` Activity in much the same way. When an Actor on a GoToSocial instance wants to `Move`, GtS will first check and validate the `Move` target, and ensure it has an `alsoKnownAs` entry equal to the Actor doing the `Move`. On successful validation, a `Move` message will be sent out to all of the moving Actor's followers, indicating the `target` of the Move. GoToSocial expects remote instances to transfer the `actor`'s followers to the `target`.
--- a/internal/federation/federatingdb/move.go
+++ b/internal/federation/federatingdb/move.go
@@ -26,15 +26,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"time"
 
 	"codeberg.org/gruf/go-logger/v2/level"
 	"github.com/superseriousbusiness/activity/streams/vocab"
 	"github.com/superseriousbusiness/gotosocial/internal/ap"
-	"github.com/superseriousbusiness/gotosocial/internal/db"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
-	"github.com/superseriousbusiness/gotosocial/internal/id"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
 	"github.com/superseriousbusiness/gotosocial/internal/messages"
 )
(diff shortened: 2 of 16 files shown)


internal/processing/workers/fromfediapi_move.go
```diff
@@ -0,0 +1,459 @@
...
+)
+
+// MoveAccount processes the given
+// Move FromFediAPI message:
+//
+//	APObjectType:     "Profile"
+//	APActivityType:   "Move"
+//	GTSModel:         *gtsmodel.Move.
+//	ReceivingAccount: Account of inbox owner receiving the Move.
+func (p *fediAPI) MoveAccount(ctx context.Context, fMsg messages.FromFediAPI) error {
```
tsmethurst, line 44: Note to reviewers: this is where the magic happens :sparkles:

internal/federation/federatingdb/move.go
```diff
@@ -0,0 +1,306 @@
...
+		gtsMove, err = f.state.DB.GetMoveByOriginTarget(ctx, objectStr, targetStr)
+		if err != nil && !errors.Is(err, db.ErrNoEntries) {
+			err := fmt.Errorf(
+				"db error retrieving Move with object %s and target %s: %w",
+				objectStr, targetStr, err,
+			)
+			return gtserror.NewErrorInternalError(err)
+		}
+
+		if gtsMove != nil {
```
tsmethurst, line 239: Note to reviewers: this part isn't really necessary, it's just a courtesy; I imagine many softwares won't store + reuse the same move ID.

docs/federation/federating_with_gotosocial.md
```diff
@@ -846,4 +846,44 @@ GoToSocial will only set `movedTo` on outgoing Actors when an account `Move` has
...
+In the above `Move`, Actor `https://example.org/users/1happyturtle` indicates that their account is moving to the URI `https://another-server.com/users/my_new_account_hurray`.
+
+#### Incoming
+
+On receiving a `Move` activity in an Actor's Inbox, GoToSocial will first validate the `Move` by making the following checks:
+
+1. Request was signed by `actor`.
+2. `actor` and `object` fields are the same (you can't `Move` someone else's account).
+3. `actor` has not already moved somewhere else.
+4. `target` is a valid Actor URI: retrievable, not suspended, not already moved, and on a domain that's not defederated by the GoToSocial instance that received the `Move`.
```
daenney, line 872: I wonder if we should relax this. For example, an admin might suspend a domain because of some moderation issues. Various users on that domain might decide to move away from that instance for that same reason. In that case we wouldn't process the Move activity, which could maybe break things?

docs/federation/federating_with_gotosocial.md
```diff
@@ -846,4 +846,44 @@ GoToSocial will only set `movedTo` on outgoing Actors when an account `Move` has
...
+5. `target` has `alsoKnownAs` set to the `actor` that sent the `Move`. In this example, `https://another-server.com/users/my_new_account_hurray` must have an `alsoKnownAs` value that includes `https://example.org/users/1happyturtle`.
+
+If checks pass, then GoToSocial will process the `Move` by redirecting followers to the new account:
+
+1. Select all followers on this GtS instance of the `actor` doing the `Move`.
+2. For each local follower selected in this way, send a follow request from that follower to the `target` of the `Move`.
+3. Remove all follows targeting the "old" `actor`.
```
daenney, line 879: Does 3 wait on 2? I imagine not and I don't think that's necessary either. Just curious.

daenney (approved):
This looks sensible to me, and the tests seem to agree!

My only thing is, in accept, announce and create we do an explicit requestingAcct.IsMoving() and I wonder if we can't move that earlier in the federatingDB, in some sort of middleware? There's a bunch of other activity types (that aren't wrapped in those 3) that someone could do and they probably shouldn't be able to do so if they're no longer on this instance.

internal/federation/federatingdb/move.go
```diff
@@ -0,0 +1,306 @@
...
+		if err := f.state.DB.PutMove(ctx, gtsMove); err != nil {
+			err := fmt.Errorf("db error storing move %s: %w", moveURIStr, err)
+			return gtserror.NewErrorInternalError(err)
+		}
+	}
+
+	// If move_id isn't set on the requesting
+	// account yet, set it so other processes
+	// know there's a Move in progress.
+	if requestingAcct.MoveID != gtsMove.ID {
```
NyaaaWhatsUpDoc, line 285: i think this function might also require a fedi lock given the setting of flags in the database all along this path that signal other attempts not to happen

NyaaaWhatsUpDoc: (this function in general, not just this part. as for where, opening the lock just before the last attempt database get call should do it imo)

tsmethurst: All these federatingdb functions get lock called on them by the go-fed stuff I believe, with the key of the lock being the uri of the activity.

internal/processing/workers/fromfediapi_move.go
```diff
@@ -0,0 +1,459 @@
...
+		return nil
+	}
+
+	// Account to which the Move is taking place.
+	var (
+		targetAcct     *gtsmodel.Account
+		targetAcctable ap.Accountable
+	)
+
+	if targetAcctURI.Host == config.GetHost() {
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: (though would it worth instead retrieving this from the federatormsg by setting target / receipient ? )

internal/processing/workers/fromfediapi_move.go
```diff
@@ -0,0 +1,459 @@
...
+	// Whatever happened above, error or
+	// not, we've just at least attempted
+	// the Move so we'll need to update it.
+	move.AttemptedAt = time.Now()
+	updateColumns := []string{"attempted_at"}
+
+	if err := errs.Combine(); err != nil {
+		// We tried to process but
+		// didn't succeed completely.
+		l.Infof("one or more errors procesing Move side effects: %v", err)
```
NyaaaWhatsUpDoc, line 286: did you mean to put error level? :p

NyaaaWhatsUpDoc: though thinking about it, i think it might be a bit easier to read in the logs if instead of using a multi error you instead just put a `l.Errorf("error .... : %v", err)`in each of the error checking clauses above so they're separate log entries

internal/processing/workers/fromfediapi_move.go
```diff
@@ -0,0 +1,459 @@
...
+	// accounts on our instance should be removed.
+	//
+	// We should rely on the target instance
+	// to send out new follows from targetAcct.
+	following, err := p.state.DB.GetAccountLocalFollows(
+		gtscontext.SetBarebones(ctx),
+		originAcct.ID,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		errs.Appendf(
```
NyaaaWhatsUpDoc, line 425: same re: above about just being a regular / single error at this check

internal/federation/federatingdb/move.go
NyaaaWhatsUpDoc: but surely we want to be using the same lock as the one used in the fediapi worker? given this function actually does most of the gatekeeping for whether the move gets queued / acted on, i think the fedi lock here is the most important location surely?

tsmethurst: Mmm, it's a bit tricky in this case to preserve the lock across functions. But I could move the gatekeeping part of it (related to last attempted, last succeeded etc) from here back to the processor. Would that make more sense?

docs/federation/federating_with_gotosocial.md
tsmethurst: Mmm, right now there's not really a good way of doing this, since http calls from blocked domains return 403 at basically the first hurdle (way up high in the middleware chain). Might be a good one for a future PR though.

internal/processing/workers/fromfediapi_move.go
```diff
@@ -0,0 +1,459 @@
...
+	// Hard force refresh Move origin account to
+	// ensure we most have up-to-date version.
+	//
+	// Note that it's possible for originAcct.MovedTo
+	// to be set to targetAcctURI already without us
+	// having actually processed the Move yet, if we
+	// manage to refresh originAcct here (or further
+	// upstream) *before* an Account Update is sent to
+	// us from their instance.
+	originAcct, _, err = p.federate.RefreshAccount(ctx,
```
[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2724/commits/c6948b8f4afbfbf4d288d4a7baae3203baf689a9

docs/federation/federating_with_gotosocial.md
NyaaaWhatsUpDoc: definitely a good one to keep in mind for the future in case this situation does come up :thinking:

internal/processing/workers/fromfediapi_move.go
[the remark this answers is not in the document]
tsmethurst: Made the changes to GetAccountByURI here: https://github.com/superseriousbusiness/gotosocial/pull/2724/commits/5a6bbe8a92f8e237fc94b889a254270ed6b684ac#diff-67cc37bfef573a1a98bb5ab3eeeeae2d398220b988cbddf88c3337ed2fa9a585R310-R314

internal/federation/federatingdb/move.go
tsmethurst: I moved some stuff around here: https://github.com/superseriousbusiness/gotosocial/pull/2724/commits/5a6bbe8a92f8e237fc94b889a254270ed6b684ac

So now most of the stateful stuff is done behind the lock by the worker function, instead of being done in the federatingdb function. I think that seems ok!

internal/processing/workers/fromfediapi_move.go
```diff
@@ -0,0 +1,459 @@
...
+	//
+	// Select local followers with barebones, since
+	// we only need follow.Account and we can get
+	// that ourselves.
+	followers, err := p.state.DB.GetAccountLocalFollowers(
+		gtscontext.SetBarebones(ctx),
+		originAcct.ID,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		errs.Appendf(
```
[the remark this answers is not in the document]
tsmethurst: Right... I'll rejig the error handling a bit here, it doesn't make a lot of sense :')

internal/processing/workers/fromfediapi_move.go
```diff
@@ -0,0 +1,459 @@
...
+	// If movedToURI is set on originAcct, make
+	// sure it's actually to the intended target.
+	//
+	// If it's not set, that's fine, we don't
+	// really need it. We know by now that the
+	// Move was really sent to us by originAcct.
+	movedToURI := originAcct.MovedToURI
+	if movedToURI != "" &&
+		movedToURI != targetAcctURIStr {
+		l.Infof(
```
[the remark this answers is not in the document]
tsmethurst: I moved this to the federatingDB function instead and now it returns Malformed, so that should do the trick I think! https://github.com/superseriousbusiness/gotosocial/pull/2724/commits/5a6bbe8a92f8e237fc94b889a254270ed6b684ac

docs/federation/federating_with_gotosocial.md
tsmethurst: Yeah, it happens in order so 3 will only happen if 2 is OK.

internal/processing/workers/fromfediapi_move.go
tsmethurst: Alright, I changed the error handling (here and elsewhere) so the called functions log the error and just return a `true` bool if it went OK, since we don't need to bubble the errors up anywhere or store them anywhere here.

internal/processing/workers/fromfediapi_move.go
tsmethurst: Ack, I changed this too (see above)

PR #2731, opened 2024-03-07
[bugfix] add workaround for Xsqlite_interrupt() permanently breaking connection

NyaaaWhatsUpDoc:
# Description

After our `modernc.org/sqlite` version bump v1.28.0 -> v1.29.2 (further narrowed to be changes in v1.29.0 -> v1.29.2) we started experiencing a regresion, in which canceling a context on an ongoing database query in the right way now causes an `Xsqlite_interrupt()` call to be made internally, which returns error code `SQLITE_INTERRUPT` on that connection's running query. Oddly, that then continually gets returned by all subsequent queries on that connection, leaving the instance in a broken state.

We still haven't narrowed this down to an SQLite library change revealing a bug on our end, or whether it's an SQLite library bug.

This is a quick workaround, which works as follows (text copied from matrix dev chat):

"okay i've found a workaround for now. so between v1.29.0 and v1.29.2 is that slightly tweaked interruptOnDone() behaviour, which causes interrupt to (imo, correctly) get called when a context is cancelled to cancel the running query. the issue is that every single query after that point seems to still then return interrupted. so as you thought, maybe that query count isn't being decremented. i don't think it's our code, but i haven't ruled it out yet.

the workaround for now is adding to our sqlite error processor to replace an SQLITE_INTERRUPTED code with driver.ErrBadConn, which hints to the golang sql package that the conn needs to be closed and a new one opened"

## Checklist

### DIFF
--- a/internal/db/bundb/errors.go
+++ b/internal/db/bundb/errors.go
@@ -18,6 +18,7 @@
 package bundb
 
 import (
+	"database/sql/driver"
 	"errors"
 
 	"github.com/jackc/pgx/v5/pgconn"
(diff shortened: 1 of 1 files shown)


daenney (approved):
Thanks for the excellent detective work!

tsmethurst:
Good workaround! :+1:

PR #2750, opened 2024-03-12
[feature] Process outgoing Move from clientAPI

tsmethurst:
# Description

>

This pull request implements actual processing functionality for accounts Moving via the GtS client API:

- Federate Move activity out when a valid Move has been submitted via the API.
- Process local side effects of a Move (redirect local followers to new target).
- Prevent Moved local accounts from taking most write/create style actions via the client API.
- Document the Move procedure.
- Indicate on the web frontend if an account has moved (see image).

Screenshot from 2024-03-11 12-18-52

Still need to write tests for this, but opening the PR already in the meantime.

## Checklist


internal/processing/workers/redirectfollowers.go
```diff
@@ -0,0 +1,129 @@
+// GoToSocial
```
tsmethurst, line 1: I didn't realize I'd need this function for processing local moves as well :facepalm: So this is just copied from the previous PR where it was introduced, and moved so that both the clientAPI and fediAPI workers can access it (similar to the wipestatus util function). No changes aside from that.

internal/state/state.go
```diff
@@ -50,6 +50,12 @@ type State struct {
 	// functions, and by the go-fed/activity library.
 	FedLocks mutexes.MutexMap
 
+	// ClientLocks provides access to this state's
+	// mutex map of per URI client locks.
+	//
+	// Used during account migration actions.
+	ClientLocks mutexes.MutexMap
```
tsmethurst, line 57: I introduced this different set of locks for processing client API actions specifically. Currently it's only used for migration, but we might want to use it for other potentially destructive / wide-ranging actions in the future.

internal/processing/account/move.go
```diff
@@ -70,83 +74,244 @@ func (p *Processor) MoveSelf(
...
 			"target account %s has already Moved somewhere else (%s); "+
 				"you will not be able to Move to that account",
-			targetAccount.URI, targetAccount.MovedToURI,
+			targetAcct.URI, targetAcct.MovedToURI,
 		)
 		return gtserror.NewErrorUnprocessableEntity(err, err.Error())
 	}
 
-	// Everything seems OK, so process the Move.
+	// If a Move has been *attempted* within last 5m,
```
tsmethurst, line 179: Changes here are mostly to just make sure that clients aren't trying to spam a move, and also to properly store a move in the db.

internal/api/client/accounts/note.go
```diff
@@ -81,6 +82,12 @@ func (m *Module) AccountNotePOSTHandler(c *gin.Context) {
 		return
 	}
 
+	if authed.Account.IsMoving() {
+		const text = "your account has Moved or is currently Moving; you cannot take create or update type actions"
```
daenney, line 86: This same conditional + text shows up in every file. Can we move that into a helper?

internal/processing/account/move.go
```diff
@@ -45,13 +48,14 @@ func (p *Processor) MoveSelf(
...
-	movedToURI, err := url.Parse(form.MovedToURI)
+	targetAcctURIStr := form.MovedToURI
+	targetAcctURI, err := url.Parse(form.MovedToURI)
 	if err != nil {
 		err := fmt.Errorf("invalid moved_to_uri provided in account Move request: %w", err)
 		return gtserror.NewErrorBadRequest(err, err.Error())
 	}
 
-	if movedToURI.Scheme != "https" && movedToURI.Scheme != "http" {
+	if targetAcctURI.Scheme != "https" && targetAcctURI.Scheme != "http" {
```
daenney, line 58: Since you can't really move to an `http` URL, as the whole fedi requires TLS, should we check it against `config.GetProtocol` instead? In case you're running in test mode or something. `GenerateURIForMove` also uses `config.GetProtocol`.

internal/uris/uri.go
```diff
@@ -108,6 +109,14 @@ func GenerateURIForBlock(username string, thisBlockID string) string {
 	return fmt.Sprintf("%s://%s/%s/%s/%s/%s", protocol, host, UsersPath, username, BlocksPath, thisBlockID)
 }
 
+// GenerateURIForBlock returns the AP URI for a new block activity -- something like:
```
daenney, line 112:
```suggestion
// GenerateURIForMove returns the AP URI for a new Move activity -- something like:
```

daenney (commented):
Did a quick first pass. Things look pretty sensible and the client API side looks a lot less complicated.

internal/api/client/accounts/note.go
daenney: Even if only moving the `const text` out, so that's defined in one place in `apiutil` or something.

internal/api/client/accounts/statuses.go
```diff
@@ -148,6 +148,13 @@ func (m *Module) AccountStatusesGETHandler(c *gin.Context) {
 		return
 	}
 
+	if authed.Account.IsMoving() && targetAcctID != authed.Account.ID {
```
NyaaaWhatsUpDoc, line 151: is this standard mastodon behaviour? i feel like statuses were still available last i remember (admittedly last i checked being quite a while ago ... :p)

internal/api/client/accounts/note.go
NyaaaWhatsUpDoc: the one part that gets largely repeated are the moving errors created. i don't really feel like the code itself is enough to separate out, but we could definitely define a `apituil.ErrForbiddenAfterMove` gtserror.WithCode type, and `apiutil.ErrNotFoundAfterMove` for the other cases where we return not found

NyaaaWhatsUpDoc: what do you think @tsmethurst ?

internal/api/client/timelines/public.go
```diff
@@ -117,6 +117,12 @@ func (m *Module) PublicTimelineGETHandler(c *gin.Context) {
 		authed, err = oauth.Authed(c, false, false, false, false)
 	} else {
 		authed, err = oauth.Authed(c, true, true, true, true)
+		if authed.Account.IsMoving() {
```
NyaaaWhatsUpDoc, line 120: i think this is probably worth doing *after* the oauth.Authed() error check below

internal/processing/account/move.go
NyaaaWhatsUpDoc: i think doing this would actually break some of our tests (as we use a mix of http / https test domains).

honestly i don't think it's much of an attack vector (if any), and the check will be faster this way than going through a read mutex lock

internal/api/client/streaming/stream.go
```diff
@@ -185,6 +186,12 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
 		account = authed.Account
 	}
 
+	if account.IsMoving() {
```
NyaaaWhatsUpDoc, line 189: would this be classed as a create / update action?

NyaaaWhatsUpDoc: or maybe just return a 404 here or something :thinking:

internal/processing/workers/redirectfollowers.go
```diff
@@ -0,0 +1,129 @@
...
+//
+// Return bool will be true if all goes OK.
+type redirectFollowers func(
+	ctx context.Context,
+	originAcct *gtsmodel.Account,
+	targetAcct *gtsmodel.Account,
+) bool
+
+// redirectFollowersF returns a redirectFollowers util function.
+func redirectFollowersF(state *state.State, account *account.Processor) redirectFollowers {
```
NyaaaWhatsUpDoc, line 47: instead of using a function ptr, is it worth instead just using a shared struct type with these methods attached that we just share a ptr to in each of the necessary worker structs requiring them?

internal/api/client/accounts/note.go
tsmethurst: Ah sure, sounds like a good idea :)

internal/processing/account/move.go
```diff
@@ -70,83 +74,244 @@ func (p *Processor) MoveSelf(
...
-			account.URI, account.MovedToURI, form.MovedToURI,
-		)
-		return gtserror.NewErrorUnprocessableEntity(err, err.Error())
-	}
+	// Next steps involve checking + setting
+	// state that might get messed up if a
+	// client triggers this function twice
+	// in quick succession, so get a lock on
+	// this account's moves.
+	lockKey := originAcct.URI + "/moves"
```
NyaaaWhatsUpDoc, line 115: it's probably just worth locking by the account URI. if we do introduce using this elsewhere, i think a move would probably constitute a whole-account lock type situation imo

internal/api/client/timelines/public.go
tsmethurst: Right! Well spotted, I'll fix.

internal/processing/workers/redirectfollowers.go
tsmethurst: Mmm yes that makes sense. I'll change it :)

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2750/commits/1b212b407288c26061893691ae91786b9bd78cfa

internal/processing/account/move.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2750/commits/f7a336205d91359bcb0a10ef907d01f66c5be9f1

internal/api/client/timelines/public.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2750/commits/f899017b6835ddf8a87ab6aa41690d13ab13d943

PR #2756, opened 2024-03-13
[docs] Add database maintenance section; update info message on ANALYZE run (sqlite)

tsmethurst:
# Description

>

~~The full ANAYLZE command we run for SQLite after each DB migration is not really necessary, and takes bloody ages (minutes and minutes on my 1cpu 1gb ram machine). This PR updates the post-migration ANALYZE command to use an analysis limit of 10000, which should be plenty (for comparison, we use 1000 for on-close callbacks).~~

Correction to the above: I've been messing about with running partial and full analyzes, and it turns out that running `PRAGMA anaylsis_limit=10000; ANALYZE;` actually doesn't cut it for the purposes of helping the query optimizer use the correct indexes.

Here's the contents of `sqlite_stat1` for statuses after a full `ANALYZE`:

image

And here's the same after a partial `ANALYZE`:

image

Huge difference in emphasis across some of the important indexes.

Running the statuses web view query after a full analyze uses the intended index:

image

Where running the same query after a partial analyze causes the query planner to not use correct indexes:

image

So in other words, we should actually keep that full analyze after all!

Also adds some v. basic docs about db maintenance, for SQLite at least.

## Checklist

### DIFF
--- a/docs/admin/database_maintenance.md
+++ b/docs/admin/database_maintenance.md
@@ -1,0 +1,35 @@
+# Database Maintenance
+
+Regardless of whether you choose to run GoToSocial with SQLite or Postgres, you may need to occasionally take maintenance steps to keep your database running well.
+
+## SQLite
+
+### Analyze / Optimize
+
+Following [SQLite best practice](https://sqlite.org/lang_analyze.html#recommended_usage_pattern), GoToSocial runs the `optimize` SQLite pragma with `analysis_limit=1000` on closing database connections to keep index information up to date.
+
+After each database migration, GoToSocial will also run `ANALYZE` with `analysis_limit=10000` to ensure that any indexes added or removed by migrations are taken into account.
+
+As such, in normal circumstances, you should not need to run manual `ANALYZE` commands against your SQLite database file.
+
+However, if you interrupted a previous `ANALYZE` command, it could be the case that query optimizer data stored in SQLite's internal tables has been removed, and you notice that queries are running remarkably slowly.
+
+If this is the case, you can try manually running an `ANALYZE` command in the SQLite CLI tool, by entering: `PRAGMA analysis_limit=10000; ANALYZE;`.
+
+It is not necessary to run a full analyze, an approximate analyze will do. [See here](https://sqlite.org/lang_analyze.html#approximate_analyze_for_large_databases) for more info.
+
+### Vacuum
+
+GoToSocial does not currently enable auto-vacuum for SQLite. As such, you may want to periodically (eg., every few months) run a `VACUUM` command on your SQLite database, to repack the database file to an optimal size.
+
+You can see lots of information about the `VACUUM` command [here](https://sqlite.org/lang_vacuum.html).
+
+The basic steps are:
+
+1. Stop GoToSocial.
+2. Run `VACUUM` using the SQLite CLI tool (this may take quite a few minutes depending on the size of your database file).
+3. Start GoToSocial.
+
+## Postgres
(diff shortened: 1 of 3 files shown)


NyaaaWhatsUpDoc:
if you're now going to _not_ limit the size of the analyze run and instead using this for documentation changes, i think it'd be worth renaming the PR

tsmethurst:
Will do yeah :)

internal/db/bundb/bundb.go
```diff
@@ -112,7 +112,10 @@ func doMigration(ctx context.Context, db *bun.DB) error {
 	log.Infof(ctx, "MIGRATED DATABASE TO %s", group)
 
 	if db.Dialect().Name() == dialect.SQLite {
-		log.Info(ctx, "running ANALYZE to update table and index statistics")
+		log.Info(ctx,
+			"running ANALYZE to update table and index statistics; this may take 10 minutes or "+
```
daenney, line 116: My database has never had this take more than 30s or so. Maybe soften this a bit?

tsmethurst: There, how about that now? My ANALYZE usually takes about 4 minutes or something (7gb db file, 1-cpu), but I know for folks on RPI or other smol machines it could take longer if they follow a lot of people, so I wanna provide an estimate that's not super fast so they don't assume something is wrong and press ctrl-c.

docs/admin/database_maintenance.md
```diff
@@ -0,0 +1,51 @@
+# Database Maintenance
+
+Regardless of whether you choose to run GoToSocial with SQLite or Postgres, you may need to occasionally take maintenance steps to keep your database running well.
+
+!!! tip
+    
+    Though the maintenance tips provided here are intended by non-destructive, you should still consider backing up your database before manually intervening to do maintenance. That way if you mistype something or accidentally run a bad command, you can simply restore your backup and try again.
```
daenney, line 7:
```suggestion
    Though the maintenance tips provided here are intended to be non-destructive, you should backup your database before manually performing maintenance. That way if you mistype something or accidentally run a bad command, you can restore your backup and try again.
```

docs/admin/database_maintenance.md
```diff
@@ -0,0 +1,51 @@
...
+The `ANALYZE` command may take ~10 minutes depending on your hardware and the size of your database file.
+
+Because of the above automated steps, in normal circumstances you should not need to run manual `ANALYZE` commands against your SQLite database file.
+
+However, if you interrupted a previous `ANALYZE` command, and you notice that queries are running remarkably slowly, it could be the case that the index metadata stored in SQLite's internal tables has been removed or undesirably altered.
+
+If this is the case, you can try manually running a full `ANALYZE` command, by doing the following:
+
+1. Stop GoToSocial.
+2. While connected to your GoToSocial database file in the `sqlite3` shell, run `PRAGMA analysis_limit=0; ANALYZE;` (this may take quite a few minutes).
```
daenney, line 32:
```suggestion
2. While connected to your GoToSocial database file in the `sqlite3` shell, run `ANALYZE;` (this may take quite a few minutes).
```

There's not much point to `PRAGMA analysys_limit=0`, it's the same thing as not setting it at all.

daenney (approved):

internal/db/bundb/bundb.go
daenney: That works! My DB is about a gig but it runs on a machine with NVMe storage so it's done in a flash.

PR #2770, opened 2024-03-19
[chore] Move local account settings to separate db table

tsmethurst:
# Description

>

This pull request migrates account settings for local accounts to a separate table, and adds functionality for selecting + updating those settings.

The reasoning for this: we were storing more and more columns for accounts which would never be used except for local accounts, which widens the accounts table pointlessly. By moving settings for local accounts to another table, we can expand the amount of settings available to users without having to worry about mashing more unused settings into the main accounts table.

Migration tested with a local postgres and a local sqlite, seems OK on both.

## Checklist


daenney (approved):
This makes sense to me. With the exception of the storing the settings separately, the rest of the code is updating how settings are fetched etc. Changes to the tests make sense in light of that.

internal/db/bundb/migrations/20240318115336_account_settings.go
```diff
@@ -0,0 +1,158 @@
...
+					return err
+				}
+			}
+
+			// Drop now unused columns from accounts table.
+			for _, column := range columns {
+				if _, err := tx.
+					NewDropColumn().
+					Table("accounts").
+					Column(column).
```
daenney, line 139: I think `Column` is variadic, so you can pass the `columns` straight in there instead of doing the loop.

internal/gtsmodel/account.go
```diff
@@ -48,45 +48,39 @@ type Account struct {
...
+	SensitizedAt            time.Time        `bun:"type:timestamptz,nullzero"`                                   // When was this account set to have all its media shown as sensitive?
+	SilencedAt              time.Time        `bun:"type:timestamptz,nullzero"`                                   // When was this account silenced (eg., statuses only visible to followers, not public)?
+	SuspendedAt             time.Time        `bun:"type:timestamptz,nullzero"`                                   // When was this account suspended (eg., don't allow it to log in/post, don't accept media/posts from this account)
+	SuspensionOrigin        string           `bun:"type:CHAR(26),nullzero"`                                      // id of the database entry that caused this account to become suspended -- can be an account ID or a domain block ID
+	SettingsID              string           `bun:"type:CHAR(26),nullzero"`                                      // If account is local, ID of the settings model for this account.
```
NyaaaWhatsUpDoc, line 82: i'm just thinking, is it worth the account_settings table having its own id? what if we just looked it up it by the account_id? since it's essentially just an extension of the accounts table only for local accounts?

NyaaaWhatsUpDoc: in PopulateAccount() we can just do: `if account.Domain == "" { /* populate account settings */ }`

also serves the nice purpose of weeding out any of those cases of empty domains when they shouldn't be, or vice-versa

daenney: That's true. We don't really have to compute an ID for it at all, unless you'd somehow want to share settings between users. Which does not seem like a thing anyone would want. Or if you do want to keep the ID, you can use the account ID as the settings ID on the settings column.

tsmethurst: Ah yes I can change that :)

PR #2773, opened 2024-03-21
[chore] bump go structr cache version -> v0.6.0

NyaaaWhatsUpDoc:
# Description

- updates `go-structr` to v0.6.0, which made some incompatible API changes that this PR fixes. this latest version adds the `Queue{}` type to the structr package which allows (in the future) for indexable queues.
- updates all the `cache.GTS{}` caches to be struct value types instead of a mixture of values + ptrs
- adds `structr.Cache{}` wrapping type to simplify updating with the incompatible API changes + allows for much simpler loading of multiple IDs from caches

## Checklist


test.out
```diff
@@ -0,0 +1,120 @@
+?   	github.com/superseriousbusiness/gotosocial/cmd/gotosocial	[no test files]
```
tsmethurst, line 1: Looks like you committed this by accident :P

internal/cache/wrappers.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// StructCache wraps a structr.Cache{} to simple index caching
+// by name (also to ease update to library version that introduced
+// this). (in the future it may be worth embedding these indexes by
+// name under the main database caches struct which would reduce
+// time required to access cached values).
+type StructCache[StructType any] struct {
+	cache structr.Cache[StructType]
+	index map[string]*structr.Index
```
tsmethurst, line 93: Just for my understanding, this index map never shrinks right? Or am I missing something there?

tsmethurst (commented):
Looks good to me! Just the one question about the index map :) And the accidental test output commit ;)

internal/cache/wrappers.go
NyaaaWhatsUpDoc: nope it stays the same size. we alloc it at init with all the indices, and stays the same so no risk of concurrency issues (since it's unprotected) with it shrinking / growing

tsmethurst: Ahhhh okay I understand, thanks :)

tsmethurst:
Squerge when you're happy!

tsmethurst (approved):

tsmethurst:
(Just merged main into this to fix the docs fail)

PR #2777, opened 2024-03-22
[feature] User-selectable preset CSS themes for accounts

tsmethurst:
# Description

>

This pull request adds functionality for user-selectable preset CSS themes, and includes three themes to start with.

Themes are parsed from `web/assets/themes` on startup, and pointers to parsed theme info are stored in the accounts processor, retrieved using the new `/api/v1/accounts/themes` endpoint, and set along with other settings using `/api/v1/accounts/update_credentials`. Theme CSS files themselves are served from the assets folder along with other assets (and therefore get etag stuff too).

Documentation added for how to select a theme, and how to add new themes (for admins).

Some screenshots!

Settings panel:

Settings panel

Dark blurple:

Dark blurple

Light blurple:

Light blurple

Midnight trip:

Midnight trip

## Checklist


daenney:
These themes are awesome! :purple_heart:

daenney:
The one thing I'm wondering is if it would make the implementation simpler if we extracted the current default theme as a "user" theme too, and then in the migration set that value. Then you don't have to deal with things being optional/unset/empty in the DB and the various structs.

That may also make it easier for folks who like the official theme but want to tweak it a bit to work from?

tsmethurst:
> extracted the current default theme as a "user" theme too

Mmm that would be a very good idea for a future change! It's quite a lot of work unfortunately because the official theme is really a combination of a bunch of the basic stuff, and the colors etc are defined quite high up in the cascade, so pulling them out will be a bit of a headache. In other words, it's something we should do as part of a proper refactor I think.

daenney (approved):
Gimme themes :rocket:

internal/api/client/accounts/themesget.go
```diff
@@ -0,0 +1,89 @@
...
+		apiutil.ErrorHandler(c, gtserror.NewErrorNotAcceptable(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	// Retrieve available themes.
+	gtsThemes := m.processor.Account().ThemesGet()
+
+	// Convert themes to apimodel.
+	themes := make([]apimodel.Theme, len(gtsThemes.SortedByTitle))
+	for i, gtsTheme := range gtsThemes.SortedByTitle {
```
NyaaaWhatsUpDoc, line 80: i think it's probably worth moving this to the typeconverter package just for conformity. though probably not something that needs to be a method receiver to the Converter{} type itself, just a regular package function

NyaaaWhatsUpDoc: and i assume probably worth making the account function instead return api theme models too, just to fit with how a lot of it is already done. what do you think??

internal/gtsmodel/accountsettings.go
```diff
@@ -29,6 +29,7 @@ type AccountSettings struct {
 	Sensitive         *bool      `bun:",nullzero,notnull,default:false"`                             // Set posts from this account to sensitive by default?
 	Language          string     `bun:",nullzero,notnull,default:'en'"`                              // What language does this account post in?
 	StatusContentType string     `bun:",nullzero"`                                                   // What is the default format for statuses posted by this account (only for local accounts).
+	Theme             string     `bun:",nullzero"`                                                   // Preset CSS theme filename selected by this Account (empty string if nothing set).
```
NyaaaWhatsUpDoc, line 32: god having this separate account settings table is so nice for these things :')

(i know you added it as a part of adding the themes support, just really appreciating it)

internal/api/client/accounts/themesget.go
tsmethurst: Mmm yes, I will change it!

internal/web/profile.go
```diff
@@ -140,16 +140,31 @@ func (m *Module) profileGETHandler(c *gin.Context) {
 		return
 	}
 
+	// Basic profile stylesheets.
+	stylesheets := []string{
```
NyaaaWhatsUpDoc, line 144: not _that_ important, but you could preallocate the slice here with `stylesheets = make([]string, 5, 6); stylesheets[0] = ...; stylesheets[1] = ...'; etc. `

internal/processing/account/themes.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// ByFileName contains themes retrievable
+	// by their filename eg., `light-blurple.css`.
+	ByFileName map[string]*Theme
+}
+
+// PopulateThemes parses available account CSS
+// themes from the web assets themes directory.
+func PopulateThemes() *Themes {
+	webAssetsAbsFilePath, err := filepath.Abs(config.GetWebAssetBaseDir())
```
NyaaaWhatsUpDoc, line 72: given where we're working from with this, is it worth moving this over to the `internal/web`package for handling?

NyaaaWhatsUpDoc:
very nice work, excited to try this :eyes:

internal/processing/account/themes.go
tsmethurst: I'm not sure. I considered it while writing the PR, but since the endpoint to get available themes is in the `accounts` part of the API, I figured it made sense to keep it here. I'm not 100% happy with it, but I think it's alright for now.

NyaaaWhatsUpDoc: what about keeping the functions for Themes() in the account section, but moving the PopulateThemes() part to the internal/web package? that way it keeps the account local stuff there and the web asset dir stuff in the internal/web package?

internal/web/profile.go
tsmethurst: Mmm I tried and it came out a bit messy, so I compromised by preallocating the total size but still appending. I think that's probably fine for this case :P

internal/web/thread.go
```diff
@@ -138,16 +138,31 @@ func (m *Module) threadGETHandler(c *gin.Context) {
 		return
 	}
 
+	// Basic thread stylesheets.
+	stylesheets := []string{
+		cssFA, cssStatus, cssThread,
```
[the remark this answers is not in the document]
tsmethurst: See above.

internal/processing/account/themes.go
tsmethurst: Hrmm but then the accounts processor would have to call into the web package to get the themes right? That doesn't feel quite right either.

NyaaaWhatsUpDoc: at *some* point maybe a re-think on this then, especially if it gets more complex or if we end up with more user stuff accessing from the assets dir :thinking:

Issue #2707, opened 2024-03-01
[feature] Provide healthcheck endpoints for uptime monitoring


NoraCodes:
### Is your feature request related to a problem ?

I am using https://nine9s.cloud to monitor my GoToSocial instances. Like many uptime monitoring services, they provide the option to use `HEAD` instead of `GET` requests to monitor the up-ness of the service, as they use less bandwidth. However, GoToSocial emits 405 Method Not Allowed for HEAD requests in most cases. (Per #2055).

### Describe the solution you'd like.

~~Ideally, GoToSocial would return `200 OK` for `HEAD` requests, even if only to / specifically.~~

A dedicated "upness" URL, or a documented API route that has a minimal impact and requires no database requests, would be equally useful. I wasn't able to find one in the documentation.

daenney:
I think we should implement this as a separate endpoint(s). The "cloud native" ecosystem seems to be standardising on `/livez` and `/readyz`, with roughly the following meaning:

* `/livez` is this process alive
* `/readyz` is this thing in a state where it should be receiving traffic

(The `z` suffixes are just to reduce the chance of naming collisions with an actual API route and have no meaning otherwise).

`/livez` could be handy since it ought to fail to respond if GoToSocial is somehow broken (like deadlocked) but would otherwise return that it's up. I think that would also solve this request, and we could have that respond to both HEAD and GET. It shouldn't return anything more than an HTTP status code, so even on just a GET it wouldn't result in any data being transferred.

On the `/readyz` endpoint we could perform a simple and very cheap DB query, like fetching the local instance account, to check we're actually in a functioning state (i.e we can handle incoming federation requests and perform our own).

Having those things would also be useful for Docker (using the [`HEALTHCHECK`](https://docs.docker.com/reference/dockerfile/#healthcheck)), as that results in some additionally useful output in the Docker CLI (the `(healthy)` between brackets):

```
[root@eu1 ~]# docker ps
CONTAINER ID   IMAGE                          COMMAND                  CREATED       STATUS                 PORTS     NAMES
xxxxxxxxxxxx   xxxxxxxxxxxx/xxxxxxx:v1.0.0   "python -m xxxxxxx.a…"   5 weeks ago   Up 5 weeks (healthy)             xxxxxxx
```

For folks using an orchestration platform this may also prove useful as those often have facilities to perform these types of checks.

tsmethurst:
[...]

PR #2783, opened 2024-03-25
[feature] Add healthcheck endpoints `/livez` and `/readyz`

tsmethurst:
# Description

>

This pull request adds some simple health check endpoints, and documentation for how to configure a health check.

Closes https://github.com/superseriousbusiness/gotosocial/issues/2707

## Checklist

### DIFF
--- a/cmd/gotosocial/action/server/server.go
+++ b/cmd/gotosocial/action/server/server.go
@@ -309,6 +309,7 @@ var Start action.GTSAction = func(ctx context.Context) error {
 		authModule        = api.NewAuth(dbService, processor, idp, routerSession, sessionName) // auth/oauth paths
 		clientModule      = api.NewClient(dbService, processor)                                // api client endpoints
 		metricsModule     = api.NewMetrics()                                                   // Metrics endpoints
+		healthModule      = api.NewHealth(dbService.Ready)                                     // Health check endpoints
 		fileserverModule  = api.NewFileserver(processor)                                       // fileserver endpoints
 		wellKnownModule   = api.NewWellKnown(processor)                                        // .well-known endpoints
 		nodeInfoModule    = api.NewNodeInfo(processor)                                         // nodeinfo endpoint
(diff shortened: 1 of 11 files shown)


NyaaaWhatsUpDoc:
this otherwise looks good and ready to go, but just one question: would it be worth instead doing a database ping instead of an actual select on a table? or instead just do a select on a table with no columns (just so we don't end up doing any actual searching)

tsmethurst:
Mm, the nice thing about doing a little select instead of a ping is that I don't think pinging catches things like database locked etc. Actually selecting something is a bit more of a representative test of "is this thing on", imo.

> or instead just do a select on a table with no columns (just so we don't end up doing any actual searching).

I'd be happy to change it to that yeah, like just a `SELECT * [...] LIMIT 1` you mean? Is that cheaper?

NyaaaWhatsUpDoc:
> Mm, the nice thing about doing a little select instead of a ping is that I don't think pinging catches things like database locked etc. Actually selecting something is a bit more of a representative test of "is this thing on", imo.
> 
> > or instead just do a select on a table with no columns (just so we don't end up doing any actual searching).
> 
> I'd be happy to change it to that yeah, like just a `SELECT * [...] LIMIT 1` you mean? Is that cheaper?

no i mean you can write a statement that basically returns just none of the columns, so for example here's:
```
sqlite> select null from instances limit 0;
Run Time: real 0.000 user 0.000075 sys 0.000000
```

and then compared to the original:
```
sqlite> select id from instances where domain = "k.iim.gay";
01A0XSYN1D4BG83VMGPEV17PH2
Run Time: real 0.000 user 0.000177 sys 0.000000
```

not a huge difference given they're both absurdly fast, but there is *some*

tsmethurst:
Got it! Changed it here https://github.com/superseriousbusiness/gotosocial/pull/2783/commits/49247130c482fea652d6debba28e8623b17552d0 and it seems to work for both sqlite and postgres :)

Issue #2410, opened 2023-12-03
[bug] Empty public timeline


filipesmedeiros:
### Describe the bug with a clear and concise description of what the bug is.

The public timeline endpoint is returning an empty array :(

### What's your GoToSocial Version?

v0.13.0-rc1

### GoToSocial Arch

amd64

### What happened?

The federated timeline was empty on the client. Went to the console and saw the request was returning `200 []`.

### What you expected to happen?

Some posts to show up (at least mine?). I am federating with other instances.

### How to reproduce it?

Maybe you can try to go to https://filipesm.eu and test it?

### Anything else we need to know?

It was still happening on the previous version (v0.12).

Related, but closed: https://github.com/superseriousbusiness/gotosocial/issues/2163

tsmethurst:
Hmmmmm, bugger. Which client are you using? In the console, do you see any Link headers in the response?

filipesmedeiros:
For some reason this is solved! Ahah At least in my case. I will close it. If it happens again, I'll reopen :)

tsmethurst:
Well, I think it's not fixed fixed, it's just that now probably there are statuses available in your public timeline. It may recur again later. I think this is basically the same issue as https://github.com/superseriousbusiness/gotosocial/issues/2273

benjaminbhollon:
Yeah, I think it's very likely the same root issue we've been having on alpha.polymaths.social. I *thought* Phanpy was one of the clients handling it correctly but don't actually use it myself so that may have been incorrectly reported to me.

filipesmedeiros:
But I don't understand how this is client reporting because the array came empty, when it shouldn't have

benjaminbhollon:
So what's happening (assuming I'm understanding everything correctly, which is not always the case 😅) is that that page of the timeline is empty (since all posts that would've been on the page were filtered out by some condition), but the `link` headers tell clients where the next page is, which *does* have content (probably). Some clients will properly realize that and continue to the next page.

The issue is that Mastodon apparently doesn't return blank pages like GtS does, so clients often ignore the `link` header (which is the technically-correct documented way of paging) and instead just page based on what posts have been loaded, in which case a blank page gets interpreted as the end of the timeline.

I *think* I did that explanation correct. XD

benjaminbhollon:
Another relevant issue: https://github.com/superseriousbusiness/gotosocial/issues/2213

filipesmedeiros:
[...]

PR #2784, opened 2024-03-25
[bugfix] Mitigate empty public/local timeline query pages

tsmethurst:
# Description

>

This pull request updates our public timeline querying to try to avoid the situation where all items from a db query get filtered out, which causes issues like https://github.com/superseriousbusiness/gotosocial/issues/2410 and https://github.com/superseriousbusiness/gotosocial/issues/2273.

Now:

- Always select a bit more than `limit` from the db when doing public timeline queries.
- Reattempt select (paging up or down) to try to get appropriate statuses if previous query resulted in all selected statuses being filtered out (limit 3 queries total).

This won't *completely* fix the problem (it can still happen in edge cases) but it will make it much, much less likely to occur at the expense of doing more/wider db queries.

## Checklist

### DIFF
--- a/internal/processing/timeline/public.go
+++ b/internal/processing/timeline/public.go
@@ -25,50 +25,112 @@ import (
 	apimodel "github.com/superseriousbusiness/gotosocial/internal/api/model"
 	"github.com/superseriousbusiness/gotosocial/internal/db"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
-	"github.com/superseriousbusiness/gotosocial/internal/oauth"
 	"github.com/superseriousbusiness/gotosocial/internal/util"
 )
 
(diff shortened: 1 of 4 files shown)


internal/processing/timeline/public.go
```diff
@@ -25,50 +25,112 @@ import (
...
 	"github.com/superseriousbusiness/gotosocial/internal/util"
 )
 
-func (p *Processor) PublicTimelineGet(ctx context.Context, authed *oauth.Auth, maxID string, sinceID string, minID string, limit int, local bool) (*apimodel.PageableResponse, gtserror.WithCode) {
-	statuses, err := p.state.DB.GetPublicTimeline(ctx, maxID, sinceID, minID, limit, local)
-	if err != nil && !errors.Is(err, db.ErrNoEntries) {
-		err = gtserror.Newf("db error getting statuses: %w", err)
-		return nil, gtserror.NewErrorInternalError(err)
-	}
+func (p *Processor) PublicTimelineGet(
```
NyaaaWhatsUpDoc, line 33: i think next change that happens to the timeline code we gotta use the new paging package, just to stop chance of further diverging

NyaaaWhatsUpDoc:
nice work :+1:

tsmethurst:
Thanks! I just realized while biking home that there's an error in it when paging up :') But I'll fix that in a separate PR.

> i think next change that happens to the timeline code we gotta use the new paging package, just to stop chance of further diverging

Agreed!

PR #2793, opened 2024-04-01
feature: filters v2 server-side warning/hiding

VyrCossont:
# Description

This PR implements Mastodon-4.0-compatible server-side filtering, aka v2 filtering. Statuses matching a filter with a `warn` action will have a `filtered` field added to them so that clients can display a click-through warning with context, or hide them entirely depending on user preferences. (Clients not aware of this field can still use v1 client-side filtering.) Statuses matching a filter with a `hide` action will not be sent to the client at all, which works for all clients regardless of supported filter APIs.

This PR doesn't implement the CRUD API for v2 filters, or allow the creation of `hide` filters through the v1 API. I'll do that in a followup PR. Our internal filter storage is already in the v2 format, so the CRUD API for v1 filters can be used to manage `warn` filters meanwhile.

## Impact

Enables filtering on Phanpy and other clients that support v2 filters but not v1 filters. (Note that Phanpy doesn't have UI for managing filters, so they have to be created through another client using the v1 API.)

Doesn't affect clients that only support v1 filters (Semaphore, Feditext).

## Implementation

Extends `typeutils.StatusToAPIStatus` to take a filter context (public or tag timeline, home or list timeline, notifications, thread, account's statuses, or none) and the filter list from the requesting user, if there is one. About half of `StatusToAPIStatus` calls are from a context where we would apply filters, so adding filters to that function makes sense to me (I could be convinced otherwise if this goes beyond the traditional scope of `typeutils`). Similarly extends `typeutils.NotificationToAPINotification` to take a filter list (the filter context is implicitly notifications).

Applying the filters generates either:
- no match: the `Status` is not modified by filtering.
- for a matching filter with the `warn` filter action, a `FilterResult`, which contains a filter that matched (even if multiple filters match, it can only hold one), and the list of the actual keywords and status IDs within that filter that matched. This gets attached to the returned `Status` so clients can put it behind a warning or hide it themselves.
- for a matching filter with the `hide` filter action, an `ErrHideStatus` error, which signals to `StatusToAPIStatus` callers that the `Status` should not be included in the results at all.

A matching `hide` filter takes action priority over a matching `warn` filter, so if we match a `warn` filter, we will record the match but keep going through the filter list in case there are any matching `hide` filters. Matching a `hide` filter exits early and doesn't bother checking the rest of the filters.

Filters don't apply to a user's own posts.

I've extended `StatusToAPIStatus` callers to:
- include the filter context if applicable: APIs like the user's bookmarks list don't filter, and use a named constant for the empty string to indicate no context.
- fetch the requesting user's filters if applicable: a few APIs don't always require auth, so sometimes there's no user.
- handle `ErrHideStatus` by returning early with no error (when adding a single status to a timeline) or by skipping a status (when returning several statuses).

## Performance

Inherently slower since we're doing some more work server-side. How much slower depends on how many filters a user has.

Filters are fetched from the existing filter cache if possible instead of fetching from the DB, like most DB objects. However, the cache/DB representation isn't quite what the filtering code uses. We could store precompiled regexps and status ID sets per filter context, as well as lookup maps for generating the `filtered` attribute, at the complexity cost of adding another kind of cache and having to invalidate all of it for a user every time any of their filters change.

Alternatively, instead of caching regexps and status IDs across requests, we could fetch the user's cached DB filters per request (as this PR already does), compile them to regexps and status IDs for a given filter context, and tweak `StatusToAPIStatus` to take those regexps and status IDs instead of a filter list. This would save a little work when filtering multiple statuses at once, for example, when returning a timeline page, and isn't as complex as the previous option since there's no second cache.

Either of these could be done in a followup PR.

Expired filters are not purged automatically, [since Mastodon intentionally doesn't do that in case the user wants to renew an expired filter later](https://docs.joinmastodon.org/api/guidelines/#filters). We could add a recurring task to purge very old ones automatically in a followup PR if user filter list size becomes a problem due to expired filters specifically.

## Checklist


VyrCossont:
One more TODO:

[ ] when filtering a status's `content`, we should first convert the HTML to plain text so filters don't match HTML tags by accident.

tsmethurst:
Will get to this as soon as 0.15.0 is out and other pending PRs are squerged :)

tsmethurst:
Apologies for leaving this in review request limbo for a while, got caught up in a necessary refactor of the frontend stuff so I haven't got around to looking at it yet!

internal/api/model/filterv2.go
```diff
@@ -0,0 +1,106 @@
...
+	//	- notifications
+	//	- public
+	//	- thread
+	//	- account
+	// Example: ["home", "public"]
+	Context []FilterContext `json:"context"`
+	// When the filter should no longer be applied. Null if the filter does not expire.
+	//
+	// Example: 2024-02-01T02:57:49Z
+	ExpiresAt *string `json:"expires_at"`
```
daenney, line 50: This should maybe be a `*time.Time` instead. It should deserialise correctly since `time.Time` implements Text(Un)Marshal.

internal/api/model/filterv2.go
```diff
@@ -0,0 +1,106 @@
...
+// FilterAction is the action to apply to statuses matching a filter.
+type FilterAction string
+
+const (
+	// FilterActionWarn filters will include this status in API results with a warning.
+	FilterActionWarn FilterAction = "warn"
+	// FilterActionHide filters will remove this status from API results.
+	FilterActionHide FilterAction = "hide"
+
+	FilterActionNumValues = 2
```
daenney, line 71: This seems unused.

internal/processing/common/status.go
```diff
@@ -184,95 +184,14 @@ func (p *Processor) GetAPIStatus(
 	apiStatus *apimodel.Status,
 	errWithCode gtserror.WithCode,
 ) {
-	apiStatus, err := p.converter.StatusToAPIStatus(ctx, target, requester)
+	apiStatus, err := p.converter.StatusToAPIStatus(ctx, target, requester, "", nil)
```
daenney, line 187:
```suggestion
	apiStatus, err := p.converter.StatusToAPIStatus(ctx, target, requester, custom.FilterContextNone, nil)
```

internal/processing/search/util.go
```diff
@@ -113,7 +113,7 @@ func (p *Processor) packageStatuses(
 			continue
 		}
 
-		apiStatus, err := p.converter.StatusToAPIStatus(ctx, status, requestingAccount)
+		apiStatus, err := p.converter.StatusToAPIStatus(ctx, status, requestingAccount, "", nil)
```
daenney, line 116:
```suggestion
		apiStatus, err := p.converter.StatusToAPIStatus(ctx, status, requestingAccount, custom.FilterContextNone, nil)
```

internal/typeutils/internaltofrontend.go
```diff
@@ -1446,7 +1606,7 @@ func (c *Converter) ReportToAdminAPIReport(ctx context.Context, r *gtsmodel.Repo
 		}
 	}
 	for _, s := range r.Statuses {
-		status, err := c.StatusToAPIStatus(ctx, s, requestingAccount)
+		status, err := c.StatusToAPIStatus(ctx, s, requestingAccount, "", nil)
```
daenney, line 1609:
```suggestion
		status, err := c.StatusToAPIStatus(ctx, s, requestingAccount, custom.FilterContextNone, nil)
```

internal/typeutils/internaltofrontend.go
```diff
@@ -1703,21 +1912,17 @@ func (c *Converter) FilterKeywordToAPIFilterV1(ctx context.Context, filterKeywor
...
-	return &apimodel.FilterV1{
-		// v1 filters have a single keyword each, so we use the filter keyword ID as the v1 filter ID.
-		ID:           filterKeyword.ID,
-		Phrase:       filterKeyword.Keyword,
-		Context:      apiContexts,
-		WholeWord:    util.PtrValueOr(filterKeyword.WholeWord, false),
-		ExpiresAt:    expiresAt,
-		Irreversible: filter.Action == gtsmodel.FilterActionHide,
-	}, nil
+	return ""
```
daenney, line 1925: It might be nice to have a `apimodel.FitlerActionNone` here and return that from the `default` in the switch statement.

internal/typeutils/internaltofrontend_test.go
```diff
@@ -427,7 +428,7 @@ func (suite *InternalToFrontendTestSuite) TestLocalInstanceAccountToFrontendBloc
 func (suite *InternalToFrontendTestSuite) TestStatusToFrontend() {
 	testStatus := suite.testStatuses["admin_account_status_1"]
 	requestingAccount := suite.testAccounts["local_account_1"]
-	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount)
+	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount, "", nil)
```
daenney, line 431:
```suggestion
	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount, custom.FilterContextNone, nil)
```

internal/typeutils/internaltofrontend_test.go
```diff
@@ -537,11 +538,186 @@ func (suite *InternalToFrontendTestSuite) TestStatusToFrontend() {
...
+	)
+	suite.ErrorIs(err, custom.ErrHideStatus)
+}
+
 func (suite *InternalToFrontendTestSuite) TestStatusToFrontendUnknownAttachments() {
 	testStatus := suite.testStatuses["remote_account_2_status_1"]
 	requestingAccount := suite.testAccounts["admin_account"]
 
-	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount)
+	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount, "", nil)
```
daenney, line 720:
```suggestion
	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount, custom.FilterContextNone, nil)
```

internal/typeutils/internaltofrontend_test.go
```diff
@@ -774,7 +950,7 @@ func (suite *InternalToFrontendTestSuite) TestStatusToFrontendUnknownLanguage()
 	*testStatus = *suite.testStatuses["admin_account_status_1"]
 	testStatus.Language = ""
 	requestingAccount := suite.testAccounts["local_account_1"]
-	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount)
+	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount, "", nil)
```
daenney, line 953:
```suggestion
	apiStatus, err := suite.typeconverter.StatusToAPIStatus(context.Background(), testStatus, requestingAccount, custom.FilterContextNone, nil)
```

internal/filter/custom/user.go
```diff
@@ -0,0 +1,45 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Package custom represents custom filters managed by the user through the API.
+package custom
```
daenney, line 19: I would probably rename this package to `user` instead of `custom`, since they're user filters. I think it's implicitly understood they're kinda custom/unique to a user. It reads a bit odd in other import paths to see stuff like `custom.FilterContextNone`, `user.FilterContextNone` makes it a bit more obvious for me as to what kind of filter this is.

daenney (commented):
Honestly this seem alright to me. The tests pass, as do the new ones too so that's also a good sign.

I've left some minor comments, mostly a few places where bare empty strings are being passed in instead of the `FilterContextNone`.

tsmethurst:
Once I've finished fiddling with this bugfix I'm working on, I'll push to this to bring it up to date with main :)

internal/typeutils/internaltofrontend.go
VyrCossont: Agreed.

internal/filter/custom/user.go
VyrCossont: Sure, no problem.

internal/api/model/filterv2.go
tsmethurst: @daenney the reason we have this as a string here is because we serialize these times following the ISO8601 (iirc) format that the Mastodon API uses, by calling this function here: https://github.com/superseriousbusiness/gotosocial/blob/a840f4d49d7d8e6957f1f848e17f4fcc06088412/internal/util/time.go#L27

We could probably get around that at some point by implementing an extension to time.Time or a custom json serializer or something, but for now we're stuck with that util function :P

internal/filter/custom/user.go
VyrCossont: `user` sounds like it could be a filter for users as well as a filter managed by a user, so we ended up going with `status`, imported as `statusfilter` to avoid conflicts with variables named `status`.

VyrCossont:
Ready for any necessary further review.

NyaaaWhatsUpDoc:
great work! thank you so much @VyrCossont !

Issue #286, opened 2021-10-19
[feature] Sign up page for new accounts


tsmethurst:
We should have a sign up page for people who want to apply for an account on an instance.

Right now the flow of creating a new account is either:

1. Ask the admin directly out of band.
2. Do the sign-up flow through the API as scripted here: https://github.com/superseriousbusiness/gotosocial/blob/main/scripts/auth_flow.sh

In addition to these methods, we should have a webpage with a form that prospective users can fill out. This would show the terms and conditions of the instance, the instance description etc, and require a prospective user to give their desired username, email address, password, locale etc. The form should be POSTed to `/api/v1/accounts`.

ForestJohnson:
I was planning on adding this as a new route here: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/api/client/auth/auth.go#L74

I haven't looked into it very far yet, my initial plan was to start by copy and pasting the SignInGETHandler / SignInPOSTHandler  into a RegisterGETHandler and RegisterPOSTHandler and work from there.

mindmeshlink:
I see this was listed as the already open version of #1027 - to be clear then, is the API signup flow one which creates users that still need manually approved by an admin to post on the instance, as a default?

Issue #703, opened 2022-07-09
[feature] Account self-registration


innogrey:
### Is your feature request related to a problem ?

If the question is absurdly stupid, please ignore it. I definitely lack experience with fediverse. The configuration file contains the "accounts-registration-open:" option. I assume that when it is set to 'true', users can self-register accounts. And now the question arises: how? where? The suggested front-ends do not provide the option 'register an account'. Can someone clarify whether the option is unimplemented, or maybe some custom api access is needed here, or have I gone very much astray. Thanks in advance.

### Describe the solution you'd like.

A way to self-register an account

### Describe alternatives you've considered.

No alternatives really

tsmethurst:
Heya! Thanks for opening this :)

The `accounts-registration-open` option is implemented! But unfortunately right now we don't have a sign-up page for new accounts. It's planned but not there yet.

In the meantime, indeed it's something you have to do via the API; check the first option under the 'accounts' section here: https://docs.gotosocial.org/en/latest/api/swagger/. There's also a script here demonstrating the account registration flow: https://github.com/superseriousbusiness/gotosocial/blob/main/scripts/auth_flow.sh

In future we'll make a nice page for this for people who want to join an instance :)

tolstoevsky:
Hi!
Any news around this feature? May we hope it's coming soon? :)

PR #2796, opened 2024-04-01
[feature] New user sign-up via web page

tsmethurst:
# Description

>

This pull request implements a new user signup / registration form, with email confirmation etc.

This is the first part of https://github.com/superseriousbusiness/gotosocial/issues/703 and closes https://github.com/superseriousbusiness/gotosocial/issues/286.

- Adds new templates and handlers in `internal/web` for signup and confirmation email things.
- Updating existing templates to add section on user registration.
- Adds the concept of an "instance application" which will be the default oauth application that new accounts registered via the web form are created by.
- Remove `accounts-approval-required` config option. Manual admin approval will always be required for new accounts.
- Move `reason` from account settings model (where it doesn't really belong) to `user` model (which makes a lot more sense!).
- Create table for denied registrations; this isn't used yet in this pull request, but will be in a subsequent PR where admins can approve/deny registration requests.
- Remove some unused fields from the `user` model.

Some screenshots:

Sign up form

Signed up confirmation

Please confirm email form

Email confirmed

Draft since we shouldn't merge this before 0.15.0 is out.

## Checklist


cmd/gotosocial/action/admin/account/account.go
```diff
@@ -185,9 +185,13 @@ var Confirm action.GTSAction = func(ctx context.Context) error {
...
 	user.Email = user.UnconfirmedEmail
 	user.ConfirmedAt = time.Now()
+	user.SignUpIP = nil
 	return state.DB.UpdateUser(
 		ctx, user,
-		"approved", "email", "confirmed_at",
+		"approved",
+		"email",
+		"confirmed_at",
+		"sign_up_ip",
```
tsmethurst, line 194: Sign up IP will now only stored for denied users (anti-abuse measure), so approving a user registration request clears the stored IP.

testrig/email.go
```diff
@@ -38,6 +39,10 @@ func NewEmailSender(templateBaseDir string, sentEmails map[string]string) email.
 		sendCallback = func(toAddress string, message string) {
 			sentEmails[toAddress] = message
 		}
+	} else {
+		sendCallback = func(toAddress string, message string) {
```
tsmethurst, line 43: Just added this so when you run the test rig you get logs of which emails would have been sent.

web/template/email_confirm.tmpl
```diff
@@ -17,12 +17,14 @@
...
 */ -}}
 
-Hello {{.Username}}!
+Hello {{ .Username -}}!
 
-You are receiving this mail because you've requested an account on {{.InstanceURL}}.
+You are receiving this mail because you've requested an account on {{ .InstanceURL -}}.
 
-We just need to confirm that this is your email address. To confirm your email, paste the following in your browser's address bar:
+To use your account, you must confirm that this is your email address.
```
tsmethurst, line 24: I got rid of the "we" language here because we don't really use it anywhere else.

web/template/finalize.tmpl
```diff
@@ -19,29 +19,35 @@
 
 {{- with . }}
 <main>
-    <form action="/oauth/finalize" method="POST">
-        <h1>Hi {{ .name -}}!</h1>
-        <p>
-            You are about to sign-up to {{ .instance.Title -}}.
-            To ensure the best experience for you, we need you to provide some additional details.
```
tsmethurst, line 26: Same here re: "we" language.

web/template/sign-up.tmpl
```diff
@@ -0,0 +1,93 @@
...
+                <label for="agreement">I have read and accept the <a href="/about#terms">terms and conditions</a> of {{ .instance.Title }}, and I agree to abide by the <a href="/about#rules">instance rules</a>.</label>
+                <input
+                    id="agreement"
+                    type="checkbox"
+                    name="agreement"
+                    required
+                    value="true"
+                >
+            </div>
+            <input type="hidden" name="locale" value="en">
```
tsmethurst, line 88: In future we can have the user select their initial locale from a dropdown, but for now I just whacked `en` in there, since it's our db default anyway.

internal/processing/workers/surfaceemail.go
```diff
@@ -158,3 +128,77 @@ func (s *surface) emailPleaseConfirm(ctx context.Context, user *gtsmodel.User, u
...
+		return gtserror.Newf("error populating user: %w", err)
+	}
+
+	newSignupData := email.NewSignupData{
+		InstanceURL:    instance.URI,
+		InstanceName:   instance.Title,
+		SignupEmail:    newUser.UnconfirmedEmail,
+		SignupUsername: newUser.Account.Username,
+		SignupReason:   newUser.Reason,
+		SignupURL:      "TODO",
```
tsmethurst, line 196: Gonna do this in the next PR when I add stuff to the settings panel for reviewing sign-ups.

example/config.yaml
```diff
@@ -411,11 +411,6 @@ instance-inject-mastodon-version: false
 # Default: true
 accounts-registration-open: true
 
-# Bool. Do sign up requests require approval from an admin/moderator before an account can sign in/use the server?
-# Options: [true, false]
-# Default: true
-accounts-approval-required: true
```
tsmethurst, line 417: Removed this option now so that approval is *always* required. We can use invites later to let admins circumvent this if they really want to, but for now I don't want anyone to end up in a situation where they accidentally leave approval required = false, and end up with people using their instance as a spam vector.

NyaaaWhatsUpDoc: i literally was just thinking about this in bed last night, same braincell :D

internal/db/bundb/migrations/20240401130338_sign_up.go
```diff
@@ -0,0 +1,122 @@
...
+				strings.Contains(e, "SQLSTATE 42701")) {
+				return err
+			}
+		}
+
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			// Get reasons from
+			// account settings.
+			type idReason struct {
+				AccountID string
```
NyaaaWhatsUpDoc, line 48: is it worth copying over the bun tags here? just to make sure all versions of bun going forwards perform this migration in the same way

internal/db/bundb/migrations/20240401130338_sign_up.go
```diff
@@ -0,0 +1,122 @@
...
+					Exec(ctx, &reasons); err != nil {
+					return err
+				}
+			}
+
+			// Remove now-unused column
+			// from account settings.
+			if _, err := tx.
+				NewDropColumn().
+				Table("account_settings").
```
NyaaaWhatsUpDoc, line 77: is it worth separating out from here into a separate statement + transaction and the same way as the above alter table? just because postgres gets fussy and fails the whole tx otherwise

NyaaaWhatsUpDoc: (which is why i'm assuming you did that with the above too)

internal/db/bundb/admin.go
```diff
@@ -331,6 +332,113 @@ func (a *adminDB) CreateInstanceInstance(ctx context.Context) error {
...
+	}
+
+	// Generate new IDs for this
+	// application and its client.
+	protocol := config.GetProtocol()
+	host := config.GetHost()
+	url := protocol + "://" + host
+
+	clientID := instanceAcct.ID
+	clientSecret := uuid.NewString()
```
NyaaaWhatsUpDoc, line 368: is a uuid cryptographically safe enough for a client secret? just double checking. if not could generate some small byte sequence using crypto/rand.Reader{} and base32 encode it using crockford's base32 set (see `internal/middleware/requestid.go` line 39)

internal/db/bundb/admin.go
```diff
@@ -331,6 +332,113 @@ func (a *adminDB) CreateInstanceInstance(ctx context.Context) error {
...
+		ID:     clientID,
+		Secret: clientSecret,
+		Domain: url,
+	}
+
+	// Store it.
+	return a.state.DB.Put(ctx, oc)
+}
+
+func (a *adminDB) GetInstanceApplication(ctx context.Context) (*gtsmodel.Application, error) {
```
NyaaaWhatsUpDoc, line 403: just a note, we need to be very careful to make sure we don't give out this application secret *ever* in the codebase since the client_id is very guessable (being the instance id), and since it has free access to write to accounts

internal/db/bundb/instance.go
```diff
@@ -380,3 +380,33 @@ func (i *instanceDB) GetInstanceModeratorAddresses(ctx context.Context) ([]strin
...
 	return addresses, nil
 }
+
+func (i *instanceDB) GetInstanceModerators(ctx context.Context) ([]*gtsmodel.Account, error) {
+	accountIDs := []string{}
+
+	// Select account IDs of approved, confirmed,
+	// and enabled moderators or admins.
+
+	q := i.db.
```
NyaaaWhatsUpDoc, line 390: just a note: might be worth caching these lists in the future on heavily used / moderated instances

internal/db/bundb/migrations/20240401130338_sign_up.go
```diff
@@ -0,0 +1,122 @@
...
+				return err
+			}
+
+			// Remove now-unused columns from users.
+			for _, column := range []string{
+				"current_sign_in_at",
+				"current_sign_in_ip",
+				"last_sign_in_at",
+				"last_sign_in_ip",
+				"sign_in_count",
```
NyaaaWhatsUpDoc, line 89: is this missing filtered and chosen languages?

internal/processing/user/email.go
```diff
@@ -28,53 +28,82 @@ import (
...
+		user.Account, err = p.state.DB.GetAccountByID(ctx, user.AccountID)
+		if !errors.Is(err, db.ErrNoEntries) {
+			// Real error.
+			return nil, gtserror.NewErrorInternalError(err)
 		}
-		user.Account = a
+
+		// No account found for this user,
+		// or error populating account.
+		return nil, gtserror.NewErrorNotFound(err)
```
NyaaaWhatsUpDoc, line 59: honestly if we can't find the account for a user we've confirmed exists they should probably all be internal errors :p

internal/db/bundb/migrations/20240401130338_sign_up.go
tsmethurst: I think there's probably not much point doing it in this specific case, since we know the `reason` column must exist, as we've just selected from it above.

internal/db/bundb/migrations/20240401130338_sign_up.go
tsmethurst: We're only using `idReason` to select, so I think it should be OK without the tags. The nullzero tag on `reason` only applies to inserts.

internal/db/bundb/admin.go
tsmethurst: It *should* be, it uses `crypto/rand` under the hood: https://pkg.go.dev/github.com/google/uuid#NewRandom

internal/db/bundb/admin.go
tsmethurst: For sure, though note also that an application token obtained using the instance application is only an application token, it would not count as a user access token, you'd have to authorize it by signing in as the instance account to do that.

internal/db/bundb/instance.go
tsmethurst: Agreed! Definitely something we will probably want to add down the line :)

internal/processing/account/create.go
```diff
@@ -32,15 +33,48 @@ import (
...
+// validated and normalized by the caller.
 func (p *Processor) Create(
 	ctx context.Context,
-	appToken oauth2.TokenInfo,
 	app *gtsmodel.Application,
 	form *apimodel.AccountCreateRequest,
-) (*apimodel.Token, gtserror.WithCode) {
+) (*gtsmodel.User, gtserror.WithCode) {
+	const (
+		usersPerDay = 10
```
NyaaaWhatsUpDoc, line 49: good idea

internal/processing/account/create.go
```diff
@@ -32,15 +33,48 @@ import (
...
+		regBacklog  = 20
```
NyaaaWhatsUpDoc, line 50: and again!!

internal/db/bundb/migrations/20240401130338_sign_up.go
tsmethurst: Whoops, yes, well spotted! Will fix that.

internal/processing/account/create.go
```diff
@@ -67,38 +101,63 @@ func (p *Processor) Create(
...
-		PreApproved: !config.GetAccountsApprovalRequired(), // Mark as approved if no approval required.
-		SignUpIP:    form.IP,
-		Locale:      form.Locale,
-		AppID:       app.ID,
+		Username: form.Username,
+		Email:    form.Email,
+		Password: form.Password,
+		Reason:   text.SanitizeToPlaintext(reason),
+		// Mark as approved if no approval required.
+		// PreApproved: !config.GetAccountsApprovalRequired(),
```
NyaaaWhatsUpDoc, line 119: did you mean to leave this commented-out? :thinking: probably worth adding a comment if there was a reason for it

internal/db/bundb/migrations/20240401130338_sign_up.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2796/commits/71c4c6beb0e25dc6bee790a9b59e9824c24f190b

internal/processing/account/create.go
tsmethurst: Nope, just a mistake :') I'll fix it now.

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2796/commits/4bcd1ffeeda746f2a83276aa7a0c439eb310d765

NyaaaWhatsUpDoc:
only the one un-addressed comment now (though not exactly high priority if you don't got the beans), otherwise consider this approved! nice work :rocket:

internal/processing/user/email.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/2796/commits/ff24ea113b106ecd5a3827480f7c65f365b4bdec kapow!

Issue #2769, opened 2024-03-19
[bug] Following Accounts Not Organized by Follow Date


CDN18:
This issue arose from yesterday's discussion in the matrix room.

### Describe the bug with a clear and concise description of what the bug is.

Accounts in my Gotosocial Following list aren't organized by the order in when I followed them. This has been somewhat problematic for me, especially when trying to ascertain whether someone has accepted my follow request or, once accepted, difficulty in locating them in my Following list to add to a specific list.

It seems like an issue about `/api/v1/accounts/{id}/following`

### What's your GoToSocial Version?

v0.14.1-SNAPSHOT+git-51e8879

### GoToSocial Arch

arm64 Docker

### What happened?

Following accounts are not organized by follow date

### What you expected to happen?

Following accounts are sorted by follow date.

### How to reproduce it?

Simply visit oneself's GotoSocial profile and check the followings.

daenney:
When loading them into the cache, we explicitly do so sorted by ID in descending order too. This should result in the newest ones always coming first: https://github.com/superseriousbusiness/gotosocial/blob/65b53660313f0faaa5d0bbb53ddcdb8ff23559d2/internal/db/bundb/relationship.go#L329-L335

Since our IDs are ULIDs and as such are temporally sortable this shouldn't be caused by clients doing some sorting either.

However, running the query myself `select * from follows where account_id = 'REDACTED' order by id DESC;`

```
id|created_at|updated_at|uri|account_id|target_account_id|show_reblogs|notify
```
```csv
FOLLOW_ID|2023-03-11 00:10:09|2023-03-11 00:10:09|URI|ACCOUNT_ID|TARGET_ACCOUNT_ID|1|0
FOLLOW_ID|2023-09-20 18:07:34|2023-09-20 18:07:34|URI|ACCOUNT_ID|TARGET_ACCOUNT_ID|1|0
FOLLOW_ID|2023-08-18 22:16:43|2023-10-14 23:39:41.903336+00:00|URI|ACCOUNT_ID|TARGET_ACCOUNT_ID|0|0
FOLLOW_ID|2023-05-04 12:39:09|2023-05-04 12:39:09|URI|ACCOUNT_ID|TARGET_ACCOUNT_ID|1|0
FOLLOW_ID|2022-11-20 22:19:59|2023-07-16 05:48:50.792319+00:00|URI|ACCOUNT_ID|TARGET_ACCOUNT_ID|0|0
[...]

PR #2801, opened 2024-04-02
[bugfix] Sort follows chronologically

daenney:
# Description

The id on the follows table is not a ULID, but a random ID. Sorting on them results in a completely random order. Instead, sort on created_at, which sould result in a stable and intended sort order.

Fixes: #2769

## Checklist

### DIFF
--- a/internal/federation/federatingdb/following_test.go
+++ b/internal/federation/federatingdb/following_test.go
@@ -47,8 +47,8 @@ func (suite *FollowingTestSuite) TestGetFollowing() {
 	suite.Equal(`{
   "@context": "https://www.w3.org/ns/activitystreams",
   "items": [
-    "http://localhost:8080/users/1happyturtle",
-    "http://localhost:8080/users/admin"
+    "http://localhost:8080/users/admin",
+    "http://localhost:8080/users/1happyturtle"
   ],
   "type": "Collection"
 }`, string(fJson))
--- a/internal/db/bundb/relationship.go
+++ b/internal/db/bundb/relationship.go
@@ -331,7 +331,7 @@ func newSelectFollows(db *bun.DB, accountID string) *bun.SelectQuery {
 		Table("follows").
 		Column("id").
 		Where("? = ?", bun.Ident("account_id"), accountID).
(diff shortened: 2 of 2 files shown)


internal/federation/federatingdb/following_test.go
```diff
@@ -47,8 +47,8 @@ func (suite *FollowingTestSuite) TestGetFollowing() {
 	suite.Equal(`{
   "@context": "https://www.w3.org/ns/activitystreams",
   "items": [
-    "http://localhost:8080/users/1happyturtle",
-    "http://localhost:8080/users/admin"
+    "http://localhost:8080/users/admin",
+    "http://localhost:8080/users/1happyturtle"
```
daenney, line 51: This actually seems wrong to me. The created_at of `1happyturtle` is later than `admin`, so I feel like the original order was correct. But the test failed. If someone knows why we have an inversion here I'm curious.

daenney: Actually helps if you fetch the column you're sorting on, which explains the inversion. Need to fix that :smile:

daenney: Wait, no, that's not necessary. Now I'm back to being confused.

daenney: Oh, is this because `GetFollowsByIDs` needs to also do the sorting?

daenney: Although, do we even want that for federation? This is just returning a Collection, so there's no guarantee of an order anyway. So maybe it's fine?

NyaaaWhatsUpDoc: it's passing the more comprehensive set of paging tests in `./internal/api/client/accounts/follow_test.go`, which creates and tests its own set of follow / following relations instead of just the pre-prepared follows from the test models, which i trust more for confirming that we're maintaining behaviour imo

NyaaaWhatsUpDoc (approved):
looks good to me! though not quite sure what is going on with the test runner... i'll restart it see if that kicks it into gear :p

PR #2804, opened 2024-04-03
[chore] Log less output on failed test

daenney:
# Description

>

This changes the testrig log level to be error by default instead of info. This makes test failures a lot easier to read, as we don't have the parade of info logs for each failure to scroll through.

It speeds up the test suite by a couple of seconds since we need to buffer and flush a lot less messages. On a clean run, so no test failures, it's about a 3s difference on my machine. Depending on the amount of test failures, total time saved can vary.

This also introduces a GTS_TESTRIG_LOG_LEVEL environment variable that we explicitly check for, making it easy to override the log level should we have a need for it. This would be primarily for running locally, and not so much as part of go test.

Lastly, it updates the syslog tests to use log.Error because if the log level is set to error but we call log.Info no message is emitted and we hang indefinitely on the channel read.

## Checklist

### DIFF
--- a/testrig/config.go
+++ b/testrig/config.go
@@ -18,6 +18,7 @@
 package testrig
 
 import (
+	"os"
 	"time"
 
 	"codeberg.org/gruf/go-bytesize"
@@ -33,8 +34,16 @@ func InitTestConfig() {
 	})
 }
 
+func logLevel() string {
+	level := "error"
+	if lv := os.Getenv("GTS_TESTRIG_LOG_LEVEL"); lv != "" {
+		level = lv
+	}
+	return level
+}
+
 var testDefaults = config.Configuration{
-	LogLevel:           "info",
+	LogLevel:           logLevel(),
 	LogTimestampFormat: "02/01/2006 15:04:05.000",
 	LogDbQueries:       true,
 	ApplicationName:    "gotosocial",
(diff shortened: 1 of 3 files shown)


daenney:
This also seems to improve how much data Go temporarily writes to your filesystem on a test failure (which seems to only get removed after the test suite ends). With the log level set to error it's about 10MB less. That's obviously not all logs since we don't log 10MB worth at the info level, but it does seem to help to set a stricter log level.

With multiple test failures this difference does add up, and it's less stuff to write to disk and then flush to a terminal.

testrig/config.go
```diff
@@ -33,8 +34,16 @@ func InitTestConfig() {
 	})
 }
 
+func logLevel() string {
+	level := "error"
+	if lv := os.Getenv("GTS_TESTRIG_LOG_LEVEL"); lv != "" {
```
tsmethurst, line 39: I think you can just change this to `GTS_LOG_LEVEL` since that's the same as the env var used in the actual server. This is similar to what we do here for some other variables: https://github.com/superseriousbusiness/gotosocial/issues/2807

daenney: Ah, fair enough. Wasn't sure if we should reuse it, since the `GTS_*` ones are normally read through Cobra/Viper etc. But I'll switch it.

daenney: Fixed in fee0a88a7900a3ae0ccba3287c307994fc283f50

PR #2810, opened 2024-04-04
[bugfix] defer rows.Close() when using QueryContext

tsmethurst:
Update our SQLite OnBusy handler to close rows when using QueryContext, ensuring that the connection gets returned properly to the pool.


tsmethurst:
Hmm nevermind, as Kim pointed out, we actually return that `rows` as part of a wrapped query, so we shouldn't have to close it ourselves (bun should close it).

Issue #2808, opened 2024-04-04
[bug/sqlite] `sqlite.db-wal` file grows indefinitely without flushing to `sqlite.db`


tsmethurst:
Someone's reported this in the Matrix already too, but didn't open an issue for it yet. Opening it now since I can see the problem on my server as well:

```
-rw-r--r--    1 gotosocial gotosocial 7.1G Apr  3 00:01 sqlite.db
-rw-r--r--    1 gotosocial gotosocial 3.4M Apr  4 16:05 sqlite.db-shm
-rw-r--r--    1 gotosocial gotosocial 1.7G Apr  4 16:05 sqlite.db-wal
```

^^ Here there's a db size of 7.1G, which is fine for an old and widely federating instance, but a wal file of 1.7G, which is much, much larger than it ought to be. This won't cause any immediate problems on this particular server, but it is likely to cause slowdown over time since reading from the wal file is less optimized than reading from the db itself.

The SQLite docs contain some details on why this situation might occur: https://www.sqlite.org/wal.html#avoiding_excessively_large_wal_files

The one that jumps out at me is checkpoint starvation, particularly:

> if a database has many concurrent overlapping readers and there is always at least one active reader, then no checkpoints will be able to complete and hence the WAL file will grow without bound.

We should see if we can figure out why this is happening; are we leaving readers open maybe? Or is there some change in the latest version of SQLite that we're using which disables the automatic checkpoint mechanism, or leaves too many readers open, or something?

We could also try enabling manual periodic checkpointing (once per 30 minutes or so), and see if this helps, though I feel like that might just mask some other underlying problem.

tsmethurst:
This looks very much like this: https://gitlab.com/cznic/sqlite/-/issues/179

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/pull/2811 (for now).

Issue #2774, opened 2024-03-22
[bug/sqlite] Instance can get in a state where every db query returns `database is locked (517)`


tsmethurst:
Difficult to reproduce this one. I noticed last week that my instance (running on https://github.com/superseriousbusiness/gotosocial/commit/0362d49da0d09784f9b974718839f5e69df3bae1 or so) had started responding to most requests with 500 errors. On checking I see lots of stuff like the following:

```
timestamp="17/03/2024 13:54:07.347" func=server.glob..func1.Logger.func13.1 level=ERROR latency="5.039544ms" userAgent="xxx" method=GET statusCode=500 path=/api/v1/accounts/verify_credentials clientIP=xxx.xxx.xxx.xxx errors="Error #01: error converting account: AccountToAPIAccountPublic: error counting statuses: database is locked (517)\n" requestID=xvz7jk4e04001z3j88p0 msg="Internal Server Error: wrote 54B"
```

This state of affairs persisted for about 30 minutes until I restarted the instance, which caused it to start behaving normally again after the restart.

517 corresponds to [sqlite busy](https://www.sqlite.org/rescode.html#busy_snapshot), which we should be handling already, so I'm really not sure what was happening there. One to investigate.

NyaaaWhatsUpDoc:
have you recently started using filtering ? i'm wondering if the new upsert type perhaps is not going the usual retryBusy route 🤔

NyaaaWhatsUpDoc:
worst case, at this point I'm getting fairly comfortable with the SQLite driver internals for easier debugging :p

tsmethurst:
> have you recently started using filtering

I didn't make any filters yet no, so I don't think that's the issue. Could it be another regression in sqlite from https://github.com/superseriousbusiness/gotosocial/commit/c27049ad15c3f4dad068bc528cab34c06653cd32 ?

NyaaaWhatsUpDoc:
it's hard to say off the top of my head. might take a look into it a bit later, for now i need to do some IRL cleaning (have people over for dinner later!) and then get to your PR :)

NyaaaWhatsUpDoc:
leaving this issue open after the above bugfix, as i'm still not 100% sure why the instance got *stuck* in this state

marmulak:
Hello, running 0.15.0-rc1+git-65b5366 (binary release) on freebsd-amd64. My instance became unresponsive after this:

[...]

Issue #2728, opened 2024-03-06
[bug] DB queries on SQLite-backed instance return error `interrupted (9)` for every query (modernc.org/sqlite > v1.29.2)


tsmethurst:
Since updating our sqlite version in https://github.com/superseriousbusiness/gotosocial/pull/2718, occasionally an instance will become unresponsive since all SQLite DB queries return `interrupted (9)`. A restart fixes this (temporarily at least).

Not yet sure what's causing the issue, we're investigating.

tsmethurst:
Workaround in https://github.com/superseriousbusiness/gotosocial/pull/2731

tsmethurst:
Steps to recreate the issue:

- Check out v0.14.0 of GtS.
- `DEBUG=1 ./scripts/build.sh`
- `DEBUG=1 GTS_DB_ADDRESS=./sqlite.db ./gotosocial testrig start`
- Open https://localhost:8080/@admin in the browser.
- Hold down ctrl+f5 in the browser for ~10 seconds to force spammy refresh of GtS.
- Watch the testrig logs.

NyaaaWhatsUpDoc:
a bug-fix upstream has now been opened: https://gitlab.com/cznic/sqlite/-/merge_requests/67

PR #2811, opened 2024-04-04
[bugfix/workaround]replace modernc.org/sqlite-v1.29.5 with gitlab.com/NyaaaWhatsUpDoc/sqlite-v1.29.5-concurrency-workaround

NyaaaWhatsUpDoc:
# Description

See for changes: https://gitlab.com/NyaaaWhatsUpDoc/sqlite/-/commit/29a5812b7c37800d643eff7fc9686e61ad3112c8

This is a temporary fix for #2808 and #2774. This works by cutting out the possible concurrent access of a singullar sqlite3 db pointer with the "interruptOnDone()" goroutine. The downside now is that context handling will now **not** be immediate, it will instead have to check at the next sqlite3 step() for a canceled context. In practice though this shouldn't cause (us at least) any issues.

You can confirm this by triggering #2774 on a testrig instance (same methodology as with #2728) then attempting to shutdown the instance. It will hang on "database locked". Switching to this branch the issue will no longer be there. Switching to this branch you'll also notice that your database WAL file gets regularly cleaned up. 

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/superseriousbusiness/gotosocial
 
 go 1.21
 
+replace modernc.org/sqlite => gitlab.com/NyaaaWhatsUpDoc/sqlite v1.29.5-concurrency-workaround
+
 toolchain go1.21.3
 
 require (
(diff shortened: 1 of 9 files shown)


NyaaaWhatsUpDoc:
marking this is draft until @daenney @tsmethurst agree (or not! in which case, closed it shall be :p) this is the best solution for now

tsmethurst:
We chatted about this in Matrix and decided it's a good fix for now because it solves our immediate problem! Ideally a fix should be done upstream so we don't have to maintain a fork, however, but solving the immediate issue is more important for now :)

PR #2812, opened 2024-04-05
[performance] massively improved ActivityPub delivery worker efficiency

NyaaaWhatsUpDoc:
# Description

This replaces our current transport delivery routine to instead push deliveries to a concurrency-safe queue, which is pulled from by a fixed-size pool of delivery workers. These delivery workers are optimized (unlike currently) to prefer new attempted deliveries over backlogged erroring (i.e. retry-backoff) deliveries, which **hugely** optimizes both the amount of time our goroutines spend and the actual number of them, so they're doing **useful** delivery work and not performing retry-backoffs in serial.

Also sneaks in a lil change to use `errorsv2.AsV2()` instead of `errors.As()` in a number of places because it's much nicer working with generic types, and likely faster instead of the existing reflect logic.

## Checklist


internal/httpclient/client.go
```diff
@@ -187,154 +188,184 @@ func New(cfg Config) *Client {
...
+		}
+	}
+}
 
-				// Get current time.
-				now := time.Now()
+// DoOnce wraps an underlying http.Client{}.Do() to perform our wrapped request type:
+// rewinding response body to permit reuse, signing request data when SignFunc provided,
+// marking erroring hosts, updating retry attempt counts and setting backoff from header.
+func (c *Client) DoOnce(r *Request) (rsp *http.Response, retry bool, err error) {
```
NyaaaWhatsUpDoc, line 248: i added this as it helps separate out the retry-backoff logic from the main Do() function, but leaves it re-usable in a way that external packages can also make use of it (e.g. the `internal/processing/workers/delivery` package which needs it during the delivery loop but with a lot of changes to the standard way)

internal/transport/delivery/delivery.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+
+		if !retry {
+			// Drop deliveries when no
+			// retry requested, or they
+			// reached max (either).
+			continue loop
+		}
+
+		// Determine next delivery attempt.
```
NyaaaWhatsUpDoc, line 239: by specifically setting the next expected delivery time instead of just relying on a backoff duration, it means that messages in the backlog we leave and come back to we may not even need to perform a backoff sleep for, enough other deliveries may have been processed that we can just come back to it and process it as-is.

this is handled in the `if d := dlv.backoff(); d > min { ... }` clause above, where we specifically only backoff sleep while there's no other new deliveries to attempt.

internal/transport/delivery/delivery_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// Start new HTTP test server listener.
+	l, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer l.Close()
+
+	// Start the HTTP server.
+	srv := new(http.Server)
```
NyaaaWhatsUpDoc, line 63: specifically not using httptest.Server{} here as httptest links that server with its own http.Client{}, whereas here we're using an httpclient.Client{} (well, the delivery routine is)

go.mod
```diff
@@ -49,7 +49,7 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/viper v1.18.2
 	github.com/stretchr/testify v1.9.0
-	github.com/superseriousbusiness/activity v1.6.0-gts.0.20240221151241-5d56c04088d4
+	github.com/superseriousbusiness/activity v1.6.0-gts.0.20240408131430-247f7f7110f0
```
NyaaaWhatsUpDoc, line 52: will make a version for this and bump it once we have it confirmed running on main for a few days :)

tsmethurst: Ack, no problem.

internal/httpclient/client.go
```diff
@@ -187,154 +188,184 @@ func New(cfg Config) *Client {
...
-		} else if errorsv2.IsV2(err,
+	// Log any errors.
+	r.Entry.Error(err)
+
+	switch {
+	case !retry:
+		// If they were told not to
+		// retry, also set number of
+		// attempts to prevent retry.
+		r.attempts = c.retries + 1
```
tsmethurst, line 279: No chance of c.retries being accessed concurrently right? Just to check.

internal/processing/workers/federate.go
```diff
@@ -75,6 +75,12 @@ func (f *federate) DeleteAccount(ctx context.Context, account *gtsmodel.Account)
 		return nil
 	}
 
+	// Drop any queued outgoing AP requests to / from account,
+	// (this stops any queued likes, boosts, creates etc).
+	f.state.Workers.Delivery.Queue.Delete("ActorID", account.URI)
```
tsmethurst, line 80: Nice one :)

tsmethurst (approved):
Looks great to me :) Good work! Once 0.15.0 is out let's squerge this.

internal/transport/delivery/delivery_test.go
NyaaaWhatsUpDoc: see: [65a54b4](https://github.com/superseriousbusiness/gotosocial/pull/2812/commits/65a54b47700075ee4ee9e77cb456a697f1ae4cf8)

PR #2826, opened 2024-04-12
[feature] Admin accounts endpoints; approve/reject sign-ups

tsmethurst:
# Description

>

This pull request implements the rest of https://github.com/superseriousbusiness/gotosocial/issues/703 by implementing functionality to allow admins to view pending account sign-ups, and approve or reject them.

- implements `/api/v1/admin/accounts` and `/api/v2/admin/accounts` endpoints, to allow admins to page through lists of accounts with various parameters (follows Mastodon API).
- implements `/api/v1/admin/accounts/{id}` endpoint, to allow admins to get the admin view of a specific account
- implements new settings panel sections for browsing through accounts, and for showing a list of pending accounts
- updates admin panel detailed account view with more stats about the account being viewed, and adds options to approve / reject sign-ups there
- documents sign-up flow for admins
- update various bits and bobs of the settings panel stuff, to use typescript a bit more

Closes https://github.com/superseriousbusiness/gotosocial/issues/703

## Checklist


docs/admin/signups.md
```diff
@@ -0,0 +1,59 @@
+# New Account Sign-Ups
+
+If you want to allow more people than just you to have an account on your instance, you can open your instance to new account sign-ups / registrations.
+
```
NyaaaWhatsUpDoc, line 4: well said!

internal/db/bundb/account.go
```diff
@@ -222,6 +225,216 @@ func (a *accountDB) GetInstanceAccount(ctx context.Context, domain string) (*gts
...
+	origin string,
+	status string,
+	mods bool,
+	invitedBy string,
+	username string,
+	displayName string,
+	domain string,
+	email string,
+	ip net.IP,
+	maxID string,
```
NyaaaWhatsUpDoc, line 239: i think it would be here using the new paging logic so we don't further stray from unifying things. (so you'd make it just a single `page *paging.Page` argument

internal/processing/admin/accountreject.go
```diff
@@ -0,0 +1,123 @@
...
+	defer unlock()
+
+	// Can't reject an account with a
+	// user that's already been approved.
+	if *user.Approved {
+		err := fmt.Errorf("account %s has already been approved", accountID)
+		return nil, gtserror.NewErrorUnprocessableEntity(err, err.Error())
+	}
+
+	// Convert to API account *before* doing the
```
NyaaaWhatsUpDoc, line 65: is it worth moving everything from this point onward to the clientAPI worker function? you can have it use the lock briefly to just check whether user is already approved, unlock, then enqueue the rest of it to the worker func, otherwise the worker function does nothing but send an email :p

NyaaaWhatsUpDoc: (oh and keeping the ToAdminAPIAccount() bit too, obviously you'll need that *before* you perform the account deletion)

NyaaaWhatsUpDoc: either that or just move the email sending over here, since worker is only sending an email.

this same thought i've gone through in this thread also applies to the accept logic too.

internal/processing/admin/accounts.go
```diff
@@ -0,0 +1,258 @@
...
+		valid := []string{"staff"}
+		if !slices.Contains(valid, v) {
+			err := fmt.Errorf("permissions %s not recognized; valid choices are %+v", v, valid)
+			return nil, gtserror.NewErrorBadRequest(err, err.Error())
+		}
+	}
+
+	var ip net.IP
+	if v := request.IP; v != "" {
+		ip = net.ParseIP(v)
```
NyaaaWhatsUpDoc, line 73: probably better to use `netip.ParseAddr()` here since that's the new, faster way of doing things in the stdlib. the type it returns is slightly different so you'll need to also quickly change the db func below too

internal/processing/admin/accounts.go
```diff
@@ -0,0 +1,258 @@
...
+	ctx context.Context,
+	request *apimodel.AdminGetAccountsRequest,
+	page *paging.Page,
+) (
+	*apimodel.PageableResponse,
+	gtserror.WithCode,
+) {
+	// Validate "origin".
+	if v := request.Origin; v != "" {
+		valid := []string{"local", "remote"}
```
NyaaaWhatsUpDoc, line 46: is it worth moving each of these sets of `valid := []string{...}` slices to be globals instead? since they're only ever checked against and never changed

internal/processing/admin/accountreject.go
NyaaaWhatsUpDoc: oh and whichever way you decide with this, i just realized the AccountAccept() function probably wants to use the client lock as well, just to prevent attempted accept / reject at the same time :p

NyaaaWhatsUpDoc:
looking gooooooooood :rocket: :eyes:

internal/processing/admin/accountreject.go
NyaaaWhatsUpDoc: what about just manually setting the approved flag to false ? or alternatively as i said sending the email from this function instead of the worker?

tsmethurst: I'll have a poke at it!

internal/processing/admin/accounts.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/2826/commits/e455c90827a223ebeb29c47b057f15385888ece2

internal/processing/admin/accounts.go
tsmethurst: Yeah! done: https://github.com/superseriousbusiness/gotosocial/pull/2826/commits/c10248be6dd7915b9276db53738d8e21c2b9b38e

internal/processing/admin/accountreject.go
tsmethurst: Done :) https://github.com/superseriousbusiness/gotosocial/pull/2826/commits/36d9b2885daeb1fc582caa45dc2b9621db997190

PR #2831, opened 2024-04-14
[feature/performance] Store account stats in separate table

tsmethurst:
# Description

>

This pull request changes the way we generate account stats. Previously, we were caching some stats in an in-memory cache. Before that, we just selected stats directly using COUNT queries every time an account needed to viewed via the API.

Instead in this PR stats are now stored in the database, and a best-effort is made to keep those stats up to date in the processing/worker functions.

This also means that we can now parse remote followers/following counts and store those, instead of having remote accounts always show counts only for local followers/following. 

Stats for local accounts are regenerated if, when fetched from the cache/db, they haven't been regenerated for 24hrs or more, to avoid drift.

~~Some new functionality has been added to the cleaner package as well: every 24hrs at 1am a new "database cleaner" function will run. Currently this just regenerates stats for local accounts to prevent possible drift due to race conditions, but in future we could also use this new cleaner functionality to clean up old statuses when we get around to implementing that.~~

Should close https://github.com/superseriousbusiness/gotosocial/issues/1887

## Checklist


internal/ap/collections.go
```diff
@@ -147,6 +157,15 @@ func (iter *orderedCollectionIterator) PrevItem() TypeOrIRI {
 	return cur
 }
 
+func (iter *orderedCollectionIterator) TotalItems() *int {
```
NyaaaWhatsUpDoc, line 160: i think since you're returning a signed int anyways, instead of making this a ptr in the case of no total it's probably worth returning -1 instead

internal/cleaner/database.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/superseriousbusiness/gotosocial/internal/db"
+	"github.com/superseriousbusiness/gotosocial/internal/gtscontext"
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/log"
+)
+
+// Database encompasses a set of
+// database cleanup / admin utils.
+type Database struct {
```
NyaaaWhatsUpDoc, line 32: i'm not sure this is something we need to regenerate on a scheduled basis given the increment / decrement functions in the processor. the onlly time it *may* fall out of sync is possibly during a restart / panic, so instead of having it be a scheduled thing i think instead it would worth adding it as an admin CLI + frontend command. otherwise this is a very likely relatively expensive for not often any change database operation we're scheduling every day. *alternatively* if we schedule it on a longer time frame than that it runs into the situation that if you restart semi often then the schedule may never get used. i feel like making it an admin command mitigates these issues. what do you think?

NyaaaWhatsUpDoc:
a few comments so far, and i need to spend some more time thinking about the client locks just to make sure there isn't an issue, but otherwise looking good so far!

internal/cleaner/database.go
NyaaaWhatsUpDoc: this also leaves the option for selecting to regenerate local-only accounts, and remote-only accounts. the latter of which allows someone that wants them to run a long operation to ensure **all** the accounts in their database are up-to-date (even if statuses is only *our* count of them)

NyaaaWhatsUpDoc: or actually, just thought of a solution that solves this situation **perfectly** :D: if we moved this operation from being a scheduled operation, to instead run on instance start. we're only running it for local accounts, so it should be a fast one, and it perfectly handles the case of increments / decrements not happening on restart due to workers shutting down!

NyaaaWhatsUpDoc: but maybe offering it as an admin CLI option (configurable to be local / remote) so admins can force recheck for their remote accounts if they want to?

tsmethurst: Hmmmm I see what you mean.... What if I were to add a field `RegeneratedAt` to AccountStats, and if, when doing a GetAccountStats, that timestamp is older than say 48 hours, it gets regenerated? For local accounts only. That's the "lazy" way of doing it, but I think it solves most of the problems above... What do you think?

tsmethurst: (For remote accounts this stuff gets pulled again when the account gets refreshed anyways, which I think is on a 6hr cooldown iirc)

NyaaaWhatsUpDoc: that would work, though i'm just a bit apprehensive to add scheduled tasks unless we *really* need them. you could just make it run on instance start, and a later PR actually expose it as an admin CLI thing?

tsmethurst: Ah no no, I don't mean running it as a scheduled task, I'll take that bit out :) I mean similar to what we do now with remote accounts, where we refresh them if FetchedAt is older than six hours, we'll regenerate local account stats when doing GetAccountStats if RegeneratedAt is older than 48 hrs or some similar number :)

internal/ap/collections.go
tsmethurst: Done!

internal/cleaner/database.go
tsmethurst: Done!

internal/federation/dereferencing/account.go
```diff
@@ -1036,6 +1041,129 @@ func (d *Dereferencer) fetchRemoteAccountEmojis(ctx context.Context, targetAccou
...
+	uri, err := url.Parse(uriStr)
+	if err != nil {
+		return -1, err
+	}
+
+	collect, err := d.dereferenceCollection(ctx, requestUser, uri)
+	if err != nil {
+		// Just log for debug, we don't
+		// mind if we can't fetch this.
+		log.Debugf(ctx,
```
NyaaaWhatsUpDoc, line 1157: i'd maybe make this a warning level error, it's something that's useful for admins to see in the logs in terms of troubleshooting

NyaaaWhatsUpDoc: or actually saying that, given there's logging in the caller of this function (and it's not handled as being a showstopper), probably just worth returning this as an error here

internal/db/bundb/account.go
```diff
@@ -1147,3 +1080,188 @@ func (a *accountDB) UpdateAccountSettings(
 		return nil
 	})
 }
+
+func (a *accountDB) GetAccountStats(ctx context.Context, accountID string) (*gtsmodel.AccountStats, error) {
```
NyaaaWhatsUpDoc, line 1084: given that this is often called to populate an account model to begin with, what do you think if you updated the function signature of this to be `GetAccountStatus(ctx context.Context, account *gtsmodel.Account) error` ? it just saves needed to do the double fetch of an account model that's likely already there.

NyaaaWhatsUpDoc (changes requested):

internal/processing/workers/util.go
```diff
@@ -238,3 +238,242 @@ func (u *utilF) redirectFollowers(
 
 	return true
 }
+
+func (u *utilF) incrementStatusesCount(
+	ctx context.Context,
+	account *gtsmodel.Account,
+	status *gtsmodel.Status,
+) error {
+	// Populate stats.
```
NyaaaWhatsUpDoc, line 247: probably worth using a client lock for each of these increment / decrement functions :)

internal/db/bundb/account.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/2831/commits/9ff5e0cbabdecd32b86fe64d0acf6ec15097cb40

internal/federation/dereferencing/account.go
tsmethurst: Alright, rejigged it here and changed it to warn logging from error level as it's not critical: https://github.com/superseriousbusiness/gotosocial/pull/2831/commits/9d15127841fa81ed1713c0c5963f2aa632f3a58f

internal/processing/workers/util.go
tsmethurst: Good idea! Changed it here: https://github.com/superseriousbusiness/gotosocial/pull/2831/commits/ceb84e55ee2facd2a7263c77c97c431e42b99f3e I also renamed "ClientLocks" to "AccountLocks" since we were using it for more than just ClientAPI actions, and also we're using it just for account URIs (I checked).

Issue #2844, opened 2024-04-16
[bug] fedilab shows error after boosting post


albertotirla:
### Describe the bug with a clear and concise description of what the bug is.

essentially, I can no longer interact with posts using fedilab. I only recently started doing so, so I don't know of previous problems, if any. I tryed to restart the docker container, that didn't help

### What's your GoToSocial Version?

latest

### GoToSocial Arch

x86_64 docker

### What happened?

fedilab throws an error when trying to boost a post, and errors appear in the log

### What you expected to happen?

boosts should just work

### How to reproduce it?

* open fedilab
* login
* when it works, because the webview likes to give a lot of an error occured messages at this stage, go to your home timeline
* try to boost somebody's post

### Anything else we need to know?

I can't think of anything else in particular, besides these logs:

```shell
gotosocial_1  | timestamp="16/04/2024 19:59:56.834" func=server.glob..func1.Logger.func13.1 level=INFO latency="54.65µs"
 userAgent=Fedilab/3.28.2/515 method=GET statusCode=404 path=/api/v2/filters clientIP=x.x.x.x errors="Error #01: Not Found\n" requestID=yyyyyyyyyyyy msg="Not Found: wrote 21B"
```

CDN18:
#2793 may close this

tsmethurst:
What errors does fedilab show (if any) when you try to boost something?

tsmethurst:
Hmm I logged in via fedilab to try to replicate this. I don't see any errors when just logging in and browsing the timeline (using android, and the version of fedilab installed via Google play), but indeed when I boost something I get an error returned. However, when I log in via another client and check, the status did get boosted. I have an idea already what the cause might be, so I'll make a fix and check.

tsmethurst:
I can't replicate with replying to one of my own statuses at least, that seems to work OK.

tsmethurst:
This is fixed by https://github.com/superseriousbusiness/gotosocial/pull/2846 (I just checked) :) So I'll close it.

PR #2846, opened 2024-04-17
[bugfix] Fix minor API issue w/ boosted statuses

tsmethurst:
# Description

>

Following the Mastodon API we should be taking interaction counts from the boosted status, rather than generating them separately on the boost wrapper status, when serializing via the API.

This PR fixes that by serializing the wrapped reblogged status first, then just taking the interactions from that.

This also means fewer DB calls in total when serializing a boosted status, which is a nice bonus :)

MIGHT be a fix for https://github.com/superseriousbusiness/gotosocial/issues/2844 but i'm not sure.

## Checklist


NyaaaWhatsUpDoc:
nice! I'm always a fan of fewer db calls 😎

PR #2849, opened 2024-04-17
[bugfix] nil panic when deleting status

NyaaaWhatsUpDoc:
ensures that requestingAccount is always set, otherwise the new status count decrementer panics


tsmethurst:
Whoops! Thanks!

PR #2863, opened 2024-04-22
[experiment] add alternative wasm sqlite3 implementation available via build-tag

NyaaaWhatsUpDoc:
# Description

- wasm sqlite3 can be accessed via `wasmsqlite3` build tag
- moves sqlite driver wrapping to separate subpkg, ~same needs to be done for postgres eventually~
- removes the sqlite3 interrupt workaround in the error processing (no longer needed)
- removes the retryBusy() handler from the modernc.org/sqlite wrapper, again no longer seems to be needed (maybe related to the custom modernc.org/sqlite branch we're on)

## Checklist


go.mod
```diff
@@ -78,7 +79,7 @@ require (
 	golang.org/x/text v0.14.0
 	gopkg.in/mcuadros/go-syslog.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.1
-	modernc.org/sqlite v1.29.8
+	modernc.org/sqlite v0.0.0-00010101000000-000000000000
```
tsmethurst, line 82: This is a bit odd, what happened here?

tsmethurst:
I would suggest that we keep the retry on busy handling, as it is still applicable to us: https://www.sqlite.org/wal.html#sometimes_queries_return_sqlite_busy_in_wal_mode

Also, we do let clients choose whether to use wal mode or not, and if they don't use wal mode for whatever reason, the busy handler will be much more necessary.

go.mod
NyaaaWhatsUpDoc: this is due to the replace line at the top of the file, since we're replacing modernc.org/sqlite with gitlab.com/NyaaaWhatsUpDoc/sqlite the version here becomes essentially useless and the version in question is actually in the replace line. the replace itself is necessary as the gitlab.com/NyaaaWhatsUpDoc/sqlite fork keeps the original Go package name

daenney:
> Also, we do let clients choose whether to use wal mode or not, and if they don't use wal mode for whatever reason, the busy handler will be much more necessary.

We should probably change that honestly. Letting clients chose the mode isn't really necessary, unless there's some evidence of WAL performing worse for people. But I suspect the evidence goes in the other direction, and it limits the amount of things we have to think about when someone reports a database performance issue.

Since we have the transpile or the WASM build we can rely on WAL always being available no matter the SQLite version any distro ships.

NyaaaWhatsUpDoc:
> > Also, we do let clients choose whether to use wal mode or not, and if they don't use wal mode for whatever reason, the busy handler will be much more necessary.
> 
> We should probably change that honestly. Letting clients chose the mode isn't really necessary, unless there's some evidence of WAL performing worse for people. But I suspect the evidence goes in the other direction, and it limits the amount of things we have to think about when someone reports a database performance issue.
> 
> Since we have the transpile or the WASM build we can rely on WAL always being available no matter the SQLite version any distro ships.

Agreed. Also doing it now will be especially easier since we still have the "alpha" tag, where changes like this are more to be expected.

internal/db/bundb/bundb.go
```diff
@@ -529,11 +522,7 @@ func buildSQLiteAddress(addr string) string {
 		// Use random name for in-memory instead of ':memory:', so
 		// multiple in-mem databases can be created without conflict.
 		addr = uuid.NewString()
-
-		// in-mem-specific preferences
-		// (shared cache so that tests don't fail)
-		prefs.Add("mode", "memory")
-		prefs.Add("cache", "shared")
+		prefs.Add("vfs", "memdb")
```
ncruces, line 525: For a `memdb` database to be shared between connections, the name needs to start with a `/`.
This is true both for my [port](https://pkg.go.dev/github.com/ncruces/go-sqlite3/vfs/memdb#pkg-overview), as well as the [original](https://github.com/sqlite/sqlite/blob/b11daa50f9ea11c332bb5913a071c5a0fd6c9993/src/memdb.c#L42-L48) SQLite implementation (which gets compiled into _modernc_).

If you're having issues with missing tables (as mentioned [here](https://github.com/ncruces/go-sqlite3/discussions/78#discussioncomment-9201818)) this is your likely culprit.

To be clear, if you want to open multiple connections to the same memory database (and `database/sql` connection pooling does this for you), your URI must have a `/`:
```
file:/rANd0m-UuiD?vfs=memdb&_pragma=busy_timeout(1000)
```

daenney:
I took a quick look at the tests, a lot of the failures are in typeutils and the only difference seems to be the resulting url starting with http vs https. Probably just a config setting?

The rest is just "database is closed" which seems to cause goroutine blockage/runtime hangs.

internal/db/bundb/bundb.go
ncruces: I also found https://github.com/ncruces/go-sqlite3/issues/87.

If you feel that the above causes you any grief, I'll take that as a strong indicator that I should fix it, for portability. So ping me.

daenney:
Only thing that fails now is the envparsing test:

```diff
--- /tmp/tmp.IALNpE
+++ /tmp/tmp.JgNPAF

@@ -74,13 +74,13 @@
     "db-tls-mode": "disable",
     "db-type": "sqlite",
     "db-user": "sex-haver",
-    "dry-run": false,
+    "dry-run": true,
     "email": "",
     "host": "example.com",
     "http-client": {
-        "allow-ips": null,
-        "block-ips": null,
-        "timeout": 0,
+        "allow-ips": [],
+        "block-ips": [],
+        "timeout": 10000000000,
         "tls-insecure-skip-verify": false
     },
     "instance-deliver-to-shared-inboxes": false,
```

I'm guessing it needs a rebase and that should be it.

NyaaaWhatsUpDoc:
> Only thing that fails now is the envparsing test:
> 
> ```diff
> --- /tmp/tmp.IALNpE
> +++ /tmp/tmp.JgNPAF
> 
> @@ -74,13 +74,13 @@
>      "db-tls-mode": "disable",
>      "db-type": "sqlite",
>      "db-user": "sex-haver",
> -    "dry-run": false,
> +    "dry-run": true,
>      "email": "",
>      "host": "example.com",
>      "http-client": {
> -        "allow-ips": null,
> -        "block-ips": null,
> -        "timeout": 0,
> +        "allow-ips": [],
> +        "block-ips": [],
> +        "timeout": 10000000000,
>          "tls-insecure-skip-verify": false
>      },
>      "instance-deliver-to-shared-inboxes": false,
> ```
> 
> I'm guessing it needs a rebase and that should be it.

t'was actually a package init function in testrig i added messing with it. a little confusing and hard to narrow down but caught it in the end :)

daenney:
All the tests pass omgomgomg :smile:

daenney (approved):
Giving @tsmethurst some time to review it next week but LGTM shipit all the things.

testrig/config.go
```diff
@@ -18,138 +18,162 @@
 package testrig
 
 import (
-	"cmp"
 	"os"
+	"strconv"
 	"time"
 
+	// linkname shenanigans
+	_ "unsafe"
```
tsmethurst, line 26: Hmm, what's this do?

tsmethurst:
This mostly looks good to me, but I would still recommend we keep the retry on busy handler, as even if we at some point force admins to always use WAL mode, it's still important to implement retry on busy, as there are cases where it will be hit. From the SQLite docs:

> The [second advantage of WAL-mode](https://www.sqlite.org/wal.html#advantages) is that writers do not block readers and readers to do not block writers. This is mostly true. But there are some obscure cases where a query against a WAL-mode database can return [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy), so applications should be prepared for that happenstance.
> 
> Cases where a query against a WAL-mode database can return [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy) include the following:
> 
>     If another database connection has the database mode open in [exclusive locking mode](https://www.sqlite.org/pragma.html#pragma_locking_mode) then all queries against the database will return [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy). Both Chrome and Firefox open their database files in exclusive locking mode, so attempts to read Chrome or Firefox databases while the applications are running will run into this problem, for example.
> 
>     When the last connection to a particular database is closing, that connection will acquire an exclusive lock for a short time while it cleans up the WAL and shared-memory files. If a separate attempt is made to open and query the database while the first connection is still in the middle of its cleanup process, the second connection might get an [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy) error.
> 
>     If the last connection to a database crashed, then the first new connection to open the database will start a recovery process. An exclusive lock is held during recovery. So if a third database connection tries to jump in and query while the second connection is running recovery, the third connection will get an [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy) error.

ncruces:
Sorry to interject, but AFAIK, those cases are still covered by [`sqlite3_busy_handler`](https://www.sqlite.org/c3ref/busy_handler.html) or [`sqlite3_busy_timeout`](https://www.sqlite.org/c3ref/busy_timeout.html) or [`PRAGMA busy_timeout`](https://www.sqlite.org/pragma.html#pragma_busy_timeout).

So, in those cases if you already have an SQLite busy timeout (or handler), you only get [`SQLITE_BUSY`](https://www.sqlite.org/rescode.html#busy) after the timeout has expired.

AFAIK, the _only_ situation where SQLite returns `SQLITE_BUSY` immediately, without waiting (if you configured a busy handler), is if it detects a dead-lock. And this only happens when you upgrade a read-lock to a write-lock in the middle of a transaction, which can _also_ be avoided if [`BEGIN IMMEDIATE`](https://www.sqlite.org/lang_transaction.html) is used for all transactions that may write.

If you do this, retrying after you get `SQLITE_BUSY` is becomes almost pointless. You can always wait a bit more, but then, why not just increase the configured timeout?

testrig/config.go
NyaaaWhatsUpDoc: leftover from when i was using one of the package internal methods from `internal/config`. i'll remove it now :)

NyaaaWhatsUpDoc:
> Sorry to interject, but AFAIK, those cases are still covered by [`sqlite3_busy_handler`](https://www.sqlite.org/c3ref/busy_handler.html) or [`sqlite3_busy_timeout`](https://www.sqlite.org/c3ref/busy_timeout.html) or [`PRAGMA busy_timeout`](https://www.sqlite.org/pragma.html#pragma_busy_timeout). So, in those cases if you already have an SQLite busy timeout (or handler), you only get [`SQLITE_BUSY`](https://www.sqlite.org/rescode.html#busy) after the timeout has expired.
> 
> AFAIK, the _only_ situation where SQLite returns `SQLITE_BUSY` immediately, without waiting (if you configured a busy handler), is if it detects a dead-lock. And this only happens when you upgrade a read-lock to a write-lock in the middle of a transaction, which can _also_ be avoided if [`BEGIN IMMEDIATE`](https://www.sqlite.org/lang_transaction.html) is used for all transactions that may write. And in this case you need to do more than just retry, you need to `ROLLBACK` the transaction before retrying (I dunno if you're doing this).
> 
> So if you use `BEGIN IMMEDIATE` for write transactions, retrying after you get `SQLITE_BUSY` is becomes almost pointless. You can always wait a bit more, but then, why not just increase the configured timeout?

Indeed. We're currently setting `_txlock=immediate` so that covers the case of deadlocks under WAL possibly returning `SQLITE_BUSY`, (and needing to rollback the tx before another attempt).

testrig/config.go
NyaaaWhatsUpDoc: removed here: a9d9058c7

tsmethurst:
Aha OK thanks for clarifying @ncruces and @NyaaaWhatsUpDoc :)

Issue #2843, opened 2024-04-16
[bug] Delivery goroutines experienced a panic with "not yet initialized" while executing CLI commands


CDN18:
### Describe the bug with a clear and concise description of what the bug is.

CLI commands are terminating with error messages indicating that a panic occurred due to not being initialized:

For example, this error appears when executing the command 'gotosocial admin account create':

```
goroutine 25 [running]:
github.com/superseriousbusiness/gotosocial/internal/transport/delivery.(*Worker).run(0x0?, {0x27eb968?, 0x400010c660?})
        /drone/src/internal/transport/delivery/delivery.go:159 +0x134
codeberg.org/gruf/go-runners.(*Service).GoRun.func1()
        /drone/src/vendor/codeberg.org/gruf/go-runners/service.go:59 +0x64
created by codeberg.org/gruf/go-runners.(*Service).GoRun in goroutine 1
        /drone/src/vendor/codeberg.org/gruf/go-runners/service.go:49 +0x8c
panic: not yet initialized

goroutine 30 [running]:
github.com/superseriousbusiness/gotosocial/internal/transport/delivery.(*Worker).run(0x0?, {0x27eb968?, 0x400010c840?})
        /drone/src/internal/transport/delivery/delivery.go:159 +0x134
codeberg.org/gruf/go-runners.(*Service).GoRun.func1()
        /drone/src/vendor/codeberg.org/gruf/go-runners/service.go:59 +0x64
created by codeberg.org/gruf/go-runners.(*Service).GoRun in goroutine 1
        /drone/src/vendor/codeberg.org/gruf/go-runners/service.go:49 +0x8c
```

### What's your GoToSocial Version?

v0.15.1-SNAPSHOT+git-6de5717

### GoToSocial Arch

arm64 Docker

### What happened?

CLI commands terminated with error

### What you expected to happen?

CLI ammnads are executed successfully

### How to reproduce it?

Run any cli command.
In my case the errored commands are `admin account create/list/confirm`

tsmethurst:
Thanks! I think we just need to initialize the delivery queues when running commands via the CLI (even if we don't end up using them). @NyaaaWhatsUpDoc do you feel like doing this one or should I?

PR #2862, opened 2024-04-22
[chore] Upgrade our Go version to 1.22

daenney:
# Description

With Go 1.22 having been released at the start of February, it's now been a few months. No major issues have shown up, and the two point release since then have primarily been security fixes plus some general bug fixing.

This sets the required Go version to 1.22, as there's nothing in 1.22.1 or 1.22.2 that we would explicitly require. It sets the toolchain to the latest point release, to ensure we pick up any fixes from there when building releases etc.

## Checklist

Please put an x inside each checkbox to indicate that you've read and followed it: `[ ]` -> `[x]`

If this is a documentation change, only the first checkbox must be filled (you can delete the others if you want).

- [x] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [ ] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [x] I/we have not leveraged AI to create the proposed changes.
- [ ] I/we have performed a self-review of added code.
- [ ] I/we have written code that is legible and maintainable by others.
- [ ] I/we have commented the added code, particularly in hard-to-understand areas.
- [ ] I/we have made any necessary changes to documentation.
- [ ] I/we have added tests that cover new code.
- [x] I/we have run tests and they pass locally with the changes.
- [ ] I/we have run `go fmt ./...` and `golangci-lint run`.

PR #2865, opened 2024-04-24
[performance] update remaining worker pools to use queues

NyaaaWhatsUpDoc:
# Description

Updates the remaining worker pools (media, client, federator) worker pools to use the new queueing system, to allow for purging of jobs where possible (e.g. for incoming deletes that render other jobs unnecessary). This will make it easier in the future to drop + serialize all queued jobs to disk, and then resume those jobs on startup.

## Checklist


tsmethurst:
Ooh this is exciting :eyes:

cmd/gotosocial/action/server/server.go
```diff
@@ -195,10 +177,27 @@ var Start action.GTSAction = func(ctx context.Context) error {
 		return fmt.Errorf("error starting list timeline: %s", err)
 	}
 
-	// Create a media cleaner using the given state.
+	// Start the job scheduler
+	// (this is required for cleaner).
+	state.Workers.StartScheduler()
```
NyaaaWhatsUpDoc, line 182: this all had to get shuffled and the scheduler start separated out as the workers now panic if they are started with a nil processing function (since they rely on a message queue, instead of a function queue). and the scheduler has to get started *before* the workers as it's needed to initialize the cleaner, which is needed to initialize the processor, which in turn is needed to start the workers :p

go.mod
```diff
@@ -16,12 +16,13 @@ require (
...
 	codeberg.org/gruf/go-iotools v0.0.0-20230811115124-5d4223615a7f
 	codeberg.org/gruf/go-kv v1.6.4
+	codeberg.org/gruf/go-list v0.0.0-20240425093752-494db03d641f
 	codeberg.org/gruf/go-logger/v2 v2.2.1
 	codeberg.org/gruf/go-mutexes v1.4.1
 	codeberg.org/gruf/go-runners v1.6.2
 	codeberg.org/gruf/go-sched v1.2.3
 	codeberg.org/gruf/go-store/v2 v2.2.4
-	codeberg.org/gruf/go-structr v0.6.2
+	codeberg.org/gruf/go-structr v0.7.0
```
NyaaaWhatsUpDoc, line 25: this update adds a QueueCtx{} type, which is basically the same as the original but its context aware  (which means less wrapping code on our end)

go.mod
```diff
@@ -16,12 +16,13 @@ require (
 	codeberg.org/gruf/go-fastcopy v1.1.2
 	codeberg.org/gruf/go-iotools v0.0.0-20230811115124-5d4223615a7f
 	codeberg.org/gruf/go-kv v1.6.4
+	codeberg.org/gruf/go-list v0.0.0-20240425093752-494db03d641f
```
NyaaaWhatsUpDoc, line 19: this is used in a SimpleQueue{} type that we create, that we use as just a function queue, without needing to faff with interfaces or anything (it uses generics)

cmd/gotosocial/action/server/server.go
tsmethurst: Don't forget to change this as needed in the other CLI commands too, re: https://github.com/superseriousbusiness/gotosocial/issues/2843

internal/federation/federatingdb/create_test.go
```diff
@@ -120,7 +121,7 @@ func (suite *CreateTestSuite) TestCreateFlag1() {
 	}
 
 	// should be a message heading to the processor now, which we can intercept here
-	msg := <-suite.fromFederator
+	msg, _ := suite.state.Workers.Federator.Queue.Pop()
```
tsmethurst, line 124: Should this also use that new getFederatorMsg function with the timeout?

internal/federation/federatingdb/delete.go
```diff
@@ -34,43 +37,117 @@ import (
...
+		err error
+	)
+
+	// Attempt to delete account.
+	ok, err = f.deleteAccount(ctx,
+		requesting,
+		receiving,
+		uriStr,
+	)
+	if err != nil || ok { // handles success
```
tsmethurst, line 63: I'm not really a fan of this syntax, because it looks like you're returning an error even if OK, just because of the variable name (and it necessitates an extra comment just to explain it). Could you split it out to:

```go
if err != nil {
	return err
}

if ok {
	// Handled.
	return nil
}
```

I know it's a bit more verbose but I think it's clearer what's happening.

internal/federation/federatingdb/delete.go
```diff
@@ -34,43 +37,117 @@ import (
...
+	bool, // success?
+	error, // any error
+) {
+	account, err := f.state.DB.GetAccountByURI(ctx, uri)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return false, gtserror.Newf("error getting account: %w", err)
+	}
+
+	if account != nil {
+		if account.ID != requesting.ID {
```
tsmethurst, line 98: Could you explicitly say something like "make sure requester is only trying to delete itself" or something, just to make clear that's what we're checking for?

internal/federation/federatingdb/delete.go
```diff
@@ -34,43 +37,117 @@ import (
...
+		requesting,
+		receiving,
+		uriStr,
+	)
+	if err != nil || ok { // handles success
+		return err
+	}
+
+	// Attempt to delete status.
+	ok, err = f.deleteStatus(ctx,
```
tsmethurst, line 68: Same comments as above re: account.

internal/federation/federatingdb/delete.go
```diff
@@ -34,43 +37,117 @@ import (
...
+	bool, // success?
+	error, // any error
+) {
+	status, err := f.state.DB.GetStatusByURI(ctx, uri)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return false, gtserror.Newf("error getting status: %w", err)
+	}
+
+	if status != nil {
+		if status.AccountID != requesting.ID {
```
tsmethurst, line 134: Same comment as above re: account.

internal/httpclient/client.go
```diff
@@ -116,6 +123,11 @@ func New(cfg Config) *Client {
 	var c Client
 	c.retries = 5
 
+	if cfg.HTTPClient != nil {
+		// Copy over existing client.
+		c.client = *cfg.HTTPClient
```
tsmethurst, line 128: Could you explain what changed here? Not in comments I mean, just here in the review :P

internal/workers/workers.go
```diff
@@ -39,48 +37,36 @@ type Workers struct {
...
-	// will not use up a worker slot.
-	//
-	// As such, you should only call them in special cases where something
-	// synchronous needs to happen before you can do something else.
-	ProcessFromClientAPI func(context.Context, messages.FromClientAPI) error
-	ProcessFromFediAPI   func(context.Context, messages.FromFediAPI) error
-
-	// Media manager worker pools.
-	Media runners.WorkerPool
+	// Client provides a worker pool that handles
```
tsmethurst, line 40: Much cleaner :)

internal/workers/worker_fn.go
```diff
@@ -0,0 +1,120 @@
...
+	// passed to each of the pool Worker{}s.
+	Queue queue.SimpleQueue[func(context.Context)]
+
+	// internal fields.
+	workers []*FnWorker
+}
+
+// Start will attempt to start 'n' FnWorker{}s.
+func (p *FnWorkerPool) Start(n int) (ok bool) {
+	if ok = (len(p.workers) == 0); ok {
```
tsmethurst, line 43: This is quite confusing to read because it's not really clear what work `ok` is doing in here. It seems like you're using it twice for two different things but I'm not really following. Also the last time you use it in the `&& ok` bit, it must be true by that point right? :thinking: Could you refactor this a bit?

internal/workers/worker_fn.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+	}
+	return
+}
+
+// Stop will attempt to stop contained FnWorker{}s.
+func (p *FnWorkerPool) Stop() (ok bool) {
+	if ok = (len(p.workers) > 0); ok {
+		for i := range p.workers {
+			ok = p.workers[i].Stop() && ok
```
tsmethurst, line 58: Same comment here re this OK variable.

internal/util/fns.go
```diff
@@ -0,0 +1,68 @@
...
+		defer func() {
+			if r := recover(); r != nil {
+				// Gather calling func frames.
+				pcs := make([]uintptr, 10)
+				n := runtime.Callers(3, pcs)
+				i := runtime.CallersFrames(pcs[:n])
+				c := gatherFrames(i, n)
+
+				const msg = "recovered panic: %v\n\n%s\n"
+				fmt.Fprintf(os.Stderr, msg, r, c.String())
```
tsmethurst, line 44: Is there a reason to print to stderr here manually instead of calling log.Error? I guess because this is quite a low-level function?

internal/util/fns.go
```diff
@@ -0,0 +1,68 @@
...
+			}
+		}()
+		fn()
+		done = true
+		return
+	}() { //nolint
```
tsmethurst, line 50: Huh, what is the lint warning being masked here? Ah is it the empty `for` body?

internal/queue/simple.go
```diff
@@ -0,0 +1,141 @@
...
+// Len returns the current length of the queue.
+func (q *SimpleQueue[T]) Len() int {
+	q.m.Lock()
+	l := q.l.Len()
+	q.m.Unlock()
+	return l
+}
+
+// alloc will allocate new list element (relying on memory pool).
+func (q *SimpleQueue[T]) alloc() *list.Elem[T] {
```
tsmethurst, line 125: Damn you know you're going deep when you're making functions called `alloc` :') You nerd.

tsmethurst (commented):
Good changes! A necessary refactor! Just got some questions about code readability because there's some stuff my pea brain can't follow properly :')

internal/federation/federatingdb/delete.go
NyaaaWhatsUpDoc: 1d2a07ca1

internal/workers/worker_fn.go
NyaaaWhatsUpDoc: yeah i'll refactor to add some comments and explain what's going on :)

internal/workers/worker_fn.go
NyaaaWhatsUpDoc: yeah i'll refactor to add some comments and explain what's going on :)

internal/federation/federatingdb/create_test.go
NyaaaWhatsUpDoc: 794ebdb87

cmd/gotosocial/action/server/server.go
NyaaaWhatsUpDoc: good point!

internal/federation/federatingdb/delete.go
```diff
@@ -34,43 +37,117 @@ import (
...
+
+	// Serialize ID URI.
+	uriStr := id.String()
+
+	var (
+		ok  bool
+		err error
+	)
+
+	// Attempt to delete account.
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: see: d9e0c7a94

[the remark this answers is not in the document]
NyaaaWhatsUpDoc: see: d9e0c7a94

internal/federation/federatingdb/delete.go
NyaaaWhatsUpDoc: see: d9e0c7a94

internal/federation/federatingdb/delete.go
NyaaaWhatsUpDoc: see: 273dcc6f6

tsmethurst:
Oeie sorry you'll have to merge main into this because of https://github.com/superseriousbusiness/gotosocial/pull/2862

Should be a painless merge though I think (fingers crossed)

internal/util/fns.go
NyaaaWhatsUpDoc: yeah it doesn't like the empty "for {}" block :sweat_smile: it says you can remove it but clearly not :p

internal/federation/federatingdb/delete.go
```diff
@@ -60,8 +60,11 @@ func (f *federatingDB) Delete(ctx context.Context, id *url.URL) error {
 		receiving,
 		uriStr,
 	)
-	if err != nil || ok { // handles success
+	if err != nil {
 		return err
+	} else if ok {
```
tsmethurst, line 66: Thank you :')

internal/util/fns.go
NyaaaWhatsUpDoc: it's just so we print without any kind of log line prefix, otherwise that can make panic stack traces harder to read. it instead just prints it as you expect a standard panic stacktrace to get outputted :)

tsmethurst:
Feel free to squerge once everything is cleaned up btw :) The code itself looked ready to me.

internal/workers/worker_fn.go
NyaaaWhatsUpDoc: see: 92b2f7ae9

internal/workers/worker_fn.go
NyaaaWhatsUpDoc: see: 92b2f7ae9

internal/httpclient/client.go
NyaaaWhatsUpDoc: i removed it here: c71099c04

basically i was messing around at some point with being able to pass in an existing client so we could more easily use the httptest package, but that's out of scope for this PR :)

cmd/gotosocial/action/server/server.go
NyaaaWhatsUpDoc: updated CLI handling here: 4a263f9e717923fa1985aadb1c7830acd4c777bb + ea3a8716895da4de7a8c2629df18da062ed9184f

also ran the `test/import-export.sh` script and all seems good still :+1:

NyaaaWhatsUpDoc: this should also fix #2843

PR #2870, opened 2024-04-26
[chore] Update setting testrig loglevel

daenney:
# Description

cmp.Or was introduced in Go 1.22 and picks the first value that's not the zero value for the type. For a string, the zero value is the empty string, which is what os.Getenv will return if the environment variable is not set. That then results in "error" being returned instead.

This allows loading an environment variable with a default without having to do the check and write out the conditional.

## Checklist

### DIFF
--- a/testrig/config.go
+++ b/testrig/config.go
@@ -18,6 +18,7 @@
 package testrig
 
 import (
+	"cmp"
 	"os"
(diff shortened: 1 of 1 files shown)


tsmethurst:
Neat :)

PR #2871, opened 2024-04-26
[chore] Bump go swagger

tsmethurst:
Bumps our go-swagger version to account for an issue with the latest go swagger release and go v1.22 (https://github.com/go-swagger/go-swagger/issues/3070)


daenney (approved):
Sorry. I tried to not touch swagger because usually upgrading it breaks things :facepalm:.

tsmethurst:
No problem at all :D

PR #2880, opened 2024-04-29
[chore] Fix conflict in workers tests

tsmethurst:
Update workers tests to ensure that each test gets its own fresh processor, so that they avoid using each others workers queues.

### DIFF
--- a/internal/processing/workers/fromclientapi_test.go
+++ b/internal/processing/workers/fromclientapi_test.go
@@ -125,15 +125,7 @@ func (suite *FromClientAPITestSuite) checkStreamed(
 	ctx, cncl := context.WithTimeout(ctx, time.Second*5)
 	defer cncl()
 
-	var (
-		msg stream.Message
-		ok  bool
-	)
-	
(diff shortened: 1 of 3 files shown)


internal/processing/workers/fromclientapi_test.go
```diff
@@ -122,7 +125,15 @@ func (suite *FromClientAPITestSuite) checkStreamed(
 	ctx, cncl := context.WithTimeout(ctx, time.Second*5)
 	defer cncl()
 
-	msg, ok := str.Recv(ctx)
+	var (
+		msg stream.Message
+		ok  bool
+	)
+	
+	testrig.WaitFor(func() bool {
```
NyaaaWhatsUpDoc, line 133: the context timeout removes the need for testrig.WaitFor() :)

tsmethurst: Ah whoops, I left this in by accident while I was trying to figure out what was causing issues, I'll remove, thanks!

PR #2881, opened 2024-04-29
[feature] Page through accounts as moderator

tsmethurst:
# Description

>

This pull request implements paging through accounts alphabetically via the admin API, as opposed to paging by creation date. To accomplish this, a new index is added on a concatenation of `[domain]/@[username]`.

Paging is also implemented in the frontend, so admins can page up and down through accounts in the settings panel.

## Checklist


internal/api/client/statuses/statushistory_test.go
```diff
@@ -50,10 +50,10 @@ func (suite *StatusHistoryTestSuite) TestGetHistory() {
 
 	// Setup request.
 	recorder := httptest.NewRecorder()
-	request :=  httptest.NewRequest(http.MethodGet, target, nil)
```
tsmethurst, line 53: This got caught by go fmt, not related to this PR.

internal/db/bundb/account.go
```diff
@@ -479,13 +502,21 @@ func (a *accountDB) GetAccounts(
...
-		q = q.Order("account.created_at DESC")
+		// It's counterintuitive because it
+		// says ASC in the query, but we're
+		// going forwards in the alphabet,
+		// and z > a in a string comparison.
+		q = q.OrderExpr("? ASC", bun.Ident("domain_username"))
 	}
 
-	if err := q.Scan(ctx, &accountIDs); err != nil {
+	if err := q.Scan(ctx, &accountIDs, new([]string)); err != nil {
```
NyaaaWhatsUpDoc, line 519: how come we need this second argument? i thought we were only selecting the one column here :thinking:

internal/processing/admin/accounts.go
```diff
@@ -115,8 +115,12 @@ func (p *Processor) AccountsGet(
 		return paging.EmptyResponse(), nil
 	}
 
-	hi := accounts[count-1].ID
-	lo := accounts[0].ID
+	var (
+		loAcct = accounts[count-1]
```
NyaaaWhatsUpDoc, line 119: i think maybe worth adding a quick note here too that the "lo" and "hi" values are flipped here compared to using actual IDs due to paging alphabetically :)

internal/db/bundb/account.go
tsmethurst: We are selecting account_id but also selecting domain_username as part of that query, and unfortunately bun doesn't really know what to do if you don't provide a second array to soak up the domain_username values, unless there's a way around it I haven't spotted.

internal/processing/admin/accounts.go
tsmethurst: Resolved! We figured it out :D

internal/db/bundb/account.go
```diff
@@ -309,32 +336,50 @@ func (a *accountDB) GetAccounts(
...
+			bun.Ident("username"),
+			bun.Ident("domain_username"),
 		)
-		if err != nil {
-			return nil, fmt.Errorf("error getting maxID account %s: %w", maxID, err)
-		}
+	} else {
+		// Create a subquery for
+		// Postgres to reuse.
+		subQ = a.db.NewRaw(
```
NyaaaWhatsUpDoc, line 354: passing in this subquery multiple times in postgres, does it calculate it each and every time? or is it possible to use a temporary variable for it or something?

tsmethurst: As far as I can tell, it's computed only once because of the expression index I also added:

> Index expressions are relatively expensive to maintain, because the derived expression(s) must be computed for each row insertion and [non-HOT update.](https://www.postgresql.org/docs/current/storage-hot.html) However, the index expressions are not recomputed during an indexed search, since they are already stored in the index. In both examples above, the system sees the query as just WHERE indexedcolumn = 'constant' and so the speed of the search is equivalent to any other simple index query. Thus, indexes on expressions are useful when retrieval speed is more important than insertion and update speed.

-- https://www.postgresql.org/docs/current/indexes-expressional.html

So I think we should be fine. I tried doing this a few different ways (using WITH, or as subqueries), but Postgres doesn't offer the same flexibility SQLite does, so this ended up being (I think) the best way of doing it, even though the queries themselves end up looking a bit silly because of the expression re-use.

internal/db/bundb/account.go
NyaaaWhatsUpDoc: i'm wondering whether using a generated column that you also create an index on might work out a bit better (e.g. https://www.sqlite.org/gencol.html)

so generate a column name `domain_username` and then in both the sqlite and postgres queries can just do `SELECT id FROM accounts WHERE (...) ORDER domain_username ASC/DESC`. it's a column that would probably come in handy in other areas too i'm sure :thinking:

PR #2882, opened 2024-04-29
[bugfix] function queue memory pools limitlessly grow

NyaaaWhatsUpDoc:
# Description

It's not quite a memory leak, as all the memory can still get used, it just currently grows during peak times and never shrinks again. This updates it to follow the logic i've used previously in https://codeberg.org/gruf/go-mutexes/src/commit/130931bd8926f053baaaefa7e51ff0a88fe79a60/map.go#L99, so every `const gcfreq = 1024` pops from the queue, it will sweep out the memory pool of list elements.

## Checklist

### DIFF
--- a/internal/queue/simple.go
+++ b/internal/queue/simple.go
@@ -24,6 +24,11 @@ import (
 	"codeberg.org/gruf/go-list"
 )
 
+// frequency of GC cycles
+// per no. unlocks. i.e.
+// every 'gcfreq' unlocks.
+const gcfreq = 1024
+
 // SimpleQueue provides a simple concurrency safe
 // queue using generics and a memory pool of list
 // elements to reduce overall memory usage.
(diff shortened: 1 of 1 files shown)


internal/queue/simple.go
```diff
@@ -121,21 +157,45 @@ func (q *SimpleQueue[T]) Len() int {
...
-// alloc will allocate new list element (relying on memory pool).
-func (q *SimpleQueue[T]) alloc() *list.Elem[T] {
-	if len(q.p) > 0 {
-		elem := q.p[len(q.p)-1]
-		q.p = q.p[:len(q.p)-1]
-		return elem
+// elemPool is a very simple
+// list.Elem[T] memory pool.
+type elemPool[T any] struct {
+	current []*list.Elem[T]
```
NyaaaWhatsUpDoc, line 163: for those interested, the design of this i basically followed how the `sync.Pool{}` does it internally

PR #2883, opened 2024-04-30
[bugfix] paging rel links

NyaaaWhatsUpDoc:
# Description

Fix paging relative links so they use the correct cursoring parameter. Also updates the ASCII paging doc to explain this cursoring and make it clear.

## Checklist

### DIFF
--- a/internal/paging/boundary.go
+++ b/internal/paging/boundary.go
@@ -23,26 +23,36 @@ package paging
 func EitherMinID(minID, sinceID string) Boundary {
 	/*
 
-	           Paging with `since_id` vs `min_id`:
+				Paging with `since_id` vs `min_id`:
 
-	                limit = 4       limit = 4
-	               +----------+    +----------+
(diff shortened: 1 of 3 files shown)


internal/paging/boundary.go
```diff
@@ -23,26 +23,36 @@ package paging
...
-	               +----------+    +----------+
-	   since_id--> |          |    |xxxxxxxxxx| <-- min_id
-	               +----------+    +----------+
-	               |          |    |          |
-	               +----------+    +----------+
+				Paging with `since_id` vs `min_id`:
+
+					limit = 4       limit = 4
+					+----------+    +----------+
+		  max_id--> |xxxxxxxxxx|    |          | <-- max_id
```
NyaaaWhatsUpDoc, line 30: this looks correct in the IDE, github is just a silly sausage

tsmethurst:
Nice one, thanks for fixing :)

PR #2924, opened 2024-05-21
[performance] update storage backend and make use of seek syscall when available

NyaaaWhatsUpDoc:
# Description

This moves us from using `codeberg.org/gruf/go-store/v2/storage` using `codeberg.org/gruf/go-storage` as our underlying storage library. Which itself is just moving the library out of the nested repo it was in, since it expanded in usage scope from its original use. This in turn trims some of the unused dependencies from the vendor/ folder, and makes it easier updating the `go-storage` library myself since i don't need to update the rest of the key-value store around it. (also the key-value store itself isn't really getting used anymore by any of my projects, so it may end up falling by the wayside, and that's fine)

In updating to this version of the underlying storage library, since there's no more wrapping of the returned reader types, we can now rely on the seek syscall in a number of places which should speed-up file serving and video decoding on those using disk based storage :)

## Checklist


cmd/gotosocial/action/admin/media/prune/common.go
```diff
@@ -80,10 +80,6 @@ func setupPrune(ctx context.Context) (*prune, error) {
 func (p *prune) shutdown() error {
 	errs := gtserror.NewMultiError(2)
 
-	if err := p.storage.Close(); err != nil {
```
NyaaaWhatsUpDoc, line 83: the new store type doesn't bother with any kind of close mechanism as it's not really needed

internal/api/fileserver/servefile.go
```diff
@@ -224,10 +224,20 @@ func serveFileRange(rw http.ResponseWriter, r *http.Request, src io.Reader, rng
 		return
 	}
 
-	// Dump the first 'start' many bytes into the void...
-	if _, err := fastcopy.CopyN(io.Discard, src, start); err != nil {
-		log.Errorf(r.Context(), "error reading from source: %v", err)
-		return
+	if rs, ok := src.(io.ReadSeeker); ok {
```
tsmethurst, line 227: Nice :)

tsmethurst:
Good one!

Issue #823, opened 2022-09-13
[feature] Allow avatar + header to be deleted/unset via the api + settings panel


tsmethurst:
Right now, users can *set* profile pic and header, but they can't *unset*. We should implement functionality that allows them to do this. Simplest way I can think of right now to is to parse an account update form with a null value set for header or avatar as a delete.

PR #2926, opened 2024-05-22
Implement profile API

VyrCossont:
# Description

This Mastodon 4.2 extension provides capabilities missing from the existing Mastodon account update API: deleting an account's avatar or header.

See: https://docs.joinmastodon.org/methods/profile/ and #823

## Review notes

This is really an extension of the account API, not a new thing, so I've added it to the account processor. That required the account processor to take a dependency on the media *processor* as well as the media *manager* (since the media manager doesn't have the delete-attachment-by-ID functionality we need).

The two new endpoints have nearly identical logic except for which attachment they affect (avatar or header).

## Checklist

### DIFF
--- a/internal/api/client/accounts/profile.go
+++ b/internal/api/client/accounts/profile.go
@@ -108,11 +108,6 @@ func (m *Module) accountDeleteProfileAttachment(c *gin.Context, processDelete fu
 		return
 	}
 
-	if authed.Account.IsMoving() {
-		apiutil.ForbiddenAfterMove(c)
-		return
-	}
-
 	if _, err := apiutil.NegotiateAccept(c, apiutil.JSONAcceptHeaders...); err != nil {
 		apiutil.ErrorHandler(c, gtserror.NewErrorNotAcceptable(err, err.Error()), m.processor.InstanceGetV1)
 		return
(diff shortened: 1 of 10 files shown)


internal/api/client/accounts/profile.go
```diff
@@ -0,0 +1,128 @@
...
+// accountDeleteProfileAttachment checks that an authenticated account is present and allowed to alter itself,
+// runs an attachment deletion processor method, and returns the updated account.
+func (m *Module) accountDeleteProfileAttachment(c *gin.Context, processDelete func(context.Context, *gtsmodel.Account) (*apimodel.Account, gtserror.WithCode)) {
+	authed, err := oauth.Authed(c, true, true, true, true)
+	if err != nil {
+		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	if authed.Account.IsMoving() {
```
tsmethurst, line 111: You can remove this check here I think. We should allow people to clear their avatar + header after a move imo.

tsmethurst:
The dependency stuff feels a bit convoluted, I wonder if it would be good enough to just update the header/avatar media attachment IDs on the account and then let the media cleanup do the actual deletion of the media from storage later on when it does cleanup? I'll double check this when we're back from holiday.

VyrCossont:
Alternatively, we could move `account.Processor.deleteProfileAttachment` into `media.Processor` itself. It's sort of on the line between media and accounts, and could fit into either processor.

I'd rather delete the deleted avatar/header media immediately, as that's what the media processor already does when deleting a regular media attachment. The user might expect this in the case of something like uploading the wrong file.

PR #2936, opened 2024-05-27
[feature] Implement Filter API v2

VyrCossont:
# Description

This pull request implements the Mastodon v2 filter API. v2 allows filters to be grouped and named, but the most important addition is the ability to filter individual statuses, which is not possible with v1. Want to hide an inflammatory, irrelevant, or unfunny post that your friends keep boosting? Now you can.

This is a follow-up to #2793, which implemented server-side filtering but only exposed it through the v1 API.

It also fixes several tests and minor bugs in the v1 filter API.

## Impact

Phanpy only supports v2 filters, so this should make Phanpy more useful as a GtS frontend. (Tusky seems to be able to fall back to v1 if v2 isn't available, Feditext doesn't support v2 yet feditext/feditext#57, and Semaphore is EOL and only supports v1.)

## Open questions

This version implements the `keywords_attributes` param for [creating](https://docs.joinmastodon.org/methods/filters/#form-data-parameters) and [updating](https://docs.joinmastodon.org/methods/filters/#form-data-parameters-1) filter keywords in the same API call as creating or updating a filter, and extends this to filter statuses with `statuses_attributes` parameters.

However, this functionality doesn't seem to be very well thought out. (More detail in https://princess.industries/@vyr/statuses/01HXDEFGK0W8R4M4X7TKB2RKNJ which is an unlisted post on a GtS instance so open that link in your Fedi client.) It duplicates separate calls for creating and updating filter keywords, I've had to make a lot of assumptions implementing it because it's based on a poorly documented Rails feature which doesn't map naturally to form parameters, and I haven't found a client that uses it to test against. Tusky actually has a comment suggesting that they avoided it on purpose:

> [This is _terrible_, but the all-in-one update filter api Just Doesn't Work](https://github.com/tuskyapp/Tusky/blob/f9b5eeaf76a512bb267220002f4b4f3f4c201852/app/src/main/java/com/keylesspalace/tusky/components/filters/EditFilterViewModel.kt#L160)

So maybe it'd be smarter to omit `keywords_attributes` and `statuses_attributes`, on the basis that they can't be incompatible with the Mastodon implementation if we don't ship them at all. However, I'd like to hear from client devs and make sure nobody is expecting to use them first.

__Resolved: we're dropping these.__

## Checklist

### DIFF
--- a/internal/processing/filters/v2/get.go
+++ b/internal/processing/filters/v2/get.go
@@ -1,0 +1,78 @@
+// GoToSocial
+// Copyright (C) GoToSocial Authors admin@gotosocial.org
+// SPDX-License-Identifier: AGPL-3.0-or-later
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package v2
+
+import (
+	"context"
+	"errors"
+	"slices"
+	"strings"
+
+	apimodel "github.com/superseriousbusiness/gotosocial/internal/api/model"
+	"github.com/superseriousbusiness/gotosocial/internal/db"
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+)
+
+// Get looks up a filter by ID and returns it with keywords and statuses.
+func (p *Processor) Get(ctx context.Context, account *gtsmodel.Account, filterID string) (*apimodel.FilterV2, gtserror.WithCode) {
+	filter, err := p.state.DB.GetFilterByID(ctx, filterID)
+	if err != nil {
+		if errors.Is(err, db.ErrNoEntries) {
+			return nil, gtserror.NewErrorNotFound(err)
+		}
+		return nil, gtserror.NewErrorInternalError(err)
+	}
+	if filter.AccountID != account.ID {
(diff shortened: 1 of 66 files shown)


tsmethurst:
Wow! Thanks for this PR :D

Re: those `_attributes` params, I'd be happy to just drop them from this PR and we can implement that functionality later if necessary (if there are clients that break without it).

VyrCossont:
@tsmethurst sounds good. I'll remove that feature from this PR and we can resurrect it later if necessary.

internal/processing/filters/v2/get.go
```diff
@@ -0,0 +1,78 @@
...
+
+// GetAll looks up all filters for the current account and returns them with keywords and statuses.
+func (p *Processor) GetAll(ctx context.Context, account *gtsmodel.Account) ([]*apimodel.FilterV2, gtserror.WithCode) {
+	filters, err := p.state.DB.GetFiltersForAccountID(
+		ctx,
+		account.ID,
+	)
+	if err != nil {
+		if errors.Is(err, db.ErrNoEntries) {
+			return nil, nil
```
tsmethurst, line 56: Does this work when there are no filters? Ie., you're returning a nil slice, but does it actually serialize as `[]` or does it serialize as `null`?

internal/processing/filters/v2/keywordcreate.go
```diff
@@ -0,0 +1,64 @@
...
+	// Check that the filter is owned by the given account.
+	filter, err := p.state.DB.GetFilterByID(gtscontext.SetBarebones(ctx), filterID)
+	if err != nil {
+		if errors.Is(err, db.ErrNoEntries) {
+			return nil, gtserror.NewErrorNotFound(err)
+		}
+		return nil, gtserror.NewErrorInternalError(err)
+	}
+	if filter.AccountID != account.ID {
+		return nil, gtserror.NewErrorNotFound(nil)
```
tsmethurst, line 44: Rather than returning nil in cases like this, I think it would be better to pass in something like "filter %s doesn't belong to accountID %s" so that we can use it in internal logging if desired.

internal/processing/filters/v2/get.go
VyrCossont: I'll look into it. Maybe worth enforcing in a test.

VyrCossont: Added a test.

tsmethurst:
Great stuff!

PR #2937, opened 2024-05-28
Support listening on unix sockets

### LINKED ISSUE #1088: [feature] Possibility to bind to unix domain socket instead of ip:port

I would like GTS to bind to a unix domain socket. But it seems this isn't supported yet.

I would like to be able to configure `config.yml` like this
`bind-address: "/foo/bar"`
and GTS would bind to the socket `/foo/bar` instead of some port on some ip.

NONE

### Additional context.

_No response_

balki:
# Description

1. bind-address now accepts unix socket paths
2. Add config option trusted-platform (passing to gin-gonic)

closes #1088

## Checklist

- [ ] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat. (**found existing issue for same exact feature. Nothing more to add**)
- [X] I/we have made any necessary changes to documentation.
- [ ] I/we have added tests that cover new code. (**Using the feature in my instance @balki@balki.me. working good**)
- [X] I/we have run tests and they pass locally with the changes.
- [X] I/we have run `go fmt ./...` and `golangci-lint run`.


daenney:
Hiya! Welcome to GoToSocial and thanks for the contribution. In the future I would suggest [bringing changes up in the Matrix channel first](https://pkg.go.dev/github.com/gin-gonic/gin#Engine.RunUnix), or as an issue, because we may not always accept a change and that would spare you the effort of implementing it.

Gin already has a [built-in ability to listen on a unix socket](https://pkg.go.dev/github.com/gin-gonic/gin#Engine.RunUnix). I think adding the capability to GtS would be great, but if we're going to do so I'd rather we do it that way.

However, there is one problem currently in the form of gin-gonic/gin#2718 that would need resolving first. That would be solved by gin-gonic/gin#3359 but it seems the Gin maintainers haven't picked that up. However, looking at [recent commit history](https://github.com/gin-gonic/gin/commits/master/) there does seem to be a decent amount of activity. So if someone could perhaps attract the maintainer's attention to that PR we can get this fixed and then upgrade to a newer version of Gin instead, unlocking that capability.

Since the functionality already exists in Gin, bar that one issue we can hopefully get resolved, I'm going to go ahead and close this as we won't be taking the feature on in this form.

balki:
Looked into gin's [RunUnix](https://github.com/gin-gonic/gin/blob/master/gin.go#L527), Few notes
1. Existing socket is not removed before creating a listener. This means if the process crashes hard, it wont be able to come back up again till the socket file is manually deleted.
2. There is no way to specify the permission mode.

The function itself is just couple of lines wrapper around `http.Serve`. There is not any significant difference if gin's version is used or anyhttp or writing those lines in Gts itself. There is no specific integration to any other part of gin framework.

### Advantages of using anyhttp
1. Above mentioned cases are handled correctly
2. Provides one transparent API to multiple types of addresses. [Link](https://github.com/balki/anyhttp?tab=readme-ov-file#address-syntax)
This means Gts can just pass the address string from config file to `anyhttp.GetListener` and will get the right listener back (as of now, unix socket, tcp address, systemd fd as integer, systemd fd as socket name. Also takes care of handling the related environment variables). (I am planning to add some enhancements like a way to specify permissions mode as part of address URI).

### Importance of unix socket support
For most other applications, unix socket is just a nice to have. But for Gts (and other fediverse servers) it is very important because they make external requests to a lot of other servers and thus leaking the source ip address to all those servers. An outbound proxy can be configured (so only proxy's address is seen by other servers), however it is not possible to be very certain, all parts of the code(and dependencies) respect the proxy setting. So I run Gts in a private network namespace, so if proxy is not used, the request would just fail. Easiest way to get incoming requests from webserver to private network is to have a unix socket listener (Without messing with iptables, ip route settings).

    [External cloudflare/VPS] --> [LocalWebserver] --unix-socket-> [Private Gts]

I recommend to support unix sockets soon using any implementation you prefer. The problem with `Context.ClientIP` in gin has a workaround using `TrustedPlatform` setting. Though not perfect, works fine with minor config changes. 

I personally can patch and build on every new release, but the feature is good for all users who selfhost from private/home network.

Apologies for not discussing in matrix channel. I have not setup a matrix account yet.

daenney:
> But for Gts (and other fediverse servers) it is very important because they make external requests to a lot of other servers and thus leaking the source ip address to all those servers. 

This is not solved by having GtS *listen* on a unix socket. Connections GtS makes to other servers can already go through a proxy by setting the `http_proxy`/`https_proxy` environment variables.

balki:
> This is not solved by having GtS listen on a unix socket

Not directly. But having a private network namespace solves it. unix sockets make it easier to send requests to private network namespace. 

> Connections GtS makes to other servers can already go through a proxy by setting the http_proxy/https_proxy environment variables.

As I mentioned already it is impossible to certain the entire app would use the setting (now and in all future releases). May be Gts codebase would always use the setting. but what about the 200+ dependencies ? and future updates to the 200+ dependencies. E.g. Lets say an external tool is used to process images. And it tries to fetch an embedded image in an svg, or a cover for an audio file. Even [logging](https://en.wikipedia.org/wiki/Log4Shell) library can break. Just setting https_proxy setting is not guaranteed cover all the requests. It does not replace the need for network level protection like firewall or namespace.

If someone does not like a post or status, Unlike facebook or twitter, the person can easily find the approximate physical location from the ip address. So it is very important to not leak local ip address in a selfhosted social media app.

May be it is just me who is paranoid. I already have a way to patch and build. So I am good. Thank you for your time and fine app!

PR #2943, opened 2024-05-29
[feature] Add from: search operator and account_id query param

VyrCossont:
# Description

This pull request implements Mastodon-compatible functionality for restricting a search to statuses authored by a specific account:

- [`account_id` parameter to search API](https://docs.joinmastodon.org/methods/search/#query-parameters)
- `from:localuser` or `from:remoteuser@domain` as an additional (or only) term in search query text, useful for the majority of clients which don't support `account_id` (also implemented in Mastodon, documented only in online help)

There are two main use cases:

- Find your most recent conversations with somebody else
- Find your own posts

## Checklist

### DIFF
--- a/internal/processing/search/get.go
+++ b/internal/processing/search/get.go
@@ -62,14 +62,15 @@ func (p *Processor) Get(
 	req *apimodel.SearchRequest,
 ) (*apimodel.SearchResult, gtserror.WithCode) {
 	var (
-		maxID     = req.MaxID
-		minID     = req.MinID
-		limit     = req.Limit
-		offset    = req.Offset
-		query     = strings.TrimSpace(req.Query)                      // Trim trailing/leading whitespace.
-		queryType = strings.TrimSpace(strings.ToLower(req.QueryType)) // Trim trailing/leading whitespace; convert to lowercase.
-		resolve   = req.Resolve
-		following = req.Following
+		maxID         = req.MaxID
+		minID         = req.MinID
+		limit         = req.Limit
+		offset        = req.Offset
+		query         = strings.TrimSpace(req.Query)                      // Trim trailing/leading whitespace.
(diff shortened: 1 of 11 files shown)


internal/processing/search/get.go
```diff
@@ -826,12 +831,38 @@ func (p *Processor) statusesByText(
...
 	offset int,
 	query string,
+	fromAccountID string,
 	appendStatus func(*gtsmodel.Status),
 ) error {
+	var err error
+
+	// If the owning account for statuses was not provided as the account_id query parameter,
+	// it may still have been provided as a search operator in the query string.
+	queryPartSeparator := " "
```
tsmethurst, line 841: You can make this a const :)

internal/processing/search/get.go
```diff
@@ -826,12 +831,38 @@ func (p *Processor) statusesByText(
...
+	queryParts := strings.Split(query, queryPartSeparator)
+	nonOperatorQueryParts := make([]string, 0, len(queryParts))
+	for _, queryPart := range queryParts {
```
tsmethurst, line 844: Probably worth separating this out into a little function imo, just a little cleaner to read that way. What do you think?

tsmethurst:
Looks deece! We should document this in the user guide somewhere. Would you mind creating a page in there and adding to mkdocs.yaml or whatever it is?

internal/processing/search/get.go
VyrCossont: Sure, I can do that.

VyrCossont:
@tsmethurst no problem. I found the docs preview workflow in `CONTRIBUTING.md` and will add something to the user guide.

internal/processing/search/get.go
VyrCossont: I've set the new method up to return a `parsedQuery` struct so that it'll be easy to add more search operators in the future.

tsmethurst:
Thank you!

PR #2951, opened 2024-06-01
Compile filter keyword regexps when touched through PutFilter or UpdateFilter

VyrCossont:
# Description

User reported a bug in chat where their home timeline was crashing while filtering due to a nil `Regexp`:

```
gotosocial_1  | timestamp="01/06/2024 15:19:54.432" func=server.init.func1.Logger.14.1 level=INFO latency="25.915262ms"userAgent="***" method=GET statusCode=200 path=/api/v1/timelines/home clientIP=x.x.x.x errors="Error #01: recovered panic: runtime error: invalid memory addresss or nil pointer dereference
" requestID=4v3p5n4f04001700hav0 msg="OK: wrote 23B"
gotosocial_1  | timestamp="01/06/2024 15:19:54.432" func=server.init.func1.Logger.14.1 level=ERROR stacktrace="()
	/usr/local/go/src/runtime/panic.go:261
runtime.sigpanic()
	/usr/local/go/src/runtime/signal_unix.go:881
regexp.(*Regexp).doExecute()
	/usr/local/go/src/regexp/exec.go:527
()
	/usr/local/go/src/regexp/exec.go:514
regexp.(*Regexp).MatchString()
	/usr/local/go/src/regexp/regexp.go:531
typeutils.(*Converter).statusToAPIFilterResults()
	/drone/src/internal/typeutils/internaltofrontend.go:750
typeutils.(*Converter).statusToFrontend()
	/drone/src/internal/typeutils/internaltofrontend.go:1124
typeutils.(*Converter).StatusToAPIStatus()
	/drone/src/internal/typeutils/internaltofrontend.go:702
server.init.func1.HomeTimelineStatusPrepare.7()
	/drone/src/internal/processing/timeline/home.go:108
" requestID=4v3p5n4f04001700hav0 msg="recovered panic: runtime error: invalid memory address or nil pointer dereference"
```

This seems likely to be the result of creating or updating a filter through the v1 filter API, which uses `PutFilter` or `UpdateFilter` to create or update a filter and filter keyword at the same time, but doesn't compile the filter keyword's regexp because we missed those paths when compiled regexp caching was added. Will mark this as ready once I repro locally.

Followup to #2903

## Checklist

### DIFF
--- a/internal/db/bundb/filter.go
+++ b/internal/db/bundb/filter.go
@@ -154,6 +154,13 @@ func (f *filterDB) populateFilter(ctx context.Context, filter *gtsmodel.Filter) 
 }
 
 func (f *filterDB) PutFilter(ctx context.Context, filter *gtsmodel.Filter) error {
+	// Pre-compile filter keyword regular expressions.
+	for _, filterKeyword := range filter.Keywords {
+		if err := filterKeyword.Compile(); err != nil {
+			return gtserror.Newf("error compiling filter keyword regex: %w", err)
+		}
+	}
+
 	// Update database.
 	if err := f.db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
 		if _, err := tx.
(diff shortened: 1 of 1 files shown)


tsmethurst:
Thanks for picking this up :)

NyaaaWhatsUpDoc:
the changes look good to me once you've confirmed it working!

VyrCossont:
Confirmed that this patch fixes the scenario I described: filter created through v1 API, attempts to retrieve filtered posts crash because the filter regex was never compiled.

tsmethurst:
Nice one :)

PR #2960, opened 2024-06-04
[feature] User muting

### LINKED ISSUE #717: [feature] Mute users
We should implement user muting, so that people can mute users to prevent posts from those users being inserted into their home timeline, without having to unfollow that user.

See https://docs.joinmastodon.org/methods/accounts/mutes/ and https://docs.joinmastodon.org/methods/accounts/

To be fully compliant with the mastodon API, we can also try to implement duration-based muting, but we'll have to look into the cleanest + most performant way of doing this.

VyrCossont:
# Description

This pull request implements Mastodon-compatible user muting, including expiring mutes.

Closes #717

## Impact

Pretty much every Mastodon client supports mutes. I've tested with Semaphore and Feditext. Note that Semaphore doesn't support creating limited-duration mutes.

## Review notes

Mutes always affect posts from muted accounts. They also affect posts *replying to muted accounts*, unless there is at least one account replied to or mentioned who is not muted or blocked. This implicitly includes the user requesting the post, since it's not possible to mute yourself, and thus you'll always see replies to posts from muted accounts that also mention you.

Muting in this patch is implemented alongside filters, in the internal to API status/notification converter. As with filters, the patch doesn't make any attempt to delete expired mutes from the database, although unlike filters, there's no way to access an expired mute, since mutes aren't an externally visible object, and there's no API expectation to be able to refresh an expired mute by editing it. We could add a background cleanup task later without breaking any expectations (in `Cleaner`?); I haven't added one in this patch because I'm not yet familiar with GtS cleanup tasks.

I have thoughts on merging mutes and filters into the visibility cache system https://gist.github.com/VyrCossont/239272d87b6dcd3b611347835267d3c0 in the near future, but as is, this patch doesn't need any of that to work.

### Stuff to watch out for

Please keep an eye out for anything that was adapted from the similar block/unblock API that I forgot to update the names on.

### Changes to nearby code

`CacheConfiguration` had the wrong tag for the block ID cache mem ratio, and in another place, the block ID cache was commented with the wrong name.

## Checklist


internal/cache/cache.go
```diff
@@ -160,5 +162,6 @@ func (c *Caches) Sweep(threshold float64) {
 	c.GTS.Token.Trim(threshold)
 	c.GTS.Tombstone.Trim(threshold)
 	c.GTS.User.Trim(threshold)
+	c.GTS.UserMute.Trim(threshold)
```
NyaaaWhatsUpDoc, line 165: you need to add trimming for UserMuteIDs here too :)

internal/cache/db.go
```diff
@@ -47,7 +47,7 @@ type GTSCaches struct {
 	// Block provides access to the gtsmodel Block (account) database cache.
 	Block StructCache[*gtsmodel.Block]
 
-	// FollowIDs provides access to the block IDs database cache.
+	// BlockIDs provides access to the block IDs database cache.
```
NyaaaWhatsUpDoc, line 50: nice catch :sweat_smile:

internal/cache/db.go
```diff
@@ -1293,6 +1299,44 @@ func (c *Caches) initUser() {
...
+	cap := calculateResultCacheMax(
+		sizeofUserMute(), // model in-mem size.
+		config.GetCacheUserMuteMemRatio(),
+	)
+
+	log.Infof(nil, "cache size = %d", cap)
+
+	copyF := func(u1 *gtsmodel.UserMute) *gtsmodel.UserMute {
+		u2 := new(gtsmodel.UserMute)
+		*u2 = *u1
```
NyaaaWhatsUpDoc, line 1312: you'll also want to ensure the ptr fields to any account / target account fields get nil'ed out here, just to make sure they're not cached

internal/db/bundb/relationship_mute.go
```diff
@@ -0,0 +1,306 @@
...
+		if _, err := r.db.
+			NewSelect().
+			TableExpr("?", bun.Ident("user_mutes")).
+			ColumnExpr("?", bun.Ident("id")).
+			Where("? = ?", bun.Ident("account_id"), accountID).
+			WhereGroup(" AND ", func(q *bun.SelectQuery) *bun.SelectQuery {
+				var notYetExpiredSQL string
+				switch r.db.Dialect().Name() {
+				case dialect.SQLite:
+					notYetExpiredSQL = "? > DATE('now')"
```
NyaaaWhatsUpDoc, line 288: (just musing here: at some point we should probably just register an sqlite function that handles `NOW()` as `DATE('now')` so we don't need this type of stuff)

internal/processing/account/mute.go
```diff
@@ -0,0 +1,195 @@
...
+	targetAccountID string,
+	form *apimodel.UserMuteCreateUpdateRequest,
+) (*apimodel.Relationship, gtserror.WithCode) {
+	targetAccount, existingMute, errWithCode := p.getMuteTarget(ctx, requestingAccount, targetAccountID)
+	if errWithCode != nil {
+		return nil, errWithCode
+	}
+
+	if existingMute != nil &&
+		existingMute.MutingNotifications() == *form.Notifications &&
```
NyaaaWhatsUpDoc, line 50: this might be an issue if form.Notifications is nil right? (or is it always provided?)

either way might be a bit more concise here to instead do `util.EqualPtrs(existingMute.Notifications, form.Notifications)`

internal/processing/account/mute.go
```diff
@@ -0,0 +1,195 @@
...
+		mute.ID = existingMute.ID
+	} else {
+		mute.ID = id.NewULID()
+	}
+	if form.Duration != nil {
+		mute.ExpiresAt = time.Now().Add(time.Second * time.Duration(*form.Duration))
+	}
+
+	if err := p.state.DB.PutMute(ctx, mute); err != nil {
+		err = fmt.Errorf("MuteCreate: error creating or updating mute in db: %w", err)
```
NyaaaWhatsUpDoc, line 74: instead of using `fmt.Errorf()` if you instead use `gtserror.Newf()` it automatically includes the calling function name as a prefix, so you won't need the "MuteCreate: " prefix :)

internal/processing/account/mute.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	if existingMute == nil {
+		// Already not muted, nothing to do.
+		return p.RelationshipGet(ctx, requestingAccount, targetAccountID)
+	}
+
+	// We got a mute, remove it from the db.
+	if err := p.state.DB.DeleteMuteByID(ctx, existingMute.ID); err != nil {
+		err := fmt.Errorf("MuteRemove: error removing mute from db: %w", err)
```
NyaaaWhatsUpDoc, line 99: see above about `gtserror.Newf()`

internal/processing/account/mute.go
```diff
@@ -0,0 +1,195 @@
...
+	ctx context.Context,
+	requestingAccount *gtsmodel.Account,
+	page *paging.Page,
+) (*apimodel.PageableResponse, gtserror.WithCode) {
+	mutes, err := p.state.DB.GetAccountMutes(ctx,
+		requestingAccount.ID,
+		page,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return nil, gtserror.NewErrorInternalError(err)
```
NyaaaWhatsUpDoc, line 117: would be good to add some contextual error info here, e.g. `err := gtserror.Newf("error getting account mutes: %w", err)`

internal/processing/account/mute.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (p *Processor) getMuteTarget(
+	ctx context.Context,
+	requestingAccount *gtsmodel.Account,
+	targetAccountID string,
+) (*gtsmodel.Account, *gtsmodel.UserMute, gtserror.WithCode) {
+	// Account should not mute or unmute itself.
+	if requestingAccount.ID == targetAccountID {
+		err := fmt.Errorf("getMuteTarget: account %s cannot mute or unmute itself", requestingAccount.ID)
```
NyaaaWhatsUpDoc, line 170: same as above re: `gtserror.Newf()`, and same goes for the rest of this function

internal/filter/usermute/usermute.go
```diff
@@ -0,0 +1,80 @@
...
+func (e *compiledUserMuteListEntry) expired(now time.Time) bool {
+	return !e.ExpiresAt.IsZero() && e.ExpiresAt.Before(now)
+}
+
+type CompiledUserMuteList struct {
+	byTargetAccountID map[string]compiledUserMuteListEntry
+}
+
+func NewCompiledUserMuteList(mutes []*gtsmodel.UserMute) (c *CompiledUserMuteList) {
+	c = &CompiledUserMuteList{byTargetAccountID: map[string]compiledUserMuteListEntry{}}
```
NyaaaWhatsUpDoc, line 57: it's probably worth preallocating this map here to be of the same length as the incoming list of mutes.

internal/gtsmodel/usermute.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Expired returns whether the mute has expired at a given time.
+// Mutes without an expiration timestamp never expire.
+func (u *UserMute) Expired(now time.Time) bool {
+	return !u.ExpiresAt.IsZero() && !u.ExpiresAt.After(now)
+}
+
+// MutingNotifications returns whether the mute applies to notifications as well as statuses.
+func (u *UserMute) MutingNotifications() bool {
```
NyaaaWhatsUpDoc, line 44: given the `Notifications` field is always going to be set coming out of the database, and the fact we have to use a ptr is a peculiarity of the orm we're using, it might be worth just getting rid of this and instead just always directly dereferencing the ptr. for new UserMute{} objects being inserted we can just always make sure we allocate a `Notifications` field value. it just reduces the number of code abstractions, keeps with how we're doing things in a lot of places and makes it simpler if we ever do move from our current orm

internal/typeutils/internaltofrontend.go
```diff
@@ -724,14 +726,48 @@ func (c *Converter) statusToAPIFilterResults(
...
 	now := time.Now()
+
+	// If the requesting account mutes the account that created this status, we're done.
+	if mutes.Matches(s.AccountID, filterContext, now) {
+		return nil, statusfilter.ErrHideStatus
+	}
+	// If this is a multi-party conversation and all of the accounts replied to or mentioned are
+	// invisible to the requesting account (due to blocks, domain blocks, moderation, etc.),
+	// or muted, do not show this status.
+	otherParties := make([]*gtsmodel.Account, 0, 1+len(s.Mentions))
```
NyaaaWhatsUpDoc, line 745: i think this block of code from here down could do with a bit of breaking-up with some comments explaining what's going on, just for easier reading in the future, if that's good please :)

NyaaaWhatsUpDoc:
a number of comments, but great work :eyes:

internal/cache/cache.go
VyrCossont: Done.

internal/cache/db.go
VyrCossont: Done.

internal/db/bundb/relationship_mute.go
VyrCossont: Good idea. I'm surprised we hadn't run into this yet, but I didn't find any other instances of date math in SQL in this codebase.

internal/gtsmodel/usermute.go
VyrCossont: Sure, let's do that. For something like `UserMute`, it's only created in a handful of places.

internal/filter/usermute/usermute.go
VyrCossont: Done.

internal/processing/account/mute.go
VyrCossont: At this point, `form` should have passed through `normalizeCreateUpdateMute`, which sets `form.Notifications` to its default if it wasn't provided by the API caller. And since we're eliminating `existingMute.MutingNotifications()`, we can write this as `*existingMute.Notifications == *form.Notifications`.

internal/processing/account/mute.go
VyrCossont: Done.

internal/processing/account/mute.go
VyrCossont: Done.

internal/processing/account/mute.go
VyrCossont: Good idea.

internal/typeutils/internaltofrontend.go
VyrCossont: Done. This iteration should be a lot easier to read.

VyrCossont:
Not sure what's up with the test timeouts. Running the full test script locally to see what I can find.

VyrCossont:
Given that several other PRs are stuck with similar errors and I can't reproduce them locally, I suspect something's up with CI. Did give me an excuse to fix some style issues though.

tsmethurst:
Looks like a CI thing yeah, I'll kick it :)

internal/api/client/accounts/mute.go
```diff
@@ -0,0 +1,163 @@
...
+//			description: not acceptable
+//		'500':
+//			description: internal server error
+func (m *Module) AccountMutePOSTHandler(c *gin.Context) {
+	authed, err := oauth.Authed(c, true, true, true, true)
+	if err != nil {
+		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
```
tsmethurst, line 94: Could you put that check for isMigrating or isMoved or whatever it's called here, so we can make sure accounts that are migrating/have migrated can't POST new mutes?

internal/api/client/accounts/mute.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func normalizeCreateUpdateMute(form *apimodel.UserMuteCreateUpdateRequest) error {
+	// Apply defaults for missing fields.
+	form.Notifications = util.Ptr(util.PtrValueOr(form.Notifications, false))
+
+	// Normalize mute duration if necessary.
+	// If we parsed this as JSON, expires_in
+	// may be either a float64 or a string.
+	if ei := form.DurationI; ei != nil {
```
tsmethurst, line 139: Since we have to do this in quite a few places now, it might be worth doing a chore PR at some point to put this parsing in apiutil. Not necessarily for this PR though, it's plenty of changes as is :)

internal/gtsmodel/usermute.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// UserMute refers to the muting of one account by another.
+type UserMute struct {
+	ID              string    `bun:"type:CHAR(26),pk,nullzero,notnull,unique"`                                           // id of this item in the database
+	CreatedAt       time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"`                        // when was item created
+	UpdatedAt       time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"`                        // when was item last updated
+	ExpiresAt       time.Time `bun:"type:timestamptz,nullzero"`                                                          // Time mute should expire. If null, should not expire.
+	AccountID       string    `bun:"type:CHAR(26),unique:user_mutes_account_id_target_account_id_uniq,notnull,nullzero"` // Who does this mute originate from?
+	Account         *Account  `bun:"rel:belongs-to"`                                                                     // Account corresponding to accountID
```
tsmethurst, line 31: We don't really use these `belongs-to` tags anymore since we do our population from caches (and separate db queries) and not from joins, so you can replace `bun:"rel:belongs-to"` with `bun:"-"` to indicate that the field shouldn't be stored in a table.

internal/gtsmodel/usermute.go
```diff
@@ -0,0 +1,41 @@
...
+	TargetAccountID string    `bun:"type:CHAR(26),unique:user_mutes_account_id_target_account_id_uniq,notnull,nullzero"` // Who is the target of this mute?
+	TargetAccount   *Account  `bun:"rel:belongs-to"`                                                                     // Account corresponding to targetAccountID
+	Notifications   *bool     `bun:",nullzero,notnull,default:false"`                                                    // Apply mute to notifications as well as statuses.
+}
+
+// Expired returns whether the mute has expired at a given time.
+// Mutes without an expiration timestamp never expire.
+func (u *UserMute) Expired(now time.Time) bool {
+	return !u.ExpiresAt.IsZero() && !u.ExpiresAt.After(now)
```
tsmethurst, line 40: I think elsewhere in this PR you use `u.ExpiresAt.Before(now)`; better to use one or the other throughout for consistency imo.

tsmethurst:
Looks great :) Couple small comments then I think this can be squerged, good work

internal/api/client/accounts/mute.go
NyaaaWhatsUpDoc: yep was just thinking this too. though yeah another PR for another day

NyaaaWhatsUpDoc: (i'm resolving this just so it's clear what is / isn't remaining work to be done on this particular PR)

internal/gtsmodel/usermute.go
VyrCossont: I think that was either a previous iteration of this PR or the filter implementation next to the mute implementation, but I'm going to take this opportunity to add `.Expired()` to filters for consistency.

internal/api/client/accounts/mute.go
VyrCossont: Done. I've left it off the *unmute* equivalent, since that's consistent with other APIs that allow deleting by moved users so they can clean up their data if they want.

internal/gtsmodel/usermute.go
VyrCossont: Done.

internal/gtsmodel/usermute.go
VyrCossont: Oh wait no I found it. It was in `compiledUserMuteListEntry`.

PR #2963, opened 2024-06-04
[chore] Exit on misconfigured journal mode

### LINKED ISSUE #2962: [bug] WASM SQLite build + default config is slow on not-Linux

When running with the WASM SQLite build on anything other than Linux, `db-sqlite-journal-mode` must not be set to `WAL` ~~as that can currently lead to database corruption~~ as otherwise performance will be atrocious.

Instead, we should use `TRUNCATE` in those cases. This has a slight performance impact, but should be fine in practice for small instances especially while this is still in the experimental stages.

main

### GoToSocial Arch

_No response_

### What happened?

_No response_

_No response_

### How to reproduce it?

_No response_

We should do two things:

* [ ] Add some documentation about this
* [ ] Update the config parser to check for `runtime.GOOS` and if that's not Linux on the WASM build throw an error if the journal mode is set to WAL

daenney:
# Description

Adds documentation and a configuration validation check to ensure that when folks run with the WASM SQLite build on anything other than Linux we don't allow them to run with journaling set to WAL. That will result in abysmal performance until ncruces/go-sqlite3#85 is resolved.

Fixes #2962

## Checklist

### DIFF
--- a/docs/configuration/database.md
+++ b/docs/configuration/database.md
@@ -12,6 +12,9 @@ To configure GoToSocial to use SQLite, change `db-type` to `sqlite`. The `addres
 
 Note that the `:memory:` setting will use an *in-memory database* which will be wiped when your GoToSocial instance stops running. This is for testing only and is absolutely not suitable for running a proper instance, so *don't do this*.
 
+!!! warning "WASM SQLite build"
+    When running the experimental WASM SQLite build on anything other than Linux, you have to set `db-sqlite-journal-mode` to `TRUNCATE` instead of `WAL`. You'll experience abysmal performance otherwise.
+
 ## Postgres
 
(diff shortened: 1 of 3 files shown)


tsmethurst:
Since a fix for WAL mode on the BSDs is now merged, maybe this PR isn't necessary, especially since it's still experimental/behind a build tag anyway. Once the BSD fix for WASM sqlite is in a release, and we can upstep to it, this documentation will become unnecessary anyway. What do you think?

daenney:
Yup. When we first discussed it, it didn't seem like there was an easy answer / implementation that we'd get to quickly. 12 hours later and Nuno fixed it :smile:.

ncruces:
https://github.com/ncruces/go-sqlite3/releases/tag/v0.16.1

PR #2970, opened 2024-06-06
[bugfix] update media if more than just url changes

### LINKED ISSUE #2851: [bug] Remote edits of media descriptions are not applied

When a remote post is edited, changes to media descriptions are not visible on Gotosocial.

v0.15.0

### GoToSocial Arch

amd64 Binary

### What happened?

A remote post which had no image description got edited to add them, on Gotosocial the image description stays absent.

The image description becomes visible as it would be on the origin instance.

### How to reproduce it?

1. Follow a test account on Mastodon
2. From the Mastodon account post an image without description / alt text
3. Edit the post to add a description / alt text
4. On Gotosocial the Post remains without image description / alt text

_No response_

NyaaaWhatsUpDoc:
# Description

Refacts much of the shared media attachment loading / updating logic in the dereferencer, both for readability and shared code. Also now handles the case of attachment descriptions / blurhashes changing, in both cases it updates the database model, but particularly in the latter it forces a recache to fetch the latest image version as the content is likely to have changed.

closes #2851

## Checklist

### DIFF
--- a/internal/federation/dereferencing/util.go
+++ b/internal/federation/dereferencing/util.go
@@ -18,10 +18,121 @@
 package dereferencing
 
 import (
+	"context"
+	"io"
+	"net/url"
 	"slices"
 
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/media"
+	"github.com/superseriousbusiness/gotosocial/internal/transport"
+	"github.com/superseriousbusiness/gotosocial/internal/util"
 )
+
+// loadAttachment handles the case of a new media attachment
+// that requires loading. it stores and caches from given data.
+func (d *Dereferencer) loadAttachment(
+	ctx context.Context,
+	tsport transport.Transport,
+	accountID string, // media account owner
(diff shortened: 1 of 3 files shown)


internal/federation/dereferencing/util.go
```diff
@@ -18,11 +18,123 @@
...
+	}
+
+	// Force load attachment recache *right now*.
+	recached, err := processing.LoadAttachment(ctx)
+
+	if recached != nil {
+		// Only set recached media
+		// file if it was at least
+		// partially downloaded.
+		existing = recached
```
tsmethurst, line 132: Does this replace the pointer to existing with the pointer to recached? This feels a bit odd to do since you're returning something anyway.

NyaaaWhatsUpDoc: oh right you mean just replace with `return recached, nil` ?

tsmethurst: I think so yeah, else it could introduce cases where the caller to this function is surprised to find that the pointer to existing that they passed in has been replaced with something else :P

NyaaaWhatsUpDoc: how about: a63de9648

tsmethurst: looks good!

PR #2972, opened 2024-06-06
Implement filters_changed stream event

### LINKED ISSUE #2965: [feature] Implement filters_changed stream event

GtS doesn't notify streaming clients when filters change (because Vyr didn't think of it when she was first implementing filters). If the user is using multiple clients (including things like [`slurp`](https://github.com/VyrCossont/slurp)'s filter import) and one client changes filters, the other ones will not be aware that they have outdated filters.

GtS should emit the [`filters_changed`](https://docs.joinmastodon.org/methods/streaming/#events) stream event when filter keywords or statuses change. (The Mastodon docs only mention it for filter keyword changes, but Mastodon also emits `filters_changed` when filter statuses change.)

Non-streaming clients that want to stay updated can periodically poll the filter list endpoint, but this causes additional load on GtS.

### Additional context.

This event seems most useful for filter API v1 clients, since they can only apply filter keywords client-side.

VyrCossont:
# Description

Implements the `filters_changed` stream event.

Closes #2965

## Checklist


tsmethurst:
Nice one :)

PR #2982, opened 2024-06-09
[bugfix] boost and account recursion

NyaaaWhatsUpDoc:
# Description

fixes a possible recursion issue with moved accounts referrring back to the original account. adds a protection for boost wrapping statuses to make we're never allowing processing of boostOf-boostOf's

- fix possible recursion issue setting Moved account in type converter
- fix possible recursion issue setting Reblogged status in type converter
- add protections for accepting boost wrapping statuses as boost targets
- add protections for move account recursion loops (closed move chains)

## Checklist

### DIFF
--- a/internal/db/bundb/account.go
+++ b/internal/db/bundb/account.go
@@ -259,7 +259,8 @@ func (a *accountDB) GetAccountsByMovedToURI(ctx context.Context, uri string) ([]
 	// given moved_to_uri column.
 	if err := a.db.NewSelect().
 		Table("accounts").
-		Column("id").Where("? = ?", bun.Ident("moved_to_uri"), uri).
+		Column("id").
+		Where("? = ?", bun.Ident("moved_to_uri"), uri).
 		Scan(ctx, &accountIDs); err != nil {
 		return nil, err
 	}
--- a/internal/db/bundb/account.go
+++ b/internal/db/bundb/account.go
@@ -252,6 +252,26 @@ func (a *accountDB) GetInstanceAccount(ctx context.Context, domain string) (*gts
 	return a.GetAccountByUsernameDomain(ctx, username, domain)
 }
 
+func (a *accountDB) GetAccountsByMovedToURI(ctx context.Context, uri string) ([]*gtsmodel.Account, error) {
+	var accountIDs []string
+
+	// Find all account IDs with
+	// given moved_to_uri column.
+	if err := a.db.NewSelect().
+		Table("accounts").
+		Column("id").Where("? = ?", bun.Ident("moved_to_uri"), uri).
+		Scan(ctx, &accountIDs); err != nil {
+		return nil, err
+	}
+
+	if len(accountIDs) == 0 {
+		return nil, nil
+	}
+
+	// Return account models for all found IDs.
+	return a.GetAccountsByIDs(ctx, accountIDs)
+}
+
 // GetAccounts selects accounts using the given parameters.
 // Unlike with other functions, the paging for GetAccounts
 // is done not by ID, but by a concatenation of `[domain]/@[username]`,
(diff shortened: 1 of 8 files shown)


internal/processing/common/status.go
```diff
@@ -169,11 +169,19 @@ func (p *Processor) UnwrapIfBoost(
...
-	return p.GetVisibleTargetStatus(ctx,
+	status, errWithCode := p.GetVisibleTargetStatus(ctx,
 		requester,
 		status.BoostOfID,
 		nil,
 	)
+	if errWithCode != nil {
+		return nil, errWithCode
+	}
+	if status.BoostOfID != "" {
```
NyaaaWhatsUpDoc, line 180: this specifically ensures anywhere we're dealing with boosted statuses and needing to unwrap them, we will always be dealing with the most unwrapped status, and not dealing with bad input data

internal/db/bundb/account.go
```diff
@@ -252,6 +252,26 @@ func (a *accountDB) GetInstanceAccount(ctx context.Context, domain string) (*gts
...
 }
 
+func (a *accountDB) GetAccountsByMovedToURI(ctx context.Context, uri string) ([]*gtsmodel.Account, error) {
+	var accountIDs []string
+
+	// Find all account IDs with
+	// given moved_to_uri column.
+	if err := a.db.NewSelect().
+		Table("accounts").
+		Column("id").Where("? = ?", bun.Ident("moved_to_uri"), uri).
```
NyaaaWhatsUpDoc, line 262: we could add an index for this, but my initial feeling is that account moves are so infrequent that i'm not sure it's worth it?

internal/processing/status/boost.go
```diff
@@ -69,7 +76,7 @@ func (p *Processor) BoostCreate(
 	}
 
 	if !boostable {
-		err := gtserror.New("status is not boostable")
+		err := errors.New("status is not boostable")
```
tsmethurst, line 79: This error message doesn't reach callers and should just be used for internal logging, so it's probably fine to leave it as gtserror no? Unless I'm missing something

internal/db/bundb/account.go
tsmethurst: We can try it for a bit and see :) if we end up with performance issues we can always add an index later.

tsmethurst:
This seems reasonable, just a few wee comments about errors.

internal/processing/status/boost.go
```diff
@@ -58,7 +59,13 @@ func (p *Processor) BoostCreate(
 		return nil, errWithCode
 	}
 
-	// Ensure valid boost target.
+	// Check is viable target.
+	if target.BoostOfID != "" {
+		err := gtserror.Newf("target status %s is boost wrapper", target.URI)
+		return nil, gtserror.NewErrorInternalError(err)
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: good idea!

internal/processing/status/boost.go
NyaaaWhatsUpDoc: oh actually i did intend for this to reach the caller (i just fucked it up :p), unless you think this shouldn't be exposed to the caller?

internal/processing/status/boost.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: see: 69af75dbb0bc6b07718c39e092d0d0467b6d2a66

internal/processing/status/boost.go
NyaaaWhatsUpDoc: ignore the previous message i sent, you're right :)

NyaaaWhatsUpDoc: see: 258bfbd4a

PR #2989, opened 2024-06-10
[feature] add worker task serialization logic

NyaaaWhatsUpDoc:
# Description

Adds necessary serialization / deserialization logic to the different worker message / task types for persisted work queues on startup / shutdown. Also adds a very barebones WorkerTask{} type that will be later filled out + used to store them in the database.

## Checklist


cmd/gotosocial/action/server/server.go
```diff
@@ -249,6 +249,11 @@ var Start action.GTSAction = func(ctx context.Context) error {
 		return fmt.Errorf("error scheduling cleaner jobs: %w", err)
 	}
 
+	// Now schedule background cleaning tasks.
```
tsmethurst, line 252: Is this repeat code? Or am i losing it?

internal/httpclient/transport.go
```diff
@@ -45,12 +44,6 @@ func (t *signingtransport) RoundTrip(r *http.Request) (*http.Response, error) {
 		r.Header.Del("Signature")
 		r.Header.Del("Digest")
 
-		// Rewind body reader and content-length if set.
```
tsmethurst, line 48: Could you explain the reason this is removed now? Not in the code I mean, just for my sake :)

internal/messages/messages.go
```diff
@@ -50,6 +56,105 @@ type FromClientAPI struct {
...
+		originID = msg.Origin.ID
+	}
+
+	// Set target account ID.
+	if msg.Target != nil {
+		targetID = msg.Target.ID
+	}
+
+	// Marshal GTS model as raw JSON block.
+	modelJSON, err := json.Marshal(msg.GTSModel)
```
tsmethurst, line 98: Oh interesting... does this work properly for all our gtsmodels? We've got some pretty funky types so I wonder if it's not a lossy translation.

internal/messages/messages.go
```diff
@@ -101,3 +333,61 @@ func FederatorMsgIndices() []structr.IndexConfig {
...
+func resolveGTSModel(typ string, data []byte) (interface{}, error) {
+	if typ == "" && data == nil {
+		// No data given.
+		return nil, nil
+	}
+
+	var value interface{}
+
+	switch typ {
+	case reflect.TypeOf((*gtsmodel.Account)(nil)).String():
```
tsmethurst, line 363: Is it worth storing these type names in memory as vars so we don't have to do this reflection every time resolveGTSModel is called? Or would that not make much difference?

internal/messages/messages_test.go
```diff
@@ -0,0 +1,166 @@
+package messages_test
```
tsmethurst, line 1: License header missing here.

internal/messages/messages_test.go
```diff
@@ -0,0 +1,166 @@
...
+			APActivityType: ap.ActivityCreate,
+			GTSModel:       &gtsmodel.Status{ID: "69", Content: "hehe"},
+			TargetURI:      "https://gotosocial.org",
+			Origin:         &gtsmodel.Account{ID: "654321"},
+			Target:         &gtsmodel.Account{ID: "123456"},
+		},
+		data: toJSON(map[string]any{
+			"ap_object_type":   ap.ObjectNote,
+			"ap_activity_type": ap.ActivityCreate,
+			"gts_model":        json.RawMessage(toJSON(&gtsmodel.Status{ID: "69", Content: "hehe"})),
```
tsmethurst, line 30: It's probably worth taking a whole status from the testrig for this, with all the funky fields a whole status includes, and ensure that serializes/deserializes as expected. What do you think?

internal/transport/deliver.go
```diff
@@ -130,25 +130,28 @@ func (t *transport) prepare(
...
-	var body byteutil.ReadNopCloser
-	body.Reset(data)
-
 	// Prepare POST signer.
 	sign := t.signPOST(data)
 
+	// Use *bytes.Reader for request body,
+	// as NewRequest() automatically will
+	// set .GetBody and content-length.
+	// (this handles necessary rewinding).
```
tsmethurst, line 139: Ahh I see...

internal/transport/deliver.go
```diff
@@ -130,25 +130,28 @@ func (t *transport) prepare(
...
 
 	// Prepare a new request with data body directed at URL.
-	r, err := http.NewRequestWithContext(ctx, "POST", url, &body)
+	r, err := http.NewRequestWithContext(ctx, "POST", to.String(), body)
 	if err != nil {
 		return nil, gtserror.Newf("error preparing request: %w", err)
 	}
 
+	// Set our predefined controller user-agent.
+	r.Header.Set("User-Agent", t.controller.userAgent)
```
tsmethurst, line 153: Oh, were we not setting this correctly before?

tsmethurst (commented):
Looks good to me, I've got a few questions about ser/des but that's it :)

cmd/gotosocial/action/server/server.go
NyaaaWhatsUpDoc: no it is! lemme get that

NyaaaWhatsUpDoc: done 7201baecb

internal/messages/messages.go
NyaaaWhatsUpDoc: my thought was that since this is only done on startup, it would be better to incur if anything a bit more of a cost here and reduce memory usage the rest of the time (however negligible) :)

internal/messages/messages_test.go
NyaaaWhatsUpDoc: done: b32969db6

internal/messages/messages.go
NyaaaWhatsUpDoc: resolving this as we can move this conversion to the thread about using more complex testrig models instead :)

internal/messages/messages_test.go
NyaaaWhatsUpDoc: updated to use testrig models here: 2fda9bc0f

NyaaaWhatsUpDoc: though had to switch to using "google/cmp" instead of "testify" for the test comparisons as testify doesn't handle complex types very well like RSA keys :p

internal/transport/deliver.go
NyaaaWhatsUpDoc: apparently not!

internal/messages/messages_test.go
NyaaaWhatsUpDoc: to be doubly sure i added a check to ensure account public / private keys are preserved (as those are the most likely to get bodged during json serialization): d4315aa35

tsmethurst:
Looks good to me now :) if you're happy then squerge away

PR #2990, opened 2024-06-10
[bugfix] Deref stats async, serve stub collections if handshaking

tsmethurst:
# Description

>

This pull request solves a bug introduced in 0.16.0-rc1 where account stats were fetched during the initial dereference attempt, leading to a potential recursive handshake loop where two GtS instances would fail to dereference each other.

The fix is to:

1. Only fetch stats asynchronously, and during a first dereference just return local stats (ie., 0 followers, 0 following, 0 statuses).
2. Serve bare stub endpoints for an Actor's collections (without even totalItems) if dereference is done during handshaking, just to get things moving along. This part isn't strictly necessary to fix this bug now, but it will ensure that future softwares trying to deref GtS collections, who withhold their collections first, can still get something in response.

The PR also stops serving totalItems in cases where the receiving account has decided to hide their collections. This brings it in line with what Mastodon glitch does.

Oh, and the PR also uses our paging code for `outbox`, in contrast to the sorta freewheeling code we had there before.

## Checklist

### DIFF
--- a/internal/federation/dereferencing/account.go
+++ b/internal/federation/dereferencing/account.go
@@ -102,10 +102,14 @@ func (d *Dereferencer) GetAccountByURI(ctx context.Context, requestUser string, 
 	}
 
 	if accountable != nil {
-		// This account was updated, enqueue re-dereference featured posts.
+		// This account was updated, enqueue re-dereference featured posts + stats.
 		d.state.Workers.Dereference.Queue.Push(func(ctx context.Context) {
 			if err := d.dereferenceAccountFeatured(ctx, requestUser, account); err != nil {
 				log.Errorf(ctx, "error fetching account featured collection: %v", err)
+			}
+
+			if err := d.dereferenceAccountStats(ctx, requestUser, account); err != nil {
+				log.Errorf(ctx, "error fetching account stats: %v", err)
 			}
 		})
 	}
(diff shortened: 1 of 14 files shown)


internal/federation/dereferencing/account.go
```diff
@@ -712,9 +728,11 @@ func (d *Dereferencer) enrichAccount(
 		log.Errorf(ctx, "error fetching remote emojis for account %s: %v", uri, err)
 	}
 
-	// Fetch followers/following count for this account.
-	if err := d.fetchRemoteAccountStats(ctx, latestAcc, requestUser); err != nil {
-		log.Errorf(ctx, "error fetching remote stats for account %s: %v", uri, err)
+	// Ensure basic account stats populated; real stats will be fetched from remote asynchronously.
+	if err := d.state.DB.PopulateAccountStats(ctx, latestAcc); err != nil {
```
NyaaaWhatsUpDoc, line 732: remember the account may not actually be inserted into the database yet, so the regenerate status transaction may fail. it feels like logically this might fit better in getAccountByURI()

internal/processing/fedi/collections.go
```diff
@@ -45,91 +47,162 @@ func (p *Processor) InboxPost(ctx context.Context, w http.ResponseWriter, r *htt
...
+		params.Query.Set("limit", "40") // enables paging
+		obj = ap.NewASOrderedCollection(params)
+
+	default:
+		// Paging enabled.
+		// Get page of full public statuses.
+		statuses, err := p.state.DB.GetAccountStatuses(
+			ctx,
+			receivingAcct.ID,
+			page.Limit,    // limit
```
NyaaaWhatsUpDoc, line 111: you might want to use `page.GetLimit()` here in the case that page object is nil (these functions specifically handle that)

NyaaaWhatsUpDoc:
just a few comments but otherwise looking good! nice work :ok_hand:

internal/processing/fedi/collections.go
tsmethurst: Ack! Thanks! https://github.com/superseriousbusiness/gotosocial/pull/2990/commits/86f9d94c17e7a5ff3d072288858d2564801c1d42

PR #3000, opened 2024-06-12
[chore] media and emoji refactoring

NyaaaWhatsUpDoc:
# Description

Refactors a bunch of the media manager and media fetching code in the dereferencer to use more shared code, follow more of our standard behaviour / current idioms and store media / emojis in the database *immediately* at the start of processing as uncached. Such that they can *always* be picked up and recached. This also removes the necessity of our media workers entirely! Which as it turns out were only ever actually being used when media processing got cancelled part-way through with a context.Canceled / context.DeadlineExceeded, which can easily just be done by the dereference worker given it's not a common occurrence. This should also result in less database calls during account / status dereferencing of remote emojis, as it now follows the same style of logic as media attachments in that it checks the existing model first for unchanged emojis.

So, some useful refactoring / yak shaving, and some possible performance benefits too :)

## Checklist


tsmethurst:
You nutter, you've only gone and done it, she's only gone and done it lads.

internal/gtsmodel/mediaattachment.go
```diff
@@ -44,22 +44,30 @@ type MediaAttachment struct {
...
 
 // Thumbnail refers to a small image thumbnail derived from a larger image, video, or audio file.
 type Thumbnail struct {
-	Path        string    `bun:",nullzero,notnull"`                                           // Path of the file in storage.
-	ContentType string    `bun:",nullzero,notnull"`                                           // MIME content type of the file.
-	FileSize    int       `bun:",notnull"`                                                    // File size in bytes
-	UpdatedAt   time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"` // When was the file last updated.
-	URL         string    `bun:",nullzero"`                                                   // What is the URL of the thumbnail on the local server
-	RemoteURL   string    `bun:",nullzero"`                                                   // What is the remote URL of the thumbnail (empty for local media)
+	Path        string `bun:",notnull"`  // Path of the file in storage.
```
tsmethurst, line 66: What's the reasoning for changing these from nullzero? Just checking, not really an issue or a problem.

internal/db/bundb/migrations/20240613091853_drop_unused_media_columns.go
```diff
@@ -0,0 +1,83 @@
...
+			return nil
+		})
+	}
+
+	if err := Migrations.Register(up, down); err != nil {
+		panic(err)
+	}
+}
+
+func doesColumnExist(ctx context.Context, tx bun.Tx, table, col string) (bool, error) {
```
tsmethurst, line 71: Might be worth moving this to a util function somewhere, as I can imagine this is something we'll end up using more for other migrations.

Also, did you test it for both SQLite and Postgres?

internal/db/bundb/migrations/20211113114307_init/emoji.go
```diff
@@ -36,7 +36,6 @@ type Emoji struct {
 	ImageStaticContentType string    `validate:"required" bun:",nullzero,notnull"`                                                            // MIME content type of the static version of the emoji image.
 	ImageFileSize          int       `validate:"required,min=1" bun:",nullzero,notnull"`                                                      // Size of the emoji image file in bytes, for serving purposes.
 	ImageStaticFileSize    int       `validate:"required,min=1" bun:",nullzero,notnull"`                                                      // Size of the static version of the emoji image file in bytes, for serving purposes.
-	ImageUpdatedAt         time.Time `validate:"-" bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"`                         // When was the emoji image last updated?
```
tsmethurst, line 39: I think I'd rather avoid touching these init models if that's alright, they're supposed to represent the state of the models at that point in time. I know it means creating + dropping the column pointlessly for new tables, but it's also nice to keep a record around in the codebase of what they looked like at init.

I hope at some point we can mark some kind of migration milestone where we can drop all these old models and just use the new ones (if you know what I mean), but in the meantime I think they should stay as-is.

internal/media/processingmedia.go
```diff
@@ -65,124 +65,102 @@ func (p *ProcessingMedia) AttachmentID() string {
...
-		// Gather errors as we proceed.
-		var errs = gtserror.NewMultiError(4)
+		// TODO: in time update this
+		// to perhaps follow a similar
+		// freshness window to statuses
+		// / accounts? But that's a big
+		// maybe, media don't change in
+		// the same way so this is largely
+		// just to slow down fail retries.
+		const maxfreq = 6 * time.Hour
```
tsmethurst, line 135: Indeed, most remotes (afaik) use cache-busting by naming different versions of a media file differently, so they can use long + reasonable cache http headers.

tsmethurst:
These look like good and sensible changes, and I don't spot any immediate issues. Just a few comments :)

internal/gtsmodel/mediaattachment.go
NyaaaWhatsUpDoc: basically allows us to input empty values for these fields without that ending up being null in the database, as with the initial create + cache stages so separated out we have to initially insert media attachments with some empty values, and it's during the actual load stage that they get populated with values :)

internal/db/bundb/migrations/20240613091853_drop_unused_media_columns.go
NyaaaWhatsUpDoc: this function is actually useable across the full range of migration files (as they're all part of the same package), though maybe worth moving to a separate `utils.go` to make it clear that it is usable in any of them.

i'm fairly certain i did at the time, but now i'm second guessing myself so gonna give it another go :')

internal/media/processingmedia.go
```diff
@@ -295,15 +262,11 @@ func (p *ProcessingMedia) store(ctx context.Context) error {
...
 	)
 
-	// Prefer discovered mime type, fall back to
-	// generic "this contains some bytes" type.
-	mime := info.MIME.Value
-	if mime == "" {
-		mime = "application/octet-stream"
-	}
+	// Prefer discovered MIME, fallback to generic data stream.
+	mime := cmp.Or(info.MIME.Value, "application/octet-stream")
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: yes it is a useful one!

internal/db/bundb/migrations/20240613091853_drop_unused_media_columns.go
NyaaaWhatsUpDoc: confirmed working :)

internal/db/bundb/migrations/20211113114307_init/emoji.go
NyaaaWhatsUpDoc: fixed that here by updating the test that was failling (basically where we re-create the emojis table) to use a snapshot of the emojis model from at that commit: 02a7dd2df9b21f6d3a453ed162089c7c8711b49d

internal/gtsmodel/mediaattachment.go
tsmethurst: Ah fair enough. I suppose for this particular thing storing null or an empty string doesn't make much difference.

tsmethurst:
Looks good now I think :) If you're happy with it then squerge I reckon.

tsmethurst:
She's squerged it lads!

PR #3001, opened 2024-06-12
[chore] Don't render sign-up form when registrations closed

tsmethurst:
Hide the sign-up form when instance is not open to new registrations, to prevent people thinking they can still submit.


NyaaaWhatsUpDoc:
looks good! squerge when ready

PR #3007, opened 2024-06-14
[bugfix] avoid v. long notification clear query

tsmethurst:
# Description

>

This pull request fixes an issue where, when clearing notifications for an account with a lot of notifications, GtS would send an absolutely enormous query to the database driver with a list of every ID to be deleted. This was preventing notifications from being cleared successfully. Instead, it now does a simpler delete that returns IDs, and uses that to do cache invalidation. Also avoids doing a bunch of unnecessary SELECTs of notifications, since we have no invalidate callbacks for notifications (I think, unless I'm missing something!).

## Checklist

### DIFF
--- a/internal/db/bundb/notification.go
+++ b/internal/db/bundb/notification.go
@@ -22,7 +22,6 @@ import (
 	"errors"
 	"slices"
 
-	"github.com/superseriousbusiness/gotosocial/internal/db"
 	"github.com/superseriousbusiness/gotosocial/internal/gtscontext"
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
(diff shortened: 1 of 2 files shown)


internal/db/bundb/notification.go
```diff
@@ -108,6 +107,11 @@ func (n *notificationDB) GetNotificationsByIDs(ctx context.Context, ids []string
 	notifs, err := n.state.Caches.GTS.Notification.LoadIDs("ID",
 		ids,
 		func(uncached []string) ([]*gtsmodel.Notification, error) {
+			// Skip query if everything was cached.
+			if len(uncached) == 0 {
```
tsmethurst, line 111: This avoids running a query that's like `SELECT blah blah blah FROM notifications WHERE id in (null)`.

internal/db/bundb/notification.go
```diff
@@ -328,61 +321,33 @@ func (n *notificationDB) DeleteNotifications(ctx context.Context, types []string
...
-
-	// Invalidate all cached notifications by IDs on return.
-	defer n.state.Caches.GTS.Notification.InvalidateIDs("ID", notifIDs)
+	var notifIDs []string
+	q = q.Returning("?", bun.Ident("id"))
 
-	// Load all notif into cache, this *really* isn't great
-	// but it is the only way we can ensure we invalidate all
-	// related caches correctly (e.g. visibility).
-	for _, id := range notifIDs {
```
tsmethurst, line 341: afaik we don't have any cache invalidation triggers for notifs, so we shouldn't need to load them all like this

NyaaaWhatsUpDoc:
nice work 😎

PR #3013, opened 2024-06-16
[feature] Conversations API

### LINKED ISSUE #1312: [feature] Implement `/api/v1/conversations` masto API endpoints for viewing direct message (dm) conversations
We can handle direct messages sending/receiving just fine, but currently there's no easy way to view all dm conversations in progress. Mastodon uses `/api/v1/conversations` (and related endpoints) for this, so we should do the same so that clients (tusky, pinafore) can use the GtS API to view conversations.

Relevant links:

- https://docs.joinmastodon.org/methods/conversations/
- https://docs.joinmastodon.org/entities/Conversation/

(draft of) Required steps to make this work:

- [ ] Add `Conversation` gts model + frontend model
- [ ] Add 'conversations' package + logic in https://github.com/superseriousbusiness/gotosocial/tree/main/internal/api/client
- [ ] Add Conversation cache in `internal/cache/gts.go`
- [ ] Add Conversation db model + migration
- [ ] Add logic for storing / updating conversations when a new message in the conversation thread is received.
- [ ] Add logic for marking conversations as 'read'

VyrCossont:
# Description

Implements a Mastodon-compatible DM conversations API.

Closes #1312

## Checklist

## TODO

- [x] Deleting conversations when the owning account is deleted
- [x] Updating conversations when the last status is deleted
- [x] Duplication of code related to filtering & muting (*best handled in a separate PR since it'd involve extending the visibility cache*)
- [x] Testing conversation updates
- [x] Interaction with updated statuses (should editing a status in the middle of a conversation mark it as unread?) (*no, other implementations don't do this*)
- [x] Test that it generates the correct notification and expected notification payload for a DM
- [x] Test that it doesn't do those things for non-DMs


VyrCossont:
There are some experimental bits in this one:
- Advanced migrations: a table that stores a pre-assigned ID, finished flag, and state JSON blob, and a point in server startup that runs them after regular migrations and most of the app setup. (I note this because it's *not* triggered by CLI tools.) Right now we have one advanced migration, which stores the progress of creating conversations for DMs that predate this feature, in case the task is interrupted before it finishes. That process requires substantial amounts of app logic so we can't run it as part of `bundb` migrations. We might want to run advanced migrations in a background goroutine?
- `db/test` contains a reusable factory struct mixin to create conversations and statuses for them, with ULIDs and timestamps based on a fake clock that we control. I'd like to expand this to reduce test copypasta in the future: our fixtures are fragile, and we should generate more things as needed rather than having a shared set of test objects.

cmd/gotosocial/action/server/server.go
```diff
@@ -290,6 +290,11 @@ var Start action.GTSAction = func(ctx context.Context) error {
 		return fmt.Errorf("error initializing metrics: %w", err)
 	}
 
+	// Run advanced migrations.
```
tsmethurst, line 293: Does this need to be added to the test rig too?

docs/api/swagger.yaml
```diff
@@ -6182,6 +6214,36 @@ paths:
...
+                "404":
+                    description: not found
+                "406":
+                    description: not acceptable
+                "500":
+                    description: internal server error
+            security:
+                - OAuth2 Bearer:
+                    - write:conversations
+            summary: Delete a single conversation with the given ID.
```
tsmethurst, line 6244: Does this also delete all DMs in the conversation? Might be worth clarifying this here just to avoid any mistaken assumptions.

internal/cache/size.go
```diff
@@ -320,6 +321,20 @@ func sizeofClient() uintptr {
...
+	return uintptr(size.Of(&gtsmodel.Conversation{
+		ID:               exampleID,
+		CreatedAt:        exampleTime,
+		UpdatedAt:        exampleTime,
+		AccountID:        exampleID,
+		OtherAccountIDs:  []string{exampleID, exampleID, exampleID},
+		OtherAccountsKey: strings.Join([]string{exampleID, exampleID, exampleID}, ","),
+		ThreadID:         exampleID,
+		LastStatusID:     exampleID,
+		Read:             func() *bool { ok := true; return &ok }(),
```
tsmethurst, line 334: You can just use util.Ptr here

internal/cache/wrappers.go
```diff
@@ -158,6 +158,34 @@ func (c *StructCache[T]) LoadIDs(index string, ids []string, load func([]string)
 	})
 }
 
+// LoadIDs2Part works as LoadIDs, except using a two-part key,
+// where the first part is an ID shared by all the objects,
+// and the second part is a list of per-object IDs.
+func (c *StructCache[T]) LoadIDs2Part(index string, id1 string, id2s []string, load func(string, []string) ([]T, error)) ([]T, error) {
```
tsmethurst, line 164: I'll defer to @NyaaaWhatsUpDoc here since they know the structcache API a lot better than me (unsurprisingly :'))

internal/config/defaults.go
```diff
@@ -156,52 +156,54 @@ var Defaults = Configuration{
...
+		AccountMemRatio:                   5,
+		AccountNoteMemRatio:               1,
+		AccountSettingsMemRatio:           0.1,
+		AccountStatsMemRatio:              2,
+		ApplicationMemRatio:               0.1,
+		BlockMemRatio:                     2,
+		BlockIDsMemRatio:                  3,
+		BoostOfIDsMemRatio:                3,
+		ClientMemRatio:                    0.1,
+		ConversationMemRatio:              2,
```
tsmethurst, line 168: This seems quite high considering how few conversations there probably are, I'd tune it down to 1 and 2 or something probably, similar to bookmarks.

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,452 @@
...
+}
+
+func (c *conversationDB) getAccountConversationLastStatusIDs(ctx context.Context, accountID string, page *paging.Page) ([]string, error) {
+	return loadPagedIDs(&c.state.Caches.GTS.ConversationLastStatusIDs, accountID, page, func() ([]string, error) {
+		var conversationLastStatusIDs []string
+
+		// Conversation last status IDs not in cache. Perform DB query.
+		if _, err := c.db.
+			NewSelect().
+			TableExpr("?", bun.Ident("conversations")).
```
tsmethurst, line 166: If you're not doing anything with TableExpr and ColumnExpr you can just do `Table("conversations")` and `Column("last_status_id")` here.

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,452 @@
...
+		// In case there were no cached conversations,
+		// explicitly invalidate the user's conversation last status ID cache.
+		c.state.Caches.GTS.ConversationLastStatusIDs.Invalidate(accountID)
+	}()
+
+	return c.db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+		// Delete conversations matching the account ID.
+		deletedConversationIDs := []string{}
+		if err := tx.NewDelete().
+			Model((*gtsmodel.Conversation)(nil)).
```
tsmethurst, line 301: Instead of specifying `Model` here you can just use `Table("conversations")` I believe.

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,452 @@
...
+			Where("? = ?", bun.Ident("account_id"), accountID).
+			Returning("?", bun.Ident("id")).
+			Scan(ctx, &deletedConversationIDs); // nocollapse
+		err != nil {
+			return gtserror.Newf("error deleting conversations for account %s: %w", accountID, err)
+		}
+
+		// Delete any conversation-to-status links matching the deleted conversation IDs.
+		if _, err := tx.NewDelete().
+			Model((*gtsmodel.ConversationToStatus)(nil)).
```
tsmethurst, line 311: Same here re: Model, you can just give the table name.

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,452 @@
...
+		log.Panicf(nil, "db conn %s was neither pg nor sqlite", c.db)
+	}
+
+	updatedConversationIDs := []string{}
+	deletedConversationIDs := []string{}
+
+	if err := c.db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+		// Delete this status from conversation-to-status links.
+		if _, err := tx.NewDelete().
+			Model((*gtsmodel.ConversationToStatus)(nil)).
```
tsmethurst, line 340: Same here re: model.

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,452 @@
...
+			RETURNING id
+			`,
+			latestConversationStatusesTempTable,
+		).Scan(ctx, &deletedConversationIDs); // nocollapse
+		err != nil {
+			return gtserror.Newf("error deleting conversation while deleting status %s: %w", statusID, err)
+		}
+
+		// Clean up.
+		if _, err := tx.NewRaw(`DROP TABLE ?`, conversationStatusesTempTable).Exec(ctx); err != nil {
```
tsmethurst, line 436: I think you can do this without having to use raw as well.

If you're just sick of writing bun stuff I can help you with it if you like :P I sense some exhaustion

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,452 @@
...
+			Exec(ctx); // nocollapse
+		err != nil {
+			return gtserror.Newf("error deleting conversation-to-status links while deleting status %s: %w", statusID, err)
+		}
+
+		// Create a temporary table with all statuses other than the deleted status
+		// in each conversation for which the deleted status is the last status
+		// (if there are such statuses).
+		conversationStatusesTempTable := bun.Ident("conversation_statuses_" + id.NewULID())
+		if _, err := tx.NewRaw(
```
tsmethurst, line 351: I'm really not very keen on these raw transactions tbh. Is there no way with bun to do this jiggery pokery? If there is, I'd much rather we use that, since it gives us all the dialect-specific escaping, sanitization, etc.

internal/db/bundb/migrations/20240611190733_add_conversations.go
```diff
@@ -0,0 +1,90 @@
...
+					Index(index).
+					Column(columns...).
+					IfNotExists().
+					Exec(ctx); err != nil {
+					return err
+				}
+			}
+
+			// Add an additional uniqueness constraint to the conversations table.
+			if _, err := tx.
```
tsmethurst, line 66: You can do this with struct tags on the model itself, no need to do it here.

internal/db/bundb/status.go
```diff
@@ -682,3 +682,35 @@ func (s *statusDB) getStatusBoostIDs(ctx context.Context, statusID string) ([]st
 		return statusIDs, nil
 	})
 }
+
+func (s *statusDB) MaxDirectStatusID(ctx context.Context) (string, error) {
+	maxID := ""
+	if err := s.db.
+		NewSelect().
+		Model((*gtsmodel.Status)(nil)).
```
tsmethurst, line 690: Same comment re: model.

internal/db/test/conversation.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/util"
+)
+
+type testSuite interface {
+	FailNow(string, ...interface{}) bool
+}
+
+// ConversationFactory can be embedded or included by test suites that want to generate statuses and conversations.
+type ConversationFactory struct {
```
tsmethurst, line 37: Oh my god, we're writing java now :') (I'm joking, it's fine, just giving me flashbacks...)

internal/processing/conversations/get.go
```diff
@@ -0,0 +1,104 @@
...
+			compiledMutes,
+		)
+		if err != nil {
+			err = gtserror.Newf(
+				"couldn't convert conversation %s to API representation for account %s: %w",
+				conversation.ID,
+				requestingAccount.ID,
+				err,
+			)
+			return nil, gtserror.NewErrorInternalError(err)
```
tsmethurst, line 91: Is there a chance this could fail if an entry or account is missing? Just wondering if we might want to just log + skip instead of returning error and no conversations. Open question, not a sneaky "please change this" comment.

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,452 @@
...
+			log.Errorf(ctx, "error populating conversation %s: %v", conversation.ID, err)
+			return true
+		}
+		return false
+	})
+
+	return conversations, nil
+}
+
+func (c *conversationDB) PutConversation(ctx context.Context, conversation *gtsmodel.Conversation, columns ...string) error {
```
tsmethurst, line 230: This is a bit different from other DB functions in that the Put does an upsert rather than just an insert, whereas other db files generally have separate Put and Update functions.

I'm not saying this is a bad thing at all, since we probably ought to be doing more upserts elsewhere to avoid getting lots of duplicate key errors logged in Postgres (for example).

However if we're going to change the way we do Put vs Update, I'd rather we save it for a separate chore PR rather than gradually introducing a mix of the two (which might be confusing for people working on the code, since they won't know if a Put is an upsert or an insert without going to look at the function).

How would you feel about, for now, either renaming this to `UpsertConversation`, which makes v. clear to callers what it does, or making separate Put and Update functions?

internal/db/bundb/status.go
```diff
@@ -682,3 +682,35 @@ func (s *statusDB) getStatusBoostIDs(ctx context.Context, statusID string) ([]st
...
+
+func (s *statusDB) GetDirectStatusIDsBatch(ctx context.Context, minID string, maxID string, count int) ([]string, error) {
+	var statusIDs []string
+	if err := s.db.
+		NewSelect().
+		Model((*gtsmodel.Status)(nil)).
+		Column("id").
+		Where("? = ?", bun.Ident("visibility"), gtsmodel.VisibilityDirect).
+		Where("? > ?", bun.Ident("id"), minID).
+		Where("? <= ?", bun.Ident("id"), maxID).
```
tsmethurst, line 708: This differs from other paging-style functions where we don't include the maxID entry, in that this *does* include the maxID entry. Is there a specific reason it needs to be this way? If so, probably worth commenting it here (and/or elsewhere) to point out how + why it differs.

internal/processing/conversations/read.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/util"
+)
+
+func (p *Processor) Read(
+	ctx context.Context,
+	requestingAccount *gtsmodel.Account,
+	id string,
+) (*apimodel.Conversation, gtserror.WithCode) {
+	// Get the conversation, including participating accounts and last status.
+	conversation, err := p.state.DB.GetConversationByID(ctx, id)
```
tsmethurst, line 39: imo this looks a bit cleaner if you just do:

```go
		err = gtserror.Newf("db error getting conversation %s: %w", id, err)
	}

	if conversation == nil {
		return nil, gtserror.NewErrorNotFound(err)
	}
```

internal/processing/conversations/read.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	mutes, err := p.state.DB.GetAccountMutes(gtscontext.SetBarebones(ctx), requestingAccount.ID, nil)
+	if err != nil {
+		err = gtserror.Newf("couldn't retrieve mutes for account %s: %w", requestingAccount.ID, err)
+		return nil, gtserror.NewErrorInternalError(err)
+	}
+	compiledMutes := usermute.NewCompiledUserMuteList(mutes)
+
+	apiConversation, err := p.converter.ConversationToAPIConversation(
```
tsmethurst, line 71: Not something we need to do right now in this PR, but since there's lots of places in the codebase now where a converter function needs filters and mutes, it might be worth shifting the logic of fetching them back to the converter itself rather than passing them in every time. This'll make the function signatures a bit tidier. Just a little reminder for a chore PR later to refactor this.

tsmethurst:
Hell yeah this is looking tasty, great work. I've got comments mostly on some of the naming and the way that bun is (or isn't) used :) But in terms of logic and structure this is looking really good.

internal/db/bundb/conversation.go
VyrCossont: The advantage of this style is that it makes it easier to find everywhere that model is used, and also harder to typo the table name.

internal/db/bundb/conversation.go
VyrCossont: That's spot on, I'll rename it to `UpsertConversation`.

internal/db/bundb/conversation.go
VyrCossont: Can do. This iteration ended up not using anything that we definitely can't do in Bun.

internal/db/bundb/migrations/20240611190733_add_conversations.go
VyrCossont: IIRC Bun can't handle multiple `unique:` options on the same field. Let me try it one more time and make sure I didn't just typo the option name.

cmd/gotosocial/action/server/server.go
VyrCossont: Yes it does.

docs/api/swagger.yaml
VyrCossont: It only deletes the conversation status itself, and it doesn't prevent a new conversation from being created later with the same thread and participants. I'll document that.

internal/cache/size.go
VyrCossont: Will do. I'll leave the existing uses of this construct in that file in place for now, and we can clean it up globally later.

internal/db/test/conversation.go
VyrCossont: Yeahhhhhh I had one of my own writing this.

internal/processing/conversations/read.go
VyrCossont: Sure, and there's an equivalent in `delete.go` that I'll update too.

internal/config/defaults.go
VyrCossont: Done.

internal/processing/conversations/read.go
VyrCossont: I think this would be best done by refactoring the visibility cache to incorporate mutes and filters: https://gist.github.com/VyrCossont/239272d87b6dcd3b611347835267d3c0

internal/db/bundb/conversation.go
VyrCossont: Sure thing.

internal/db/bundb/migrations/20240611190733_add_conversations.go
VyrCossont: Must have been a mistake on my part. Tried it again and confirmed that Bun does support multiple `unique:` options on the same field.

internal/db/bundb/status.go
VyrCossont: Done. I've renamed the param to `maxIDInclusive` and added further comments.

internal/processing/conversations/get.go
VyrCossont: I don't think that can happen when converting the status, because the attached accounts and statuses would have already been loaded before we start rendering the response. However, on the off chance that changes, I don't see any problem with logging and skipping; we've definitely got other endpoints which do that, such as `internal/processing/account.BlocksGet`.

internal/db/bundb/conversation.go
VyrCossont: Done.

internal/db/bundb/conversation.go
VyrCossont: Not quite true, turns out Bun's `NewCreateTable` can't do `CREATE TABLE … AS SELECT …` because it assumes that there's always a parenthesized column list, so we still have to use raw queries to create these tables, or do even more work to rewrite them as the more portable `CREATE TABLE` followed by [`INSERT … SELECT`](https://bun.uptrace.dev/guide/query-insert.html#insert-select). But we can use a very short raw query parameterized with a Bun subquery for the `SELECT` part, and I think that's a decent tradeoff.

I also think we're not getting anything in terms of escaping or type checking by using Bun's query builders that pure raw queries can't also do: raw queries can still use placeholders and escaping, and Bun's query builders *can't* tell me if I spelled a column name wrong at compile time, so the only benefit is that a subset of SQL keywords get checked at compile time, and the cost is that it's a lot less readable, IMO, especially with our frequently-unnecessary house style use of `bun.Ident`.

Take a look, and let me know if I'm missing any tricks.

VyrCossont: If we end up using this pattern again, we can add a small `CreateTableAsQuery`/`NewCreateTableAs` helper along the lines of `UpsertQuery`/`NewUpsert`, but it doesn't seem worth it for two usages in the same part of the same file.

cmd/gotosocial/action/server/server.go
```diff
@@ -290,6 +290,11 @@ var Start action.GTSAction = func(ctx context.Context) error {
...
+	if err := processor.Conversations().MigrateDMsToConversations(ctx); err != nil {
```
NyaaaWhatsUpDoc, line 294: i feel like this might make a bit more sense semantically if we add a `migrations` sub-processor and move this to there. otherwise we'll end up with a long list of `processor.SubProcessor().DoMigration()` we need to maintain in the server / testrig / tests. and whether it's worth having a `server_migrations` or `advanced_migrations` table to mark when we've done these :thinking:

VyrCossont: How about building that list into `processor` itself? Future advanced migrations might need access to multiple subprocessors, and it would have everything needed to construct them.

This patch does include an `advanced_migrations` table for tracking whether advanced migrations have been applied already.

VyrCossont: Nah, you're right, cleaner to have a separate processor. Adding one.

internal/db/bundb/status.go
NyaaaWhatsUpDoc, line 690: does this definitely not panic in practice? while i did used to use nil models in the past there can be situations where it panics expecting there to be a model set. in these cases i just tend to use `Table("statuses")`. but also if this definitely works then don't worry about it

(same goes for below too)

internal/processing/conversations/read.go
NyaaaWhatsUpDoc: i might take a look at this at some point :thinking:

internal/processing/conversations/delete.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (p *Processor) Delete(
+	ctx context.Context,
+	requestingAccount *gtsmodel.Account,
+	id string,
+) gtserror.WithCode {
+	// Get the conversation so that we can check its owning account ID.
+	conversation, err := p.state.DB.GetConversationByID(gtscontext.SetBarebones(ctx), id)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return gtserror.NewErrorInternalError(err)
```
NyaaaWhatsUpDoc, line 38: i would probably update the error returns here to include some more context, e.g. 

```golang
    err := gtserror.Newf("error getting conversation from db: %w", err)
    return gtserror.NewErrorInternalError(err)
}

if conversation == nil {
    const text = "conversation not found"
    return gtserror.NewErrorNotFound(errors.New(text), text)
}

if conversation.AccountID != requestingAccount.ID {
    const text = "conversation not found"
    err := errors.New("conversation not owned by requester")
    return gtserror.NewErrorNotFound(err, text)
}

// ... etc
```

noting some useful things (if you haven't already picked up on them):
- `gtserror.Newf()` follows the same argument / return semantics of `fmt.Errorf()`, except that it prefixes the error string with the current calling function. useful for tracking error propagation :), though also worth noting we usually don't use that if generating an error message to be returned at the frontend just for neater API responses
- `gtserror.NewError ${HttpErrorType} (err, text)` obviously wraps it with that particular http error type's status code, but unless you provide a 2nd argument it will just provide the standard status text for that code as the API response. so depending on what / where / how the error is wrapped we'll sometimes provide separate help text, or just use the error's text itself as the API response text.

internal/processing/conversations/get.go
```diff
@@ -0,0 +1,105 @@
...
+	requestingAccount *gtsmodel.Account,
+	page *paging.Page,
+) (*apimodel.PageableResponse, gtserror.WithCode) {
+	conversations, err := p.state.DB.GetConversationsByOwnerAccountID(
+		ctx,
+		requestingAccount.ID,
+		page,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return nil, gtserror.NewErrorInternalError(err)
```
NyaaaWhatsUpDoc, line 48: same as above re: errors

internal/processing/conversations/migrate.go
```diff
@@ -0,0 +1,135 @@
...
+
+type AdvancedMigrationState struct {
+	MinID          string
+	MaxIDInclusive string
+}
+
+func (p *Processor) MigrateDMsToConversations(ctx context.Context) error {
+	advancedMigration, err := p.state.DB.GetAdvancedMigration(ctx, advancedMigrationID)
+	if err != nil {
+		if !errors.Is(err, db.ErrNoEntries) {
```
NyaaaWhatsUpDoc, line 43: might be a bit easier to read here if you just collapse this into a single if clause i.e. `if err != nil && !errors.Is(err, db.ErrNoEntries)`. the logic shouldn't be any different

internal/processing/conversations/read.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (p *Processor) Read(
+	ctx context.Context,
+	requestingAccount *gtsmodel.Account,
+	id string,
+) (*apimodel.Conversation, gtserror.WithCode) {
+	// Get the conversation, including participating accounts and last status.
+	conversation, err := p.state.DB.GetConversationByID(ctx, id)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return nil, gtserror.NewErrorInternalError(err)
```
NyaaaWhatsUpDoc, line 41: same as above re: errors in this function

internal/processing/conversations/update.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+// UpdateConversationsForStatus updates all conversations related to a status,
+// and returns a map from local account IDs to conversation notifications that should be sent to them.
+func (p *Processor) UpdateConversationsForStatus(ctx context.Context, status *gtsmodel.Status) (map[string]*apimodel.Conversation, error) {
+	notifications := map[string]*apimodel.Conversation{}
+
+	// We need accounts to be populated for this.
+	if err := p.state.DB.PopulateStatus(ctx, status); err != nil {
+		return nil, err
```
NyaaaWhatsUpDoc, line 42: some added error context here would probably be useful

internal/processing/conversations/update.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if status.Visibility != gtsmodel.VisibilityDirect {
```
NyaaaWhatsUpDoc, line 45: in the negative cases at least, it's probably better to perform these checks *before* performing `PopulateStatus()` as otherwise we might end up doing database calls to populate something we never use :p

internal/typeutils/internaltofrontend.go
```diff
@@ -1679,6 +1679,54 @@ func (c *Converter) NotificationToAPINotification(
...
+) (*apimodel.Conversation, error) {
+	apiConversation := &apimodel.Conversation{
+		ID:     conversation.ID,
+		Unread: !*conversation.Read,
+	}
+	for _, account := range conversation.OtherAccounts {
+		var apiAccount *apimodel.Account
+		blocked, err := c.state.DB.IsEitherBlocked(ctx, requestingAccount.ID, account.ID)
+		if err != nil {
+			return nil, err
```
NyaaaWhatsUpDoc, line 1700: same here re: adding context error to errors

internal/gtsmodel/advancedmigration.go
```diff
@@ -0,0 +1,48 @@
...
+	ID        string    `bun:",pk,nullzero,notnull,unique"`                                 // id of this migration (preassigned, not a ULID)
+	CreatedAt time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"` // when was item created
+	UpdatedAt time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"` // when was item last updated
+	StateJSON string    `bun:",nullzero"`                                                   // JSON dump of the migration state
+	Finished  *bool     `bun:",nullzero,notnull,default:false"`                             // has this migration finished?
+}
+
+func AdvancedMigrationLoad[State any](a *AdvancedMigration) (State, error) {
+	var state State
+	err := json.Unmarshal([]byte(a.StateJSON), &state)
```
NyaaaWhatsUpDoc, line 37: is it worth having these functions given they're just a thin wrapper around `json.Marshal()` and `json.Unmarshal()`? imo it might be a bit easier to understand if we just used `json.Marshal()` and `json.Unmarshal()` in the advanced migrations function. the database column is already called `StateJSON` so we know it's always gonna be JSON :p

also given that JSON likes to work in terms of byte slices, `StateJSON` might be better off as a `[]byte` type column

internal/cache/wrappers.go
NyaaaWhatsUpDoc: function signature wise, this is great! it's just the name i'm sticking a bit on, but also i can't really think of a better name :p. maybe we leave the name for now and at some point in the future i'll get hit with an inspiration for a better name

internal/db/bundb/conversation.go
```diff
@@ -0,0 +1,489 @@
...
+}
+
+func (c *conversationDB) getAccountConversationLastStatusIDs(ctx context.Context, accountID string, page *paging.Page) ([]string, error) {
+	return loadPagedIDs(&c.state.Caches.GTS.ConversationLastStatusIDs, accountID, page, func() ([]string, error) {
+		var conversationLastStatusIDs []string
+
+		// Conversation last status IDs not in cache. Perform DB query.
+		if _, err := c.db.
+			NewSelect().
+			Model((*gtsmodel.Conversation)(nil)).
```
NyaaaWhatsUpDoc, line 166: same here re: nil model (if this doesn't panic feel free to just resolve this)

internal/db/bundb/status.go
VyrCossont: This is [documented Bun functionality for selects that don't return model structs](https://bun.uptrace.dev/guide/golang-orm.html#scanning-query-results). It definitely works.

internal/processing/conversations/update.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/id"
+	"github.com/superseriousbusiness/gotosocial/internal/log"
+	"github.com/superseriousbusiness/gotosocial/internal/util"
+)
+
+// UpdateConversationsForStatus updates all conversations related to a status,
+// and returns a map from local account IDs to conversation notifications that should be sent to them.
+func (p *Processor) UpdateConversationsForStatus(ctx context.Context, status *gtsmodel.Status) (map[string]*apimodel.Conversation, error) {
+	notifications := map[string]*apimodel.Conversation{}
```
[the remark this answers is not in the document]
VyrCossont: The output of this function should be de-duped already through `allParticipantsSet`, so we don't need to use a map type here. I'll replace it with a list of structs.

internal/processing/conversations/delete.go
VyrCossont: Extracted this to a common processor function and added error context.

NyaaaWhatsUpDoc:
fantastic work on this PR :ok_hand:

PR #3032, opened 2024-06-23
[docs] traefik redirect regex documentation

coxde:
# Description

This pull request fixes the Traefik redirect regex issue in the [split domain documentation](https://docs.gotosocial.org/en/latest/advanced/host-account-domain/). The current regex won't keep the query string, e.g. `?resource=acct:example@example.org`, which would cause federation issues with other instances.

I've met the federation issues on my own instance. After checking Traefik logs (e.g. `GET /.well-known/webfinger?resource=acct:example@example.org 502`), I found out it's due to the redirection. And changed the regex which fixed the problem for me. Maybe some further checking of the regex would be better (I am really bad at regex :(

## Checklist

### DIFF
--- a/docs/advanced/host-account-domain.md
+++ b/docs/advanced/host-account-domain.md
@@ -98,8 +98,8 @@ myservice:
   labels:
     - 'traefik.http.routers.myservice.rule=Host(`example.org`)'                                                                # account-domain
     - 'traefik.http.middlewares.myservice-gts.redirectregex.permanent=true'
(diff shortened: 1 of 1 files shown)


daenney:
That's definitely a bug, that query string needs to be kept. Thanks for catching and fixing this!

daenney (approved):

tsmethurst:
Whoops! Thank you for the fix! <3

PR #3046, opened 2024-06-26
[feature] Object store custom URL (S3)

### LINKED ISSUE #2574: [feature] Object store custom URL (S3)

For the purpose of being able to use a separate caching proxy, CDN acceleration and other things together with object storage for media and emoji, GoToSocial needs to construct a URL with a different base than the bucket URL/endpoint that it can return to clients.

Imagine for example GoToSocial using an S3 bucket, but the owner wanting the data itself to be accessed through a Cloudfront distribution, a Fastly CDN, or even just their own caching proxy. This can also facilitate migration between different object store backends, as the URL constructed and given to clients by GoToSocial in API responses would not have to change.

We should add a configuration option to GoToSocial that lets people define what the "external" URL should be for their bucket and use that when constructing media references.

It is paramount that this is implemented in such a way that the bucket name is not automatically interpolated into the constructed URL, as that's often not valid when using a CDN distribution. The endpoint should be taken as is, with only the path to the media object within the bucket appended to it.

NONE

### Additional context.

We have a whole host of issues on the issue tracker that are variations of this one. I'm going to link those into here and close them so we have a single tracking issue for this feature request.

* #1026
* #1632
* #2155
* #2549

CDN18:
# Description

This PR tries to implement basic support for Object Store custom URLs.

1. Added a config field named `storage-s3-custom-url` in `internal/config`, along with a default value (empty string), helper methods, and validation logic. The current validation logic requires storage-s3-custom-url to be a string that starts with `http(s)://` and end without a `/`. I noticed this format might be inconsistent with the format of storage-s3-endpoint, and if necessary, I will adjust it in future commits.
2. Added additional handling in storage for cases where storage-s3-custom-url is set. When storage-s3-custom-url is set, it will return a link in the format of `<customURL> + / + <key>`. The customURL will also be used as the CSPURI.

- If users wish to return a presigned URL while using a custom URL, they should modify `storage-s3-endpoint` instead of `storage-s3-custom-url`.
- If users want to use a custom URL with a path (e.g., retaining the bucket name when using a custom URL), they can append the desired base path directly to `storage-s3-custom-url`. ~~In this case, the logic for returning the CSP URI needs to be slightly adjusted to return only the Host and Scheme of the custom URL when `storage-s3-custom-url` is set.~~
- `storage-s3-proxy` takes precedence over `storage-s3-custom-url`

If this PR is acceptable, I will further supplement the relevant documentation and any additional necessary content.

closes #2574

*Edit: Apologies, I forgot to commit the configuration changes earlier. ~~The test result was cached thus displayed as falling.~~

## Checklist


internal/storage/storage.go
```diff
@@ -133,12 +134,22 @@ func (d *Driver) URL(ctx context.Context, key string) *PresignedURL {
...
-	u, err := s3.Client().PresignedGetObject(ctx, d.Bucket, key, urlCacheTTL, url.Values{
-		"response-content-type": []string{mime.TypeByExtension(path.Ext(key))},
-	})
-	if err != nil {
-		// If URL request fails, fallback is to fetch the file. So ignore the error here
-		return nil
+	var u *url.URL
+	var err error
+	if d.CustomURL != "" {
+		u, err = url.Parse(d.CustomURL + "/" + key)
```
daenney, line 140: I don't think the extra `/` here is necessary. We've checked the custom URL ends with a `/` in the config validation.

internal/storage/storage.go
```diff
@@ -133,12 +134,22 @@ func (d *Driver) URL(ctx context.Context, key string) *PresignedURL {
...
+		if err != nil {
+			// If URL request fails, fallback is to fetch the file. So ignore the error here
```
daenney, line 142: You've not done a request here, only tried to parse a URL and see if it's valid. This comment needs adjusting.

daenney (commented):
I think this ought to work. You also need to update https://github.com/superseriousbusiness/gotosocial/blob/main/test/envparsing.sh, it's an extra test to make sure the value is picked up correctly from the environment.

internal/storage/storage.go
CDN18: Hmm, I think this `/` is necessary. Currently during config validation, an error is triggered if `storage-s3-custom-url` ends with a `/`. see [intrnal/config/validate.go#L133-L138](https://github.com/superseriousbusiness/gotosocial/pull/3046/files#diff-9abd7a0266df367fa6ee3ecb0c4b3d5046d678313073c1e152e305f267ec2825R133-R138)

CDN18:
Thanks for the review! I've made the remaining changes according to your suggestion.

- updated the envparsing script and corrected the comment. 
- made a minor adjustment in ProbeCSPUri to address https://github.com/superseriousbusiness/gotosocial/pull/3046#discussion_r1657056845 
- add the corresponding documentation

internal/storage/storage.go
daenney: Ah you're right, I read it the other way around.

test/envparsing.sh
```diff
@@ -168,6 +168,7 @@ EXPECT=$(cat << "EOF"
     "storage-s3-bucket": "gts",
     "storage-s3-endpoint": "localhost:9000",
     "storage-s3-proxy": true,
+    "storage-s3-custom-url": "",
```
daenney, line 171: These keys are alphabetically ordered, makes the test a bit easier. Just needs to move up a few lines, between `-bucket` and `-endpoint`. Tests should pass then.

internal/config/validate.go
```diff
@@ -118,6 +118,26 @@ func Validate() error {
 		errf("%s must be set", WebAssetBaseDirFlag())
 	}
 
+	// `storage-s3-custom-url`
+	switch s3CustomURL := GetStorageS3CustomURL(); s3CustomURL {
+	case "":
+		// No problem.
+
+	default:
+		if s3CustomURL[:8] != "https://" && s3CustomURL[:7] != "http://" {
```
tsmethurst, line 127: This panics if s3CustomURL is less than 7 characters. You can just use strings.HasPrefix() for this instead :)

tsmethurst:
Hiya! Sorry for taking so long to get to this. It looks good overall, thanks for doing the work :)

If it's alright with you I might still push some small commits to it just to refactor some wording a tiny bit, but otherwise I think this is alright and should do what it needs to do!

CDN18:
> Hiya! Sorry for taking so long to get to this. It looks good overall, thanks for doing the work :)
> 
> 
> 
> If it's alright with you I might still push some small commits to it just to refactor some wording a tiny bit, but otherwise I think this is alright and should do what it needs to do!

No worries, I don't mind at all. Thanks for reviewing it and feel free to push any changes you want :P

internal/config/validate.go
CDN18: Oh, hadn't considered that. Will make the change as soon as I'm back at my computer

NyaaaWhatsUpDoc:
thanks for the contribution @CDN18! nice work :). @tsmethurst once you've pushed the changes you were thinking of, @ me in the matrix and i'll do a final lookover over this!

tsmethurst:
Hey @CDN18, I tidied this up and renamed `storage-s3-custom-url` to `storage-s3-redirect-url` because I felt it was a bit more descriptive. This is now ready to go I think :) Thanks again for your work!

internal/config/validate.go
```diff
@@ -118,6 +120,28 @@ func Validate() error {
...
+				"%s invalid: %w",
+				StorageS3RedirectURLFlag(), err,
+			)
+		} else if url.Scheme != "https" && url.Scheme != "http" {
+			errf(
+				"%s scheme must be https or http",
+				StorageS3RedirectURLFlag(),
+			)
+		}
+	}
```
CDN18, line 143: The variable name here is still using 's3CustomURL', we'd better rename it to coincide with the name change

example/config.yaml
```diff
@@ -551,11 +551,42 @@ storage-local-base-path: "/gotosocial/storage"
...
+# "normal" s3 usage, and most admins can happily just ignore this setting.
+#
+# If set, then media fileserver requests to your instance will be redirected
+# to this URL instead of your bucket URL, preserving relevant path parts.
+#
+# This is useful if you are using a CDN proxy in front of your S3 bucket, and you
+# want to serve media from the CDN rather than serving from your S3 bucket directly.
+#
+# For example, if you have your storage-s3-endpoint value set to "s3.my-storage.example.org",
+# and you have a CDN set up to proxy your bucket, serving from "dn.some-fancy-host.org",
```
CDN18, line 578: There's a typo: the beginning of `"dn.some-fancy-host.org"` is missing a `c`

CDN18 (commented):
@tsmethurst Thanks for tidying up :) The docs and conf comments look much more detailed now!

example/config.yaml
tsmethurst: Thanks, fixed! https://github.com/superseriousbusiness/gotosocial/pull/3046/commits/95b99ba871d186ab40ebb01cbadac26cdc712b54

internal/config/validate.go
tsmethurst: Thanks, fixed! https://github.com/superseriousbusiness/gotosocial/pull/3046/commits/95b99ba871d186ab40ebb01cbadac26cdc712b54

PR #3056, opened 2024-06-29
[fix] use instance configured emoji maxsize on emoji update validation

f0x52:
Found by @Sentynel, fixes the TODO on the emoji details form to actually use the instance's configured max size for emoji, like what was already implemented for new emoji uploads.


tsmethurst:
Ah nice one, thanks!

PR #3071, opened 2024-07-04
[bugfix] Handle ErrHideStatus when preparing timeline statuses

VyrCossont:
# Description

The included tests weren't successful as regression tests for this bug, but describe behavior that works now and should continue to work in the future, so I'm leaving them in.

Reported by @tsmethurst. The symptom is the log message `db error while trying to prepare`, which occurs in three places, one of which already handles `ErrHideStatus`, and the other two of which I've added handling for in this patch.

## Checklist


daenney (approved):

tsmethurst:
Thanks for fixing :)

Issue #3057, opened 2024-06-30
[feature] Limit interactions (likes, replies, boosts) on a status with per-status settable "interaction policy"


tsmethurst:
We didn't have an issue for this yet, so I'm just creating one here to track this.

Currently working on a branch to introduce the concept of an interaction policy into GoToSocial. Users ought to be able to set a default interaction policy per visibility level of a status, and also when creating a new status they should be able to specify an interaction policy for just that status.

Interaction policy will let users decide who can reply / like / boost a status without needing approval, who can do with while requiring approval, and who can't do it at all. For example, you could set "canReply" without approval to followers, and set "canReply" with approval to everyone else, etc.

On a federation level, this will be sent out as an `interactionPolicy` property on a Note with this form:

```json
{
  "@context": [
    "https://gotosocial.org/ns",
    "https://www.w3.org/ns/activitystreams"
  ],
  "content": "hey @f0x and @dumpsterqueer",
  "contentMap": {
    "en": "hey @f0x and @dumpsterqueer",
    "fr": "bonjour @f0x et @dumpsterqueer"
  },
  "interactionPolicy": {
    "canAnnounce": {
      "always": [
        "https://gts.superseriousbusiness.org/users/dumpsterqueer",
        "https://gts.superseriousbusiness.org/users/f0x"
      ],
      "approvalRequired": [
        "https://www.w3.org/ns/activitystreams#Public"
      ]
    },
    "canLike": {
      "always": [
        "https://www.w3.org/ns/activitystreams#Public"
      ],
      "approvalRequired": []
    },
    "canReply": {
      "always": [
        "https://www.w3.org/ns/activitystreams#Public"
      ],
      "approvalRequired": []
    }
  },
  "tag": [
    {
      "href": "https://gts.superseriousbusiness.org/users/dumpsterqueer",
      "name": "@dumpsterqueer@superseriousbusiness.org",
      "type": "Mention"
    },
    {
      "href": "https://gts.superseriousbusiness.org/users/f0x",
      "name": "@f0x@superseriousbusiness.org",
      "type": "Mention"
    }
  ],
  "type": "Note"
}
```

[...]

PR #3075, opened 2024-07-06
[chore] Add interaction policy gtsmodels

tsmethurst:
# Description

>

This pull request adds `gtsmodel`s and database migrations to drop the hardly-used "replyable", "likeable", and "boostable" flags, in preparation for adding proper interaction policies in a subsequent PR (relates to https://github.com/superseriousbusiness/gotosocial/issues/3057).

I'd have done this all in one PR, but it was becoming monstrous, so this is a wee interim PR to avoid overwhelming reviewers (and myself).

## Checklist


daenney:
I would prefer if we remove the commented out blocks of code instead of leaving them with a todo for later. It seems like the body of those tests would need rewriting to for interaction policies. We can always retrieve the code from git if we want to restore it.

tsmethurst:
> I would prefer if we remove the commented out blocks of code instead of leaving them with a todo for later. It seems like the body of those tests would need rewriting too for interaction policies. We can always retrieve the code from git if we want to restore it.

I can do that if you prefer -- I've already got those tests fixed up and uncommented on the main interaction policies branch I'm working on, so commenting them out is really just an interim thing between this getting squerged, and the next PR being opened.

daenney:
Nah in that case it's fine :+1:

internal/db/bundb/interaction.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, err
+	}
+
+	return approval, nil
+}
+
+func (r *interactionDB) PopulateInteractionApproval(ctx context.Context, approval *gtsmodel.InteractionApproval) error {
+	var (
+		err  error
+		errs = gtserror.NewMultiError(4)
```
NyaaaWhatsUpDoc, line 106: there's not 4 possible errors to be appended here :P

internal/db/bundb/migrations/20240620074530_interaction_policy.go
```diff
@@ -0,0 +1,244 @@
...
+		log.Info(ctx, "migrating statuses and account settings to interaction policy model, please wait...")
+
+		// Add interaction_policy
+		// column to statuses table.
+		_, err := db.ExecContext(ctx,
+			"ALTER TABLE ? ADD COLUMN ? JSONB",
+			bun.Ident("statuses"),
+			bun.Ident("interaction_policy"),
+		)
+		if err != nil {
```
NyaaaWhatsUpDoc, line 43: for each of these parts instead of checking the error afterward (at which point a postgres tx will be totally failed anyways), you could just use the `doesColumnExist()` function which should be in the migrations namespace :)

internal/db/bundb/migrations/20240620074530_interaction_policy.go
```diff
@@ -0,0 +1,244 @@
...
+)
+
+func init() {
+	up := func(ctx context.Context, db *bun.DB) error {
+		log.Info(ctx, "migrating statuses and account settings to interaction policy model, please wait...")
+
+		// Add interaction_policy
+		// column to statuses table.
+		_, err := db.ExecContext(ctx,
+			"ALTER TABLE ? ADD COLUMN ? JSONB",
```
NyaaaWhatsUpDoc, line 39: do we use embedded JSON type for any other columns in the database? not something that needs changing here, just musing on whether we might want to update other places that just take blob data to use jsonb instead as we have quite a places that take json column values

tsmethurst: Not yet, for sqlite at least! I'm using the new stuff :sunglasses:

internal/db/bundb/migrations/20240620074530_interaction_policy.go
```diff
@@ -0,0 +1,244 @@
...
+			},
+			{
+				table:      "status_faves",
+				column:     "approved_by_uri",
+				columnType: "varchar",
+				defaultVal: "",
+			},
+		} {
+			_, err := db.ExecContext(ctx,
+				"ALTER TABLE ? ADD COLUMN ? ? ?",
```
NyaaaWhatsUpDoc, line 87: does `bun.Safe()` definitely serialize empty string to nothing for all dialects? i'm just wondering how it handles the `defaultVal` column value for the final two types and whether it ends up adding in empty quotes

internal/db/bundb/interaction.go
tsmethurst: Whoops! Copy paste error, thanks, will fix!

internal/gtsmodel/interactionpolicy.go
```diff
@@ -0,0 +1,314 @@
...
+			v == VisibilityPublic
+
+	// Any other combo
+	// is probably fine.
+	default:
+		return true
+	}
+}
+
+type PolicyValues []PolicyValue
```
NyaaaWhatsUpDoc, line 112: is it worth having these type aliases? i only really tend to make slice type aliases if there's method receivers attached to them

internal/gtsmodel/interactionapproval.go
```diff
@@ -0,0 +1,48 @@
...
+	InteractingAccount   *Account        `bun:"-"`                                                           // account corresponding to targetAccountID
+	InteractionURI       string          `bun:",nullzero,notnull"`                                           // URI of the target like, reply, or announce
+	InteractionType      InteractionType `bun:",nullzero,notnull"`                                           // One of Like, Reply, or Announce.
+	URI                  string          `bun:",nullzero,notnull,unique"`                                    // ActivityPub URI of the Accept.
+}
+
+type InteractionType string
+
+const (
+	InteractionLike     InteractionType = "Like"
```
NyaaaWhatsUpDoc, line 45: might be worth making these integer type to save space in the database. can always have a `func (t InteractionType) String() string` receiver method so they appear nicely in debug logs (or can be converted easily for other API usage)

NyaaaWhatsUpDoc:
finished reviewing now. great stuff!! :sunglasses:

internal/db/bundb/interaction.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3075/commits/d4f8f8405afc2be625d6f808d1d0bf0902c32e31

internal/db/bundb/migrations/20240620074530_interaction_policy.go
tsmethurst: When I tested it it seemed to handle it OK, but I removed it here just to be safe: https://github.com/superseriousbusiness/gotosocial/pull/3075/commits/fef8043d5ccfb657a72c611ea4eb5175f3af2ce4

internal/db/bundb/migrations/20240620074530_interaction_policy.go
tsmethurst: Done! Thanks, forgot about that function! Now I can do the whole migration inside a transaction :) https://github.com/superseriousbusiness/gotosocial/pull/3075/commits/fef8043d5ccfb657a72c611ea4eb5175f3af2ce4

internal/gtsmodel/interactionapproval.go
tsmethurst: Ah nice idea, done :) https://github.com/superseriousbusiness/gotosocial/pull/3075/commits/34db81f35b2aa23c37bb55837c9e5bdeb6754653

internal/gtsmodel/interactionpolicy.go
tsmethurst: I sorta like it, and might add to it in the next PR (i'm not sure yet) so I'd rather keep it if that's OK. Unless there's performance implications of course, then I can change it!

internal/gtsmodel/accountsettings.go
```diff
@@ -21,15 +21,20 @@ import "time"
...
+	CustomCSS                      string             `bun:",nullzero"`                                                   // Custom CSS that should be displayed for this Account's profile and statuses.
+	EnableRSS                      *bool              `bun:",nullzero,notnull,default:false"`                             // enable RSS feed subscription for this account's public posts at [URL]/feed
+	HideCollections                *bool              `bun:",nullzero,notnull,default:false"`                             // Hide this account's followers/following collections.
+	InteractionPolicyDirect        *InteractionPolicy `bun:""`                                                            // Interaction policy to use for new direct visibility statuses by this account. If null, assume default policy.
+	InteractionPolicyMutualsOnly   *InteractionPolicy `bun:""`                                                            // Interaction policy to use for new mutuals only visibility statuses. If null, assume default policy.
```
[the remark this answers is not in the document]
tsmethurst: It should be OK without, since a nil pointer will convert to null in the database. nullzero is, I think, only really useful when you're dealing with non-pointer structs and basic types. https://bun.uptrace.dev/guide/models.html#nulls

Issue #3073, opened 2024-07-05
[bug] turnoff autocapitalize and spellcheck for some input tags


coxde:
### Describe the bug with a clear and concise description of what the bug is. Please include screenshots of any visual issues.

Hiya! Maybe this is more of a suggestion than a bug? Anyway, because some input tags don't set `autocapitalize=none` and `spellcheck=false`, when I use a virtual keyboard (on a phone for example), the auto capitalization and spell checking is really annoying. And those inputs don't really need capitalize and spell check (also a potential privacy issue). For example: username, domain, IP don't need those (and maybe more!).

![image](https://github.com/superseriousbusiness/gotosocial/assets/63153334/fc78550d-5282-4a4f-9c1a-5c23a55df699)

### What's your GoToSocial Version?

v0.16.0+git-f1cbf6f

### Browser version

Firefox 127.0.2

### Anything else we need to know?

MDN:

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck

tsmethurst:
Good idea

PR #3077, opened 2024-07-07
[frontend] Better autocapitalize/spellcheck settings on forms

tsmethurst:
# Description

>

This PR does some usability tweaks on the frontend:

- Use autocapitalize/spellcheck more appropriately for different form fields.
- Use `type="url"` where appropriate.
- Do domain name validation in form fields that require a domain name.
- Use monospace for more "codelike" fields.

closes https://github.com/superseriousbusiness/gotosocial/issues/3073

## Checklist

### DIFF
--- a/web/source/settings/views/admin/actions/keys/expireremote.tsx
+++ b/web/source/settings/views/admin/actions/keys/expireremote.tsx
@@ -22,9 +22,33 @@ import { TextInput } from "../../../../components/form/inputs";
 import MutationButton from "../../../../components/form/mutation-button";
 import { useTextInput } from "../../../../lib/form";
 import { useInstanceKeysExpireMutation } from "../../../../lib/query/admin/actions";
+import isValidDomain from "is-valid-domain";
 
 export default function ExpireRemote({}) {
-	const domainField = useTextInput("domain");
+	const domainField = useTextInput("domain", {
+		validator: (v: string) => {
+			if (v.length === 0) {
+				return "";
+			}
+
+			if (v[v.length-1] === ".") {
+				return "invalid domain";
+			}
+
+			const valid = isValidDomain(v, {
+				subdomain: true,
+				wildcard: false,
+				allowUnicode: true,
+				topLevel: false,
+			});
+
+			if (valid) {
+				return "";
+			}
+
+			return "invalid domain";
+		}
+	});
 
 	const [expire, expireResult] = useInstanceKeysExpireMutation();
 
@@ -52,11 +76,13 @@ export default function ExpireRemote({}) {
 			<TextInput
 				field={domainField}
 				label="Domain"
-				type="string"
+				type="text"
+				autoCapitalize="none"
+				spellCheck="false"
 				placeholder="example.org"
 			/>
 			<MutationButton
-				disabled={!domainField.value}
+				disabled={!domainField.value || !domainField.valid}
 				label="Expire keys"
 				result={expireResult}
 			/>
--- a/web/source/settings/views/admin/debug/apurl/index.tsx
+++ b/web/source/settings/views/admin/debug/apurl/index.tsx
@@ -53,6 +53,9 @@ export default function ApURL() {
 				<TextInput
 					field={urlField}
 					label="URL"
+					type="url"
+					pattern="(http|https):\/\/.+"
+					placeholder="https://example.org/users/someone"
 				/>
 				<MutationButton
 					disabled={!urlField.value}
--- a/web/source/settings/views/admin/emoji/remote/steal-this-look.tsx
+++ b/web/source/settings/views/admin/emoji/remote/steal-this-look.tsx
@@ -52,9 +52,10 @@ export default function StealThisLook({ emojiCodes }) {
 					</label>
 					<div className="row">
 						<input
-							type="text"
 							id="url"
 							name="url"
+							type="url"
+							pattern="(http|https):\/\/.+"
 							onChange={urlField.onChange}
 							value={urlField.value}
 						/>
--- a/web/source/settings/views/admin/instance/settings.tsx
+++ b/web/source/settings/views/admin/instance/settings.tsx
@@ -79,7 +79,10 @@ function InstanceSettingsForm({ data: instance }: InstanceSettingsFormProps) {
 	const [submitForm, result] = useFormSubmit(form, useUpdateInstanceMutation());
 
 	return (
-		<form onSubmit={submitForm}>
+		<form
+			onSubmit={submitForm}
+			autoComplete="none"
+		>
 			<h1>Instance Settings</h1>
 
 			<div className="form-section-docs">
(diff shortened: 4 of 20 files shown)


web/source/settings/views/admin/actions/keys/expireremote.tsx
```diff
@@ -22,9 +22,33 @@ import { TextInput } from "../../../../components/form/inputs";
...
 
 export default function ExpireRemote({}) {
-	const domainField = useTextInput("domain");
+	const domainField = useTextInput("domain", {
+		validator: (v: string) => {
+			if (v.length === 0) {
+				return "";
+			}
+
+			if (v[v.length-1] === ".") {
```
daenney, line 34: In theory, a fully qualified domain is `domain.com.`, as `.` is the root. But in practice I doubt anyone would enter it that way. Though it depends a bit where the user is copy-pasting from.

It might be nice to do a `length==1` and `not a dot` instead.

web/source/settings/views/admin/debug/apurl/index.tsx
```diff
@@ -53,6 +53,9 @@ export default function ApURL() {
 				<TextInput
 					field={urlField}
 					label="URL"
+					type="url"
+					pattern="(http|https):\/\/.+"
+					placeholder="https://example.org/users/someone"
```
daenney, line 58: Do we want to set a `minlength` on these too?

web/source/settings/views/admin/emoji/remote/steal-this-look.tsx
```diff
@@ -52,9 +52,10 @@ export default function StealThisLook({ emojiCodes }) {
 					</label>
 					<div className="row">
 						<input
-							type="text"
 							id="url"
 							name="url"
+							type="url"
+							pattern="(http|https):\/\/.+"
```
daenney, line 58: Same here, maybe a `minlength`?

web/source/settings/views/admin/instance/settings.tsx
```diff
@@ -172,12 +180,15 @@ function InstanceSettingsForm({ data: instance }: InstanceSettingsFormProps) {
...
 				placeholder="admin"
+				autoCapitalize="none"
+				spellCheck="false"
 			/>
 
 			<TextInput
 				field={form.contactEmail}
 				label="Contact email"
 				placeholder="admin@example.com"
+				type="email"
```
daenney, line 191: :pray:

web/source/settings/views/moderation/accounts/pending/index.tsx
```diff
@@ -44,7 +44,23 @@ export default function AccountsPending() {
...
+					You can see a list of pending account sign-ups below.
+					<br/>
+					To approve or reject a sign-up, click on the account's name in the
+					list, and use the controls at the bottom of the account detail view.
+				</p>
+				<a
+					href="https://docs.gotosocial.org/en/latest/admin/signups/"
+					target="_blank"
+					className="docslink"
+					rel="noreferrer"
```
daenney, line 59: Since we set a `Referrer-Policy` header now with `same-origin` the `rel="noreferrer"` here shouldn't be necessary any more. But might be good to have.

web/source/settings/views/moderation/accounts/search/index.tsx
```diff
@@ -42,7 +43,31 @@ export function AccountSearchForm() {
...
-		by_domain: useTextInput("by_domain", { defaultValue: urlQueryParams.get("by_domain") ?? ""}),
+		by_domain: useTextInput("by_domain", {
+			defaultValue: urlQueryParams.get("by_domain") ?? "",
+			validator: (v: string) => {
+				if (v.length === 0) {
+					return "";
+				}
+
+				if (v[v.length-1] === ".") {
+					return "invalid domain";
```
daenney, line 54: Same thing here with the final dot. Probably not an issue but in theory we should accept `domain.com.`.

web/source/settings/views/admin/debug/apurl/index.tsx
daenney: Do we need to set the `'u'` flag here to allow for unicode?

web/source/settings/views/admin/actions/keys/expireremote.tsx
daenney: We can also add some more validation. Any domain should at least be of the shape `domain.com`, so checking for at least one dot may also be good.

tsmethurst: Right! Indeed, the domain validation library actually allows domains ending with `.` which is why I added that cheeky little check to see if the last character is `.`, because I don't think our backend supports that properly.

I think it checks for at least one dot, I'll validate it now and change it if not.

web/source/settings/views/admin/debug/apurl/index.tsx
tsmethurst: Ah good question, I'll check...

web/source/settings/views/moderation/accounts/pending/index.tsx
tsmethurst: Mmm, I'll leave it in just to make it explicit :)

web/source/settings/views/admin/actions/keys/expireremote.tsx
tsmethurst: Alright, it does seem to check for at least one dot :) So `test` is not valid, `test.` is not valid (with the ends-with-a-dot check) but `test.com` is valid.

web/source/settings/views/admin/debug/apurl/index.tsx
tsmethurst: Seems to work OK with unicode characters already :) `https;//üxample.org` is valid.

Instead of a min-length I guess what we could do is try parsing the field value into a URL, and fail validation if it throws, similar to what we do for RE2 regexes now in the header permissions form. What do you think?

tsmethurst: Oof OK I tried it with parsing a URL but that's somehow even clunkier and doesn't always return invalid even for really wonky URLs, so I think probably just keeping the regex in this PR is good enough for now..

daenney: The only reason I suggest `minlength` was to ensure we got at least `http://` without even bothering with the pattern match since anything less than 7 characters won't match anyway. It's probably fine as is though.

tsmethurst: Aha okay! Yeah this is probably fine then, and a bit more explicit about what we're looking for :)

web/source/settings/views/admin/emoji/remote/steal-this-look.tsx
tsmethurst: Resolved above

web/source/settings/views/moderation/accounts/search/index.tsx
tsmethurst: Resolved above :sunglasses:

Issue #75, opened 2021-06-28
[feature] Avatar & profile header descriptions


ArnoVdH:
Image descriptions are useful, accessibility-wise. They are used in posts, but there's at the moment no way of describing profile avatars or head images. Adding these would be a nice change.

rkingett:
Agreed. Adding these would be a very nice change and a very welcome edition.

PR #3086, opened 2024-07-08
[feature/frontend] Allow setting alt-text for avatar + header

tsmethurst:
# Description

>

This PR allows users to set alt text on their headers + avatars, which will then show up on the web frontend for their profile (falls back to previous alt text if no alt text set).

Also uses a photoswipe gallery view for the avatar in the profile header, so it can be peeped at more closely without having to open another tab (falls back to previous behavior if no javascript allowed on page).

Alt text isn't federated at this point, it's just stored locally.

closes https://github.com/superseriousbusiness/gotosocial/issues/75

## Checklist


ThisIsMissEm:
Is the header/avatar alt text federated at all?

tsmethurst:
Not yet! Some time soon we'll probably start federating it in the title or summary or whatever it is that's used for attachments currently, should be a relatively small change I think.

ThisIsMissEm:
Alright, because just the other night I was thinking about "do we have alt text for these?" and realised we didn't in Mastodon. (we apparently use the "name" property for alt text)

ThisIsMissEm:
We currently use `name` but are discussing using `summary` (which we also prefer in versions greater than 3.2.0 for media attachments), however there is a concern about `summary` being able to contain HTML, and hence needing to be sanitised.

Hopefully we'll also support sending both for at least one minor relesse.

https://github.com/mastodon/mastodon/pull/32538

We've also marked implementation of alt text on avatars & headers as a "help wanted" ticket, which means that the team wants to implement, it's small enough that someone could implement with guidance, and guidance/help will be provided.

https://github.com/mastodon/mastodon/issues/10313#issuecomment-2416000335

tsmethurst:
That's great, thanks for the updates :) I'll make an issue for GtS specifically to federate these fields out, and link back here.

Issue #2058, opened 2023-08-02
[feature] Store status tombstones to avoid breaking threads


tsmethurst:
Right now if a status parent is deleted, child statuses end up 'orphaned' and will return an error during status visibility checks.

To avoid breaking threads, we should consider storing tombstones for deleted statuses. The ID of the tombstone could be the ID of the tombstoned (yes that's a word now) status, and it could contain fields like AccountID, InReplyToID, InReplyToURI, etc. Something along those lines. When doing visibility checks, if a parent status for the given status cannot be found, the visibility checker could look in tombstones to see if the status is orphaned, and alter visibility accordingly.

To avoid clogging up the db with unchecked tombstones, tombstones older than a month (?) could be periodically cleared out.

tsmethurst:
Well, I think any tombstoning we would do to avoid breaking threads would also be accompanied by some tweaks to the visibility logic. So yes and no!

mirabilos:
A wild idea I just had is to mangle the `in_reply_to_id`, `in_reply_to_uri` and `in_reply_to_account_id` fields of all `statuses` whose parent gets deleted (e.g. in a blocking or defederating situation) to the next retained parent (or to the top, if none… for full effect, this might need thread-to-the-top traversal and fetching of missing toots if not yet present, but if not, emptying those fields might be better than nothing).

The `thread_id` could be kept (unless made a top-level post, I assume?).

This would not only avoid #2011 but also retain the thread information for forever-ish (otherwise, status tombstones ought to NOT be deleted (per <https://github.com/superseriousbusiness/gotosocial/issues/2058#issuecomment-1814386837>) as long as the thread is still present in `statuses`), modulo refetches I guess…

… status tombstones are probably the better solution.

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/2058.

PR #3088, opened 2024-07-09
[bugfix] Don't throw error when parent statuses are missing (#2011)

### LINKED ISSUE #2011: [bug] unable to interact (including delete) post after delete-and-redraft parent + 500 ISE

I’ve written a post (in a direct message), then replied to that, then later delete-and-redraft-ed the first one.

This now leads to severe misbehaviour:

One, it’s still in my Semaphore timeline, but I cannot delete it (“404”, but the log does not show the ID of the post itself but probably the one of its now-gone parent).

Two, if I zoom on it in Semaphore, I get “internet access failed, showing offline content”.

Three, if I start a new Semaphore in a new tab, then search for the toot URL copied from the other tab, I get a 500:

    timestamp="21/07/2023 00:22:39.122" func=middleware.Logger.func1.1 level=ERROR latency=28.552173ms userAgent="Mozilla/5.0 (X11; Linux x86_64; rv:102.0) Gecko/20100101 Firefox/102.0" method=GET statusCode=500 path=/api/v2/search error="Error #01: packageStatuses: error checking visibility of status 01H5TPKZC51NA28WE3822PS62F for account 01GS55SW7VWYN3BFRYCX5NW526: isStatusVisible: error populating status 01H5TPKZC51NA28WE3822PS62F: \"error populating status parent: no entries\",\"error populating status parent: no entries\"\n" requestID=yq0d6xc90400048tbnhg msg="Internal Server Error: wrote 54B"

0.10.0-rc3 git-f431974

### GoToSocial Arch

amd64 binary

### What happened?

_No response_

I’m **fully** expecting to be able to do **any and all** interaction with any status I posted myself, independent of what happened to any other toot, related or not.

This includes parent being gone (deleted (redrafted or not)), parent not being visible to me any more because the parent’s author blocked or fediblocked me or I blocked or fediblocked them, or because the instance of the parent is gone or no longer reachable or suddenly serves http on the https port (don’t laugh, it happened to one (`hubzilla.eu`) which I saw watching the log today).

### How to reproduce it?

I assume posting, replying, then deleting the first post will do.

_No response_

snowkat:
# Description

>

This pull request allows retrieving posts that reply to a parent status that isn't in the database, by ending the traversal below the missing parent. A missing parent post can happen for a few reasons: the calling user being blocked, a deleted post, or federation/remote server issues come to mind.

closes #2011

## Checklist

- [X] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [X] I/we have made any necessary changes to documentation.
- [X] I/we have run tests and they pass locally with the changes.
- [X] I/we have run `go fmt ./...` and `golangci-lint run`.

### DIFF
--- a/internal/db/bundb/status.go
+++ b/internal/db/bundb/status.go
@@ -561,14 +561,13 @@ func (s *statusDB) GetStatusParents(ctx context.Context, status *gtsmodel.Status
 
 	for id := status.InReplyToID; id != ""; {
 		parent, err := s.GetStatusByID(ctx, id)
-		if err != nil {
-			if errors.Is(err, db.ErrNoEntries) {
-				// This can be caused by a few cases: user has been blocked,
-				// status has been deleted, federation issues...
-				log.Debug(ctx, "status parent not found, returning up to this point")
-				break
-			}
+		if err != nil && !errors.Is(err, db.ErrNoEntries) {
 			return nil, err
+		}
+
+		if parent == nil {
+			// Parent status not found (e.g. deleted)
+			break
 		}
 
 		// Append parent status to slice
--- a/internal/db/bundb/status.go
+++ b/internal/db/bundb/status.go
@@ -202,7 +202,7 @@ func (s *statusDB) PopulateStatus(ctx context.Context, status *gtsmodel.Status) 
 				gtscontext.SetBarebones(ctx),
 				status.InReplyToID,
 			)
-			if err != nil {
+			if err != nil && !errors.Is(err, db.ErrNoEntries) {
 				errs.Appendf("error populating status parent: %w", err)
 			}
 		}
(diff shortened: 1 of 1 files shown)


snowkat:
As an aside: I tried adding a test to status_test.go, but that requires updating testrig/testmodels.go in a way that breaks multiple other tests. If there's better way to go about adding statuses outside of `NewTestStatuses()`, I can build a test for this case.

tsmethurst:
I think this is probably OK as a *temporary* fix until we have #2058 in place to handle this properly. @NyaaaWhatsUpDoc do you have any onions?

internal/db/bundb/status.go
```diff
@@ -562,6 +562,12 @@ func (s *statusDB) GetStatusParents(ctx context.Context, status *gtsmodel.Status
 	for id := status.InReplyToID; id != ""; {
 		parent, err := s.GetStatusByID(ctx, id)
 		if err != nil {
+			if errors.Is(err, db.ErrNoEntries) {
```
NyaaaWhatsUpDoc, line 565: alternatively how do you feel about doing:

```golang
parent, err := s.GetStatusByID(ctx, id)
    return nil, err
}

if parent == nil {
    break
}
```

it just makes it a bit clearer this isn't so much an error condition and more to be expected, as yeah we should have been handling this given how we handle deleting statuses

NyaaaWhatsUpDoc:
great work, thanks! just the one comment from me :)

internal/db/bundb/status.go
snowkat: Makes sense to me!

PR #3090, opened 2024-07-09
[feature] support processing of (many) more media types

NyaaaWhatsUpDoc:
# Description

This replaces our media processing pipeline with the previous Go code we were using, to instead use embedded WebAssembly binaries of [ffmpeg and ffprobe](https://codeberg.org/gruf/go-ffmpreg). As a result we can now easily support a multitude of extra image and video types, and adds audio file support too. This also allows us to ensure *all* media we processing (minus emojis) gets its metadata stripped, whereas previously it was only select types in select situations.

This also removes media file size enforcement out of the media manager itself, and instead to be handled by callers of the various media processing functions. (while still adding a check to the media manager for whether a "limit" was reached, for nicer error output). This makes the code in the media manager itself a little neater, and as part of this change minimizes the amount to which we wrap http response bodies, to ensure we can rely on Linux's sendfile syscall for optimal performance draining incoming media to /tmp. (as part of this change, i also got rid of the unnecessary separate image / video size config flags, which weren't actually being used, and replaced them with general media max local / remote media sizes).

Performance impact is yet to be measured, but given we're not doing any expensive transcodes it hopefully won't be *too* big.

## Todo

- [x] update tests to pass given the big overhaul of media processing code
- [x] add tests for some of the new processable media filetypes
- [x] finish documenting the code, there's still some placeholder `// ...` comments

## Checklist


internal/media/util.go
```diff
@@ -17,25 +17,134 @@
...
+import (
+	"cmp"
+	"errors"
+	"fmt"
+	"image"
+	"image/jpeg"
+	"io"
+	"os"
+
+	"codeberg.org/gruf/go-mimetypes"
```
NyaaaWhatsUpDoc, line 29: this is an embedded map of mimetypes by file extension: https://codeberg.org/gruf/go-mimetypes

tsmethurst:
Beautiful :heart_eyes:

PR #3102, opened 2024-07-14
[chore] add step-by-step logging to long migration

tsmethurst:
This PR just adds some logging to the very long interaction policy migration, so that users know stuff is happening and they oughtn't be jumping the gun and pressing ctrl-c on it.

Looks like this:

image


VyrCossont (approved):
good idea, i would not have guessed that this could take an hour plus

PR #3108, opened 2024-07-16
[feature] Allow users to set default interaction policies per status visibility

tsmethurst:
# Description

This PR adds api endpoints for interaction policies:

- `GET /api/v1/interaction_policies/defaults` -- get current default policies for requesting account. 
- `PATCH /api/v1/interaction_policies/defaults` -- set default policies for requesting account.

New statuses created by the account from that point onwards will use whatever visibility policy they had set when the status was created.

Actual enforcement of these policies will be done in a subsequent PR.

Also adds frontend stuff for setting / viewing policies.

image

Part of #3057 

## Checklist


internal/processing/account/interactionpolicies.go
```diff
@@ -0,0 +1,201 @@
...
+			)
+			return nil, gtserror.NewErrorInternalError(err)
+		}
+	}
+
+	// Take set "direct" policy
+	// or global default.
+	direct := cmp.Or(
+		settings.InteractionPolicyDirect,
+		gtsmodel.DefaultInteractionPolicyDirect(),
```
NyaaaWhatsUpDoc, line 52: i think it might be worth either updating these defaults to be global variables so they aren't instantiated every time, or updating these uses here to be `if direct = settings.InteractionPolicyDirect; direct == nil { direct = gtsmodel.DefaultInteractionPolicyDirect() }`, as the only downside to using `cmp.Or()` here is that since it's a function call we always end up allocating new default policies, whether they get used or not.

same goes for the other defaults checking below

internal/processing/account/interactionpolicies.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (p *Processor) DefaultInteractionPoliciesGet(
+	ctx context.Context,
+	requestingAccount *gtsmodel.Account,
+) (*apimodel.DefaultPolicies, gtserror.WithCode) {
+	// Get settings for this account.
+	settings := requestingAccount.Settings
+	if settings == nil {
+		var err error
+		settings, err = p.state.DB.GetAccountSettings(ctx, requestingAccount.ID)
```
NyaaaWhatsUpDoc, line 38: given how often we end up doing this i think it might be worth separating this out into a `PopulateAccountSettings()` function, what do you think?

internal/processing/account/interactionpolicies.go
```diff
@@ -0,0 +1,201 @@
...
+		// Set new default policy.
+		settings.InteractionPolicyPublic = policy
+	}
+
+	if err := p.state.DB.UpdateAccountSettings(ctx, settings); err != nil {
+		err := gtserror.Newf("db error updating setttings: %w", err)
+		return nil, gtserror.NewErrorInternalError(err, err.Error())
+	}
+
+	requestingAccount.Settings = settings
```
NyaaaWhatsUpDoc, line 199: you shouldn't need to do this given the settings is a ptr :)

internal/processing/status/create.go
```diff
@@ -281,26 +288,81 @@ func (p *Processor) processMediaIDs(ctx context.Context, form *apimodel.Advanced
...
+	// What? Fall back to global default.
 	default:
-		vis = gtsmodel.VisibilityDefault
+		status.Visibility = gtsmodel.VisibilityDefault
 	}
 
-	// Todo: sort out likeable/replyable/boostable in next PR.
+	// Set federated flag to form value
+	// if provided, or default to true.
+	status.Federated = cmp.Or(
```
NyaaaWhatsUpDoc, line 312: again the issue here is that we're allocating even in the case that form.Federated ends up being used. i know it's only a boolean ptr allocation but i think it's not the best habit we get into for our codebase. i think doing `util.Ptr(util.PtrValueOr(form.Federated, true))`would probably do what you wanted inline here

internal/typeutils/internaltofrontend.go
```diff
@@ -2256,3 +2271,96 @@ func (c *Converter) ThemesToAPIThemes(themes []*gtsmodel.Theme) []apimodel.Theme
...
+//
+// RequestingAccount can also be nil for
+// unauthorized requests (web, public api etc).
+func (c *Converter) InteractionPolicyToAPIInteractionPolicy(
+	ctx context.Context,
+	policy *gtsmodel.InteractionPolicy,
+	status *gtsmodel.Status,
+	requestingAccount *gtsmodel.Account,
+) (*apimodel.InteractionPolicy, error) {
+	convertURIs := func(policyURIs gtsmodel.PolicyValues) (apiPolicyValues []apimodel.PolicyValue) {
```
NyaaaWhatsUpDoc, line 2289: given this function doesn't actually use any internal typeutils.Converter{} fields, it's probably worth just making this a global scope function instead of a receiver method to the converter.

also given the complexity of the `convertURIs` function (so it won't get [inlined](https://dave.cheney.net/2020/04/25/inlining-optimisations-in-go) and will instead need to be allocated as a separate internal function), i think this is probably worth separating out into a separate function that gets called, instead of being defined within `InteractionPolicyToAPIInteractionPolicy()`. what do you think?

NyaaaWhatsUpDoc: also whilst i'm here, are the status and account actually needed to be passed here since they're not used?

tsmethurst: > also whilst i'm here, are the status and account actually needed to be passed here since they're not used?

They will be in the upcoing PR :sunglasses:

Elsewise, I'll see what I can pull out.

internal/processing/account/interactionpolicies.go
tsmethurst: Got it! Done: https://github.com/superseriousbusiness/gotosocial/pull/3108/commits/dae10f538420f9d96a937260289407217c34fa06

internal/processing/account/interactionpolicies.go
tsmethurst: It was only called twice in this package as far as I could see, elsewhere it's just called as part of PopulateAccount, so I just added a little utility function here to avoid some code repetition: https://github.com/superseriousbusiness/gotosocial/pull/3108/commits/eec181d9eeec2b507e80f7c6de8e57510896809d

Happy to move that to a more generally accessible util function if we end up needing it more though :)

internal/api/model/status.go
```diff
@@ -102,6 +102,8 @@ type Status struct {
 	Text string `json:"text,omitempty"`
 	// A list of filters that matched this status and why they matched, if there are any such filters.
 	Filtered []FilterResult `json:"filtered,omitempty"`
+	// The interaction policy for this status, as set by the status author.
+	InteractionPolicy InteractionPolicy `json:"interaction_policy"`
```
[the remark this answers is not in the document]
tsmethurst: Hrmn maybe in the future, I'd rather leave it explicit for now though I think.

internal/processing/account/interactionpolicies.go
tsmethurst: Got it, silly. Resolved it in this one: https://github.com/superseriousbusiness/gotosocial/pull/3108/commits/eec181d9eeec2b507e80f7c6de8e57510896809d

internal/typeutils/frontendtointernal.go
```diff
@@ -57,3 +62,160 @@ func APIFilterActionToFilterAction(m apimodel.FilterAction) gtsmodel.FilterActio
...
+		&canAnnounceAlways,
+	} {
+		if !slices.ContainsFunc(
+			*uris,
+			func(uri gtsmodel.PolicyValue) bool {
+				return uri == gtsmodel.PolicyValuePublic ||
+					uri == gtsmodel.PolicyValueAuthor
+			},
+		) {
+			*uris = *util.Ptr(append(
```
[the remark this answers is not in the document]
tsmethurst: Right yeah I see what you mean, it's a bit mental. I'll see if I can make more explicit what it's doing.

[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3108/commits/c0046ffb6459c7971e58e50214e347611ece1b68 should be a bit more readable :')

internal/processing/status/create.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3108/commits/f093c479613812dabe8709c0e1efd0c0440fcdbc

internal/api/model/status.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: yeah was thinking this a future thing :)

internal/typeutils/internaltofrontend.go
tsmethurst: Pulled it out here! https://github.com/superseriousbusiness/gotosocial/pull/3108/commits/ec54eb11ff2b81344e2ce3bea2bd022b8d4f0780

Also did deduplication during iteration to avoid having to iterate through the slice twice.

internal/typeutils/frontendtointernal.go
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: moving that previous comment here:

only thing i'm unsure of here is the slices.Grow() 🤔 under the hood it's basically doing return append(slice, "")[:len-1] to guarantee the extra slot but reslicing to hide the extra length, which you're then filling with append(slice, PolicyValue). or am i understanding this wrong?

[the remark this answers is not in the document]
tsmethurst: I'll change it! (for context to anyone reading, me and kim just discussed this on matrix too)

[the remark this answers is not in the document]
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3108/commits/d68ca8d7f245fd6d1ae6d4a86290d5da016ce3f7

PR #3110, opened 2024-07-17
[chore] media pipeline improvements

NyaaaWhatsUpDoc:
# Description

This updates our media processing to only store attachment path details on a successful store. This makes it easier to see attachments that are stored locally. It also tweaks the wording of the placeholder attachments warning such that it works for users that say don't want to store any local media whatsoever.

In the name of storage space / efficiency, the media attachments filetype column has also been updated from being a string, to be an integer that only gets stringified for the frontend. And the ffmpeg / ffprobe instance pools are updated such that instances are only allocated when they are needed now, so startup memory usage should be a bit lower. (though note that it will not "size back" after being allocated but during periods of low usage, that would end up adding significant complexity that's not worth it for server software).

## Checklist


internal/typeutils/util.go
```diff
@@ -104,62 +125,50 @@ func misskeyReportInlineURLs(content string) []*url.URL {
...
 		return "", arr
 	}
 
-	// Plural / singular.
-	var (
-		attachments string
-		links       string
-	)
-
-	if unknownsLen == 1 {
```
tsmethurst, line 139: I'd rather we keep this bit, as it reads a bit nicer for screenreader users.

internal/util/ptr.go
```diff
@@ -43,6 +43,15 @@ func PtrIf[T comparable](t T) *T {
 	return &t
 }
 
+// PtrOr returns ptr value only if 't' non-nil.
+func PtrOr[T any](t *T) T {
```
tsmethurst, line 47: Might be better to call this PtrValueOrZero or something like that, what do you think?

tsmethurst:
This seems pretty reasonable to me :) Can't say no to cheeky savings in the database.

internal/typeutils/util.go
NyaaaWhatsUpDoc: oh that's a good point! thanks for pointing that out :) have added back singular / plural word forms 4c186693f895de12829689eefc83edf616a4afa9

internal/util/ptr.go
NyaaaWhatsUpDoc: oh i see what you mean given its similarity to the PtrValueOr() function. i ended up updating this to be `PtrOrZero()` given it "takes ptr value or zero", and updating the other to `PtrOrValue()` so they both match (+ both read a bit better) given it "takes ptr value, or value"

NyaaaWhatsUpDoc: what do you think?

tsmethurst: That looks fine yeah :)

NyaaaWhatsUpDoc:
the tests were still failing as the new lazy failing of the ffmpeg / ffprobe webassembly modules was causing the tests to timeout that were expecting some things to be processed within a certain timeframe. so i updated the lazy initialization to store just 1 open instance at testrig init, and all other instances it will initialize as it needs them :)

PR #3111, opened 2024-07-17
[chore] Add interaction filter to complement existing visibility filter

tsmethurst:
# Description

>

This PR is another intermediate PR and the next part of https://github.com/superseriousbusiness/gotosocial/issues/3057. ~~It depends on https://github.com/superseriousbusiness/gotosocial/pull/3108 to be squerged first.~~

The PR adds a new "interaction filter" alongside the existing visibility filter. The interaction filter is used to check if a given interaction is permitted against a certain status.

In this PR only "permitted" and "forbidden" are used, but in the next PR (coming up after this one), "with approval" is also added.

## Checklist


internal/federation/dereferencing/status_permitted.go
```diff
@@ -0,0 +1,216 @@
...
+
+	if replyable.Forbidden() {
+		// Replier is not permitted
+		// to do this interaction.
+		return onFalse()
+	}
+
+	// TODO in next PR: check conditional /
+	// with approval and deref Accept.
+	if !replyable.Permitted() {
```
tsmethurst, line 152: This (and the matching code in `isPermittedBoost`) looks a bit silly right now because it's basically a stub. In the next PR (which adds federation of interaction policies and Accepts), this will check if approvedByURI is set on the status, and go deref the Accept it references etc.

internal/processing/status/create.go
```diff
@@ -191,6 +193,45 @@ func (p *Processor) processInReplyTo(ctx context.Context, requester *gtsmodel.Ac
...
+		// as pending approval and wait for an accept.
+		pendingApproval = true
+
+	case policyResult.Permitted():
+		// We're permitted to do this
+		// based on another kind of match.
+		pendingApproval = false
+	}
+
+	status.PendingApproval = &pendingApproval
```
tsmethurst, line 233: This pending approval stuff will also be handled properly in the next PR, it doesn't do much now.

internal/processing/processor.go
```diff
@@ -176,7 +177,8 @@ func NewProcessor(
 ) *Processor {
 	var (
 		parseMentionFunc = GetParseMentionFunc(state, federator)
-		filter           = visibility.NewFilter(state)
+		visFilter        = visibility.NewFilter(state)
```
NyaaaWhatsUpDoc, line 180: these should be updated to be arguments passed in to `NewProcessor()`, just so they're all using the same instances

Issue #3041, opened 2024-06-26
[feature] Encode thumbnails into format that keeps transparency


tsmethurst:
Currently I think we encode thumbnails to .jpg format, which is fine for static images that don't have an alpha/transparency layer, but is a bit of a shame for images that use transparency.

We might consider encoding (gif, webp, and png) image thumbnails to png instead, to try to keep transparency, but we'll have to investigate the best way of doing this.

Re: https://github.com/superseriousbusiness/gotosocial/issues/3040

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/pull/3116

PR #3116, opened 2024-07-19
[feature] use webp for thumbnails

NyaaaWhatsUpDoc:
# Description

Updates our media processor to use webp for generating thumbnails instead of jpeg. This allows us to maintain transparency layers while generating thumbnails of similar quality that work out *a little bit* smaller in practice.

## Checklist


NyaaaWhatsUpDoc:
tests still need updating

internal/media/ffmpeg.go
```diff
@@ -64,23 +75,48 @@ func ffmpegClearMetadata(ctx context.Context, filepath string, ext string) error
...
+			"scale="+scale+","+
+
+			// YUVA 4:2:0 colorspace
+			"format=yuva420p",
+
+		// Only one frame
 		"-frames:v", "1",
+
+		// ~40% webp quality
+		"-qscale:v", "40",
```
tsmethurst, line 114: I think you need to use `-quality` for the webp codec, see https://ffmpeg.org/ffmpeg-codecs.html#Options-36

internal/media/ffmpeg.go
```diff
@@ -100,10 +136,21 @@ func ffmpegGenerateStatic(ctx context.Context, filepath string) (string, error)
 	// Generate static with ffmpeg.
 	if err := ffmpeg(ctx, dirpath,
 		"-loglevel", "error",
+
```
tsmethurst, line 139: I think we can now just drop this `ffmpegGenerateStatic` function and just call `ffmpegGenerateThumb` for everything instead :)

internal/media/ffmpeg.go
NyaaaWhatsUpDoc: qscale is the global way to set quality regardless of codec :)

NyaaaWhatsUpDoc: so it passes it to the appropriate quality parameter of the codec

internal/media/ffmpeg.go
NyaaaWhatsUpDoc: ooh good point. that'll require more tests updating, you good with that?

NyaaaWhatsUpDoc: though wait actually, at the moment for static emoji we're generating full quality (lossless) forms of them, but then we'd be making a significant quality drop from png -> webm 40% quality

internal/media/ffmpeg.go
tsmethurst: Ah okay, are you sure it works for web piss too?

internal/media/ffmpeg.go
tsmethurst: Mm right. Maybe make this do `-lossless` webp then? Or web pee with default quality or something?

tsmethurst: And yes no problem re: tests :)

NyaaaWhatsUpDoc: I'm thinking maybe leave this one for another PR. still need to check what webp lossless will turn out like vs. png

internal/media/ffmpeg.go
NyaaaWhatsUpDoc: yeee, I tried both ways and both files had the same sha256 sum :)

internal/media/ffmpeg.go
tsmethurst: Alright no probs, I'll just update the non-emoji media for now

tsmethurst:
Closes https://github.com/superseriousbusiness/gotosocial/issues/3041

PR #3121, opened 2024-07-20
[feature] add flac support

NyaaaWhatsUpDoc:


internal/media/manager.go
```diff
@@ -38,8 +38,9 @@ var SupportedMIMETypes = []string{
 	"image/gif",  // .gif
 	"image/webp", // .webp
 
-	"audio/mp2", // .mp2
-	"audio/mp3", // .mp3
+	"audio/mp2",    // .mp2
+	"audio/mp3",    // .mp3
+	"audio/x-flac", // .flac
```
daenney, line 43: This should include `audio/flac` too. A bunch of software has been using that and it's officially incoming with https://datatracker.ietf.org/doc/draft-ietf-cellar-flac/14/

NyaaaWhatsUpDoc: useful to know, thank you!

PR #3124, opened 2024-07-21
[bugfix] update common get target account / status doing refresh async

NyaaaWhatsUpDoc:
fixes media 404s sometimes, as currently the async refresh causes the original media files to get invalidated while the client is still requesting them


tsmethurst:
Hurray!

PR #3133, opened 2024-07-23
[bugfix] Use punycode for `host` part of `resource` query param when doing webfinger requests

tsmethurst:
# Description

>

This pull request fixes an issue where we were using url-escaped unicode for account domains in webfinger queries. According to the webfinger RFC, the resource value must be a valid URI, and according to the URI RFC, the `host` part of a URI must contain only ASCII (I know, I know...).

To support this change, the check for dereferenced final account / status URIs has been updated to account for passing in unicode search queries, as otherwise the status or account will be rejected.

## Checklist


tsmethurst:
Ah hang on, found one thing that still needs to change.

tsmethurst:
I'm gonna squerge this one for now cuz I think it's fairly uncontroversial.

In a follow-up PR I wanna update the GetAccountByURI, GetAccountByURL, GetStatusByURI, and GetStatusByURL functions to *also* check in the database for both punycoded and unicoded URIs (preferring punycoded), but that doesn't belong here.

PR #3134, opened 2024-07-23
[chore] Add some log lines to recent migrations warning not to interrupt

tsmethurst:
Just a chore PR to add some logging to some of the lengthy recent migrations, so admins know that stuff is actually happening and won't be tempted to press ctrl-c.


NyaaaWhatsUpDoc:
squerge when ready :ok_hand:

PR #3136, opened 2024-07-24
[feature] Implement Mastodon-compatible roles

### LINKED ISSUE #3115: [feature] Admin role permission bitmask

This is extremely low priority.

Mastodon 4.0.0 added role permissions, a bitmask of individual permissions the role has. Apps that have admin tools, can utilize them to conditionally show them, e.g. a Moderator might not have permission to block federation and a client could just not show them the block federation tools at all or provide a helpful message rather than let it fail

https://docs.joinmastodon.org/entities/Role/#permissions

Since, from my understanding, GoToSocial has a preset of roles, admins should probably just have the 'Administrator' permission and mods should have the individual ones that best describe what they have access to

From my side, I can skip the permission checks or use the role names since they are static

### Additional context.

This is very niche and experimental and should really be treated as very low priority, as nobody but a handful of clients have admin tools at all (Though I will open more issues for the admin API in the future :3)

VyrCossont:
# Description

- `Account.role` should only be available through `verify_credentials` for checking current user's permissions
- `Account.role` now carries a Mastodon-compatible permissions bitmap and a marker for whether it should be shown to the public
- `Account.roles` added for *public* display roles (undocumented but stable since Mastodon 4.1)
- Web template now uses only public display roles (no user-visible change here, we already special-cased the `user` role)

Closes #3115

## Comments

We don't have or need Mastodon's elaborate custom permissions, but it costs us very little to be API-compatible with it in this case.

Implementing `Account.roles` is not strictly necessary for the feature request, but is closely adjacent and will make it easier for me to implement role badges in Feditext.

## Checklist


internal/api/model/account.go
```diff
@@ -305,6 +339,102 @@ const (
...
+type AccountRolePermissions int
+
+// MarshalJSON serializes an AccountRolePermissions as a numeric string.
+func (a *AccountRolePermissions) MarshalJSON() ([]byte, error) {
+	return json.Marshal(strconv.Itoa(int(*a)))
+}
+
+// UnmarshalJSON deserializes an AccountRolePermissions from a number or numeric string.
+func (a *AccountRolePermissions) UnmarshalJSON(b []byte) error {
+	if bytes.Equal(b, []byte("null")) {
```
NyaaaWhatsUpDoc, line 355: it would be better to do `if string(b) == "null"` here. converting to string only for the scope of a comparison doesn't end up doing any allocation :). also under the hood `bytes.Equal()` takes advantage of this so the inlined actual output of this call here will functionally be `if string(b) == string([]byte("null"))` :p

internal/api/model/account.go
```diff
@@ -305,6 +339,102 @@ const (
...
+		return nil
+	}
+
+	i := 0
+	if err := json.Unmarshal(b, &i); err != nil {
+		s := ""
+		if err := json.Unmarshal(b, &s); err != nil {
+			return gtserror.Newf("not a number or string")
```
NyaaaWhatsUpDoc, line 363: can just use `.New()` here instead of `Newf()`

internal/api/model/account.go
```diff
@@ -305,6 +339,102 @@ const (
...
+		}
+
+		i, err = strconv.Atoi(s)
+		if err != nil {
+			return gtserror.Newf("not a numeric string")
```
NyaaaWhatsUpDoc, line 368: same again

internal/api/model/account.go
NyaaaWhatsUpDoc: though given this gets returned to the API, it's probably better to return using `errors.New()` so it doesn't have our function name prefix

NyaaaWhatsUpDoc:
a few comments but otherwise great :)

PR #3141, opened 2024-07-26
[feature] Implement following hashtags

### LINKED ISSUE #2326: [feature] Would be great to implement the Tags endpoints (for following hashtags and whatnot)

Currently not able to follow a hashtag.

If we implement the [tags](https://docs.joinmastodon.org/methods/tags/) endpoints, this should be solved.

I believe there is currently no replacement for this. There is the tag timeline, but that is not the same since it doesn't allow to follow a tag, only to see its timeline.

### Additional context.

Love GTS, have a self-hosted version :) Thanks for all the work. I can try to help implement this if no one else has time!

VyrCossont:
# Description

- Adds API methods for following, unfollowing, and getting tags, and listing followed tags
- Sends new or updated public statuses with tags to the home timelines of local accounts following those tags

Closes #2326

## Comments

Followed tags have no state of their own, so we only store the association between a tag ID and an account ID. Caches mappings from a tag to accounts that follow it, and from an account to all the tags that it follows.

## Checklist


VyrCossont:
Hashtags I thought I'd followed on my test instance aren't showing up in my home timeline when they come in on remote statuses. Switching this to draft while I investigate.

VyrCossont:
It was boost-related. Here's the situation:
- I follow someone but I hide their boosts because they're too noisy
- I still want to see any post that hits my instance with, say, `#retrocomputing`
- they boost a post from an account that I don't follow, original post has `#retrocomputing`
- the status actually being timelined in `timelineAndNotifyStatus` is the boost, not the original
- I don't see boosts from the account that I follow, so I don't see the boost
- the boost itself doesn't have any tags, so I don't see it due to following the hashtag

Fix: unwrap boosts when timelining based on post content.

VyrCossont:
Apparently I need to explicitly check user-user blocks; the normal followers path checks them implicitly, because there can't be a follow relationship where a block exists. Admin-level domain blocks for remote statuses should be covered by `enrichStatus`.

VyrCossont:
Fix is working. Hopefully that was the last missing bit.

internal/db/bundb/tag.go
```diff
@@ -131,3 +135,162 @@ func (t *tagDB) PutTag(ctx context.Context, tag *gtsmodel.Tag) error {
...
+	}
+
+	// Invalidate account ID caches for the account and those tags.
+	t.state.Caches.DB.TagIDsFollowedByAccount.Invalidate(accountID)
+	t.state.Caches.DB.AccountIDsFollowingTag.Invalidate(tagIDs...)
+
+	return nil
+}
+
+func (t *tagDB) GetFollowerAccountIDsForTagIDs(ctx context.Context, tagIDs []string) ([]string, error) {
```
NyaaaWhatsUpDoc, line 285: subjective suggestion: `GetAccountIDsFollowingTagIDs()` reads a bit nicer imo

internal/db/bundb/tag.go
```diff
@@ -131,3 +135,162 @@ func (t *tagDB) PutTag(ctx context.Context, tag *gtsmodel.Tag) error {
...
+			Exec(ctx, &accountIDs); // nocollapse
+		err != nil && !errors.Is(err, db.ErrNoEntries) {
+			return nil, gtserror.Newf("error getting account IDs following tag %s: %w", tagID, err)
+		}
+
+		return accountIDs, nil
+	})
+}
+
+func (t *tagDB) DoesAccountFollowTag(ctx context.Context, accountID string, tagID string) (bool, error) {
```
NyaaaWhatsUpDoc, line 197: subjective suggestion: `IsAccountFollowingTag()` would fit better with our database functions are typically named, we have quite a few `Is___()` already

internal/gtsmodel/tag.go
```diff
@@ -28,4 +28,14 @@ type Tag struct {
 	Useable   *bool     `bun:",nullzero,notnull,default:true"`                              // Tag is useable on this instance.
 	Listable  *bool     `bun:",nullzero,notnull,default:true"`                              // Tagged statuses can be listed on this instance.
 	Href      string    `bun:"-"`                                                           // Href of the hashtag. Will only be set on freshly-extracted hashtags from remote AP messages. Not stored in the database.
+	Following *bool     `bun:"-"`                                                           // Whether the current user is following this tag, if there is a current user. Stored per-user, not in this table.
+}
+
+// FollowedTag represents a user following a tag.
+type FollowedTag struct {
+	// ID of the account that follows the tag.
```
NyaaaWhatsUpDoc, line 36: i was about to ask if it's worth adding a unique constraint to account_id + tag_id, but is that what specifying both as the primary key essentially does?

internal/api/client/followedtags/get.go
```diff
@@ -0,0 +1,139 @@
...
+		apiutil.ErrorHandler(c, gtserror.NewErrorUnauthorized(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	if _, err := apiutil.NegotiateAccept(c, apiutil.JSONAcceptHeaders...); err != nil {
+		apiutil.ErrorHandler(c, gtserror.NewErrorNotAcceptable(err, err.Error()), m.processor.InstanceGetV1)
+		return
+	}
+
+	page, errWithCode := paging.ParseIDPage(c,
```
NyaaaWhatsUpDoc, line 114: i'm so glad this paging code is getting such good use now. i literally spent an entire month last year obsessing over exactly how to write the API in a way it could be threaded through the codebase, so seeing it used always makes me smile :blush:

internal/processing/tags/get.go
```diff
@@ -0,0 +1,60 @@
...
+// Get gets the tag with the given name, including whether it's followed by the given account.
+func (p *Processor) Get(
+	ctx context.Context,
+	account *gtsmodel.Account,
+	name string,
+) (*apimodel.Tag, gtserror.WithCode) {
+	// Try to get an existing tag with that name.
+	tag, err := p.state.DB.GetTagByName(ctx, name)
+	if err != nil {
+		if errors.Is(err, db.ErrNoEntries) {
```
NyaaaWhatsUpDoc, line 40: subjective suggestion: in the codebase we often instead of handling `ErrNoEntries` as a proper error in the first error block, instead do something like:
```golang
tag, err := GetTagByName(ctx, name)
    return nil, gtserror.NewInternalError(...)
}

if tag == nil {
    // handle not found case
}
```

again though a very subjective point, feel free to ignore :p

internal/gtsmodel/tag.go
```diff
@@ -28,4 +28,14 @@ type Tag struct {
 	Useable   *bool     `bun:",nullzero,notnull,default:true"`                              // Tag is useable on this instance.
 	Listable  *bool     `bun:",nullzero,notnull,default:true"`                              // Tagged statuses can be listed on this instance.
 	Href      string    `bun:"-"`                                                           // Href of the hashtag. Will only be set on freshly-extracted hashtags from remote AP messages. Not stored in the database.
+	Following *bool     `bun:"-"`                                                           // Whether the current user is following this tag, if there is a current user. Stored per-user, not in this table.
```
NyaaaWhatsUpDoc, line 31: instead of adding a following field here, which i worry has a risk of becoming confusing if we're dealing with tags of multiple users at a time (if that's ever a situation). I wonder if instead we add `Tag *Tag` as a field to the `FollowedTag` type, and then maybe a separate typeutils.Converter function that wraps the current gtsmodel->frontend tag function that just sets the following flag? this may not be the best solution, totally open to others and maybe if @tsmethurst has some ideas they can chime in too. i just worry the following field on Tag{} could potentially cause confusion given until now it has never been a per-account structure if you get what i mean

tsmethurst: Agreed yeah; I understand why this is here, but I think this should be done in the typeconverter instead of being pinned to the model.

internal/processing/workers/surfacetimeline.go
```diff
@@ -122,16 +133,11 @@ func (s *Surface) timelineAndNotifyStatusForFollowers(
...
-		filters, err := s.State.DB.GetFiltersForAccountID(ctx, follow.AccountID)
-		if err != nil {
-			return gtserror.Newf("couldn't retrieve filters for account %s: %w", follow.AccountID, err)
-		}
-
-		mutes, err := s.State.DB.GetAccountMutes(gtscontext.SetBarebones(ctx), follow.AccountID, nil)
+		filters, mutes, err := s.getFiltersAndMutes(ctx, follow.AccountID)
 		if err != nil {
-			return gtserror.Newf("couldn't retrieve mutes for account %s: %w", follow.AccountID, err)
+			errs.Append(err)
```
NyaaaWhatsUpDoc, line 138: i think it's worth adding some context to the error here i.e. `errs.Appendf(...)`

internal/processing/workers/surfacetimeline.go
```diff
@@ -391,6 +414,133 @@ func (s *Surface) timelineStatus(
...
+		// Unwrap boost and work with the original status.
+		status = status.BoostOf
+	}
+
+	// Insert the status into the home timeline of each tag follower.
+	errs := gtserror.MultiError{}
+	for _, tagFollowerAccount := range tagFollowerAccounts {
+		filters, mutes, err := s.getFiltersAndMutes(ctx, tagFollowerAccount.ID)
+		if err != nil {
+			errs.Append(err)
```
NyaaaWhatsUpDoc, line 440: same goes for the error here

internal/gtsmodel/tag.go
VyrCossont: Works for me. Let's do a `FollowingTag` struct that embeds the `Tag` model.

internal/processing/tags/get.go
VyrCossont: Ah, I keep forgetting this bit of house style, thanks for the reminder.

internal/processing/workers/surfacetimeline.go
```diff
@@ -596,23 +753,66 @@ func (s *Surface) listTimelineStatusUpdateForFollow(
...
+		// Unwrap boost and work with the original status.
+		status = status.BoostOf
+	}
+
+	// Stream the update to the home timeline of each tag follower.
+	errs := gtserror.MultiError{}
+	for _, tagFollowerAccount := range tagFollowerAccounts {
+		filters, mutes, err := s.getFiltersAndMutes(ctx, tagFollowerAccount.ID)
+		if err != nil {
+			errs.Append(err)
```
NyaaaWhatsUpDoc, line 802: same for this error

internal/processing/workers/surfacetimeline.go
```diff
@@ -596,23 +753,66 @@ func (s *Surface) listTimelineStatusUpdateForFollow(
...
+
+		if _, err := s.timelineStreamStatusUpdate(
+			ctx,
+			tagFollowerAccount,
+			status,
+			stream.TimelineHome,
+			filters,
+			mutes,
+		); err != nil {
+			errs.Append(err)
```
NyaaaWhatsUpDoc, line 814: same for this error

internal/processing/tags/unfollow.go
```diff
@@ -0,0 +1,60 @@
...
+	tag, err := p.state.DB.GetTagByName(ctx, name)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return nil, gtserror.NewErrorInternalError(
+			gtserror.Newf("DB error getting tag with name %s: %w", name, err),
+		)
+	}
+
+	if tag == nil {
+		// There is no need to create a tag just to unfollow it.
+		tag = &gtsmodel.Tag{Name: name}
```
NyaaaWhatsUpDoc, line 48: do we not want to return a 404 here for tags that don't exist?

internal/processing/workers/surfacetimeline.go
```diff
@@ -495,16 +653,11 @@ func (s *Surface) timelineStatusUpdateForFollowers(
...
-		filters, err := s.State.DB.GetFiltersForAccountID(ctx, follow.AccountID)
-		if err != nil {
-			return gtserror.Newf("couldn't retrieve filters for account %s: %w", follow.AccountID, err)
-		}
-
-		mutes, err := s.State.DB.GetAccountMutes(gtscontext.SetBarebones(ctx), follow.AccountID, nil)
+		filters, mutes, err := s.getFiltersAndMutes(ctx, follow.AccountID)
 		if err != nil {
-			return gtserror.Newf("couldn't retrieve mutes for account %s: %w", follow.AccountID, err)
+			errs.Append(err)
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: scratch that just realized it's returning the wrapped error info from getFiltersAndMutes(), ignore this :)

internal/processing/workers/surfacetimeline.go
NyaaaWhatsUpDoc: scratch that just realized it's returning the wrapped error info from getFiltersAndMutes(), ignore this :)

internal/processing/workers/surfacetimeline.go
NyaaaWhatsUpDoc: scratch that just realized it's returning the wrapped error info from getFiltersAndMutes(), ignore this :)

internal/processing/workers/surfacetimeline.go
NyaaaWhatsUpDoc: scratch that just realized it's returning the wrapped error info from getFiltersAndMutes(), ignore this :)

internal/processing/workers/surfacetimeline.go
```diff
@@ -391,6 +414,133 @@ func (s *Surface) timelineStatus(
...
+			ctx,
+			s.State.Timelines.Home.IngestOne,
+			tagFollowerAccount.ID, // home timelines are keyed by account ID
+			tagFollowerAccount,
+			status,
+			stream.TimelineHome,
+			filters,
+			mutes,
+		); err != nil {
+			errs.Append(err)
```
NyaaaWhatsUpDoc, line 454: i think would be useful to add some error context here e.g. with `errs.Appendf()`

internal/processing/tags/get.go
NyaaaWhatsUpDoc: i haven't heard the term house style before but that's a good way of putting it :grin:

internal/gtsmodel/tag.go
VyrCossont: Actually, we don't even need that; `Tag`s with `Following` info attached are very short-lived. `TagToAPITag` is used so rarely that I'm just going to add an optional `following *bool` param to it.

internal/processing/tags/unfollow.go
VyrCossont: The Mastodon API we're copying specifies `follow` and `unfollow` as idempotent, so we can't do that.

VyrCossont: Sorry, read that completely wrong. Yeah we could 404 here.

NyaaaWhatsUpDoc:
looks good to me! great work again :sunglasses:

PR #3147, opened 2024-07-28
[bugfix] take into account rotation when generating thumbnail

NyaaaWhatsUpDoc:
# Description

- takes into account rotation data when generating output thumbnail
- simplifies ffprobe output to only show the fields we actually need

## Checklist

### DIFF
--- a/internal/media/ffmpeg.go
+++ b/internal/media/ffmpeg.go
@@ -535,7 +535,6 @@ type ffprobePacketOrFrame struct {
 
 type ffprobeSideData struct {
 	Rotation float64 `json:"rotation"`
-	// + unused fields.
 }
 
 type ffprobeStream struct {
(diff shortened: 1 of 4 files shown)


internal/media/ffmpeg.go
```diff
@@ -456,15 +480,62 @@ func (res *ffprobeResult) Process() (*result, error) {
 		}
 	}
 
+	// Check extra packet / frame information
+	// for provided orientation (not always set).
+	for _, pf := range res.PacketsAndFrames {
```
NyaaaWhatsUpDoc, line 485: this is specifically where we pull out the rotation value, simplified to be +- 0,90,180,270

internal/media/util.go
```diff
@@ -37,12 +37,23 @@ import (
...
 		maxThumbWidth  = 512
 		maxThumbHeight = 512
 	)
+
+	// If image is rotated by
+	// any odd multiples of 90,
+	// flip width / height to
+	// get the correct scale.
+	switch rotation {
+	case -90, 90, -270, 270:
```
NyaaaWhatsUpDoc, line 52: here is where we apply the correct scaling depending on rotation

tsmethurst:
Nice one!

PR #3161, opened 2024-08-01
[chore] add back exif-terminator and use only for jpeg,png,webp

NyaaaWhatsUpDoc:
# Description

So it turns out that `ffmpeg -map_metadata -1` can indeed clean all metadata, BUT if you're not performing any encoding then the most metadata it can clean is at the global level. Any stream metadata (which is often the case with EXIF) cannot be removed since it's not stepping through the stream data.

This PR adds back `exif-terminator` to be used exclusively for "jpeg", "png", "webp". All other filetypes fall back to best-effort exif data removal using ffmpeg.

- [x] waiting on merge + release of: https://codeberg.org/superseriousbusiness/exif-terminator/pulls/11

## Checklist


internal/media/ffmpeg.go
```diff
@@ -390,18 +373,33 @@ func (res *result) GetFileType() (gtsmodel.FileType, string) {
...
 				framerate = fr
 			}
 		}
 	}
+
+	// If image is rotated by
+	// any odd multiples of 90,
+	// flip width / height to
+	// get the correct scale.
+	switch res.rotation {
```
NyaaaWhatsUpDoc, line 398: this moves this here, as the rotation tag may not apply to every single stream individually afaik, as it could technically be a tag attached to multiple streams, so the best guess we can do is apply it to the end result

tsmethurst:
Took the liberty of fixing the test failure :)

NyaaaWhatsUpDoc:
@tsmethurst  you're an angel :O

tsmethurst:
i can be ur angle... or ur deville

NyaaaWhatsUpDoc:
https://www.youtube.com/watch?v=oz4MDhBjsFQ

PR #3169, opened 2024-08-03
[bugfix] send back Sec-Websocket-Protocol header for streaming WebSocket

leahneukirchen:
# Description

Chrome expects the selected Sec-Websocket-Protocol to be sent back on the WebSocket upgrade request (RFC6455 1.9).

This makes GTS live update on e.g. elk.zone with a Chrome-based browser.

Solution adapted from https://stackoverflow.com/questions/58417479/sec-websocket-protocol-issues

Fixes #2427.

## Checklist

### DIFF
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -152,15 +152,21 @@ import (
 //			description: bad request
 func (m *Module) StreamGETHandler(c *gin.Context) {
 	var (
-		account     *gtsmodel.Account
-		errWithCode gtserror.WithCode
+		token         string
+		tokenInHeader bool
+		account       *gtsmodel.Account
+		errWithCode   gtserror.WithCode
 	)
 
-	// Try query param access token.
-	token := c.Query(AccessTokenQueryKey)
-	if token == "" {
-		// Try fallback HTTP header provided token.
-		token = c.GetHeader(AccessTokenHeader)
+	if t := c.Query(AccessTokenQueryKey); t != "" {
+		// Token was provided as
+		// query param, no problem.
+		token = t
+	} else if t := c.GetHeader(AccessTokenHeader); t != "" {
+		// Token was provided as
+		// header, no problem.
+		token = t
+		tokenInHeader = true
 	}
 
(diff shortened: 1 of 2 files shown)


internal/api/client/streaming/stream.go
```diff
@@ -231,6 +232,12 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
 	// If the upgrade fails, then Upgrade replies to the client
 	// with an HTTP error response.
 	wsConn, err := m.wsUpgrade.Upgrade(c.Writer, c.Request, nil)
+	h := http.Header{}
+	if c.GetHeader(AccessTokenHeader) != "" {
+		// Send it back, else Chrome fails to connect
+		h.Set(AccessTokenHeader, token)
+	}
+	wsConn, err := m.wsUpgrade.Upgrade(c.Writer, c.Request, h)
```
NyaaaWhatsUpDoc, line 240: we could probably just simplify this here to instead be:
```
wsConn, err := m.wsUpgrade.Upgrade(c.Writer, c.Request, http.Header{
    AccessTokenHeader: []string{token},
})
```

if the token was never set then it'll just be an empty value (which i'm fairly certain net/http won't end up serializing), and if it was set as a query token hopefully this will encourage the client to instead use the header instead

tsmethurst:
Do you mind if I push a couple small changes to this branch? Then it can be merged I think :)

leahneukirchen:
Yes, feel free!

internal/api/client/streaming/stream.go
tsmethurst: Tbh I don't think we should encourage clients to use this header. The only reason we accept it is because (iirc) Mastodon also uses it as a workaround to avoid having clients pass tokens in query strings. But it's not designed for passing tokens at all, it's meant to indicate sub-protocols: https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#sec-websocket-protocol. Ideally we could remove it altogether, but since some clients now use it, we're more or less stuck with it :P

PR #3173, opened 2024-08-05
[chore/performance] Reuse existing pix_fmt where possible when encoding to webp thumb

tsmethurst:
# Description

>

Update our decoding to read pix_fmt into the result, and reuse that same pix_fmt where possible to avoid expensive filter operations.

## Checklist


internal/media/ffmpeg.go
```diff
@@ -274,7 +281,8 @@ type result struct {
 }
 
 type stream struct {
-	codec string
+	codec  string
+	pixFmt string
```
NyaaaWhatsUpDoc, line 285: pix_fmt isn't useful for both audio and video streams, it would be worth moving it in the videoStream type

NyaaaWhatsUpDoc:
though before merging this as i said it's worth seeing if this actually does effect memory usage or not. all we have are 2 different 4 hour windows, one during peak and one during minimum usage

internal/media/ffmpeg.go
```diff
@@ -266,6 +272,7 @@ func ffprobe(ctx context.Context, filepath string) (*result, error) {
 // data in a more useful data format.
 type result struct {
 	format   string
+	pixFmt   string
```
NyaaaWhatsUpDoc, line 275: it's not worth storing this in the result given this is a per-stream thing.

internal/media/processingmedia.go
```diff
@@ -229,10 +229,23 @@ func (p *ProcessingMedia) store(ctx context.Context) error {
 		p.media.FileMeta.Small.Size = (thumbWidth * thumbHeight)
 		p.media.FileMeta.Small.Aspect = aspect
 
-		// Generate a thumbnail image from input image path.
+		// Generate a thumbnail image from input image path,
+		// reusing existing pixel format for types that likely
+		// don't have transparency, or using yuva420p for types
+		// that can (to preserve alpha layer).
+		var pixFmt string
+		if ext == "jpeg" ||
```
NyaaaWhatsUpDoc, line 237: it might be simpler here to just do` pixFmt = result.PixFmt(); !strings.Contains(pixFmt, "a")` { pixFmt = "yuva420p" }` to ensure it's always a pixel format with transparency layer

NyaaaWhatsUpDoc: (this is based on that link you sent from stackoverflow in the matrix, just looking for a pixel format that mentions an alpha layer)

internal/media/ffmpeg.go
NyaaaWhatsUpDoc: if anything i'd just add a `func (r *result) PixFmt() string` function that returns the first non-empty stream pixel format

internal/media/ffmpeg.go
```diff
@@ -411,6 +411,84 @@ func (res *result) ImageMeta() (width int, height int, framerate float32) {
...
+// PixelFormat returns the pix_fmt value of the
+// first video stream of res, or an empty string.
+func (res *result) PixelFormat() string {
+	if len(res.video) == 0 {
+		return ""
+	}
+
+	// Return pixel format
+	// of first video stream.
+	return res.video[0].pixFmt
```
NyaaaWhatsUpDoc, line 423: i'd maybe update this to be:
```golang
for _, stream := range res.video {
    if stream.pixFmt != "" {
        return stream.pixFmt
    }
}
```

this way we return the first valid pixel format

tsmethurst:
Gonna close this for now, as based on some discussion in Matrix we may handle thumbnailing a bit differently, making this PR not relevant for now.

PR #3183, opened 2024-08-08
[performance] move thumbnail generation to go code where possible

NyaaaWhatsUpDoc:
# Description

While ffmpeg is great for thumbnail generation, and we do like using webps (pee hehe), it comes with a lot of extra overhead. This updates our media processing code to use ffmpeg as a fallback, which ends up being with the majority of media types, but primarily resorting to our own Go thumbnail generation code if possible. this results in JPEG, PNG, GIF, Webps without transparency layers being thumbnailed using Go code, and all else resorting to ffmpeg. unfortunately there is no Go support for Webp encoding, so those thumbnailed with Go code are encoded to JPEGs, hence the restriction on transparency layers.

This also updates ffmpeg to maintain the pixel colorspace where possible (credit largely to @tsmethurst for this, even if the investigation ended a bit oddly :smiling_face_with_tear:), updates our metadata cleaning to fallback to ffmpeg when exif-terminator fails (thank you @tsmethurst), updates our handling of image rotation to instead use the "orientation" flag which is both much more flexible (supporting mirroring) and much easier to work with, reduces the number of frames we page through with the ffmpeg thumbnail filter (which reduces memory usage, marginally), and finally updates our thumbnail quality setting to the defaults for both webp / jpeg (75%) which looks much nicer on the frontend while not causing much of a size increase (once again, thank you @tsmethurst)

This PR is a result of multiple days of digging and experimentation by both myself and @tsmethurst, this absolutely was a joint effort :p

## Checklist


internal/media/ffmpeg.go
```diff
@@ -99,34 +86,33 @@ func ffmpegGenerateThumb(ctx context.Context, filepath string, width, height int
 
 		// Select thumb from first 10 frames
```
tsmethurst, line 87: This comment can be changed now. Also, 7 was shown to reduce memory usage a bit right? Maybe it's worth commenting on that.

internal/media/ffmpeg.go
```diff
@@ -99,34 +86,33 @@ func ffmpegGenerateThumb(ctx context.Context, filepath string, width, height int
...
-		"-filter:v", "thumbnail=n=10,"+
+		"-filter:v", "thumbnail=n=7,"+
 
 			// scale to dimensions
 			// (scale filter: https://ffmpeg.org/ffmpeg-filters.html#scale)
-			"scale="+scale+","+
+			"scale="+strconv.Itoa(width)+
+			":"+strconv.Itoa(height)+","+
 
 			// YUVA 4:2:0 pixel format
```
tsmethurst, line 96: This comment can change too.

internal/media/thumbnail.go
```diff
@@ -0,0 +1,379 @@
...
+
+	// Done with file.
+	_ = infile.Close()
+
+	if err != nil {
+		return "", gtserror.Newf("error decoding file %s: %w", inpath, err)
+	}
+
+	// Apply orientation
+	// BEFORE any resize.
```
tsmethurst, line 213: What's the reason for this btw? Cheaper?

internal/media/thumbnail.go
```diff
@@ -0,0 +1,379 @@
...
+		return "", gtserror.Newf("error generating blurhash: %w", err)
+	}
+
+	return blurhash, nil
+}
+
+// List of pixel formats that have an alpha layer.
+// Derived from the following very messy command:
+//
+//	for res in $(ffprobe -show_entries pixel_format=name:flags=alpha | grep -B1 alpha=1 | grep name); do echo $res | sed 's/name=//g' | sed 's/^/"/g' | sed 's/$/",/g'; done
```
tsmethurst, line 315: love that we're keeping this horror show :D

tsmethurst:
Looks good :) :) :)

internal/media/thumbnail.go
NyaaaWhatsUpDoc: it's just a case of our image dimensions already taking orientation into account, so we need to reorient the image first before applying the new dimensions :)

NyaaaWhatsUpDoc: i added a code comment to make it a bit clearer too

PR #3186, opened 2024-08-09
[feature] idempotency header middleware

### LINKED ISSUE #2723: [feature] Implement `Idempotency-Key` for POSTing statuses

As a user on a phone with a bad connection, I want to avoid both losing my post if my network connection drops, and double-posting if it's not clear if my post went through.

As a mobile client dev, I want to be able to retry creating a status if I get an error indicating that it might have failed.

Implement the [`Idempotency-Key`](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/) header for the [`POST /api/v1/statuses` endpoint](https://docs.joinmastodon.org/methods/statuses/#headers). This header can be optionally set by clients to a UUID, ULID, or other short random string, and is used to de-duplicate operations requested with the same idempotency key in a given time window, an hour in Mastodon's case.

For complete Mastodon API compatibility, if we ever implement scheduled statuses, this would also involve adding an [`idempotency`](https://docs.joinmastodon.org/entities/ScheduledStatus/#params-idempotency) field to the `ScheduledStatus` type returned from [scheduled status APIs](https://docs.joinmastodon.org/methods/scheduled_statuses/).

Storing the status text, CW, poll, media attachment IDs, etc. client-side, retrieving the user's recent statuses after a potentially failed POST (and possibly a short delay), and checking for duplicates. This requires extra network traffic and extra work for the client, but doesn't require any server-side changes.

### Additional context.

Mastodon only implements it for creating statuses, but the same mechanism could be used to make any API call idempotent.

It could be implemented as an in-memory per-user set of idempotency keys, as an extra column on the `statuses` table, or as a separate table. Some databases have performance issues with tables that are frequently appended to and deleted from; partitioning such a table and dropping old partitions perodically can mitigate that problem, removing the need to vacuum in PG.

NyaaaWhatsUpDoc:
# Description

Adds a piece of middleware supporting the idempotency key. Though probably worth you looking through this PR first before we do anything with it @VyrCossont as you have more experience working with this client-side. I specifically wanted to avoid caching anything more than the status code for each response, as it would either require buffering and caching of response bodies which could end up quite memory intensive, or it will require a unique solution for every single endpoint which i'm also really not a fan of. Is responding with the previous status code + a "this already happened" http response body good enough for clients to work with? :thinking: 

closes #2723

## Checklist

### DIFF
--- a/cmd/gotosocial/action/server/server.go
+++ b/cmd/gotosocial/action/server/server.go
@@ -441,17 +441,20 @@ var Start action.GTSAction = func(ctx context.Context) error {
 
 	gzip := middleware.Gzip() // applied to all except fileserver
 
+	// Idempotency applied only to client / AP.
+	idempotency := middleware.Idempotency()
+
 	// these should be routed in order;
 	// apply throttling *after* rate limiting
 	authModule.Route(route, clLimit, clThrottle, gzip)
-	clientModule.Route(route, clLimit, clThrottle, gzip)
+	clientModule.Route(route, clLimit, clThrottle, idempotency, gzip)
 	metricsModule.Route(route, clLimit, clThrottle, gzip)
 	healthModule.Route(route, clLimit, clThrottle)
 	fileserverModule.Route(route, fsMainLimit, fsThrottle)
 	fileserverModule.RouteEmojis(route, instanceAccount.ID, fsEmojiLimit, fsThrottle)
 	wellKnownModule.Route(route, gzip, s2sLimit, s2sThrottle)
 	nodeInfoModule.Route(route, s2sLimit, s2sThrottle, gzip)
-	activityPubModule.Route(route, s2sLimit, s2sThrottle, gzip)
+	activityPubModule.Route(route, s2sLimit, s2sThrottle, idempotency, gzip)
 	activityPubModule.RoutePublicKey(route, s2sLimit, pkThrottle, gzip)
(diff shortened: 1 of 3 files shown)


VyrCossont:
So unfortunately (and this is on me, I didn't fully investigate in the feature request) we do need to store the original response data or have some way of fetching it on demand. If we send the previous response code and the "already happened" JSON message instead of response data in the expected format (or an error code and response), no existing client is going to understand what to do with that data. If we send data in the expected format, there's no such issue.

The Mastodon implementation is status-specific and makes a DB query for whatever status or scheduled status was created with that idempotency key. We could do the same (requiring specific support for each idempotency-key-enabled endpoint to cache the relevant model type and ID for each idempotency key for future lookup), or we could cache the entire response data as well as the response code.

I think the scenario where a client makes a request with an idempotency key *and* it has already made that request before will be rare, compared to the case where a client makes a request with an idempotency key and it succeeds and is not repeated, so it would make more sense to *not* cache the full response data, just enough info that we can look it up later if needed.

NyaaaWhatsUpDoc:
hmmm i shall have a think about how best to do this :thinking:

daenney:
We've migrated to Codeberg. Your PR has been imported as: https://codeberg.org/superseriousbusiness/gotosocial/pulls/3186.

PR #3198, opened 2024-08-13
[security] harden account update logic

NyaaaWhatsUpDoc:
# Description

A very quick change. It just ensures that on account refresh that the public keys do not change. It ensures that we don't accidentally end up changing the account's public key outside of `AuthenticateFederatedRequest()`.

Though exploring this in more detail, and filling my head once more with the intricasies of AP + http signatures has reminded me how the current implementation only protects against unauthorized GET requests. If a server doesn't sever ties before it goes dark, the domain lapses and someone else sets up a server on that domain, if they set-up a dummy account model that mirrors exactly what we expect (which is possible since there's no signature attached to an account model to verify that the GET request was *responded to* by the expected party), then we may end up sending them followers-only posts, DMs etc without knowing that its actually a compromised server. Though they won't be able to successfully fetch anything, since they won't be able to correctly sign an HTTP signature header.

All of the above, while unfortunate, does make sense given it's called "authorized fetch".

## Checklist

### DIFF
--- a/internal/federation/dereferencing/authenticate.go
+++ b/internal/federation/dereferencing/authenticate.go
@@ -25,7 +25,15 @@ import (
 
 // verifyAccountKeysOnUpdate verifies that account's public key hasn't changed on update from
 // our existing stored representation, UNLESS the key has been explicitly expired (i.e. key rotation).
-func verifyAccountKeysOnUpdate(existing, latest *gtsmodel.Account, now time.Time) bool {
+func verifyAccountKeysOnUpdate(existing, latest *gtsmodel.Account, now time.Time, federated bool) bool {
+	if federated {
+		// If this data was federated
+		// to us then we implicitly trust
+		// it on the grounds that it
+		// passed any signature checks.
+		return true
+	}
+
 	if existing.PublicKey == nil {
 		// New account which has been
 		// passed as a placeholder.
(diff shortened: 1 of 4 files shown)


internal/federation/dereferencing/authenticate.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"time"
+
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+)
+
+// verifyAccountKeysOnUpdate verifies that account's public key hasn't changed on update from
+// our existing stored representation, UNLESS the key has been explicitly expired (i.e. key rotation).
+func verifyAccountKeysOnUpdate(existing, latest *gtsmodel.Account, now time.Time, federated bool) bool {
```
NyaaaWhatsUpDoc, line 28: i put this in a new file "authenticate.go" as my plan is in time to move some of the dereferencing logic we end up performing in the federatingDB for `AuthenticateFederatedRequest()` into the dereferencer. this might then open us up for performing some of the httpsig authentication flow in a more flexible manner in other places too

tsmethurst:
These changes look reasonable to me! I'll add the docs in a separate PR, so if you're happy with this feel free to squerge imo.

PR #3215, opened 2024-08-19
[feature] Interaction requests client api + settings panel

tsmethurst:
# Description

>

This pull request adds review and rejection / approval of interaction requests via the client API and settings panel.

Screenshot from 2024-08-19 14-32-41

Screenshot from 2024-08-19 14-32-51

To support this, requests and rejections are now stored in the database in addition to approvals, as that was just the easiest way of making it possible to page through requests (and, in future, page through approvals and rejections).

The structure of the approvals table has also changed. To make things a bit easier for myself, the current table is just dropped and recreated with the new schema, which means existing approvals will be lost. Since approvals were not in a release yet, I don't think this is a big deal, since only people running snapshots had interaction policies up to now (and therefore stored Accepts), but I welcome feedback on this and can make more of an effort to preserve existing Accepts if we decide it's necessary to do this.

~~Still draft because I'd still like to do the following:~~

- ~~[ ] Clean up any lingering unique indexes from dropped table (if db type = postgres)~~ Not necessary.
- [x] Add some more tests.
- ~~[ ] Have the interaction filter check if a previous InteractionRejection exists for a fave or boost of a status by a given account, and automatically reject reattempts (so people can't spam interaction requests at you).~~ I'll do this in a separate PR + it might not be necessary or desirable anyway, not sure yet.

Part of #3057 

## Checklist

- [ ] I/we have made any necessary changes to documentation. -- will do this in a separate PR


internal/api/client/interactionrequests/authorize.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/gin-gonic/gin"
+	apiutil "github.com/superseriousbusiness/gotosocial/internal/api/util"
+	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
+	"github.com/superseriousbusiness/gotosocial/internal/oauth"
+)
+
+// InteractionRequestAuthorizePOSTHandler swagger:operation POST /api/v1/interaction_requests/{id}/authorize authorizeInteractionRequest
+//
+// Accept/authorize/approve an interaction request with the given ID.
```
tsmethurst, line 31: Used the same verbiage here (authorize) as we use for follow requests, for consistency.

testrig/teststructs.go
```diff
@@ -0,0 +1,121 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package testrig
```
tsmethurst, line 18: Moved this stuff to a separate package to avoid repetition + make our lives a bit easier when testing things that rely on workers for side effects.

internal/processing/workers/workers_test.go
```diff
@@ -132,63 +114,3 @@ func (suite *WorkersTestSuite) openStreams(ctx context.Context, processor *proce
 
 	return streams
 }
-
-func (suite *WorkersTestSuite) SetupTestStructs() *TestStructs {
```
tsmethurst, line 136: Moved this to the testrig as it's useful elsewhere.

internal/db/bundb/timeline.go
```diff
@@ -137,6 +124,22 @@ func (t *timelineDB) GetHomeTimeline(ctx context.Context, accountID string, maxI
 		bun.In(targetAccountIDs),
 	)
 
+	// Only include statuses that aren't pending approval.
+	q = q.Where("NOT ? = ?", bun.Ident("status.pending_approval"), true)
```
tsmethurst, line 128: Added this clause here and elsewhere in this package so that timelines don't have pending statuses put in them. Let me know if you'd prefer this in a separate PR.

internal/db/bundb/instance.go
```diff
@@ -76,6 +76,9 @@ func (i *instanceDB) CountInstanceStatuses(ctx context.Context, domain string) (
 			Where("? = ?", bun.Ident("account.domain"), domain)
 	}
 
+	// Ignore statuses that are currently pending approval.
+	q = q.Where("NOT ? = ?", bun.Ident("status.pending_approval"), true)
```
tsmethurst, line 80: See https://github.com/superseriousbusiness/gotosocial/pull/3215/files#r1723451114, another thing I can put in a separate PR if you prefer.

internal/db/bundb/account.go
```diff
@@ -1285,34 +1285,40 @@ func (a *accountDB) RegenerateAccountStats(ctx context.Context, account *gtsmode
 	if err := a.db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
 		var err error
 
-		// Scan database for account statuses.
+		// Scan database for account statuses, ignoring
+		// statuses that are currently pending approval.
 		statusesCount, err := tx.NewSelect().
-			Table("statuses").
-			Where("? = ?", bun.Ident("account_id"), account.ID).
+			TableExpr("? AS ?", bun.Ident("statuses"), bun.Ident("status")).
```
tsmethurst, line 1291: See https://github.com/superseriousbusiness/gotosocial/pull/3215/files#r1723451114, another thing I can put in a separate PR if you prefer.

internal/db/bundb/migrations/20240716151327_interaction_policy.go
```diff
@@ -1,71 +0,0 @@
...
-package migrations
-
-import (
-	"context"
-
-	gtsmodel "github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
-	"github.com/uptrace/bun"
-)
-
-func init() {
```
tsmethurst, line 27: Nuked this migration since the one i added replaces everything in it.

internal/db/bundb/migrations/20240809134448_interaction_requests_client_api.go
```diff
@@ -0,0 +1,207 @@
...
+				IfNotExists().
+				Exec(ctx); err != nil {
+				return err
+			}
+
+			for idx, col := range map[string]string{
+				"interaction_rejections_status_id_idx":              "status_id",
+				"interaction_rejections_account_id_idx":             "account_id",
+				"interaction_rejections_interacting_account_id_idx": "interacting_account_id",
+				"interaction_rejections_interaction_uri_idx":        "interaction_uri",
```
tsmethurst, line 181: Please note that unlike with the other interaction models,  `*gtsmodel.InteractionRejection.InteractionURI` is NOT unique, and so an index needs to be created for it. It's possible that more than one person might want to reject an interaction.

internal/db/bundb/interaction_test.go
```diff
@@ -0,0 +1,264 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/paging"
+	"github.com/superseriousbusiness/gotosocial/internal/typeutils"
+	"github.com/superseriousbusiness/gotosocial/internal/util"
+)
+
+type InteractionTestSuite struct {
+	BunDBStandardTestSuite
+}
+
+func (suite *InteractionTestSuite) markIntsPending(
```
NyaaaWhatsUpDoc, line 38: i think it might read a bit better if this was just "markPending". on first read i was wondering where the integers were :p

internal/processing/workers/util.go
```diff
@@ -488,128 +487,143 @@ func (u *utils) decrementFollowRequestsCount(
 	return nil
 }
 
-// approveFave stores + returns an
-// interactionApproval for a fave.
-func (u *utils) approveFave(
+// pendFave stores an interaction request for the given
+// fave, and notifies the interactee of a pending fave.
+func (u *utils) pendFave(
```
NyaaaWhatsUpDoc, line 492: i think these might read a bit better if they were `setPendingFave()`, `setPendingReply()` etc. what do you think?

internal/processing/workers/fromclientapi.go
```diff
@@ -1050,16 +1053,206 @@ func (p *clientAPI) RejectUser(ctx context.Context, cMsg *messages.FromClientAPI
...
+	if !ok {
+		return gtserror.Newf("%T not parseable as *gtsmodel.InteractionRejection", cMsg.GTSModel)
+	}
+
+	// At this point the InteractionRejection should already
+	// be in the database, we just need to do side effects.
+
+	// Send out the rejection as Reject.
+	if err := p.federate.RejectInteraction(ctx, rejection); err != nil {
+		return gtserror.Newf("error federating rejection of reply: %w", err)
```
NyaaaWhatsUpDoc, line 1197: i would instead just make this a `log.Error(...)` to ensure we always end up reaching the stages below to wipe the status.

internal/processing/workers/fromclientapi.go
```diff
@@ -1050,16 +1053,206 @@ func (p *clientAPI) RejectUser(ctx context.Context, cMsg *messages.FromClientAPI
...
+	if !ok {
+		return gtserror.Newf("%T not parseable as *gtsmodel.InteractionRejection", cMsg.GTSModel)
+	}
+
+	// At this point the InteractionRejection should already
+	// be in the database, we just need to do side effects.
+
+	// Send out the rejection as Reject.
+	if err := p.federate.RejectInteraction(ctx, rejection); err != nil {
+		return gtserror.Newf("error federating rejection of like: %w", err)
```
NyaaaWhatsUpDoc, line 1160: i would instead just make this a `log.Error(...)` to ensure we always end up reaching the stages below to delete the fave.

internal/processing/workers/fromclientapi.go
```diff
@@ -1050,16 +1053,206 @@ func (p *clientAPI) RejectUser(ctx context.Context, cMsg *messages.FromClientAPI
...
+	if err := p.federate.RejectInteraction(ctx, rejection); err != nil {
+		return gtserror.Newf("error federating rejection of reply: %w", err)
+	}
+
+	// Get the rejected status.
+	status, err := p.state.DB.GetStatusByURI(
+		gtscontext.SetBarebones(ctx),
+		rejection.InteractionURI,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
```
NyaaaWhatsUpDoc, line 1205: i think it's probably worth *not* checking for errnoentries here. in this case if the status is already deleted it actually is an error we want to return with.

internal/processing/workers/fromclientapi.go
```diff
@@ -1050,16 +1053,206 @@ func (p *clientAPI) RejectUser(ctx context.Context, cMsg *messages.FromClientAPI
...
+		return gtserror.Newf("error federating rejection of announce: %w", err)
+	}
+
+	// Get the rejected boost.
+	boost, err := p.state.DB.GetStatusByURI(
+		gtscontext.SetBarebones(ctx),
+		rejection.InteractionURI,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		return gtserror.Newf("db error getting rejected announce: %w", err)
```
NyaaaWhatsUpDoc, line 1243: i think it's probably worth *not* checking for errnoentries here. in this case if the status is already deleted it actually is an error we want to return with.

NyaaaWhatsUpDoc: (also simpler code)

internal/processing/workers/fromclientapi.go
NyaaaWhatsUpDoc: (also simpler code)

internal/processing/workers/fromclientapi.go
```diff
@@ -1050,16 +1053,206 @@ func (p *clientAPI) RejectUser(ctx context.Context, cMsg *messages.FromClientAPI
...
+	if err := p.federate.RejectInteraction(ctx, rejection); err != nil {
+		return gtserror.Newf("error federating rejection of like: %w", err)
+	}
+
+	// Get the rejected fave.
+	fave, err := p.state.DB.GetStatusFaveByURI(
+		gtscontext.SetBarebones(ctx),
+		rejection.InteractionURI,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
```
NyaaaWhatsUpDoc, line 1168: i think it's probably worth not checking for errnoentries here. in this case if the status is already deleted it actually is an error we want to return with. (also simpler code)

internal/processing/workers/fromclientapi.go
```diff
@@ -1050,16 +1053,206 @@ func (p *clientAPI) RejectUser(ctx context.Context, cMsg *messages.FromClientAPI
...
+	)
+
+	// Update stats for the reply author account.
+	if err := p.utils.incrementStatusesCount(ctx, interactingAcct, reply); err != nil {
+		log.Errorf(ctx, "error updating account stats: %v", err)
+	}
+
+	// Interaction counts changed on the replied status;
+	// uncache the prepared version from all timelines.
+	p.surface.invalidateStatusFromTimelines(ctx, reply.InReplyToID)
```
NyaaaWhatsUpDoc, line 1096: does it matter that this `invalidateStatusFromTimelines()` call here relative to the other similar calls in both `AcceptAnnounce()` and `AcceptReply()` are in different locations?

NyaaaWhatsUpDoc: and it's probably worth standardising the behaviour in all these functions e.g. invalidate steps always go *after* <some particular logic>. if you get what i mean :)

internal/processing/workers/fromfediapi.go
```diff
@@ -228,24 +230,24 @@ func (p *fediAPI) CreateStatus(ctx context.Context, fMsg *messages.FromFediAPI)
 		return nil
 	}
 
+	// We're going to possibly modify
+	// the status further, lock on its URI.
+	unlock := p.state.ProcessingLocks.Lock(status.URI)
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: also given we're doing these in the `pend___()` functions i think we're doubling up on locks here

internal/processing/interactionrequests/reject.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/messages"
+)
+
+// Reject rejects an interaction request with the given ID,
+// on behalf of the requester (whose post it must target).
+func (p *Processor) Reject(
+	ctx context.Context,
+	requester *gtsmodel.Account,
+	intReqID string,
```
NyaaaWhatsUpDoc, line 37: i think this would be a bit easier to read if it was just `requestID`, or `id`. `intReqID` reads like there should be an integer in there :p

internal/processing/interactionrequests/accept.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return apiApproval, nil
+}
+
+// Package-internal convenience function to accept a like.
+func (p *Processor) acceptLike(
+	ctx context.Context,
+	req *gtsmodel.InteractionRequest,
+) (*gtsmodel.InteractionApproval, gtserror.WithCode) {
+	if req.Like == nil {
```
NyaaaWhatsUpDoc, line 117: i don't think we need to check for these here. if they were nil but the ID for the object was there as it should be, then the initial GetInteractionRequest() to the database should have returned error in the final stage populating it. Adding these nil checks here makes it seem like this is an (at least somewhat) expected situation. Absolute worst-case it'll just panic on attempting to deref ptr below.

internal/processing/interactionrequests/accept.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return approval, nil
+}
+
+// Package-internal convenience function to accept a reply.
+func (p *Processor) acceptReply(
+	ctx context.Context,
+	req *gtsmodel.InteractionRequest,
+) (*gtsmodel.InteractionApproval, gtserror.WithCode) {
+	if req.Reply == nil {
```
NyaaaWhatsUpDoc, line 150: same as above re: nil check.

internal/processing/interactionrequests/accept.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return approval, nil
+}
+
+// Package-internal convenience function to accept an announce.
+func (p *Processor) acceptAnnounce(
+	ctx context.Context,
+	req *gtsmodel.InteractionRequest,
+) (*gtsmodel.InteractionApproval, gtserror.WithCode) {
+	if req.Announce == nil {
```
NyaaaWhatsUpDoc, line 183: same as above re: nil check.

internal/processing/interactionrequests/reject.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return apiRejection, nil
+}
+
+// Package-internal convenience function to reject a like.
+func (p *Processor) rejectLike(
+	ctx context.Context,
+	req *gtsmodel.InteractionRequest,
+) (*gtsmodel.InteractionRejection, gtserror.WithCode) {
+	if req.Like == nil {
```
NyaaaWhatsUpDoc, line 117: same as above re: nil check.

internal/processing/interactionrequests/reject.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return rejection, nil
+}
+
+// Package-internal convenience function to reject a reply.
+func (p *Processor) rejectReply(
+	ctx context.Context,
+	req *gtsmodel.InteractionRequest,
+) (*gtsmodel.InteractionRejection, gtserror.WithCode) {
+	if req.Reply == nil {
```
NyaaaWhatsUpDoc, line 150: same as above re: nil check

internal/processing/interactionrequests/reject.go
```diff
@@ -0,0 +1,209 @@
...
+
+	return rejection, nil
+}
+
+// Package-internal convenience function to reject an announce.
+func (p *Processor) rejectAnnounce(
+	ctx context.Context,
+	req *gtsmodel.InteractionRequest,
+) (*gtsmodel.InteractionRejection, gtserror.WithCode) {
+	if req.Announce == nil {
```
NyaaaWhatsUpDoc, line 183: same as above re: nil check.

internal/processing/interactionrequests/accept.go
NyaaaWhatsUpDoc: also if that does happen, it could potentially be a serious issue. a not-found is likely to end up getting masked and we won't end up dealing with it, whereas a panic and unexpected 500 definitely does get our attention :p

internal/api/client/interactionrequests/authorize.go
NyaaaWhatsUpDoc: what does the authorize here mean then? i'm not actually sure what the authorize means in that context either.

internal/api/client/interactionrequests/getpage.go
```diff
@@ -0,0 +1,211 @@
...
+	includeBoosts, errWithCode := apiutil.ParseInteractionReblogs(
+		c.Query(apiutil.InteractionReblogsKey), true,
+	)
+	if errWithCode != nil {
+		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
+		return
+	}
+
+	if !includeLikes && !includeReplies && !includeBoosts {
+		const text = "at least one of favourites, replies, or boosts must be true"
```
NyaaaWhatsUpDoc, line 176: i think the api response may make a bit more sense if this was `must be set`

internal/db/bundb/account.go
NyaaaWhatsUpDoc: nah 'tis all good :+1:

internal/api/model/interaction.go
```diff
@@ -0,0 +1,82 @@
...
+	// The type of interaction that this interaction request pertains to.
+	//
+	//	`favourite` - Someone favourited a status.
+	//	`reply` - Someone replied to a status.
+	//	`reblog` - Someone reblogged / boosted a status.
+	Type string `json:"type"`
+	// The timestamp of the interaction request (ISO 8601 Datetime)
+	CreatedAt string `json:"created_at"`
+	// The account that performed the interaction.
+	InteractingAccount *Account `json:"interacting_account"`
```
NyaaaWhatsUpDoc, line 35: very subjective one here, but i think these would make more sense as just `account` and `status` respectively (same goes for the structs below). it fits a bit better with the `reply` json key imo

NyaaaWhatsUpDoc: (also makes json a little easier to read with keys that aren't too wordy)

internal/db/bundb/account.go
```diff
@@ -1285,34 +1285,40 @@ func (a *accountDB) RegenerateAccountStats(ctx context.Context, account *gtsmode
...
+		// Scan database for pinned statuses, ignoring
+		// statuses that are currently pending approval.
 		statusesPinnedCount, err := tx.NewSelect().
-			Table("statuses").
-			Where("? = ?", bun.Ident("account_id"), account.ID).
-			Where("? IS NOT NULL", bun.Ident("pinned_at")).
+			TableExpr("? AS ?", bun.Ident("statuses"), bun.Ident("status")).
+			Where("? = ?", bun.Ident("status.account_id"), account.ID).
+			Where("? IS NOT NULL", bun.Ident("status.pinned_at")).
+			Where("NOT ? = ?", bun.Ident("status.pending_approval"), true).
```
NyaaaWhatsUpDoc, line 1306: is this actually possible for a pinned status?

internal/api/client/interactionrequests/authorize.go
tsmethurst: Authorize = accept, basically. Not sure why "authorize/reject" is used in the Masto follow requests API instead of just "accept/reject", but that's life!

internal/db/bundb/migrations/20240716151327_interaction_policy.go
tsmethurst: Oh actually you know, I'm not sure. I'll just make it an empty migration then, just in case.

internal/processing/workers/util.go
tsmethurst: I'm not sure... At this point the fave (or whatever) is already marked as pending, so nothing is being set on that fave itself, this is more about handling the side effects of a pending fave (ie., putting an interaction request in the db), notifying the target. Maybe handlePendingFave is better?

internal/db/bundb/interaction.go
```diff
@@ -128,22 +142,445 @@ func (r *interactionDB) PopulateInteractionApproval(ctx context.Context, approva
 		}
 	}
 
+	switch approval.InteractionType {
+
+	case gtsmodel.InteractionLike:
+		approval.Like, err = i.state.DB.GetStatusFaveByURI(ctx, approval.InteractionURI)
+		if err != nil {
+			errs.Appendf("error populating interactionApproval Like")
```
NyaaaWhatsUpDoc, line 150: i assume you meant to wrap the error here too :p

internal/db/bundb/interaction_test.go
tsmethurst: Got ya, yeah that's a bit weird, I'll change it.

internal/db/bundb/account.go
tsmethurst: Yeah, I think you could create a reply to someone, and then try to pin your reply. Or something daft like that. I suppose we could just make it so that you can't pin pending statuses, instead.

internal/api/model/interaction.go
tsmethurst: Mmm that's what I originally had but it felt a bit weird from a client perspective so I changed it to this. I'll change it back though (it's more consistent with the Masto API anyways).

internal/db/bundb/interaction.go
tsmethurst: Yes, thanks, fixed! https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/9fb07cb443cf55e0dd4eab287f6ddfcc7b4cff8b

internal/api/client/interactionrequests/getpage.go
tsmethurst: I'd rather leave it as is, since "must be set" would also suggest you could set them all, but to false.

internal/processing/interactionrequests/accept.go
tsmethurst: Resolving + using above comment for discussion.

internal/processing/interactionrequests/accept.go
tsmethurst: Resolving + using above comment for discussion.

internal/processing/interactionrequests/reject.go
tsmethurst: Resolving + using above comment for discussion.

internal/processing/interactionrequests/reject.go
tsmethurst: Resolving + using above comment for discussion.

internal/processing/interactionrequests/reject.go
tsmethurst: Resolving + using above comment for discussion.

internal/processing/workers/fromclientapi.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/73a7845919f0d0039698c2efcd9cb0bda360cff4

internal/processing/workers/fromclientapi.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/73a7845919f0d0039698c2efcd9cb0bda360cff4

internal/processing/workers/fromclientapi.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/73a7845919f0d0039698c2efcd9cb0bda360cff4

internal/processing/interactionrequests/accept.go
tsmethurst: I updated the db functions to be OK with NotFound errors when trying to fetch a Like, Reply, or Announce to populate an interaction request. Especially now that rejections are using the same model as accepts and requests, there's a fair chance that if someone tries to fetch an interaction request from the db, and it's for something that's been rejected, the Like, Reply, or Announce won't be in the database anymore, so it is a semi-expected error. Given that, I think it's sensible to have nil checking here. I've changed it now so that an error not found is returned if the Like (or whatever) can't be found, which should be warning enough I think.

internal/processing/interactionrequests/reject.go
tsmethurst: You're right! changed: https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/5a127b64c4384793b12148c17e1d721acf8e3eff

internal/processing/workers/util.go
tsmethurst: I changed them to "requestFave" etc. I think that makes a bit more sense!

internal/processing/workers/fromclientapi.go
tsmethurst: p.federate.RejectInteraction

internal/processing/workers/fromclientapi.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/1542bfcc99a0e010749ad2e8527d79819740ba07

internal/api/model/interaction.go
tsmethurst: Done :)

internal/db/bundb/migrations/20240716151327_interaction_policy.go
tsmethurst: Done!

docs/api/swagger.yaml
```diff
@@ -7521,6 +7562,177 @@ paths:
...
+                - application/json
+            responses:
+                "200":
+                    description: The now-rejected interaction request.
+                    schema:
+                        $ref: '#/definitions/interactionRequest'
+                "400":
+                    description: bad request
+                "401":
+                    description: unrejectd
```
VyrCossont, line 7723: Typo, probably should be "unauthorized"

tsmethurst: Ah thanks, search+replace error :')

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/7f9f8c6a11c19a49c6e6574ff9c3c6827ccb5b9a

internal/processing/workers/fromfediapi.go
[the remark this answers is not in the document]
tsmethurst: Right! Changed it :') https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/30dddd22a868b5cc0c8f4dcaf00536a61a3c9b75

internal/processing/workers/fromclientapi.go
tsmethurst: I get ya! Changed it here https://github.com/superseriousbusiness/gotosocial/pull/3215/commits/e62f6d5aef37430621babfc97bce172a9c1eeb35. I think functionally it doesn't matter that much, but it is neater to have it always be the last step indeed.

Issue #1035, opened 2022-11-12
[feature] API does not expose whether a status is boostable / replyable / local-only


zxzzxv:
### Is your feature request related to a problem ?

Right now, there is no API exposed so that a user status may be queried for permissions. This is handy for clients to properly disable those options.

### Describe the solution you'd like.

The underlying data structures store this data, but code that converts between the API and the internal representation still needs to be updated.

### Describe alternatives you've considered.

None considered.

tsmethurst:
Yep good one, we need to add this indeed.

tsmethurst:
For replyable/boostable/likeable, the API has changed thanks to the concept of interaction policies introduced by https://github.com/superseriousbusiness/gotosocial/pull/3108. So that just leaves the "federated" field, which I'll add in a separate PR.

PR #3222, opened 2024-08-22
[feature] Use `local_only` field, deprecate `federated` field

tsmethurst:
# Description

>

This pull request updates the status create API and status API model to use the `local_only` field to indicate non-federated statuses. The `local_only` field is already in use by implementations such as GlitchSoc, so this should make things a little easier for client app devs. To avoid breakage, the `federated` field has been deprecated. It still works, but only if `local_only` is not set, and `local_only` takes priority.

Along the way, I also added checks for LocalOnly statuses to the visibility filter. We were already not serving unfederated statuses via the web frontend, but references to local-only posts were still being included in outboxes (not the posts themselves), and I *think* Public local-only posts were included in API responses if `instance-expose-public-timeline` was set to true. Not sure since I don't think anyone was actually using this combination of features.

Closes https://github.com/superseriousbusiness/gotosocial/issues/1035

## Checklist


internal/api/model/status.go
```diff
@@ -209,6 +211,12 @@ type StatusCreateRequest struct {
 	SpoilerText string `form:"spoiler_text" json:"spoiler_text" xml:"spoiler_text"`
 	// Visibility of the posted status.
 	Visibility Visibility `form:"visibility" json:"visibility" xml:"visibility"`
+	// Set to "true" if this status should not be federated, ie. it should be a "local only" status.
+	LocalOnly *bool `form:"local_only"`
+	// ***DEPRECATED***.
+	//
+	// Only used if LocalOnly is not set.
+	DeprecatedFederated *bool `form:"federated"`
```
NyaaaWhatsUpDoc, line 219: i think you can probably leave this as `Federated *bool` but if you prefix the comment with `// Deprecated: ` then the linter should always indicate it as deprecated to us developers.

NyaaaWhatsUpDoc: an example if you want to see:

```golang
	// your linter will strike-through
	// your access of r.Cancel, a
	// deprecated field.
	r := http.Request{}
	r.Cancel
```

NyaaaWhatsUpDoc: in action:
deprecate_go_field

tsmethurst: Oh neat!!!

Issue #2973, opened 2024-06-06
[feature] Option to disable SMTP TLS cert verificiation


tvierling:
### Is your feature request related to a problem ?

In an environment where internal container-to-container networking is used to connect to the SMTP server, the certificate being presented will be for *external* use, not within-Docker-network use. So you'll get an error such as

`Code 422 Unprocessable Entity: tls: failed to verify certificate: x509: cannot validate certificate for 172.30.42.6 because it doesn't contain any IP SANs`

Or if you use the named Docker network DNS you'll get

`Code 422 Unprocessable Entity: tls: failed to verify certificate: x509: certificate is valid for <domains>, not <smtp_host>`

The point of using internal networking is specifically to allow trusted relay on the SMTP server side, but a verified cert can't have the internal IP in its SANs for the obvious reason that the IP can't be verified by external entities.

### Describe the solution you'd like.

Need an option to turn off TLS cert validation for SMTP when using a trusted host.

Alternatively, if the smtp_host is an IP in the trusted_proxies network, that could be another way to declare the trusted nature without another option. (But this feels like overloading an option too much with hidden behavior.)

### Describe alternatives you've considered.

Hairpinning through the host's networking using socat proxy sort-of works but it's really a nasty hack.

tsmethurst:
Oh strange I thought we had this already but I see now indeed it only works if the mail server seems to be on localhost, which doesn't work with containers. After 0.16.0 is out we can add an option for this :)

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/2973.

PR #3240, opened 2024-08-26
SMTPEncryption config option to support SMTPS email transport

ForestJohnson:
# Description

Support email servers that only support SMTPS (smtp inside TLS). 

STARTTLS  (TLS inside SMTP) is considered by  some to be deprecated and shouldn't be required IMO.

Here are the configs I tested and the corresponding email server logs:

```
      GTS_SMTP_HOST: 'smtp.nullhex.com'
      GTS_SMTP_PORT: '587'
      GTS_SMTP_USERNAME: 'forest@sequentialread.com'
      GTS_SMTP_PASSWORD: '${NULLHEX_PASSWORD}'
      GTS_SMTP_FROM: 'forest@sequentialread.com'
```

```
Aug 26 06:16:18 domechild mail.info smtpd[16098]: 0d672a6feb14295b smtp connected address=198.74.6.203 host=198-74-6-203.fttp.usinternet.com
Aug 26 06:16:18 domechild mail.info smtpd[16098]: 0d672a6feb14295b smtp tls ciphers=TLSv1.3:TLS_AES_256_GCM_SHA384:256
Aug 26 06:16:19 domechild mail.info smtpd[16098]: 0d672a6feb14295b smtp authentication user=forest@sequentialread.com result=ok
Aug 26 06:16:19 domechild mail.info smtpd[16098]: 0d672a6feb14295b smtp message msgid=de188435 size=1701 nrcpt=1 proto=ESMTP
Aug 26 06:16:19 domechild mail.info smtpd[16098]: 0d672a6feb14295b smtp envelope evpid=de1884357e21d03b from=<forest@sequentialread.com> to=<test4@nullhex.com>
Aug 26 06:16:19 domechild mail.info dovecot: lmtp(16120): Connect from local
Aug 26 06:16:19 domechild mail.info smtpd[16098]: 0d672a6feb14295b smtp disconnected reason=quit
Aug 26 06:16:19 domechild mail.info dovecot: lmtp(test4@nullhex.com)<16120><jeFCFrMdzGb4PgAAEh31eQ>: msgid=<59f6fe49-a2e0-4c56-b1cd-ce55c2185748@gotosocial.sequentialread.com>: saved mail to INBOX
Aug 26 06:16:19 domechild mail.info smtpd[16098]: 0d672a70179aef0d mda delivery evpid=de1884357e21d03b from=<forest@sequentialread.com> to=<test4@nullhex.com> rcpt=<test4@nullhex.com> user=vmail delay=0s result=Ok stat=Delivered

```

```
      GTS_SMTP_HOST: 'smtp.nullhex.com'
      GTS_SMTP_PORT: '465'
      GTS_SMTP_ENCRYPTION: 'SMTPS'
      GTS_SMTP_USERNAME: 'forest@sequentialread.com'
      GTS_SMTP_PASSWORD: '${NULLHEX_PASSWORD}'
      GTS_SMTP_FROM: 'forest@sequentialread.com'
```

```
Aug 26 06:08:28 domechild mail.info smtpd[15931]: 9567b6bf3b10535d smtp connected address=198.74.6.203 host=198-74-6-203.fttp.usinternet.com
Aug 26 06:08:28 domechild mail.info smtpd[15931]: 9567b6bf3b10535d smtp tls ciphers=TLSv1.3:TLS_AES_256_GCM_SHA384:256
Aug 26 06:08:28 domechild mail.info smtpd[15931]: 9567b6bf3b10535d smtp authentication user=forest@sequentialread.com result=ok
Aug 26 06:08:28 domechild mail.info smtpd[15931]: 9567b6bf3b10535d smtp message msgid=12b645e9 size=1701 nrcpt=1 proto=ESMTP
Aug 26 06:08:28 domechild mail.info smtpd[15931]: 9567b6bf3b10535d smtp envelope evpid=12b645e9ed22b4cb from=<forest@sequentialread.com> to=<test3@nullhex.com>
Aug 26 06:08:28 domechild mail.info dovecot: lmtp(15955): Connect from local
Aug 26 06:08:28 domechild mail.info smtpd[15931]: 9567b6bf3b10535d smtp disconnected reason=quit
Aug 26 06:08:28 domechild mail.info dovecot: lmtp(test3@nullhex.com)<15955><gwe4I9wbzGZTPgAAEh31eQ>: msgid=<e3e0d659-716e-4cd8-a885-b233d4bb75ac@gotosocial.sequentialread.com>: saved mail to INBOX
Aug 26 06:08:28 domechild mail.info smtpd[15931]: 9567b6c01d7c2c33 mda delivery evpid=12b645e9ed22b4cb from=<forest@sequentialread.com> to=<test3@nullhex.com> rcpt=<test3@nullhex.com> user=vmail delay=0s result=Ok stat=Delivered
```

```
      GTS_SMTP_PORT: '587'
      GTS_SMTP_ENCRYPTION: 'SMTPS'
```

```
Aug 26 05:58:36 domechild mail.info smtpd[15617]: b848de53739f113f smtp connected address=198.74.6.203 host=198-74-6-203.fttp.usinternet.com
Aug 26 05:58:36 domechild mail.info smtpd[15617]: b848de53739f113f smtp bad-input result="500 5.5.1 Invalid command: Pipelining not supported"
Aug 26 05:58:36 domechild mail.info smtpd[15617]: b848de53739f113f smtp disconnected reason=quit
```

## Checklist

- [X] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [Not Yet] I/we have made any necessary changes to documentation.
- [No] I/we have added tests that cover new code.
- [Nope] I/we have run tests and they pass locally with the changes.
- [Nope] I/we have run `go fmt ./...` and `golangci-lint run`.

### DIFF
--- a/internal/config/gen/gen.go
+++ b/internal/config/gen/gen.go
@@ -52,16 +52,17 @@ func main() {
 	var out string
 
 	// Load runtime config flags
-	flag.StringVar(&out, "out", "", "Generated file output path")
+	flag.StringVar(&out, "out", "./internal/config/helpers.gen.go", "Generated file output path")
 	flag.Parse()
 
 	// Open output file path
 	output, err := os.OpenFile(out, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0o644)
 	if err != nil {
-		panic(err)
+		panic(fmt.Errorf("%+v. Try running this from the repository root or recreating that file if it got deleted.", err))
 	}
 
 	fmt.Fprint(output, "// THIS IS A GENERATED FILE, DO NOT EDIT BY HAND\n")
+	fmt.Fprint(output, "// edit internal/config/gen/gen.go instead.\n")
 	fmt.Fprint(output, license)
 	fmt.Fprint(output, "package config\n\n")
 	fmt.Fprint(output, "import (\n")
--- a/internal/email/common.go
+++ b/internal/email/common.go
@@ -19,8 +19,10 @@ package email
 
 import (
 	"bytes"
+	"crypto/tls"
 	"errors"
 	"fmt"
+	"net"
 	"net/smtp"
 	"os"
 	"path/filepath"
@@ -44,7 +46,7 @@ func (s *sender) sendTemplate(template string, subject string, data any, toAddre
 		return err
 	}
 
-	if err := smtp.SendMail(s.hostAddress, s.auth, s.from, toAddresses, msg); err != nil {
+	if err := SendMail(s.hostname, s.port, s.encryptionMode, s.auth, s.from, toAddresses, msg); err != nil {
 		return gtserror.SetSMTP(err)
 	}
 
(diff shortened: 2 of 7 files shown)


ForestJohnson:
BTW if you want to also close https://github.com/superseriousbusiness/gotosocial/issues/2973,  the InsecureSkipVerify could be added to the TLS Configs here maybe?  Unsure.

daenney:
The current defaults are for STARTTLS, but I don't think we should do that. SMTP with implicit TLS, or SMTP over TLS or SMTPS, is widely deployed and available and should be our defaults.

That would also match the official recommendation since 2018 in RFC 8314:

> Previous standards for the use of email protocols with TLS used the
   STARTTLS mechanism: [[RFC2595](https://datatracker.ietf.org/doc/html/rfc2595)], [[RFC3207](https://datatracker.ietf.org/doc/html/rfc3207)], and [[RFC3501](https://datatracker.ietf.org/doc/html/rfc3501)].  With
   STARTTLS, the client establishes a cleartext application session and
   determines whether to issue a STARTTLS command based on server
   capabilities and client configuration.  If the client issues a
   STARTTLS command, a TLS handshake follows that can upgrade the
   connection.  Although this mechanism has been deployed, an alternate
   mechanism where TLS is negotiated immediately at connection start on
   a separate port (referred to in this document as "Implicit TLS") has
   been deployed more successfully.  **To encourage more widespread use of
   TLS and to also encourage greater consistency regarding how TLS is
   used, this specification now recommends the use of Implicit TLS for
   POP, IMAP, SMTP Submission, and all other protocols used between an
   MUA and an MSP.**

I would actually argue for dropping support for STARTTLS entirely, unless some strong evidence emerges for supporting STARTTLS. But given SMTPS is more widely deployed, that seems unlikely to happen.

ForestJohnson:
Yes agreed SMTPS is better, but if we change the default now, it will break every single user's email setup.  Is that really worth it? 

At very least if we are going to break every single users email-sending, we should work on better error messages.

1. The error message when it fails is really hard to see / hard to identify
2. There is no success message which gets logged when email gets sent -- this makes it nearly impossible  to troubleshoot email issues in GTS without having access to the email server to view the logs and create an unlimited # of test accounts. 


Is there some system in GTS which could allow it to use a different default depending on when the email server host/port was last updated?

internal/config/gen/gen.go
```diff
@@ -52,16 +52,17 @@ func main() {
 	var out string
 
 	// Load runtime config flags
-	flag.StringVar(&out, "out", "", "Generated file output path")
+	flag.StringVar(&out, "out", "./internal/config/helpers.gen.go", "Generated file output path")
```
tsmethurst, line 55: I don't think there's any reason to be messing about in this file when adding a config option. Changes here would be better added in a separate chore PR.

internal/email/common.go
```diff
@@ -118,3 +120,102 @@ func assembleMessage(mailSubject string, mailBody string, mailFrom string, msgID
 
 	return msg.Bytes(), nil
 }
+
+// Slightly modified version of SendMail() from `net/smtp` standard library, in order to support encryptionMode STARTTLS vs. SMTPS
+func SendMail(hostname string, port int, encryptionMode string, a smtp.Auth, from string, to []string, msg []byte) error {
```
tsmethurst, line 125: I'm not totally sold on pulling this logic out from `net/smtp` and monkey-patching our TLS logic into it. `net/smtp` [is now frozen](https://pkg.go.dev/net/smtp) anyway, so we might be better off just switching to a different library that supports the functionality we need. [This one here](https://github.com/emersion/go-smtp) seems quite widely used and up to date, so it might give us what we need.

tsmethurst:
Reviewed! Thanks for the PR. While this does work, I have a feeling we might be better off switching to a different library for our mailing needs, like the one I mentioned [here](https://github.com/superseriousbusiness/gotosocial/pull/3240/files/2667bde73c84bb24512d45e8186f0a61b8500483#diff-208fecaee0dd5b43edeea0da4128255f854fc06c1614598085e48b3f56c315c8), since it means less code for us to maintain, and provides a bunch of extensions we might be interested in using. Would you be up for having a crack at rewriting the PR to use that library instead? Based on the examples [here](https://github.com/emersion/go-smtp/blob/master/example_test.go) it doesn't look like a very different API from what we currently use, so hopefully it won't be too much of a faff. Understand if you'd rather not though, in which case I'd be happy to write a PR for this myself :)

daenney:
> Yes agreed SMTPS is better, but if we change the default now, it will break every single user's email setup. Is that really worth it?

Why would it break everyone's setup? The `SMTPEncryption` setting is new. We don't have to turn encryption on by default, but our encryption defaults should be SMTPS.

We're also pre-1.0 and we have done previous releases with breaking configuration changes.

ForestJohnson:
It would break peoples setups because right now it defaults to STARTTLS. So the email client will start behaving differently ( sending TLS ClientHello first, before SMTP `HELO` )

ForestJohnson:
Yeah maybe I can try adding the emersion package .  i never knew about the test email thing because I've never been to the gotosocial admin panel before. I don't know if I even have an admin user account 😅  I figured it wasn't a feature yet because it wasn't in the admin CLI. Maybe I can add it to the admin CLI as well.

tsmethurst:
> Maybe I can add it to the admin CLI as well.

There's probably not much point really, it's not likely to be something that people will ever need to script against.

PR #3241, opened 2024-08-26
[chore] Bump all otel deps

tsmethurst:
Bump all our opentelemetry dependencies at once, since they have to be in sync.


NyaaaWhatsUpDoc:
it's a pain we have to do it this way. i wish dependabot could do this :')

tsmethurst:
Me too, me too

PR #3242, opened 2024-08-26
[bugfix] move WASM compilation stage much later in server init

NyaaaWhatsUpDoc:
# Description

This reduces the amount of memory required on startup as some of the database migrations can end up storing quite a lot of data in memory too. Instead we compile the WebAssembly binaries into usable form later in the server init process.

This also removes the compilation of the sqlite3 binary early on, as by the time we reach this function call it will have already been compiled in order to make database calls.

## Checklist


tsmethurst:
Looks good, squerge when you're happy :)

PR #3247, opened 2024-08-29
[chore] Only call `imaging.Resize when necessary`, use even tinier blurhashes

tsmethurst:
~~Use imaging.Fit rather than imaging.Resize, as the former doesn't bother resizing the image if it's already inside the target dimensions, whereas the latter does, if I'm reading the code correctly.~~

~~EDIT: Actually I'm not quite right about that. `Fit` will leave the image alone if it's *at or below* the target dimensions. `Resize` will leave the image alone if it already *equals* the target dimensions, but scale it up or down if it's not. But in this case I don't think it makes a difference, because we calculate the thumb dimensions beforehand and pass it through to here, and if the image is already in-bounds of our thumbnail dimensions we just pass the original image dimensions. So no need for this change actually.~~

Only call imaging.Resize when necessary, to avoid unnecessary clone operation.

Also use an even tinier resize for blurhashes. Both of these changes just revert to behavior we had in 0.16.0 already, I think we just included these changes more or less by accident.

~~Draft because I need to update blurhashes in tests.~~


internal/media/thumbnail.go
```diff
@@ -255,7 +255,7 @@ func generateNativeThumb(
 	if needBlurhash {
 		// for generating blurhashes, it's more cost effective to
 		// lose detail since it's blurry, so make a tiny version.
-		tiny := imaging.Resize(img, 64, 64, imaging.NearestNeighbor)
+		tiny := imaging.Resize(img, 32, 0, imaging.NearestNeighbor)
```
NyaaaWhatsUpDoc, line 258: surely for blurhashes we want to use 32x32? i think that zero means it will either end up *not* resizing the height, or it will end up with zero height (which will give us useless blurhashes)

NyaaaWhatsUpDoc: or 64x64. i figured there shouldn't be much of a difference between 32x32 vs 64x64 but i'm really not fussed either way

tsmethurst: Leaving one of those as 0 just means it preserves aspect ratio, so it should be OK. We'll get an image of width 32px and height whatever.

PR #3255, opened 2024-08-30
[chore] status dereferencing improvements

NyaaaWhatsUpDoc:
# Description

- in the case that mention only provides a namestring, search for existing mention by username,domain
- handle definitely deleted statuses (i.e. they return 401) in the dereferencer by deleting them

## Checklist


internal/federation/dereferencing/status.go
```diff
@@ -307,44 +299,54 @@ func (d *Dereferencer) enrichStatusSafely(
...
 
 	// Unlock now
 	// we're done.
 	unlock()
 
 	if errors.Is(err, db.ErrAlreadyExists) {
-		// Ensure AP model isn't set,
-		// otherwise this indicates WE
-		// enriched the status.
-		apubStatus = nil
```
NyaaaWhatsUpDoc, line 346: this is handled above now

internal/federation/dereferencing/status.go
```diff
@@ -172,13 +173,6 @@ func (d *Dereferencer) getStatusByURI(ctx context.Context, requestUser string, u
 		nil,
 	)
 
-	if err != nil {
```
tsmethurst, line 175: Since we're now not doing this check here you can just return `d.enrichStatusSafely(...)` instead now

tsmethurst:
Seems reasonable :) Just one small comment but not an important one, squerge away when you're happy.

internal/federation/dereferencing/status.go
```diff
@@ -448,11 +454,14 @@ func (d *Dereferencer) enrichStatus(
 	// Ensure we have the author account of the status dereferenced (+ up-to-date). If this is a new status
 	// (i.e. status.AccountID == "") then any error here is irrecoverable. status.AccountID must ALWAYS be set.
 	if _, _, err := d.getAccountByURI(ctx, requestUser, attributedTo); err != nil && status.AccountID == "" {
-		return nil, nil, gtserror.Newf("failed to dereference status author %s: %w", uri, err)
+
+		// Note that we specifically DO NOT wrap the error, instead collapsing it as string.
+		// Errors fetching an account do not necessarily relate to dereferencing the status.
+		return nil, nil, gtserror.Newf("failed to dereference status author %s: %v", uri, err)
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: :D

internal/federation/dereferencing/status.go
NyaaaWhatsUpDoc: 213c65712

Issue #3249, opened 2024-08-29
[feature] Add option to hide even "public" posts from web view of profile


lilithserenada:
### Is your feature request related to a problem ?

I would prefer my instance not to be viewed by non-ActivityPub users to cut down on this trivial form of block evasion and anonymous monitoring.

### Describe the solution you'd like.

Make DISALLOW_UNAUTHENTICATED_API_ACCESS available via config.yaml

### Describe alternatives you've considered.

I noticed that GoToSocial defaults to Unlisted visibility, and through an opt-in mindset on a global network this is awesome! A limiting factor of this is that only my Followers (and theirs, if they repost) can see my posts. I would like my posts to be able to take advantage of built-in network discovery tools only within my block/allowlist specifications. I believe adding functionality for DISALLOW_UNAUTHENTICATED_API_ACCESS will enable this.

tsmethurst:
I'm not really sure what you're asking for here, can you be more explicit about what API(s) you'd like disallowed?

tsmethurst:
Ah, GoToSocial already does this by default :) You can't access any of the client APIs (except the /api/v1/instance endpoint) without being authorized.

tsmethurst:
Oh okay I see what you're getting at. The difference there is that when you open a profile page in your browser, Mastodon serves a React web app that calls the backend API to get its data. When access to the backend is disallowed, you then get that cavalcade of 401 errors, and the app sort of breaks when you try to scroll down.

In GoToSocial's case, the web view of a profile is just an HTML web page that's templated by the backend; it doesn't make any API calls. So having some kind of `DISALLOW_UNAUTHENTICATED_API_ACCESS` option doesn't make any sense here, because it wouldn't change anything.

So -- correct me if I'm wrong -- I think what you're asking for here is some kind of setting that hides your profile page? Or just doesn't show any posts, even if you've made public posts? Is that right?

tsmethurst:
Got ya :) If it's alright with you, I'll rename this issue to better reflect the feature request.

Issue #900, opened 2022-10-09
[feature] Allow users to opt in to showing their Unlisted posts via the web


tsmethurst:
Right now we only ever show Public statuses on the web ui (whether on the account's profile, or via a status link).

This behavior can be a bit confusing for users coming over from Mastodon, which shows both Public and Unlisted posts via the web. Oftentimes, users expect to be able to open Unlisted links in their browser and see the status, whereas now the page returns a 404 and an explanatory message.

To alleviate this, it might be nice to create a setting so that users can choose to have their Unlisted posts rendered via status links, as though they were Public posts. This should be off by default to keep the existing behavior.

NyaaaWhatsUpDoc:
So would the usage here be that unlisted posts don't show on the profile, but if they opt-in via account setting then the individual URL by status ID would work?

I have definitely fallen into this trap of attempting to share a status only to remember it won't be visible 😅

daenney:
There's a pattern of usage where someone starts with a public post, but then makes the follow-up posts unlisted. This is a nice way of doing threads without overloading the public timeline. Folks can see the opening post and follow along, but if they're not interested the rest doesn't show up in their home timeline.

For that pattern, it would be nice if we showed unlisted replies by the original author (that are chained to the opening post) in our web view too. That lets you view the complete thread, as it were. It might be nice to have a separate user preference for that behaviour though and not roll that into the same setting.

Personally I use unlisted because for a bunch of things I don't want them to be that level of public by default. But in this particular case I'd like the whole thread to be viewable as the reason I'm using unlisted is to not overwhelm other's timelines, not to limit hot-linking.

PR #3272, opened 2024-09-06
[feature/frontend] Add options to include Unlisted posts or hide all posts

tsmethurst:
# Description

>

This pull request adds a new account setting that determines the visibility of posts that will be included in an account's web view and RSS feeds.

Closes https://github.com/superseriousbusiness/gotosocial/issues/3249
Closes https://github.com/superseriousbusiness/gotosocial/issues/900

## Checklist


internal/processing/account/update.go
```diff
@@ -330,28 +278,175 @@ func (p *Processor) Update(ctx context.Context, account *gtsmodel.Account, form
...
+		return nil, gtserror.NewErrorInternalError(err)
 	}
 
-	p.state.Workers.Client.Queue.Push(&messages.FromClientAPI{
-		APObjectType:   ap.ActorPerson,
-		APActivityType: ap.ActivityUpdate,
-		GTSModel:       account,
-		Origin:         account,
-	})
+	if sendUpdate {
```
tsmethurst, line 311: I can change this in a separate PR if desired. I was just getting a bit annoyed with changing an internal setting and having my instance send out a bunch of unnecessary Updates.

internal/processing/account/update.go
```diff
@@ -330,28 +278,175 @@ func (p *Processor) Update(ctx context.Context, account *gtsmodel.Account, form
...
+		return nil, gtserror.NewErrorInternalError(err)
 	}
+
 	return acctSensitive, nil
 }
 
+// updateFields sets FieldsRaw on the given
+// account, and resets account.Fields to an
+// empty slice, ready for further processing.
+func (p *Processor) updateFields(
```
tsmethurst, line 333: No logic changes here (and below), just extracted to a separate function for readability.

docs/user_guide/settings.md
```diff
@@ -76,6 +76,26 @@ Some examples:
...
+
+!!! warning
+    Be aware that changes to this setting also apply retroactively.
+    
+    That is, if you previously made a post on Unlisted visibility, while set to show only Public posts on your profile, and you change this setting to show Public and Unlisted, then the Unlisted post you previously made will be visible on your profile alongside your Public posts.
+    
+    Likewise, if you change this setting to show no posts, then all your posts will be hidden from your profile, regardless of when you created them, and what this option was set to at the time. This will apply until you change this setting again.
+
+!!! tip
+    Alongside (domain-)blocking, this is a good "emergency" setting to use if you're facing harassment from people trawling through your public posts. It won't hide your posts from people who can see them in their clients, via ActivityPub, but it will at least prevent them from being able to click through your posts in their browser with no authentication, and easily share them with others with a URL.
```
NyaaaWhatsUpDoc, line 97: the way this is worded, this makes it sound like the default behaviour is that we allow unlisted posts to be shown in the web view. i'd maybe reword the beginning to say something along the lines of "if you haven't already (this is the default behaviour)"

internal/db/bundb/migrations/20240906144432_unauthed_visibility.go.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/uptrace/bun"
+)
+
+func init() {
+	up := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+
+			// If column already exists we don't need to do anything.
+			exists, err := doesColumnExist(ctx, tx,
```
NyaaaWhatsUpDoc, line 32: this is a usefullittle function isn't it :D

internal/gtsmodel/accountsettings.go
```diff
@@ -32,9 +35,44 @@ type AccountSettings struct {
...
 	InteractionPolicyMutualsOnly   *InteractionPolicy `bun:""`                                                            // Interaction policy to use for new mutuals only visibility statuses. If null, assume default policy.
 	InteractionPolicyFollowersOnly *InteractionPolicy `bun:""`                                                            // Interaction policy to use for new followers only visibility statuses. If null, assume default policy.
 	InteractionPolicyUnlocked      *InteractionPolicy `bun:""`                                                            // Interaction policy to use for new unlocked visibility statuses. If null, assume default policy.
 	InteractionPolicyPublic        *InteractionPolicy `bun:""`                                                            // Interaction policy to use for new public visibility statuses. If null, assume default policy.
 }
+
+type ShowWebStatuses int
+
+const (
+	ShowWebStatusesPublicOnly = iota
```
NyaaaWhatsUpDoc, line 49: i'd make this to be specified integer values here. otherwise with an iota we risk the integer values changing (and so risk changing behaviour based on existing stored values) if a new-one is inserted earlier in the list.

NyaaaWhatsUpDoc: (depends on the response to the above comment)

internal/gtsmodel/accountsettings.go
```diff
@@ -32,9 +35,44 @@ type AccountSettings struct {
 	CustomCSS                      string             `bun:",nullzero"`                                                   // Custom CSS that should be displayed for this Account's profile and statuses.
 	EnableRSS                      *bool              `bun:",nullzero,notnull,default:false"`                             // enable RSS feed subscription for this account's public posts at [URL]/feed
 	HideCollections                *bool              `bun:",nullzero,notnull,default:false"`                             // Hide this account's followers/following collections.
+	ShowWebStatuses                ShowWebStatuses    `bun:",notnull,default:0"`                                          // Visibility level of statuses that visitors can view via the web profile.
```
NyaaaWhatsUpDoc, line 38: alternatively instead of `ShowWebStatuses`, what if we used say `WebVisibility` that uses a `StatusVisibility` as the type? means less new types, works with existing API concepts (including in the frontend for users working with our API), and means we don't need overly wordy (imo) value types like "public_and_unlisted". it also allows flexibility in the future if we add any other visibility types between public and unlisted since it basically just works as a "all statuses >= this visibility are visible in the web frontend". though obviously we probably want to set "unlisted" as the max allowable limit.

what do you think?

internal/processing/account/update.go
NyaaaWhatsUpDoc: i think maybe a separate PR might be more useful for this if that's good. i have some other qualms with us needing to invalidate caches outside of invalidation hooks and the database itself (as the less siloed we keep that cache invalidation logic, the more likely we are to forget to invalidate a cache as it could be done *anywhere* in the codebase)

tsmethurst: Alright, I removed it for now, and moved invalidation back to the database (but only calling it if either all columns change, or that column changes): https://github.com/superseriousbusiness/gotosocial/pull/3272/commits/438224113c333895d9d1dba18e9078d7b91dec36

internal/db/bundb/migrations/20240906144432_unauthed_visibility.go.go
tsmethurst: It really is!

internal/gtsmodel/accountsettings.go
tsmethurst: Right, how's this? https://github.com/superseriousbusiness/gotosocial/pull/3272/commits/db141f95d78516416d266f5285317a73a05a130d

internal/gtsmodel/accountsettings.go
tsmethurst: Changed it now to use Visibility so this isn't relevant anymore :)

PR #3280, opened 2024-09-08
[feature] Implement exclusive lists

### LINKED ISSUE #2616: [feature] Exclusive lists

It's difficult to manage a bunch of news feeds/posts from bot accounts in my timeline. It would be cool to have a feature to help with that.

Hometown/Mastodon have a feature that excludes posts that appear in the list from the user's timeline
> Hometown introduces a new kind of exclusive list. It's a toggle available for each of your lists that says "any post that appears in this list should not appear in my home timeline".

Mastodon implementation [issue](https://github.com/mastodon/mastodon/issues/16556) 

None

### Additional context.

It should be noted that a couple of clients support the feature as implemented in Mastodon.

VyrCossont:
# Description

This pull request implements Hometown-style exclusive lists, compatibly with vanilla Mastodon's current API.

Fixes #2616

## Checklist

- [x] I/we have tested this end to end in a prod environment with a Fedi client that supports exclusive lists

### DIFF
--- a/internal/api/model/list.go
+++ b/internal/api/model/list.go
@@ -32,7 +32,7 @@ type List struct {
 	RepliesPolicy string `json:"replies_policy"`
 	// Exclusive setting for this list.
 	// If true, hide posts from members of this list from your home timeline.
-	Exclusive bool `form:"exclusive" json:"exclusive" xml:"exclusive"`
+	Exclusive bool `json:"exclusive"`
 }
 
 // ListCreateRequest models list creation parameters.
--- a/internal/api/model/list.go
+++ b/internal/api/model/list.go
@@ -30,6 +30,9 @@ type List struct {
 	//	list = Show replies to members of the list
 	//	none = Show replies to no one
 	RepliesPolicy string `json:"replies_policy"`
+	// Exclusive setting for this list.
+	// If true, hide posts from members of this list from your home timeline.
+	Exclusive bool `form:"exclusive" json:"exclusive" xml:"exclusive"`
 }
 
 // ListCreateRequest models list creation parameters.
@@ -53,6 +56,11 @@ type ListCreateRequest struct {
 	//	- list
 	//	- none
 	RepliesPolicy string `form:"replies_policy" json:"replies_policy" xml:"replies_policy"`
+	// Exclusive setting for this list.
+	// If true, hide posts from members of this list from your home timeline.
+	// default: false
+	// in: formData
+	Exclusive bool `form:"exclusive" json:"exclusive" xml:"exclusive"`
 }
 
 // ListUpdateRequest models list update parameters.
@@ -70,6 +78,10 @@ type ListUpdateRequest struct {
 	// Sample: list
 	// in: formData
 	RepliesPolicy *string `form:"replies_policy" json:"replies_policy" xml:"replies_policy"`
+	// Exclusive setting for this list.
+	// If true, hide posts from members of this list from your home timeline.
+	// in: formData
+	Exclusive *bool `form:"exclusive" json:"exclusive" xml:"exclusive"`
 }
 
 // ListAccountsChangeRequest is a list of account IDs to add to or remove from a list.
--- a/internal/db/bundb/migrations/20240903212842_add_exclusive_lists.go
+++ b/internal/db/bundb/migrations/20240903212842_add_exclusive_lists.go
@@ -1,0 +1,47 @@
+// GoToSocial
+// Copyright (C) GoToSocial Authors admin@gotosocial.org
+// SPDX-License-Identifier: AGPL-3.0-or-later
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package migrations
+
+import (
+	"context"
+	"strings"
+
+	"github.com/uptrace/bun"
+)
+
+func init() {
+	up := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			_, err := tx.ExecContext(ctx, "ALTER TABLE ? ADD COLUMN ? BOOLEAN DEFAULT false", bun.Ident("lists"), bun.Ident("exclusive"))
+			if err != nil && !(strings.Contains(err.Error(), "already exists") || strings.Contains(err.Error(), "duplicate column name") || strings.Contains(err.Error(), "SQLSTATE 42701")) {
+				return err
+			}
+			return nil
+		})
+	}
+
+	down := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
+			return nil
+		})
+	}
+
+	if err := Migrations.Register(up, down); err != nil {
+		panic(err)
+	}
+}
--- a/internal/db/bundb/migrations/20240903212842_add_exclusive_lists.go
+++ b/internal/db/bundb/migrations/20240903212842_add_exclusive_lists.go
@@ -19,7 +19,6 @@ package migrations
 
 import (
 	"context"
-	"strings"
 
 	"github.com/uptrace/bun"
 )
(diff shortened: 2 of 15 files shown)


VyrCossont:
Testing this end-to-end is a bit tricky: of the clients I tried today, Semaphore doesn't support lists at all, Elk doesn't support setting any list options other than the title, and Phanpy supports lists and the exclusive flag, but doesn't support setting the exclusive flag for GtS instances yet.

in Feditext, you can test exclusive lists by scrolling down to its advanced preferences and set "API compatibility mode" to "Enable all API calls, attempt error recovery", which disables most feature gating. (You'll want to turn that API compatibility mode off later, because Feditext will also call API endpoints that don't exist, which can trip wary WAFs like CrowdSec when they assume the site's being mapped for an attack.) Future Feditext builds will enable exclusive lists for whatever GtS version includes it.

VyrCossont:
Confirmed that you can't follow yourself in GtS, and thus can't add yourself to a list (internally, this requires that a follow exists), so you can't add yourself to an exclusive list and thus prevent yourself from seeing your own posts in your home timeline.

internal/api/model/list.go
```diff
@@ -30,6 +30,9 @@ type List struct {
 	//	list = Show replies to members of the list
 	//	none = Show replies to no one
 	RepliesPolicy string `json:"replies_policy"`
+	// Exclusive setting for this list.
+	// If true, hide posts from members of this list from your home timeline.
+	Exclusive bool `form:"exclusive" json:"exclusive" xml:"exclusive"`
```
tsmethurst, line 35: This is just for marshalling not unmarshalling, so you don't need the form and xml tags.

internal/db/bundb/migrations/20240903212842_add_exclusive_lists.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+	"strings"
+
+	"github.com/uptrace/bun"
+)
+
+func init() {
+	up := func(ctx context.Context, db *bun.DB) error {
+		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
```
tsmethurst, line 29: Now that we have a function to check if a column exists, you can do this a little bit neater, for example: https://github.com/superseriousbusiness/gotosocial/blob/2bf566dea6c6d962436d088fcbd21601519e366a/internal/db/bundb/migrations/20240906144432_unauthed_visibility.go.go#L29-L58

internal/processing/workers/surfacetimeline.go
```diff
@@ -259,11 +266,60 @@ func (s *Surface) listTimelineStatusForFollow(
...
+	listEntries, err := s.State.DB.GetListEntriesForFollowID(
+		// We only need the list IDs.
+		gtscontext.SetBarebones(ctx),
+		follow.ID,
+	)
+	if err != nil && !errors.Is(err, db.ErrNoEntries) {
+		errs.Appendf("error getting list entries: %w", err)
+		return nil, false
+	}
+
```
tsmethurst, line 299: You could probs return early here if listEntries is empty.

tsmethurst:
This looks great :) Nice simple PR that will be really useful. Just a couple comments from me. Once those are resolved we can squerge this!

internal/processing/workers/surfacetimeline.go
```diff
@@ -259,11 +266,60 @@ func (s *Surface) listTimelineStatusForFollow(
 			stream.TimelineList+":"+listEntry.ListID, // key streamType to this specific list
 			filters,
 			mutes,
-		); err != nil {
+		)
+		if err != nil {
 			errs.Appendf("error adding status to timeline for list %s: %w", listEntry.ListID, err)
 			// implicit continue
 		}
+		if timelined {
```
NyaaaWhatsUpDoc, line 274: this might be a little more neat to do as `listTimelined = listTimelined || timelined`

internal/processing/workers/surfacetimeline.go
```diff
@@ -677,21 +738,22 @@ func (s *Surface) timelineStatusUpdateForFollowers(
```
NyaaaWhatsUpDoc, line 738: i think this might be a bit neater if you do `if exclusive { continue }` here instead of the nested if below

internal/processing/workers/surfacetimeline.go
```diff
@@ -259,11 +266,60 @@ func (s *Surface) listTimelineStatusForFollow(
...
+			listTimelined = true
+		}
+	}
+
+	return exclusive, listTimelined
+}
+
+// getListEntriesAndExclusive returns list entries for a given follow,
+// and whether any of them belong to exclusive lists.
+func (s *Surface) getListEntriesAndExclusive(
```
NyaaaWhatsUpDoc, line 284: i think it would be more idiomatic go code here if we didn't pass the multierror and instead returned an error here that is appended by callers of this function. it ends up making the logic flow harder to ascertain imo

internal/processing/workers/surfacetimeline.go
```diff
@@ -700,31 +762,25 @@ func (s *Surface) timelineStatusUpdateForFollowers(
...
+// (in which case the status shouldn't be added to the home timeline).
 func (s *Surface) listTimelineStatusUpdateForFollow(
 	ctx context.Context,
 	status *gtsmodel.Status,
 	follow *gtsmodel.Follow,
 	errs *gtserror.MultiError,
 	filters []*gtsmodel.Filter,
 	mutes *usermute.CompiledUserMuteList,
-) {
+) bool {
```
NyaaaWhatsUpDoc, line 775: same here regarding returning an error

NyaaaWhatsUpDoc:
have left a handful of comments but other than that looks good :)

internal/db/bundb/migrations/20240903212842_add_exclusive_lists.go
VyrCossont: Sweet. Upgrading to this.

internal/processing/workers/surfacetimeline.go
VyrCossont: Done. I'll also do that if `lists` is empty a little later, although that should be much less likely.

internal/processing/workers/surfacetimeline.go
VyrCossont: Sure. (Oh, if only Go had `||=`. [Alas.](https://go.dev/ref/spec#Assignment_statements))

internal/processing/workers/surfacetimeline.go
VyrCossont: I did this for consistency with the functions that call it, `listTimelineStatusForFollow` and `listTimelineStatusUpdateForFollow`. Happy to follow the usual Go error style instead.

My issue with your suggested name change is that the function's name no longer describes what it does; I wouldn't myself expect a function named `getListEntries` to also return an `exclusive` flag even if the return tuple fields were named. However, since it's only called in two places, and since it doesn't have any internal state, how would you feel about splitting it into `getListEntries` and `isOnExclusiveList(…, listEntries, …)`?

internal/processing/workers/surfacetimeline.go
VyrCossont: Yep, can do.

internal/processing/workers/surfacetimeline.go
VyrCossont: Done.

internal/api/model/list.go
VyrCossont: Oops. Fixed.

internal/processing/workers/surfacetimeline.go
```diff
@@ -208,33 +219,36 @@ func (s *Surface) timelineAndNotifyStatusForFollowers(
...
-		// We only need the list IDs.
-		gtscontext.SetBarebones(ctx),
-		follow.ID,
-	)
-	if err != nil && !errors.Is(err, db.ErrNoEntries) {
-		errs.Appendf("error getting list entries: %w", err)
-		return
+	listEntries, err := s.getListEntries(ctx, follow)
+	if err != nil {
+		errs.Append(err)
```
NyaaaWhatsUpDoc, line 243: i think we should probably return at these points (here and the 1 check below) if an error is returned.

internal/processing/workers/surfacetimeline.go
```diff
@@ -700,30 +770,31 @@ func (s *Surface) timelineStatusUpdateForFollowers(
...
```
NyaaaWhatsUpDoc, line 793: same goes for returning with errors here.

NyaaaWhatsUpDoc: (and the 1 below)

internal/processing/workers/surfacetimeline.go
VyrCossont: Done (thanks for the clarification in chat)

internal/processing/workers/surfacetimeline.go
VyrCossont: and done.

internal/processing/workers/surfacetimeline.go
VyrCossont: Handled, with different names.

PR #3290, opened 2024-09-12
[performance] cache more database calls, reduce required database calls overall

NyaaaWhatsUpDoc:
# Description

- updates many of the delete-model database functions to instead of preloading the model, delete the model returning the necessary details for cache invalidation and calling the invalidation hook manually
- updates the list accounts endpoint to use our new style paging logic (needed for cache change below)
- updates `go-structr` to v0.8.9 which handles needing to check for an empty slice of uncached keys
- updates a number of the list entry related database functions to instead rely on more caches
- consolidates some caches and removes some unused functions

## Checklist


NyaaaWhatsUpDoc:
description now accurately described this PR, only thing that needs fixing is updating the list related tests :)

internal/cache/db.go
```diff
@@ -103,6 +100,12 @@ type DBCaches struct {
 	// - '<'  for follower IDs
 	FollowRequestIDs SliceCache[string]
 
+	// FollowingTagIDs provides access to account IDs following / tag IDs followed by
+	// account db cache. THIS CACHE IS KEYED AS THE FOLLOWING {prefix}{id} WHERE:
+	// - '>{$accountID}' for tag IDs followed by account
+	// - '<{$tagIDs}' for account IDs following tag
+	FollowingTagIDs SliceCache[string]
```
tsmethurst, line 107: I think for these caches with more interesting keys, it would be good to have an example or two in the comments as well, just for illustration purposes.

internal/cache/db.go
```diff
@@ -860,7 +869,6 @@ func (c *Caches) initList() {
 		// will be populated separately.
 		// See internal/db/bundb/list.go.
 		l2.Account = nil
-		l2.ListEntries = nil
```
tsmethurst, line 863: How come this was removed? Not saying it shouldn't be, it's just not clear to me why.

internal/cache/invalidate.go
```diff
@@ -121,6 +118,15 @@ func (c *Caches) OnInvalidateFollow(follow *gtsmodel.Follow) {
 		">"+follow.TargetAccountID,
 		"l>"+follow.TargetAccountID,
 	)
+
+	// Invalidate source account's lists
+	// and destination account's lists, and
+	// those specifically for this follow.
+	c.DB.ListIDs.Invalidate(
+		"a"+follow.AccountID,
```
tsmethurst, line 126: I think it'd be nice to just add a very brief few words comment on each of these invalidated keys just so it's possible to see at a glance what they are, without having to go look at the caches file for reference.

internal/db/bundb/account.go
```diff
@@ -822,44 +809,19 @@ func (a *accountDB) DeleteAccount(ctx context.Context, id string) error {
 		// delete the account
 		_, err := tx.
 			NewDelete().
-			TableExpr("? AS ?", bun.Ident("accounts"), bun.Ident("account")).
-			Where("? = ?", bun.Ident("account.id"), id).
+			Model(&deleted).
+			Where("? = ?", bun.Ident("id"), id).
+			Returning("?", bun.Ident("uri")).
```
tsmethurst, line 814: Clever to use returning statements for this :) Can't believe we didn't think of this before.

vendor/codeberg.org/gruf/go-structr/cache.go
```diff
@@ -375,6 +375,11 @@ func (c *Cache[T]) Load(index *Index, keys []Key, load func([]Key) ([]T, error))
 	// the lock.
 	unlock()
 
+	if len(keys) == 0 {
```
tsmethurst, line 378: Nice

internal/util/unique.go
```diff
@@ -17,12 +17,23 @@
...
+	m := make(map[T]S, len(in))
+	for _, v := range in {
+		m[key(v)] = v
+	}
+	return m
+}
+
 // Set represents a hashmap of only keys,
 // useful for deduplication / key checking.
 type Set[T comparable] map[T]struct{}
```
tsmethurst, line 34: Oh cool I didn't even know we had such a type, I should be using it elsewhere.

VyrCossont: I feel like every Go codebase grows this one eventually, like `util.Ptr`.

NyaaaWhatsUpDoc: hehe :D

vendor/codeberg.org/gruf/go-structr/cache.go
NyaaaWhatsUpDoc: thank you for noticing it to begin with!

internal/cache/invalidate.go
NyaaaWhatsUpDoc: good idea!

internal/cache/db.go
NyaaaWhatsUpDoc: good idea!

internal/cache/db.go
NyaaaWhatsUpDoc: basically in cases where lists can end up with a lot of entries (e.g. i have one that's over 100 entries), it feels like it's worth keeping separate just in case we ever end up forgetting how much data is embedded in the list. (also in a lot of cases it's more efficient to fetch the items we need directly instead of dealing with the intermedia that is the list entry).

tsmethurst: Ah, what I meant was, aren't we now storing these ListEntries inside the cache, because we're not nilling the slice? Or am I missing some logic from somewhere else? I'll keep reading through the PR and come back to this.

internal/gtsmodel/list.go
```diff
@@ -27,7 +27,6 @@ type List struct {
 	Title         string        `bun:",nullzero,notnull,unique:listaccounttitle"`                   // Title of this list.
 	AccountID     string        `bun:"type:CHAR(26),notnull,nullzero,unique:listaccounttitle"`      // Account that created/owns the list
 	Account       *Account      `bun:"-"`                                                           // Account corresponding to accountID
-	ListEntries   []*ListEntry  `bun:"-"`                                                           // Entries contained by this list.
```
tsmethurst, line 30: Ahh I see!

internal/gtsmodel/list.go
```diff
@@ -38,8 +37,9 @@ type ListEntry struct {
 	CreatedAt time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"` // when was item created
 	UpdatedAt time.Time `bun:"type:timestamptz,nullzero,notnull,default:current_timestamp"` // when was item last updated
 	ListID    string    `bun:"type:CHAR(26),notnull,nullzero,unique:listentrylistfollow"`   // ID of the list that this entry belongs to.
-	FollowID  string    `bun:"type:CHAR(26),notnull,nullzero,unique:listentrylistfollow"`   // Follow that the account owning this entry wants to see posts of in the timeline.
-	Follow    *Follow   `bun:"-"`                                                           // Follow corresponding to followID.
+	//	List      *List     `bun:"-"`                                                           //
```
tsmethurst, line 40: This an editing remnant?

internal/cache/db.go
tsmethurst: Ah nvm I see now, ListEntries has been removed from *gtsmodel.List :) Got it.

internal/api/client/lists/listaccounts.go
```diff
@@ -129,42 +130,27 @@ func (m *Module) ListAccountsGETHandler(c *gin.Context) {
...
 		apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
 		return
 	}
 
-	var (
-		ctx = c.Request.Context()
+	page, errWithCode := paging.ParseIDPage(c,
+		1,  // min limit
+		80, // max limit
+		0,  // no paging allowed
```
tsmethurst, line 142: This comment is a bit confusing because it implies that paging is not allowed, could you change it to something like `// default = disabled` which we use elsewhere, or `// default = paging disabled` maybe?

internal/db/bundb/list.go
```diff
@@ -85,39 +86,52 @@ func (l *listDB) getList(ctx context.Context, lookup string, dbQuery func(*gtsmo
...
-	if len(listIDs) == 0 {
-		return nil, nil
-	}
+func (l *listDB) CountListsByAccountID(ctx context.Context, accountID string) (int, error) {
+	listIDs, err := l.getListIDsByAccountID(ctx, accountID)
+	return len(listIDs), err
+}
 
-	// Return lists by their IDs.
+func (l *listDB) GetListsWithFollowID(ctx context.Context, followID string) ([]*gtsmodel.List, error) {
```
tsmethurst, line 102: While we're renaming + shuffling stuff around, I think probably `GetListsContainingFollowID` might make more sense here, what do you think?

internal/processing/workers/surfacetimeline.go
```diff
@@ -227,107 +230,59 @@ func (s *Surface) listTimelineStatusForFollow(
...
+		return false, false, gtserror.Newf("error getting lists for follow: %w", err)
 	}
 
-	// Check eligibility for each list entry (if any).
-	listTimelined := false
-	for _, listEntry := range listEntries {
-		eligible, err := s.listEligible(ctx, listEntry, status)
+	for _, list := range lists {
+		// Update exclusive flag if list is so.
+		exclusive = exclusive || *list.Exclusive
```
tsmethurst, line 250: Hmm shouldn't you be updating exclusive *after* checking eligibility for a list? Else you might set exclusive to true, skip this list because it's not relevant, then insert it into a non-exclusive list and return exclusive=true.

internal/processing/workers/surfacetimeline.go
```diff
@@ -418,22 +356,13 @@ func (s *Surface) listEligible(
...
 		return follows, nil
 
 	default:
-		// HUH??
-		err := gtserror.Newf(
-			"reply policy '%s' not recognized on list %s",
-			list.RepliesPolicy, list.ID,
-		)
-		return false, err
+		panic("unknown reply policy: " + list.RepliesPolicy)
```
tsmethurst, line 365: Maybe better to use log.Panicf or log.Panic here?

internal/processing/workers/surfacetimeline.go
```diff
@@ -779,58 +713,59 @@ func (s *Surface) listTimelineStatusUpdateForFollow(
...
 	}
 
-	// Check eligibility for each list entry (if any).
-	for _, listEntry := range listEntries {
-		eligible, err := s.listEligible(ctx, listEntry, status)
+	var exclusive, timelined bool
+	for _, list := range lists {
+
+		// Update exclusive flag if list is so.
+		exclusive = exclusive || *list.Exclusive
```
tsmethurst, line 735: Same comment here re: exclusivity.

internal/db/bundb/statusbookmark.go
```diff
@@ -264,60 +257,85 @@ func (s *statusBookmarkDB) PutStatusBookmark(ctx context.Context, bookmark *gtsm
...
 func (s *statusBookmarkDB) DeleteStatusBookmarksForStatus(ctx context.Context, statusID string) error {
-	q := s.db.
-		NewDelete().
+	// Delete status bookmarks
+	// from database by status ID.
+	q := s.db.NewDelete().
 		TableExpr("? AS ?", bun.Ident("status_bookmarks"), bun.Ident("status_bookmark")).
 		Where("? = ?", bun.Ident("status_bookmark.status_id"), statusID)
-	if _, err := q.Exec(ctx); err != nil {
+	if _, err := q.Exec(ctx); err != nil && !errors.Is(err, db.ErrNoEntries) {
```
tsmethurst, line 326: Does a delete with no change actually return ErrNoEntries btw? I can't remember if it actually does.

internal/db/bundb/relationship_follow.go
```diff
@@ -252,139 +244,155 @@ func (r *relationshipDB) UpdateFollow(ctx context.Context, follow *gtsmodel.Foll
...
+	targetAccountID string,
+) error {
+
+	// Gather necessary fields from
+	// deleted for cache invaliation.
+	var deleted gtsmodel.Follow
+	deleted.AccountID = sourceAccountID
+	deleted.TargetAccountID = targetAccountID
+
+	// Delete all blocks either from
```
tsmethurst, line 259: All follows you mean :D

internal/db/bundb/tag.go
```diff
@@ -234,9 +215,15 @@ func (t *tagDB) PutFollowedTag(ctx context.Context, accountID string, tagID stri
 		return nil
 	}
 
-	// Otherwise, this is a new followed tag, so we invalidate caches related to it.
-	t.state.Caches.DB.AccountIDsFollowingTag.Invalidate(tagID)
-	t.state.Caches.DB.TagIDsFollowedByAccount.Invalidate(accountID)
+	// If we deleted anything, invalidate caches.
```
tsmethurst, line 218: Why would we have deleted something here?

internal/db/bundb/poll.go
```diff
@@ -391,148 +391,44 @@ func (p *pollDB) PutPollVote(ctx context.Context, vote *gtsmodel.PollVote) error
 	})
 }
 
-func (p *pollDB) DeletePollVotes(ctx context.Context, pollID string) error {
```
tsmethurst, line 394: Where do we now handle deleting poll votes? I'm a bit befuddled by this bit.

internal/gtsmodel/list.go
NyaaaWhatsUpDoc: indeed! removed: 6bd75b8a2

internal/api/client/lists/listaccounts.go
NyaaaWhatsUpDoc: good idea, that does make it a lot more readable :') 071172ea9

internal/db/bundb/tag.go
NyaaaWhatsUpDoc: comment error from copy-paste. have fixed those here :) 57f268870

internal/db/bundb/relationship_follow.go
NyaaaWhatsUpDoc: indeed, another copy-paste issue :') fixed here b65fe3ff7

internal/db/bundb/relationship_follow_req.go
```diff
@@ -293,124 +286,131 @@ func (r *relationshipDB) RejectFollowRequest(ctx context.Context, sourceAccountI
...
+	targetAccountID string,
+) error {
+
+	// Gather necessary fields from
+	// deleted for cache invaliation.
+	var deleted gtsmodel.FollowRequest
+	deleted.AccountID = sourceAccountID
+	deleted.TargetAccountID = targetAccountID
+
+	// Delete all blocks either from
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: indeed, another copy-paste issue :') fixed here b65fe3ff7

internal/db/bundb/list.go
NyaaaWhatsUpDoc: makes sense to me! have done that here: 1e7ad8e5c

internal/db/bundb/statusbookmark.go
NyaaaWhatsUpDoc: it doesn't you're right! basically db.ErrNoEntries is only ever returned by a delete if it's *returning* something. have updated that here (and the call above it): 3c1fb1a0f

internal/processing/workers/surfacetimeline.go
NyaaaWhatsUpDoc: have updated here: 381478499

internal/processing/workers/surfacetimeline.go
NyaaaWhatsUpDoc: Good point! have shifted that here: 381478499

internal/processing/workers/surfacetimeline.go
NyaaaWhatsUpDoc: Good point! have shifted that here: 381478499

internal/db/bundb/poll.go
NyaaaWhatsUpDoc: ah yep i moved this to the DeletePoll function since we don't actually have a reason anywhere to delete all votes for a poll separately to the poll itself. though i forgot to commit it :') added it back here: 921b64f5558824305264b90642d952285e63df9b

NyaaaWhatsUpDoc: and nearly used the incorrect column, fix here: 0978b0116

PR #3291, opened 2024-09-13
[chore] Reject replies to rejected replies

tsmethurst:
# Description

>

This pull request updates our permissivity checks in the dereferencer to store a pre-rejected interaction request for rejected replies, and to check if a status tries to reply to a previously-rejected reply. This should help to avoid situations where someone replies to a status, their reply is rejected, and then they reply *to their own reply*, which is allowed through, causing all sorts of mess.

Closes https://github.com/superseriousbusiness/gotosocial/issues/3057 as this is the last part.

## Checklist

### DIFF
--- a/internal/federation/dereferencing/status.go
+++ b/internal/federation/dereferencing/status.go
@@ -373,7 +373,7 @@ func (d *Dereferencer) enrichStatus(
 	requestUser string,
 	uri *url.URL,
 	status *gtsmodel.Status,
-	apubStatus ap.Statusable,
+	statusable ap.Statusable,
 ) (
 	*gtsmodel.Status,
 	ap.Statusable,
@@ -393,7 +393,7 @@ func (d *Dereferencer) enrichStatus(
 		return nil, nil, gtserror.SetUnretrievable(err)
 	}
 
-	if apubStatus == nil {
+	if statusable == nil {
 		// Dereference latest version of the status.
 		rsp, err := tsport.Dereference(ctx, uri)
 		if err != nil {
@@ -402,7 +402,7 @@ func (d *Dereferencer) enrichStatus(
 		}
 
 		// Attempt to resolve ActivityPub status from response.
-		apubStatus, err = ap.ResolveStatusable(ctx, rsp.Body)
+		statusable, err = ap.ResolveStatusable(ctx, rsp.Body)
 
 		// Tidy up now done.
 		_ = rsp.Body.Close()
@@ -444,7 +444,7 @@ func (d *Dereferencer) enrichStatus(
 	}
 
 	// Get the attributed-to account in order to fetch profile.
-	attributedTo, err := ap.ExtractAttributedToURI(apubStatus)
+	attributedTo, err := ap.ExtractAttributedToURI(statusable)
 	if err != nil {
 		return nil, nil, gtserror.New("attributedTo was empty")
 	}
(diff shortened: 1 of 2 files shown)


internal/federation/dereferencing/status.go
```diff
@@ -373,7 +373,7 @@ func (d *Dereferencer) enrichStatus(
 	requestUser string,
 	uri *url.URL,
 	status *gtsmodel.Status,
-	apubStatus ap.Statusable,
+	statusable ap.Statusable,
```
tsmethurst, line 376: Just some renaming in here to bring this in line with naming we use elsewhere, while I was dicking about in this package anyway.

internal/federation/dereferencing/status_permitted.go
```diff
@@ -101,8 +116,85 @@ func (d *Dereferencer) isPermittedReply(
...
+		id := id.NewULID()
+
+		// To ensure the Reject chain stays coherent,
+		// borrow fields from the up-thread rejection.
+		// This collapses the chain beyond the first
+		// rejected reply and allows us to avoid derefing
+		// further replies we already know we don't want.
+		statusID := req.StatusID
+		targetAccountID := req.TargetAccountID
+		uri := strings.ReplaceAll(req.URI, req.ID, id)
```
NyaaaWhatsUpDoc, line 168: why doing this like this instead of manually building the URI as in other places?

NyaaaWhatsUpDoc: also i'm generally a bit confused here. this is the parent status that you've fetched an interaction request for, why are you storing a second interaction request for it here? did you mean for this logic to happen if the request actually turns out to be nil?

NyaaaWhatsUpDoc: surely this would also just be `return false, nil` in place of this whole block?

tsmethurst: The second interaction request is being stored here for the child of the rejected parent. This is the part of the logic that prevents replies to rejected replies from being stored. If the parent is rejected, any followup replies/children of that parent should also be rejected. To prevent subsequent replies to *this* child from being accepted, then, we need to store a rejected interaction request for this child, too.

For example, User A (someone on our instance), writes Post 1 that has replies disabled entirely. User B (from our instance or elsewhere, doesn't matter), writes Post 2 in reply to it that is automatically rejected and therefore not stored on the instance. User B could then write a follow-up reply to their own post, call it Post 3, that is not rejected, because there's no trace of Post 2 on the instance. They could then also write another reply to Post 3, Post 4, that would also not be rejected. Or someone else could reply to Post 3, and that would also not be rejected.

The code here resolves this by ensuring that some kind of record is kept for automatically rejected replies, even when they're not stored in the `statuses` table. So children of rejected replies don't get stored, and children of those children also don't get stored.

The reason the URI is constructed the way that it is is because we don't want to use the creator of Post 2 as the owner of the Reject activity -- it should be the creator of Post 1 who owns the Reject. Essentially, we recycle the URL of the first Reject, which was created by User A, to create rejects for all subsequent children-of-children.

tsmethurst: Essentially we're treating all replies in a rejected sub-thread as being replies to the OG status, and rejecting them on that basis, even though technically they're replies to each other.

NyaaaWhatsUpDoc: okay reasoning makes sense :ok_hand: :ok_hand: though for the URI generation, could this not be something that may have been created by a remote instance, so the URI may not conform to our method of generation? i.e. the strings.ReplaceAll() may not work?

tsmethurst: Ah balls yeah you could be right there, I'll have a look.

tsmethurst: Yep you were right. Got around this by just not setting a URI at all for these implicit rejects, we don't need it anyway and we can't be sure of what it should actually be.

Issue #3298, opened 2024-09-14
[bug] `SanitizeToPlaintext` concatenates text in unexpected ways


VyrCossont:
For example, `<p>as</p><p>df</p>` sanitizes to `asdf`. This would cause a false positive for a filter with the keyword `asdf`, and a false negative for a whole-word filter with the keyword `df`. I'd expect output more like `as\ndf\n`.

Likewise, `<br>` tags are dropped, not converted to `\n`. Guessing the same holds for `<wbr>`, and any other tags closely equivalent to characters. (Fun exercise: what would we expect from `<hr>`?)

Not familiar with the [BlueMonday](https://github.com/microcosm-cc/bluemonday) sanitizer we use, so not sure how hard this would be to fix.

Discovered while investigating #3128.

tsmethurst:
Mmm, maybe instead of BlueMonday we can use some other method of converting HTML to plaintext for the purposes of matching against filters. In the frontend we now use https://www.npmjs.com/package/html-to-text for converting the HTML representation of statuses into text for showing in certain places. If there's a Go equivalent we could look at that, perhaps.

tsmethurst:
Something like this -- https://pkg.go.dev/github.com/k3a/html2text -- or this -- https://github.com/jaytaylor/html2text -- perhaps? But then if these operations are expensive (and I'd imagine they're not as cheap as BlueMonday) we probably also want to be storing those results in the 'text' field of the `*gtsmodel.Status` model, ie., in the database. Not 100% sure.

Issue #3128, opened 2024-07-22
[bug] Filters not always filtering (boosts)


mirabilos:
### Describe the bug with a clear and concise description of what the bug is.

I’m seeing posts containing a filtered hashtag in FediText and Phanpy, but not in Semaphore. Vyr asked me to report it as server issue. Details below.

### What's your GoToSocial Version?

0.16.0-rc3+git-db80361 🦥

### GoToSocial Arch

amd64 binary

### Anything else we need to know?

The post in question is:

```
gotosocial=> \pset null '\\N'
Null display is "\N".
gotosocial=> \pset pager 0
Pager usage is off.
gotosocial=> SELECT * FROM statuses WHERE id='01J3AEHK78GWZB0F6SDQFY0SFE';
             id             |       created_at       |          updated_at           |                                  uri                                  |                           url                           |                                                                                                        content                                                                                                         |         attachments          |             tags             | mentions | emojis | local |         account_id         |                account_uri                | in_reply_to_id | in_reply_to_uri | in_reply_to_account_id | boost_of_id | boost_of_account_id | content_warning | visibility | sensitive | language | created_with_application_id | activity_streams_type | text | federated | boostable | replyable | likeable | pinned_at |          fetched_at           | poll_id | thread_id 
----------------------------+------------------------+-------------------------------+-----------------------------------------------------------------------+---------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+------------------------------+------------------------------+----------+--------+-------+----------------------------+-------------------------------------------+----------------+-----------------+------------------------+-------------+---------------------+-----------------+------------+-----------+----------+-----------------------------+-----------------------+------+-----------+-----------+-----------+----------+-----------+-------------------------------+---------+-----------
[...]

PR #3301, opened 2024-09-15
[bugfix] Use better plaintext representation of status for filtering

tsmethurst:
# Description

>

This pull request updates our filtering logic to not use our `SanitizeToPlaintext` function for reducing status HTML content to plaintext, but instead use https://github.com/k3a/html2text, which doesn't cause weird line concatenation, and can competently extract links, mentions, and hashtags properly from the text.

To avoid re-parsing a status from HTML every time we want to filter it, a TTLCache has been added to the converter which stores the parsed-to-text version of statuses.

Also some minor fixes to our filter regexes, to include whitespace and start/end line in our whole word match.

closes https://github.com/superseriousbusiness/gotosocial/issues/3298
closes https://github.com/superseriousbusiness/gotosocial/issues/3128

## Checklist


internal/gtsmodel/filter.go
```diff
@@ -61,14 +63,23 @@ type FilterKeyword struct {
...
 func (k *FilterKeyword) Compile() (err error) {
-	var wordBreak string
-	if k.WholeWord != nil && *k.WholeWord {
-		wordBreak = `\b`
+	var (
+		wordBreakStart string
+		wordBreakEnd   string
+	)
+
+	if util.PtrOrZero(k.WholeWord) {
```
tsmethurst, line 71: These weren't working properly without specifically including whitespace and end/start of line.

https://regex101.com/r/7RUq96/1

VyrCossont: Yeah that's better than omitting the subpattern entirely and doesn't require checking the start and end of the keyword.

internal/typeutils/util.go
```diff
@@ -284,3 +287,75 @@ func ContentToContentLanguage(
...
 }
+
+// StatusHash returns an xxhash of text
+// from a status, taking account of:
+//
+//   - content warning
+//   - content
+//   - media IDs + descriptions
+//   - poll options
+func StatusHash(s *gtsmodel.Status) string {
```
VyrCossont, line 298: Could we do this just based on status ID and created/edited timestamps? Less to hash that way.

internal/typeutils/util.go
```diff
@@ -284,3 +287,75 @@ func ContentToContentLanguage(
...
+	for _, attachment := range s.Attachments {
+		fields = append(fields, attachment.Description)
+	}
+
+	// Poll options.
+	if s.Poll != nil {
+		fields = append(fields, s.Poll.Options...)
+	}
+
+	return strings.Join(fields, " ")
```
VyrCossont, line 360: I'd prefer this not concatenate unrelated fields since it could lead to false-positive matches for multi-word filters. Note that `filterableTextFields` returned multiple fields.

internal/typeutils/internaltofrontend_test.go
```diff
@@ -1063,15 +1063,22 @@ func (suite *InternalToFrontendTestSuite) TestHideFilteredBoostToFrontend() {
 
 // Test that a hashtag filter for a hashtag in Mastodon HTML content works the way most users would expect.
 func (suite *InternalToFrontendTestSuite) testHashtagFilteredStatusToFrontend(wholeWord bool, boost bool) {
-	testStatus := suite.testStatuses["admin_account_status_1"]
+	testStatus := new(gtsmodel.Status)
+	*testStatus = *suite.testStatuses["admin_account_status_1"]
 	testStatus.Content = `<p>doggo doggin' it</p><p><a href="https://example.test/tags/dogsofmastodon" class="mention hashtag" rel="tag nofollow noreferrer noopener" target="_blank">#<span>dogsofmastodon</span></a></p>`
+	testStatus.Text = "doggo doggin' it\n\n#dogsofmastodon"
```
VyrCossont, line 1069: See comment about not special-casing statuses when `.Text` is available.

internal/typeutils/internaltofrontend.go
```diff
@@ -956,14 +965,7 @@ func (c *Converter) statusToAPIFilterResults(
 		// List all matching keywords.
 		keywordMatches := make([]string, 0, len(filter.Keywords))
 		for _, filterKeyword := range filter.Keywords {
-			var isMatch bool
-			for _, field := range fields {
```
VyrCossont, line 960: See comment on `filterText` concatenating unrelated fields. There's a reason I did it this way.

internal/typeutils/util.go
```diff
@@ -284,3 +287,75 @@ func ContentToContentLanguage(
...
+//   - poll options
+func filterableText(s *gtsmodel.Status) string {
+	fields := []string{}
+
+	// Content warning / title.
+	fields = append(fields, s.ContentWarning)
+
+	// Status content; use raw text if available,
+	// else use text parsed from content HTML.
+	if s.Text != "" {
```
VyrCossont, line 339: We discussed this when I first implemented v2 filters and I still think it's a mistake to treat local statuses differently and that we should not start doing it now. Raw source text is usually not equivalent to what rendering to status HTML and then to plain text would give us, but *especially* with Markdown posts.

internal/typeutils/util.go
tsmethurst: Right! I'll change it back and add a note so I don't get the same idea again, I've the memory of a sieve.

internal/typeutils/util.go
tsmethurst: Got you :+1: I totally forgot about this. I'll change it back + put a comment in the code here so I don't forget.

internal/typeutils/util.go
tsmethurst: Got it!

internal/typeutils/util.go
```diff
@@ -284,3 +285,64 @@ func ContentToContentLanguage(
...
+	if s.ContentWarning != "" {
+		fields = append(fields, s.ContentWarning)
+	}
+
+	// Status content. Though we have raw text
+	// available for statuses created on our
+	// instance, use the html2text version to
+	// remove markdown-formatting characters
+	// and ensure more consistent filtering.
+	if s.Content != "" {
```
NyaaaWhatsUpDoc, line 320: do we want to handle content map here too? ideally not if we can get away with it as iterating maps isn't particularly fast, but just double checking

internal/typeutils/internaltofrontend.go
```diff
@@ -939,32 +939,47 @@ func (c *Converter) statusToAPIFilterResults(
 		return nil, nil
 	}
 
-	// Extract text fields from the status that we will match filters against.
-	fields := filterableTextFields(s)
+	// Key this status based on ID + last updated time,
+	// to ensure we always filter on latest version.
+	statusKey := fmt.Sprintf("%s%d", s.ID, s.UpdatedAt.Unix())
```
NyaaaWhatsUpDoc, line 944: it would probably be faster here to do `s.ID + strconv.FormatInt()` here given this is quite performance sensitive code, (though may want to wrap it in a func for neatness).

internal/typeutils/converter.go
```diff
@@ -31,13 +34,25 @@ type Converter struct {
 	randAvatars    sync.Map
 	visFilter      *visibility.Filter
 	intFilter      *interaction.Filter
+
+	// TTL cache of statuses -> filterable text fields.
+	// To ensure up-to-date fields, cache is keyed as:
+	// [status.ID][status.UpdatedAt.Unix()]`
+	statusesFilterableFields cache.TTLCache[string, []string]
```
NyaaaWhatsUpDoc, line 41: i think it would probably be better to store this in the `state.State.Caches` itself, just to keep all the caches in one location.

internal/typeutils/util.go
tsmethurst: Nahhh, it's probably fine for now.

PR #3356, opened 2024-09-26
[chore] bump go-byteutil v1.2.0 -> v1.3.0

NyaaaWhatsUpDoc:
see: https://codeberg.org/gruf/go-byteutil/releases/tag/v1.3.0


NyaaaWhatsUpDoc:
unfortunately codeberg seems to be down right now so can't link to the release in the description. will do that when codeberg is back up :')

PR #3357, opened 2024-09-26
[chore] update go-sched pkg

NyaaaWhatsUpDoc:
see: https://codeberg.org/gruf/go-sched/releases/tag/v1.2.4


NyaaaWhatsUpDoc:
unfortunately codeberg seems to be down right now so can't link to the release in the description. will do that when codeberg is back up :')

PR #3361, opened 2024-09-27
[chore] update gruf / {go-cache, go-maps, go-kv}

NyaaaWhatsUpDoc:
see: https://codeberg.org/gruf/go-cache/releases/tag/v3.6.0
see: https://codeberg.org/gruf/go-maps/releases/tag/v1.0.4
see: https://codeberg.org/gruf/go-kv/releases/v1.6.5


vendor/codeberg.org/gruf/go-maps/common.go
```diff
@@ -54,15 +54,17 @@ func (m *ordered[K, V]) Delete(key K) bool {
 
 // Range passes given function over the requested range of the map.
 func (m *ordered[K, V]) Range(start, length int, fn func(int, K, V)) {
+	// Nil check
+	if fn == nil {
```
NyaaaWhatsUpDoc, line 58: it's a very simple fix (i wasn't correctly doing a nil check), but it should help in doing very large ranges on these ordered map types (used within our ttl caches, of which we still have a few), as the compiler should now know it doesn't need to do a nil check on every loop iteration.

PR #3367, opened 2024-09-27
[chore] update modernc/sqlite to v1.33.1 (with our concurrency workaround)

NyaaaWhatsUpDoc:
The version hadn't been bumped in a while. Since it seems some people are still using it, thought it would be useful to keep it up-to-date. (note this still has our concurrency workaround applied, so it should be safe)


NyaaaWhatsUpDoc:
you can run all the tests with this library just fine still with `-tags=moderncsqlite` :)

PR #3370, opened 2024-09-28
[bugfix] visibility after implicit approval not getting invalidated

### LINKED ISSUE #3343: [bug/v0.17.0-rc1] Approved replies don't show immediately on web frontend
I *think* this is something to do with visibility cacheing, so gotta double check that we're invalidating the visibility cache when a status gets approved.

NyaaaWhatsUpDoc:
# Description

c-c-c-combo! test by @tsmethurst, fix in https://codeberg.org/gruf/go-structr (our cache library) by myself

closes #3343 

## Checklist


internal/cache/visibility.go
```diff
@@ -48,9 +48,15 @@ func (c *Caches) initVisibility() {
 			{Fields: "RequesterID", Multiple: true},
 			{Fields: "Type,RequesterID,ItemID"},
 		},
-		MaxSize:   cap,
-		IgnoreErr: ignoreErrors,
-		Copy:      copyF,
+		MaxSize: cap,
+		IgnoreErr: func(err error) bool {
```
NyaaaWhatsUpDoc, line 52: basically it's too difficult otherwise making sure we have all errors possibly stored under Type,RequesterID,ItemID combinations cleared out in the case of the visibility cache, so instead it's just easier to not cache any errors.

internal/cache/wrappers.go
```diff
@@ -240,3 +240,7 @@ func (c *StructCache[T]) Len() int {
 func (c *StructCache[T]) Cap() int {
 	return c.cache.Cap()
 }
+
+func (c *StructCache[T]) Debug() map[string]any {
```
NyaaaWhatsUpDoc, line 244: oh whoops i'll remove this (though at some point i'll properly plumb this through for metrics or something)

PR #3258, opened 2024-09-01
[chore] Bump tooling versions, bump go -> v1.23.0

tsmethurst:
# Description

> If this is a code change, please include a summary of what you've coded, and link to the issue(s) it closes/implements.
>
> If this is a documentation change, please briefly describe what you've changed and why.

This pull request bumps the following:

- go v1.22 -> v1.23
- golangci-lint v1.57.2 -> v1.62.0 (and gosec linting)
- alpine v3.19.1 ->  v3.20
- gotosocial-drone-build v0.6.0 -> v0.7.0, which includes:
  - docker buildx v0.11.2 -> v0.16.2
  - goreleaser v1.25.1 -> v2.3.2

## Checklist

Please put an x inside each checkbox to indicate that you've read and followed it: `[ ]` -> `[x]`

If this is a documentation change, only the first checkbox must be filled (you can delete the others if you want).

- [x] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [ ] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [x] I/we have not leveraged AI to create the proposed changes.
- [x] I/we have performed a self-review of added code.
- [x] I/we have written code that is legible and maintainable by others.
- [x] I/we have commented the added code, particularly in hard-to-understand areas.
- [ ] I/we have made any necessary changes to documentation.
- [ ] I/we have added tests that cover new code.
- [ ] I/we have run tests and they pass locally with the changes.
- [x] I/we have run `go fmt ./...` and `golangci-lint run`.

PR #3420, opened 2024-10-12
[chore] Upgrade golangci-lint, ignore existing int overflow warnings

untitaker:
There is a new lint for unchecked int casts. Integer overflows are bad,
but the old code that triggers this lint seems to be perfectly fine.
Instead of disabling the lint entirely for new code as well, grandfather
in existing code. Maybe new code can be written in such a way to avoid
excessive casting.

Also fix a broken link in CONTRIBUTING


untitaker:
oops, this is a duplicate of https://github.com/superseriousbusiness/gotosocial/pull/3258 -- though that one is stuck. can port over the nolint annotations from there.

tsmethurst:
Porting over the nolint annotations and math.MaxWhatever checks is probably a good idea :+1:

NyaaaWhatsUpDoc:
having this linter is a good idea, though replacing api model ints with uints i'm not sure about. the reason that go returns signed integer types for lengths, silces, capacities etc is that a signed integer overflow specifically causes a panic as a negative length is invalid. but an unsigned integer just silently overflows to a valid result which makes hunting down bugs much harder.

internal/processing/account/update.go
```diff
@@ -462,11 +462,11 @@ func (p *Processor) UpdateAvatar(
...
 ) {
 	// Get maximum supported local media size.
-	maxsz := config.GetMediaLocalMaxSize()
+	maxsz := int64(config.GetMediaLocalMaxSize()) // #nosec G115 -- Already validated.
 
 	// Ensure media within size bounds.
-	if avatar.Size > int64(maxsz) {
-		text := fmt.Sprintf("media exceeds configured max size: %s", maxsz)
+	if avatar.Size > maxsz {
+		text := fmt.Sprintf("media exceeds configured max size: %d", maxsz)
```
NyaaaWhatsUpDoc, line 469: bytesize.Size specifically implements `String()` for nicer log output, this will only make error responses harder to read by returning a number bytes in the order of millions vs simplified size output.

leaving it as it was ensures we keep that nicer log output both here and below.

internal/processing/admin/emoji.go
```diff
@@ -45,11 +44,11 @@ func (p *Processor) EmojiCreate(
...
 
 	// Get maximum supported local emoji size.
-	maxsz := config.GetMediaEmojiLocalMaxSize()
+	maxsz := int64(config.GetMediaEmojiLocalMaxSize()) // #nosec G115 -- Already validated.
 
 	// Ensure media within size bounds.
-	if form.Image.Size > int64(maxsz) {
-		text := fmt.Sprintf("emoji exceeds configured max size: %s", maxsz)
+	if form.Image.Size > maxsz {
+		text := fmt.Sprintf("emoji exceeds configured max size: %d", maxsz)
```
NyaaaWhatsUpDoc, line 51: same again re: output

internal/processing/admin/emoji.go
```diff
@@ -441,11 +440,11 @@ func (p *Processor) emojiUpdateModify(
...
 
 		// Get maximum supported local emoji size.
-		maxsz := config.GetMediaEmojiLocalMaxSize()
+		maxsz := int64(config.GetMediaEmojiLocalMaxSize()) // #nosec G115 -- Already validated.
 
 		// Ensure media within size bounds.
-		if image.Size > int64(maxsz) {
-			text := fmt.Sprintf("emoji exceeds configured max size: %s", maxsz)
+		if image.Size > maxsz {
+			text := fmt.Sprintf("emoji exceeds configured max size: %d", maxsz)
```
NyaaaWhatsUpDoc, line 447: same again re: output

internal/processing/media/create.go
```diff
@@ -35,11 +35,11 @@ import (
...
 
 	// Get maximum supported local media size.
-	maxsz := config.GetMediaLocalMaxSize()
+	maxsz := int64(config.GetMediaLocalMaxSize()) // #nosec G115 -- Already validated.
 
 	// Ensure media within size bounds.
-	if form.File.Size > int64(maxsz) {
-		text := fmt.Sprintf("media exceeds configured max size: %s", maxsz)
+	if form.File.Size > maxsz {
+		text := fmt.Sprintf("media exceeds configured max size: %d", maxsz)
```
NyaaaWhatsUpDoc, line 42: same again re: output

internal/media/ffmpeg.go
```diff
@@ -556,10 +557,18 @@ func (res *ffprobeResult) Process() (*result, error) {
 				if p := strings.SplitN(str, "/", 2); len(p) == 2 {
 					n, _ := strconv.ParseUint(p[0], 10, 32)
 					d, _ := strconv.ParseUint(p[1], 10, 32)
-					num, den = uint32(n), uint32(d)
+
+					if n > math.MaxUint32 || d > math.MaxUint32 {
```
NyaaaWhatsUpDoc, line 561: this isn't necessary. the 3rd argument to ParseUint() specifically tells it to parse a maximum size of 32bit unsigned integer. the return type is just always a 64bit type so they can get away with only a single function for both.

internal/media/ffmpeg.go
```diff
@@ -556,10 +557,18 @@ func (res *ffprobeResult) Process() (*result, error) {
...
+						return nil, gtserror.Newf("overflowed numerator or denominator")
+					}
+					num, den = uint32(n), uint32(d) // #nosec G115 -- Just checked.
 				} else {
 					n, _ := strconv.ParseUint(p[0], 10, 32)
-					num = uint32(n)
+
+					if n > math.MaxUint32 {
```
NyaaaWhatsUpDoc, line 568: same again

internal/processing/account/update.go
untitaker: I'll revert those things then. I suggest to change go-bytesize to use the same type as [file sizes in the rest of std](https://pkg.go.dev/io/fs#FileInfo). this conversion shouldn't be necessary at all if everybody just agrees on a type.

NyaaaWhatsUpDoc:
just realized these changes are from the go1.23 PR, but i hadn't gotten around to reading that one yet :sweat_smile: 

given these weren't your changes either myself or @tsmethurst can make them (though totally i'm happy to), if so just let me know :)

untitaker:
np im doing it

internal/processing/media/create.go
```diff
@@ -35,11 +35,12 @@ import (
...
 
 	// Get maximum supported local media size.
-	maxsz := int64(config.GetMediaLocalMaxSize()) // #nosec G115 -- Already validated.
+	maxsz := config.GetMediaLocalMaxSize()
+	maxszInt64 := int64(maxsz) // #nosec G115 -- Already validated.
 
 	// Ensure media within size bounds.
-	if form.File.Size > maxsz {
-		text := fmt.Sprintf("media exceeds configured max size: %d", maxsz)
+	if form.File.Size > maxszInt64 {
```
untitaker, line 42: gosec annotation is ineffective when placed next to an opening curly brace, for some reason i can't figure out. so now the code looks like this.

untitaker:
this is also ready to review IMO

tsmethurst:
Hell yeah nice, thanks! I'll take a look tomorrow :)

tsmethurst:
Squerged, thank you very much!

Issue #3233, opened 2024-08-25
[feature/docs] Document `GTS_WAZERO_COMPILATION_CACHE` for caching pre-compiled WASM to disk


gmemstr:
### Is your feature request related to a problem ?

When starting up the server binary, especially on lower power machines, the `precompiling WebAssembly` step can take a while, even if the version of the binary remains unchanged.

### Describe the solution you'd like.

It may be worthwhile to look at caching the compiled WASM to disk rather than just in memory, which can then be loaded into memory on subsequent starts. This could be to a directory in `/tmp` to that reboots clear out old versions automatically. A hash/commit/revision tag could be added to the cache file to ensure when upgrading the new version is compiled and placed on disk.

### Describe alternatives you've considered.

NONE

tsmethurst:
You can do this by setting `GTS_WAZERO_COMPILATION_CACHE` to some directory :) (eg., `GTS_WAZERO_COMPILATION_CACHE=~/.cache/wazero`. We still need to document that env var.

PR #3425, opened 2024-10-13
[chore/docs] Add `/gotosocial/.cache` to Docker container, document `GTS_WAZERO_COMPILATION_CACHE`

tsmethurst:
# Description

>

This pull request adds the optional `/gotosocial/.cache` volume in the Dockerfile, and documents usage of `GTS_WAZERO_COMPILATION_CACHE` to set the compilation cache dir, for both Docker and binary deployments (with systemd).

Closes https://github.com/superseriousbusiness/gotosocial/issues/3233

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,9 +39,10 @@ USER 1000:1000
 #
 # See https://docs.docker.com/engine/reference/builder/#workdir
 #
-# First make sure storage exists + is owned by 1000:1000, then go back
-# to just /gotosocial, where we'll run from
+# First make sure storage + cache exist and are owned by 1000:1000,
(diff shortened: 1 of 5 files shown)


NyaaaWhatsUpDoc:
i think it would be better to further specify what this cache is for in this directory, in case we ever have further need for using `/gotosocial/.cache`, so making `GTS_WAZERO_COMPILATION_CACHE=/gotosocial/.cache/wazero`

tsmethurst:
Mm, well wazero creates directories of its own inside that cache directory like `wazero_vx.x.x` or whatever, so I don't think it's really necessary, then it would end up being `~/gotosocial/.cache/wazero/wazero_vx.x.x`

PR #3429, opened 2024-10-13
[feature/OFFICIALLY UNSUPPORTED] add nowasm build tag to disable building with WebAssembly

NyaaaWhatsUpDoc:
This adds a `nowasm` build tag that allows building GoToSocial without any WebAssembly modules. This will instead rely on modernc.org/sqlite and locally installed versions of ffmpeg / ffprobe. It prints a warning on server start telling the user not to make any bug reports for media processing issues, and specifically warns them that it may be less secure is it enforces no version checks.

THIS IS AN OFFICIALLY UNSUPPORTED BUILD CONFIGURATION.

This probably needs updating with some improved documentation for this, not specifically to tell people about it, but to emphasise specifically how it is unsupported and we are not interested in bug reports produced by the products of these build changes, since we have no idea what their version of ffmpeg / ffprobe could be compiled with support for etc.

### DIFF
--- a/internal/media/ffmpeg/exec_nowasm.go
+++ b/internal/media/ffmpeg/exec_nowasm.go
@@ -33,7 +33,8 @@ import (
 
 func init() {
 	fmt.Println("!! you are using an unsupported build configuration of gotosocial with WebAssembly disabled !!")
-	fmt.Println("!! please do not file bug reports, this build configuration is provided as a nice-to-have !!")
+	fmt.Println("!! please do not file bug reports regarding media processing with this configuration !!")
+	fmt.Println("!! it is also less secure; this does not enforce version checks on ffmpeg / ffprobe versions !!")
 }
 
 // runCmd will run 'name' with the given arguments, returning exit code or error.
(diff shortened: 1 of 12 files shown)


tsmethurst:
Changed the title to add `feature` first, just to make this easier to sort when writing release notes :P

internal/media/ffmpeg/exec_nowasm.go
```diff
@@ -0,0 +1,142 @@
...
+
+	"codeberg.org/gruf/go-ffmpreg/wasm"
+	"github.com/tetratelabs/wazero"
+	"github.com/tetratelabs/wazero/sys"
+)
+
+func init() {
+	fmt.Println("!! you are using an unsupported build configuration of gotosocial with WebAssembly disabled !!")
+	fmt.Println("!! please do not file bug reports regarding media processing with this configuration !!")
+	fmt.Println("!! it is also less secure; this does not enforce version checks on ffmpeg / ffprobe versions !!")
```
tsmethurst, line 37: I'd add here as well that it doesn't sandbox ffmpeg / ffprobe, as that's another quite nice advantage of the path of WASM

tsmethurst: Actually sod it, I'll just squerge it now, we can add this in later.

PR #3441, opened 2024-10-14
[bugfix] set the memory limit for ncruces/go-sqlite3 according to 32 / 64 bit platform

NyaaaWhatsUpDoc:
This sets the WebAssembly runtime memory limit for ncruces/go-sqlite3 according to it being a 32 / 64 bit platform. This should help in the cases of https://github.com/ncruces/go-sqlite3/issues/168#issuecomment-2412429221 also just means we don't end up over-reserving memory when we don't need to.

This also raises the question of what we can do about ffmpeg in the future, but we'll need to do some serious testing to figure out exactly what limit we can set that won't cause things to blow up :')

### DIFF
--- a/internal/db/sqlite/driver.go
+++ b/internal/db/sqlite/driver.go
@@ -23,12 +23,43 @@ import (
 	"context"
 	"database/sql/driver"
 
+	"codeberg.org/gruf/go-bytesize"
 	"github.com/superseriousbusiness/gotosocial/internal/db"
+	"github.com/tetratelabs/wazero"
 
+	"github.com/ncruces/go-sqlite3"
 	sqlite3driver "github.com/ncruces/go-sqlite3/driver"
 	_ "github.com/ncruces/go-sqlite3/embed"     // embed wasm binary
 	_ "github.com/ncruces/go-sqlite3/vfs/memdb" // include memdb vfs
 )
+
+func init() {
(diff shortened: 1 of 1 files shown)


tsmethurst:
Grand! Let's merge this after 0.17.0 is out, so we can test it for a bit longer and then do 0.17.1 if necessary.

ncruces:
You may find [`bits.UintSize`](https://pkg.go.dev/math/bits#pkg-constants) more readable.

ncruces:
If you want to wait for 0.17.1, I'll probably merge https://github.com/ncruces/go-sqlite3/pull/170 by then.

I hate how huge and confusing my `//go:build` lines are becoming, but this is a better default configuration, and WAL seems to work fine with `386` and  `arm` (although the interpreter will be slower than modernc for many use cases).

ncruces:
I consider this a "compatibility" change, so ti will be a `v1.20.0`.

I had planed to release it ["today"](https://github.com/ncruces/go-sqlite3/issues/175), but I'm a bit [stuck](https://github.com/ncruces/go-sqlite3/issues/172).

I doubt that the "compatibility" breaks will be an issue to anyone, but since they're there, I'd rather bundle them together, in a single release notes.

ncruces:
See: https://github.com/ncruces/go-sqlite3/releases/tag/v0.20.0

PR #3448, opened 2024-10-16
[chore] Thumbnail only first frame of animated media

tsmethurst:
# Description

>

Updates our ffmpeg call to just take the first frame and not use the thumbnail filter, which was giving unexpected results for thumbnails of short media.

We could probably do some more nuanced logic here (check the first 1% of frames if media is longer than x amount of time) but I don't think it's that necessary right now.

Closes https://github.com/superseriousbusiness/gotosocial/issues/3447

## Checklist


Ryuno-Ki:
I'm curious: If the animated started black, would it pick those?

There's a filter in regular ffmpeg: https://scribe.rip/@georgechmr/skipping-black-frame-when-generating-video-thumbnail-with-ffmpeg-9e26bf2f67e3

tsmethurst:
We could use something like that black filter in future maybe, but I don't think it's urgent right now. We also have to consider processing + memory costs when doing that kind of analysis, so something to benchmark a little bit (even anecdotally).

PR #3477, opened 2024-10-22
[bugfix] incorrect /api/v_/instance domain uri fields

### LINKED ISSUE #3467: [bug] `/api/v*/instance` `domain` and `uri` properties conflict with Mastodon’s usage of those property names

Mastodon uses the `uri` (in v1) and `domain` (in v2) properties for the domain name that gets added to account handles.

GoToSocial uses `uri` (in v1) for a link to the home page, and `domain` (in v2) for the domain where the API is hosted, with a separate `account_domain` property that serves the same purpose as Mastodon’s `domain`.

Mastodon API clients (in my case using the open source [TootSDK](https://github.com/TootSDK/TootSDK) library) that expect `uri` or `domain` to contain the account handle domain will have issues connecting to GoToSocial and/or displaying full account handles if a server is using separate domains for these.

0.17.0 (tested on various instances running this version listed on FediDB)

### GoToSocial Arch

_No response_

### What happened?

GoToSocial’s `uri` property in `/api/v1/instance` and `domain` in `/api/v2/instance` conflict with Mastodon’s properties of the same name.

Mastodon API compatibility as described in the [docs](https://docs.gotosocial.org/en/latest/#mastodon-api-compatibility)

### How to reproduce it?

call `/api/v2/instance` on a GoToSocial server

I’m wondering if this is intended behavior so I can handle it appropriately in a client app I’m developing.

NyaaaWhatsUpDoc:
# Description

closes #3467 

## Checklist


tsmethurst (approved):

tsmethurst:
Looks good. We should probably think about removing `account_domain` at some point then since it turns out it's superfluous.

NyaaaWhatsUpDoc:
> Looks good. We should probably think about removing `account_domain` at some point then since it turns out it's superfluous.

yeah i was unsure of that... at the moment it can be used to tell if a separate `account_domain` exists, but then i suppose they can already tell that as they know the host to be dialling the instance endpoint :p

tsmethurst:
Yeah indeed :thinking: Could you make an issue for removing it? We just need to investigate what uses it (can't remember why we added it) and change it to use `uri` instead, basically.

PR #3485, opened 2024-10-27
[chore] pull in go-ffmpreg v0.4.1

NyaaaWhatsUpDoc:
see: https://codeberg.org/gruf/go-ffmpreg/releases/tag/v0.4.0
also: https://codeberg.org/gruf/go-ffmpreg/releases/tag/v0.4.1


internal/media/ffmpeg/wasm.go
```diff
@@ -91,38 +85,10 @@ func compileFfprobe(ctx context.Context) error {
...
 // if already initialized this function is a no-op.
-func initRuntime(ctx context.Context) error {
+func initRuntime(ctx context.Context) (err error) {
 	if runtime != nil {
 		return nil
 	}
-
-	var cache wazero.CompilationCache
-
-	if dir := os.Getenv("GTS_WAZERO_COMPILATION_CACHE"); dir != "" {
```
tsmethurst, line 101: Does removing `GTS_WAZERO_COMPILATION_CACHE` not cause issues for caching artifacts? Or do you read that env var somewhere else now?

NyaaaWhatsUpDoc: oh that's a good point! will add that part back in :)

PR #3529, opened 2024-11-08
[performance] minimise log field allocations

NyaaaWhatsUpDoc:
- moves slice functions to separate internal/util/xslices package to fix an import cycle
- adds new xslices.AppendJust() and xslices.GrowJust() functions to grow a slice in similar manners to append() and slices.Grow() but ONLY by the necessary number of elements, not any extra as normally happens in those functions typically
- makes use of xslices.AppendJust() and xslices.GrowJust() in our log package


tsmethurst:
Looks good to me :)

tsmethurst (approved):

PR #3547, opened 2024-11-18
[feature] Add domain permission drafts and excludes

tsmethurst:
# Description

> If this is a code change, please include a summary of what you've coded, and link to the issue(s) it closes/implements.
>
> If this is a documentation change, please briefly describe what you've changed and why.

This is the first part of https://github.com/superseriousbusiness/gotosocial/issues/3. This PR adds domain permission drafts and domain permission excludes to the db, processing functions, and frontend, in preparation for adding + managing domain permission subscriptions.

Some screenshots of the frontend:

Screenshot from 2024-11-18 14-33-11
Screenshot from 2024-11-18 14-33-06
Screenshot from 2024-11-18 14-32-59
Screenshot from 2024-11-18 14-32-29
Screenshot from 2024-11-18 14-31-56

## Checklist

Please put an x inside each checkbox to indicate that you've read and followed it: `[ ]` -> `[x]`

If this is a documentation change, only the first checkbox must be filled (you can delete the others if you want).

- [x] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [x] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [x] I/we have not leveraged AI to create the proposed changes.
- [x] I/we have performed a self-review of added code.
- [x] I/we have written code that is legible and maintainable by others.
- [x] I/we have commented the added code, particularly in hard-to-understand areas.
- [ ] I/we have made any necessary changes to documentation -- still need to do this, will do it in the final PR.
- [x] I/we have added tests that cover new code.
- [x] I/we have run tests and they pass locally with the changes.
- [x] I/we have run `go fmt ./...` and `golangci-lint run`.

PR #3558, opened 2024-11-21
[performance] convert enum strings to ints

NyaaaWhatsUpDoc:
# Description

This helps improve performance by reducing the amount that these enums take-up in disk-space in their respective tables, and improves performance (however minorly) by reducing memory impact of models requiring only int enums instead of strings and also simplifying the comparisons between values of these types.

However, on larger databases this will likely be a monstrously large migration given the nature of it and it acting on the largest table we have, statuses :')

## Checklist

### DIFF
--- a/internal/db/bundb/instance.go
+++ b/internal/db/bundb/instance.go
@@ -104,7 +104,7 @@ func (i *instanceDB) CountInstanceStatuses(ctx context.Context, domain string) (
 	q = q.Where("NOT ? = ?", bun.Ident("status.pending_approval"), true)
 
 	// Ignore statuses that are direct messages.
-	q = q.Where("NOT ? = ?", bun.Ident("status.visibility"), "direct")
+	q = q.Where("NOT ? = ?", bun.Ident("status.visibility"), gtsmodel.VisibilityDirect)
 
 	count, err := q.Count(ctx)
 	if err != nil {
--- a/internal/db/bundb/migrations/20240620074530_interaction_policy.go
+++ b/internal/db/bundb/migrations/20240620074530_interaction_policy.go
@@ -20,6 +20,7 @@ package migrations
 import (
 	"context"
 
+	new_gtsmodel "github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
 
 	oldmodel "github.com/superseriousbusiness/gotosocial/internal/db/bundb/migrations/20240620074530_interaction_policy"
@@ -165,7 +166,16 @@ func init() {
 			// to new version of interaction policy.
 			for _, oldStatus := range oldStatuses {
 				// Start with default policy for this visibility.
-				v := gtsmodel.Visibility(oldStatus.Visibility)
+				v := func() gtsmodel.Visibility {
+					return map[oldmodel.Visibility]gtsmodel.Visibility{
+						oldmodel.VisibilityNone:          new_gtsmodel.VisibilityNone,
+						oldmodel.VisibilityPublic:        new_gtsmodel.VisibilityPublic,
+						oldmodel.VisibilityUnlocked:      new_gtsmodel.VisibilityUnlocked,
+						oldmodel.VisibilityFollowersOnly: new_gtsmodel.VisibilityFollowersOnly,
+						oldmodel.VisibilityMutualsOnly:   new_gtsmodel.VisibilityMutualsOnly,
+						oldmodel.VisibilityDirect:        new_gtsmodel.VisibilityDirect,
+					}[oldStatus.Visibility]
+				}()
 				policy := gtsmodel.DefaultInteractionPolicyFor(v)
 
 				if !*oldStatus.Likeable {
--- a/internal/db/bundb/migrations/20240620074530_interaction_policy.go
+++ b/internal/db/bundb/migrations/20240620074530_interaction_policy.go
@@ -20,7 +20,6 @@ package migrations
 import (
 	"context"
 
-	new_gtsmodel "github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
 	"github.com/superseriousbusiness/gotosocial/internal/log"
 
 	oldmodel "github.com/superseriousbusiness/gotosocial/internal/db/bundb/migrations/20240620074530_interaction_policy"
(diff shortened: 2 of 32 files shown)


internal/db/bundb/instance.go
```diff
@@ -104,7 +104,7 @@ func (i *instanceDB) CountInstanceStatuses(ctx context.Context, domain string) (
 	q = q.Where("NOT ? = ?", bun.Ident("status.pending_approval"), true)
 
 	// Ignore statuses that are direct messages.
-	q = q.Where("NOT ? = ?", bun.Ident("status.visibility"), "direct")
```
NyaaaWhatsUpDoc, line 107: trying to narrow down this test failure was a nightmare :')

tsmethurst: Oof ya I bet!!

internal/db/bundb/migrations/20240620074530_interaction_policy.go
```diff
@@ -165,7 +166,16 @@ func init() {
 			// to new version of interaction policy.
 			for _, oldStatus := range oldStatuses {
 				// Start with default policy for this visibility.
-				v := gtsmodel.Visibility(oldStatus.Visibility)
+				v := func() gtsmodel.Visibility {
```
tsmethurst, line 169: Does this instantiate a map for each status? Seems better to do this outside of the loop.

internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
```diff
@@ -0,0 +1,232 @@
...
+					Index(index.name).
+					Exec(ctx); err != nil {
+					return err
+				}
+			}
+
+			// Convert all visibility tables.
+			for _, table := range visTables {
+				if err := convertEnums(ctx, tx, table.Table, table.Column,
+					map[old_gtsmodel.Visibility]new_gtsmodel.Visibility{
```
tsmethurst, line 84: Same comment as above, maybe it's worth extracting this map instantiation somewhere.

internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
```diff
@@ -0,0 +1,232 @@
...
+				if index.order != "" {
+					q = q.ColumnExpr(index.order)
+				}
+				if _, err := q.Exec(ctx); err != nil {
+					return err
+				}
+			}
+
+			// Migrate over old notifications table column over to new column type.
+			if err := convertEnums(ctx, tx, "notifications", "notification_type", //nolint:revive
```
tsmethurst, line 111: Did we not have an index on notification type? Dang

internal/gtsmodel/notification.go
```diff
@@ -34,20 +34,51 @@ type Notification struct {
 	Read             *bool            `bun:",nullzero,notnull,default:false"`                             // Notification has been seen/read
 }
 
-// NotificationType describes the reason/type of this notification.
-type NotificationType string
+// NotificationType describes the
+// reason/type of this notification.
+type NotificationType int
```
tsmethurst, line 39: does it make any difference in the DB if we store these as uint8 or uint16 vs int? Just wondering if we could save even more space by using `smallint` instead of `integer` as the database storage type (assuming `smallint` is supported by both PG and SQLite) -- https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-NUMERIC-TABLE

tsmethurst:
Looking good!

internal/gtsmodel/notification.go
daenney: For SQLite this doesn't matter. Any column type with `INT` in it becomes an column with integer affinity. For storage, SQLite picks the smallest representation for the row value (so the column isn't "fixed width" based on the largest value).

> The value is a signed integer, stored in 0, 1, 2, 3, 4, 6, or 8 bytes depending on the magnitude of the value.

It'll get converted to 8-byte / int64 when reading it into memory.

internal/db/bundb/migrations/20240620074530_interaction_policy.go
NyaaaWhatsUpDoc: fixed: 79c0c3a4a

internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
NyaaaWhatsUpDoc: fixed: 79c0c3a4a

internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
NyaaaWhatsUpDoc: apparently not :grimacing:

internal/gtsmodel/notification.go
NyaaaWhatsUpDoc: fixed: 336db3d6ca223b73459639c7abbbfa00dc58b4ff

NyaaaWhatsUpDoc:
@tsmethurst re: conversation we had the other day here: https://github.com/superseriousbusiness/gotosocial/pull/3547#discussion_r1851939726

we definitely need to move models in *at the time we create them*, so we have a snapshot of the model when it was created. i just spent ages tracking down an issue with why postgres was angry at my visibility column type, but it's because the sinbinstatus type was creating the table from the *current* model type instead of a snapshot. it's much easier to just copy in a model at creation-time for a snapshot than it is hunting down the exact commit it was added at to snapshot it later.

internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
```diff
@@ -0,0 +1,249 @@
...
+	// Ensure a default value.
+	if defaultValue == nil {
+		var zero NewType
+		defaultValue = &zero
+	}
+
+	// Add new column to database.
+	if _, err := tx.NewAddColumn().
+		Table(table).
+		ColumnExpr("? SMALLINT NOT NULL DEFAULT ?",
```
mirabilos, line 163: Why `SMALLINT` and not an actual Enum type? (or does SQLite not support them?)

That being said, enum types consume double the space per row :/

From a referential integrity PoV and for making analytics queries easier, it would be interesting to have a table with just `id SMALLINT` and `label TEXT` (e.g. `visibility_types` and then have `statuses.visibility` have an FK reference to `visibility_types.id`). I understand that GtS does not consider working directly on the DB supported, though, so it’s okay if nothing comes from this.

(Just, from experience, in 10+-year-running projects, things, including later modelling that starts from the DB, work better with these things clearly structured in the SQL.)

internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
```diff
@@ -0,0 +1,249 @@
...
+		Count(ctx)
+	if err != nil {
+		return gtserror.Newf("error selecting total count: %w", err)
+	}
+
+	var updated int
+	for old, new := range mapping {
+
+		// Update old to new values.
+		res, err := tx.NewUpdate().
```
mirabilos, line 182: As I wrote on Fedi, on the phone so possible typos and not tested, this could probably have better been something like `UPDATE statuses SET visibility_new = CASE visibility WHEN 'public' THEN 1 WHeN 'next one' THEN 2 […] CASE ELSE THEN NULL (or some other default value) END CASE;`

Not sure if SQLite could do that, though.

PR #3560, opened 2024-11-22
[bugfix] Allow unsetting filter expiration dates

### LINKED ISSUE #3497: [bug] sending empty expires_in to filter v2 endpoints sets expiration to now

When creating or editing a filter v2, it is not possible to set the duration to indefinite.

0.17.2-SNAPSHOT+git-1954ccb

### GoToSocial Arch

goblin.technology

### What happened?

I created a Filter like this:
```
POST https://goblin.technology/api/v2/filters
Content-Type: application/x-www-form-urlencoded
Content-Length: 63
Authorization: Bearer <>
User-Agent: Tusky/26.2-7008c48a Android/13 OkHttp/4.12.0
title=test123&context%5B%5D=home&filter_action=warn&expires_in=
```

I got the following response:
```
{
    "id": "01JBCS3BBG8QJAEQ2XCYTACXPC",
    "title": "test123",
    "context": [
        "home"
    ],
    "expires_at": "2024-10-29T18:37:35.984Z",
    "filter_action": "warn",
    "keywords": [],
    "statuses": []
}
```

As you can see, it has the expiration set to now. 
The same happens when editing the filter with PUT /api/v2/filters

As per [api docs](https://docs.gotosocial.org/en/latest/api/swagger/) and Mastodon behavior, I would expect the expiration to be indefinite.

### How to reproduce it?

Create or edit a filter.

_No response_

VyrCossont:
# Description

There are two related problems here:
1. handling the empty string, which caused the linked bug; Tobi fixed this for v1 in #3513, but v2 has more optional fields and was handling them differently.
2. handling a literal `null`, which is the correct behavior as far as Mastodon compatibility (I checked); this hasn't led to any *reported* problems yet, and is something Go does not do well.

Fixes #3497

## Checklist


internal/api/client/filters/v2/filterput.go
```diff
@@ -180,12 +183,43 @@ func (m *Module) FilterPUTHandler(c *gin.Context) {
 		return
 	}
 
+	// If the request is JSON:
+	// Explicitly check whether `expires_in` is a null literal (vs. not being set at all).
+	hasNullExpiresIn := false
+	if c.ContentType() == gin.MIMEJSON {
```
NyaaaWhatsUpDoc, line 189: i think you could probably do this more concisely without any of this reading and replacing the reader if you just replaced the `FilterUpdateRequestV2{}.ExpiresIn` field with a `json.RawMessage{}` and then have a `FilterUpdateRequestV2{}.GetExpiresIn()` function that parses the appropriate value :)

NyaaaWhatsUpDoc:
just the one comment but otherwise looking good :)

PR #3569, opened 2024-11-25
[chore] bumps uptrace/bun dependencies to v1.2.6

NyaaaWhatsUpDoc:
see: https://github.com/uptrace/bun/blob/master/CHANGELOG.md#126-2024-11-20


NyaaaWhatsUpDoc:
so it seems this relies on Go 1.23 so will keep this open but wait for #3258

PR #3570, opened 2024-11-25
[chore] bump to Go1.23

NyaaaWhatsUpDoc:
this is a copy of #3258 but rebased on current main and updated to latest golangci-lint version, and with docker image version strings updated to just use the latest minor version of a particular major image version (this ensures we stay up-to-date with bugfixes but without breaking changes).

i made this a separate PR just because i wasn't totally comfortable doing a force push to something that wasn't my PR :p

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -389,6 +389,8 @@ checksum:
 # https://goreleaser.com/customization/snapshots/
 snapshot:
   version_template: "{{ incpatch .Version }}-SNAPSHOT"
+
+# https://goreleaser.com/customization/source/
 source:
   enabled: true
   name_template: "{{ .ProjectName }}-{{ .Version }}-source-code"
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -7,7 +7,7 @@ version: 2
 before:
   hooks:
   # generate the swagger.yaml file using go-swagger and bundle it into the assets directory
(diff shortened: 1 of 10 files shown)


test/swagger.sh
```diff
@@ -5,7 +5,7 @@
 set -eu
 
 swagger_cmd() {
-  go run github.com/go-swagger/go-swagger/cmd/swagger "$@"
+  go run ./vendor/github.com/go-swagger/go-swagger/cmd/swagger "$@"
```
NyaaaWhatsUpDoc, line 8: always tests swagger using the vendored version

.goreleaser.yml
```diff
@@ -1,12 +1,13 @@
...
+version: 2
 project_name: gotosocial
 version: 2
 
 # https://goreleaser.com/customization/hooks/
 before:
   hooks:
   # generate the swagger.yaml file using go-swagger and bundle it into the assets directory
-  - swagger generate spec --scan-models --exclude-deps -o web/assets/swagger.yaml
+  - go run ./vendor/github.com/go-swagger/go-swagger/cmd/swagger generate spec --scan-models --exclude-deps -o web/assets/swagger.yaml
```
NyaaaWhatsUpDoc, line 10: always generates swagger using the vendored version

NyaaaWhatsUpDoc:
moved back to the original PR #3258 now

Issue #3576, opened 2024-11-27
[bug] Some clients cannot retrieve notifications after the enum conversion


CDN18:
### Describe the bug with a clear and concise description of what the bug is.

After updating to snapshot build cac9d650 and later, certain clients can't correctly retrieve notifications from gotosocial. Generally, these clients specify notification types in their requests, but some of these types aren't defined by gotosocial. There weren't any issues when the notification types in the database were set as `varchar`, but the problem arose after converting them to `smallint`.

The clients known to have issues:

- Mastodon Android and its forks (Megadon, Moshidon)
- Mastodon for iOS
- Soapbox
- Some closed-source clients: Mona (iOS), Trunks…

Logs:

Mastodon for Android and its' forks:

```
gotosocial  | timestamp="27/11/2024 11:29:09.077" func=api.(*Client).Route.TokenCheck.func1 level=TRACE requestID=tqbp2vck04001ce712cg msg="authenticated user 01GZJBZFQYY9NBFCWBVCR2KJQ7 with bearer token, scope is read write follow push"
gotosocial  | timestamp="27/11/2024 11:29:09.077" func=api.(*Client).Route.TokenCheck.func1 level=TRACE requestID=tqbp2vck04001ce712cg msg="authenticated client 01PSQDSD8EP6EJJBDMSCZYZBS8 with bearer token, scope is read write follow push"
gotosocial  | timestamp="27/11/2024 11:29:09.078" func=server.init.func1.Logger.13.1 level=INFO latency="412.24µs" userAgent=MastodonAndroid/2.9.3 method=GET statusCode=404 path=/api/v1/notifications/policy clientIP=192.168.2.1 errors="Error #01: sql: no rows in result set\n" requestID=tqbp2vck04001ce712cg msg="Not Found: wrote 45B"
gotosocial  | timestamp="27/11/2024 11:29:09.080" func=api.(*Client).Route.TokenCheck.func1 level=TRACE requestID=v3bp2vck04000cba6r40 msg="authenticated user 01GZJBZFQYY9NBFCWBVCR2KJQ7 with bearer token, scope is read write follow push"
gotosocial  | timestamp="27/11/2024 11:29:09.082" func=api.(*Client).Route.TokenCheck.func1 level=TRACE requestID=v3bp2vck04000cba6r40 msg="authenticated client 01PSQDSD8EP6EJJBDMSCZYZBS8 with bearer token, scope is read write follow push"
[...]

PR #3577, opened 2024-11-27
[bugfix] Log + ignore unknown notification types

tsmethurst:
# Description

>

We were being a bit too strict in returning bad request for requests for notification types we don't know about yet. This PR updates our logic to just debug log + ignore unknown notification types when requested via `/api/v1/notifications`, instead of returning 400 Bad Request.

Closes https://github.com/superseriousbusiness/gotosocial/issues/3576

## Checklist

### DIFF
--- a/internal/api/util/parsequery.go
+++ b/internal/api/util/parsequery.go
@@ -18,13 +18,13 @@
 package util
 
 import (
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/superseriousbusiness/gotosocial/internal/gtserror"
 	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/log"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/api/util/parsequery.go
```diff
@@ -218,51 +216,6 @@ func ParseInteractionReblogs(value string, defaultValue bool) (bool, gtserror.Wi
 	return parseBool(value, defaultValue, InteractionReblogsKey)
 }
 
-func ParseNotificationType(value string) (gtsmodel.NotificationType, gtserror.WithCode) {
```
NyaaaWhatsUpDoc, line 221: the semantics feels a bit strange to move this out of the apiutil package and rename the function to 'New___()', since this function *is* actually parsing a notification type from a string value. though i suppose if it gets used elsewhere you could just lift this into the typeutils package.

internal/gtsmodel/notification.go
```diff
@@ -40,6 +43,7 @@ type NotificationType enumType
 
 const (
 	// Notification Types
+	NotificationUnknown       NotificationType = 0  // NotificationUnknown -- unknown notification type, error if this occurs
```
NyaaaWhatsUpDoc, line 46: if you're adding this it would be worth adding a case for it to the String() function below so we can catch in logs when an unknown type is returned

internal/api/util/parsequery.go
tsmethurst: Ah I just did it to make it similar to other similar functions in the gtsmodel package, like `NewDomainPermissionType` and `NewAdminActionCategory`, else we're splitting the instantiation-from-string functionality across two different packages. I don't mind if you wanna move all of the `NewXYZ` functions to apiutil from gtsmodel though, but that should probably be a separate chore PR.

NyaaaWhatsUpDoc: even if we move that to a different package in a separate PR, i think the name should still be `Parse___()` (so remain as it is), as you're parsing an integer value from a string value.

internal/gtsmodel/notification.go
NyaaaWhatsUpDoc: actually saying this, the panic is probably a better indicator for us if we end up trying to serialize an unknown value (that's something that definitely shouldn't happen, it means we're storing unknown values)

internal/api/util/parsequery.go
tsmethurst: I'm cool with us changing all that yeah, Parse definitely makes more sense :) Just didn't feel like doing it in a bugfix PR, but I'm happy to do it in a chore PR a bit later if you want. You want me to put ParseNotificationType back in apiutil for now then?

NyaaaWhatsUpDoc: nah it's all good i'll put together a chore PR after this updating them :)

PR #3587, opened 2024-12-01
[feature] Push notifications

### LINKED ISSUE #1350: [feature] WebPush Support

Client that use the Mastodon WebPush API for notifications currently aren't able to send push notifications since GtS doesn't support the WebPush API.

GtS could implement the Mastodon push API (https://docs.joinmastodon.org/methods/push/) which would allow notifications to work in these clients.

I think this would involve something along the lines of:
 * Adding endpoints to subscribe to notifications, get, edit and remove subscriptions
 * Storing the current subscriptions in the database along with what they're subscribed to
 * A component that sends pokes to the endpoint given by the client in the subscription payload
 * Encrypting the payloads of those pokes with ECDH

Resources:
 * WebPush RFC: https://datatracker.ietf.org/doc/html/rfc8030
 * WebPush encryption RFC: https://datatracker.ietf.org/doc/html/rfc8291
 * Summary article: https://web.dev/push-notifications-web-push-protocol/

GtS supports pull-based notifications currently, do any apps that use this by polling will be able to support notifications without issue.

### Additional context.

This would fix https://github.com/superseriousbusiness/gotosocial/issues/1139

VyrCossont:
# Description

This pull request implements Web Push notifications.

Closes #1350

## Checklist

## TODO

- [ ] Notify admins when report is created
- [ ] Implement notification policy part of API
- [x] Test with stub Web Push server
- [x] Test with Feditext (ID decoding fixed, but other problems exist, see below)
- [x] Test with Phanpy
  - [x] Safari
  - [x] Firefox
  - [x] Chrome
  - [x] Arc
  - [x] Edge for Mac (Phanpy Web Push doesn't work, seems like a Microsoft problem)
  - [ ] Edge for Windows
- [ ] Test with Tusky
- [x] Fix notification titles
- [x] Cap notification body size
- [x] Stop sending to a Web Push server if HTTP 4xx status codes not related to rate limit or timeout are received (Mastodon and Akkoma both do this)
- [x] Remove cleartext HTTP URL scheme
- [ ] Try using [`Topic`](https://datatracker.ietf.org/doc/html/rfc8030#section-5.4) header for multiple updates to the same status (probably pointless until edits are implemented)


VyrCossont:
In process of testing with Phanpy (using Safari with Apple's Web Push server):

```
timestamp="01/12/2024 20:53:13.122" func=webpush.(*realSender).Send.func1 level=ERROR msg="error sending Web Push notification for subscription with token ID 01DAD4B0R9CHH0GRSS4RDGG4NV: sendToSubscription: unexpected HTTP status 403 Forbidden received when sending Web Push notification: {\"reason\":\"BadJwtToken\"}"
```

and Feditext:

```
timestamp="01/12/2024 20:53:13.048" func=httpclient.(*Client).DoOnce level=ERROR method=POST url=https://apns.feditext.com/push/67b68e16b1ec74cb00502345f647e11a3184de1ac6b90468fa4bb405585aafa2/61A54241-6446-4A26-A334-921B611E52F5 contentType=application/octet-stream attempt="1" msg="http response: 500 Internal Server Error"
```

VyrCossont:
Handling errors from Safari's Web Push server: https://developer.apple.com/documentation/usernotifications/sending-web-push-notifications-in-web-apps-and-browsers#Review-responses-for-push-notification-errors

internal/gtsmodel/webpushsubscription.go
```diff
@@ -0,0 +1,67 @@
...
+
+	// Auth is a Base64-encoded authentication secret.
+	Auth string `bun:",nullzero,notnull"`
+
+	// P256dh is a Base64-encoded Diffie-Hellman public key on the P-256 elliptic curve.
+	P256dh string `bun:",nullzero,notnull"`
+
+	// NotifyFollow and friends control which notifications are delivered to a given subscription.
+	// Corresponds to NotificationType and model.PushSubscriptionAlerts.
+	NotifyFollow           *bool `bun:",nullzero,notnull,default:false"`
```
tsmethurst, line 54: I wonder if we could save space by storing this as one integer column instead and having bits represent notification types, similar to what Mastodon does for roles? https://docs.joinmastodon.org/entities/Role/#permission-flags It means a bit of work parsing the number back and forth but it could be worthwhile. What do you think? No worries if this is a faff and you don't feel like it, we can always change it later if we really want it. I suppose not that many web push subscriptions will end up getting stored anyway so it might be a false economy.

internal/gtsmodel/vapidkeypair.go
```diff
@@ -0,0 +1,28 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package gtsmodel
+
+// VAPIDKeyPair represents the instance's VAPID keys (stored as Base64 strings).
+// This table should only ever have one entry, with a known ID of 0.
+//
+// See: https://datatracker.ietf.org/doc/html/rfc8292
+type VAPIDKeyPair struct {
+	ID      int    `bun:",pk,notnull"`
```
tsmethurst, line 25: Is the id field necessary at all here if there's only ever one entry?

cmd/gotosocial/action/server/server.go
```diff
@@ -242,6 +245,25 @@ var Start action.GTSAction = func(ctx context.Context) error {
 		}
 	}
 
+	// Get or create a VAPID key pair.
+	vapidKeyPair, err := dbService.GetVAPIDKeyPair(ctx)
```
tsmethurst, line 249: You could probably just shift this creation logic back to the db function and call it "GetOrCreateVAPIDKeyPair" imo. I don't think there's any cases where you'd call the function to get it and not want to create it afterwards if it didn't exist. Or maybe there are? I'm not 100% sure.

tsmethurst:
Looking great so far, very exciting :)

internal/gtsmodel/webpushsubscription.go
NyaaaWhatsUpDoc: exactly what i was thinking. though maybe worth specifically using a set-size integer just to sure we always have the expected number of bits available to set

VyrCossont:
So here's a fun thing I just learned:
- Mastodon [sends extra HTTP headers called `Encryption` and `Crypto-Key`](https://github.com/mastodon/mastodon/blob/aeae1520068c54cb51dd3a350276df393ef43149/app/workers/web/push_notification_worker.rb#L32-L33) that existed in [early versions of RFC 8188](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-encryption-encoding-03#section-7.2) Encrypted Content-Encoding for HTTP but do not exist in [the final version of RFC 8188](https://datatracker.ietf.org/doc/html/rfc8188). It also uses the draft `Content-Encoding: aesgcm` header instead of the final `Content-Encoding: aes128gcm`
- Feditext's APNS relay [expects those headers](https://github.com/feditext/feditext-apns/blob/0b5adf5df1c26a78815885c9b9a53d9a4e12719b/app.rb#L43-L44)
- [`webpush-go`](https://github.com/SherClockHolmes/webpush-go) does not send those headers
- The headers may be redundant with the header fields at the start of the binary content blob, in which case the bug is arguably with Feditext's APNS relay, but other Web Push -> native relays may also expect the headers and the draft `Content-Encoding: aesgcm`, definitely including [Toot!'s](https://github.com/DagAgren/toot-relay/blob/0154b4297d1acf399c1c5a22dcd65310959955b6/toot-relay.go#L115-L133)
    - Tusky apparently does Web Push support through [UnifiedPush](https://unifiedpush.org/). [`PushNotificationHelper.kt`](https://github.com/tuskyapp/Tusky/blob/50ca44a5f6fa0a4733b5500376ed8ef4ff723189/app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/PushNotificationHelper.kt#L182-L194) suggests that UnifiedPush doesn't work with Mastodon's obsolete version of Web Push encryption, in which case implementing Web Push correctly might allow Tusky to do less network I/O when working with GtS in the future.

internal/gtsmodel/webpushsubscription.go
VyrCossont: Sure, why not. Use a 64-bit integer and we'll have enough PN type bits forever.

VyrCossont:
- [Diff of draft 3 and final versions of RFC 8188](https://author-tools.ietf.org/iddiff?url1=draft-ietf-httpbis-encryption-encoding-03&url2=rfc8188&difftype=--html)
- [Current version of gem used for Mastodon web push encryption](https://github.com/ClearlyClaire/webpush/blob/f14a4d52e201128b1b00245d11b6de80d6cfdcd9/lib/webpush/encryption.rb)
- [RFC 8291: Message Encryption for Web Push](https://datatracker.ietf.org/doc/html/rfc8291) and its older drafts may also have some bearing on this whole deal

Horrible thought: we could have both standards-compliant Web Push encryption and Mastodon Clown Car Web Push encryption implementations and send both for every Web Push notification, or at least until we get an error back from the Web Push server that implies we should use one or the other

leo60228:
This is not a Mastodon specific issue. The [web-push JS library](https://github.com/web-push-libs/web-push) which is cited by both Mozilla and Google and is seemingly some sort of reference implementation still defaults to using these headers and requires explicit opt-in (`{contentEncoding: 'aes128gcm'}`) to use the final encryption standard. [Google's documentation](https://web.dev/articles/push-notifications-web-push-protocol) also says to include them, despite citing the final specs.

VyrCossont:
Sounds like I need to check whether Mozilla and Google Web Pusb servers can take `aes128gcm`. Hopefully they do, in which case we don't have to implement `aesgcm`.

cmd/gotosocial/action/server/server.go
VyrCossont: I guess as long as we use `INSERT … ON CONFLICT IGNORE` followed by re-fetching from the DB on failure to insert, we don't have to worry about accidentally overwriting VAPID keys already in use.

internal/gtsmodel/vapidkeypair.go
VyrCossont: Two reasons, neither of which I'm actually using right now:
1. We need an ID column to use `ON CONFLICT` clauses
2. The single-row constraint can be enforced by the DB if we add a `CHECK (id = 0)` clause when creating the table, although it's not clear to me how to do that in Bun

tsmethurst:
Let me know if you need any help with this / want an NTFY set up on superseriousbusiness.org or anything like that :) You seem to be pushing through it fine but I'm happy to help if I can!

VyrCossont:
So we apparently don't need to worry about Chrome or Firefox (at least the name-brand versions, as their push servers handle `aes128gcm` just fine.

Safari is reporting problems with the (VAPID) JWT, and it looks like `webpush-go` always prepends `"mailto:"` to whatever string you give it for a subject. [This is used for contact information](https://datatracker.ietf.org/doc/html/rfc8292#section-2.1) and is supposed to either take a `mailto:` or `https:` URL. I hate email and my GtS server doesn't have a contact email configured, so I'd elected to provide the instance's `https:` URL as contact info, which would result in a malformed `mailto:` URL actually being used for the `sub` claim by `webpush-go`. I'll try giving it a dummy email if no contact email is available, and if that works, I'll open a `webpush-go` bug to allow `https:` subjects somehow.

VyrCossont:
Yep, that worked, we're good on Safari now.

internal/gtsmodel/webpushsubscription.go
VyrCossont: Done.

cmd/gotosocial/action/server/server.go
VyrCossont: Implemented.

internal/gtsmodel/vapidkeypair.go
VyrCossont: Addendum: Bun doesn't support `CHECK` clauses and I couldn't figure out a way to wedge them in easily, but as long as our ID is always 0, the PK constraint will have the same effect.

VyrCossont:
Mastodon just added support for modern Web Push (what I'm implementing here) with mastodon/mastodon#33528 which should improve the state of the ecosystem a bit.

VyrCossont:
We really should be doing some kind of Unicode bidirectional text isolation when building notification titles from display names, and indeed anywhere that we build a string from partially attacker-controlled text, but we should probably try to solve that for all of GtS and add it to this later. It's a huge pain in the ass, and there are bugs in the nascent Go library for dealing with bidi text:
- https://meta.stackoverflow.com/questions/310158/right-to-left-marker-in-comment-shouldnt-cause-the-rest-of-the-line-to-change/310228#310228
- https://github.com/golang/go/issues/69819

VyrCossont:
OK, this is ready for review. I'll handle the policy, admin report, and `standard` compatibility in followup PRs.

internal/db/bundb/migrations/20241124012636_add_web_push_subscriptions.go
```diff
@@ -0,0 +1,66 @@
...
+				NewCreateTable().
+				Model(&gtsmodel.WebPushSubscription{}).
+				IfNotExists().
+				Exec(ctx); err != nil {
+				return err
+			}
+
+			for index, columns := range map[string][]string{
+				"web_push_subscriptions_account_id_idx": {"account_id"},
+				"web_push_subscriptions_token_id_idx":   {"token_id"},
```
tsmethurst, line 40: Since `token_id` is already unique in the table you don't need an index for it here :)

internal/webpush/realsender.go
```diff
@@ -0,0 +1,315 @@
...
+	// Load subscriptions.
+	subscriptions, err := r.state.DB.GetWebPushSubscriptionsByAccountID(ctx, notification.TargetAccountID)
+	if err != nil {
+		return gtserror.Newf(
+			"error getting Web Push subscriptions for account %s: %w",
+			notification.TargetAccountID,
+			err,
+		)
+	}
+	if len(subscriptions) == 0 {
```
tsmethurst, line 85: I think you can drop this check and just keep the check below after filtering out subs for irrelevant notif types.

internal/webpush/realsender.go
```diff
@@ -0,0 +1,315 @@
...
+			Subscriber:      vapidSubjectEmail,
+			VAPIDPublicKey:  vapidKeyPair.Public,
+			VAPIDPrivateKey: vapidKeyPair.Private,
+			TTL:             int(TTL.Seconds()),
+		},
+	)
+	if err != nil {
+		return gtserror.Newf("error sending Web Push notification: %w", err)
+	}
+	defer func() {
```
tsmethurst, line 259: You can just `defer resp.Body.Close()`

internal/webpush/realsender.go
```diff
@@ -0,0 +1,315 @@
...
+		_ = resp.Body.Close()
+	}()
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
```
tsmethurst, line 263: I think this logic could be a bit easier to read if you turn it into a switch statement, something like:

```go
switch {
	// All good, delivered.
	case resp.StatusCode >= 200 && resp.StatusCode <= 299:

	// Temporary outage or some other delivery issue.
	case resp.StatusCode == http.StatusRequestTimeout ||
		resp.StatusCode == http.StatusRequestEntityTooLarge ||
		resp.StatusCode == http.StatusTooManyRequests ||
		resp.StatusCode == http.StatusServiceUnavailable:
		
		// Try to get the response body.
		bodyBytes, err := io.ReadAll(io.LimitReader(resp.Body, responseBodyMaxLen))
		if err != nil {
			return gtserror.Newf("error reading Web Push server response: %w", err)
		}

		// Return the error with its response body.
		return gtserror.Newf(
			"unexpected HTTP status %s received when sending Web Push notification: %s",
			resp.Status,
			string(bodyBytes),
		)

	// Some serious error that indicates auth problems.
	// We should not send any more notifications to this subscription. Try to delete it.
	default:
		err := r.state.DB.DeleteWebPushSubscriptionByTokenID(ctx, subscription.TokenID)
		if err != nil {
			return gtserror.Newf(
				"received HTTP status %s but failed to delete subscription: %s",
				resp.Status,
				err,
			)
		}
			
		log.Infof(
			ctx,
			"Deleted Web Push subscription with token ID %s because push server sent HTTP status %s",
			subscription.TokenID, resp.Status,
		)
		return nil
}
```

tsmethurst: This way you can also drop the final "return" statement that comes afterwards.

internal/webpush/realsender.go
```diff
@@ -0,0 +1,315 @@
...
+// This should only be used by NewSender and in tests.
+func NewRealSender(httpClient *http.Client, state *state.State) Sender {
+	return &realSender{
+		httpClient: httpClient,
+		state:      state,
+		tc:         typeutils.NewConverter(state),
+	}
+}
+
+// TTL is an arbitrary time to ask the Web Push server to store notifications
```
tsmethurst, line 59: If I'm reading this correctly, these consts are only used inside the `sendToSubscription` function, so you could put them in there as a const block instead like this:

```go
// sendToSubscription sends a notification to a single Web Push subscription.
func (r *realSender) sendToSubscription(
	ctx context.Context,
	vapidKeyPair *gtsmodel.VAPIDKeyPair,
	vapidSubjectEmail string,
	subscription *gtsmodel.WebPushSubscription,
	notification *gtsmodel.Notification,
	apiNotification *apimodel.Notification,
) error {
	const (
		// An arbitrary time to ask the Web Push server to store
		// notifications while waiting for the client to retrieve them.
		TTL = 48 * time.Hour

		// Limits how much of the Web Push server
		// response we read for error messages.
		responseBodyMaxLen = 1024

		// A polite maximum length for a Web Push
		// notification's body text, in bytes.
		//
		// Note that this isn't limited per se, but
		// Web Push servers may reject anything
		// with a total request body size over 4k.
		bodyMaxLen = 3000
	)
}
```

internal/processing/push/create.go
```diff
@@ -0,0 +1,67 @@
...
+		AccountID:         accountID,
+		TokenID:           tokenID,
+		Endpoint:          request.Subscription.Endpoint,
+		Auth:              request.Subscription.Keys.Auth,
+		P256dh:            request.Subscription.Keys.P256dh,
+		NotificationFlags: alertsToNotificationFlags(request.Data.Alerts),
+	}
+
+	if err := p.state.DB.PutWebPushSubscription(ctx, subscription); err != nil {
+		return nil, gtserror.NewErrorInternalError(
```
tsmethurst, line 61: This is just a preference thing so feel free to ignore it, but I think it's generally a bit nicer when wrapping errors to do it like this:

```go
if err := p.state.DB.PutWebPushSubscription(ctx, subscription); err != nil {
	err := gtserror.Newf("couldn't create Web Push subscription for token ID %s: %w", tokenID, err)		
}

return p.apiSubscription(ctx, subscription)
```

If you do decide to change it, could you change it in the other places in the code as well? If you prefer not to, also fine :P Like I said, this is just a v. minor preference thing.

tsmethurst:
This is looking great! I have a few comments but nothing major, almost ready to squerge imo :) Great work @VyrCossont !

internal/db/bundb/migrations/20241124012636_add_web_push_subscriptions.go
VyrCossont: Fixed.

internal/webpush/realsender.go
VyrCossont: Sure, I guess it's idiomatic enough that we don't need to explicitly ignore the error.

internal/webpush/realsender.go
VyrCossont: Yeah, agreed, that'll be more readable.

internal/processing/push/create.go
VyrCossont: Sure. Saves us a line when creating a wrapped error like this.

internal/webpush/realsender.go
VyrCossont: Done. Also broke out the title and body formatting to separate functions so this one doesn't get too big.

VyrCossont:
One thing I've noted while testing this patch in prod is that Phanpy can successfully subscribe to push notifications, but reloading Phanpy's setting page doesn't _show_ that it's subscribed, even though it is. This seems likely to be Phanpy expecting the `policy` field, so I may stub that once I confirm the cause, though I still plan to implement it completely in a separate PR.

VyrCossont:
Partially fixed by stubbing `policy`. Phanpy's subscription status is now correct, but notification type checkboxes always show checked regardless of whether the type is set or not. This looks like a Phanpy bug: it seems to be looking at keys but not values in the `alerts` dict, and Mastodon only sends notification type keys for enabled notification types (a mismatch with [its own docs](https://docs.joinmastodon.org/entities/WebPushSubscription/)). I'll poke Phanpy later.

internal/webpush/realsender.go
```diff
@@ -0,0 +1,341 @@
...
+// realSender is the production Web Push sender, backed by an HTTP client, DB, and worker pool.
+type realSender struct {
+	httpClient *http.Client
+	state      *state.State
+	tc         *typeutils.Converter
+}
+
+// NewRealSender creates a Sender from an http.Client instead of an httpclient.Client.
+// This should only be used by NewSender and in tests.
+func NewRealSender(httpClient *http.Client, state *state.State) Sender {
```
NyaaaWhatsUpDoc, line 52: i think we should pass in a type-converter instance instead of instantiating it here, just so they all refer to the same instance. just so we don't say add some internal structure to it and don't wonder why they aren't all in sync :p

internal/webpush/mocksender.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+
+	"github.com/superseriousbusiness/gotosocial/internal/filter/usermute"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+)
+
+// MockSender collects a map of notifications sent to each account ID.
+// This should only be used in tests.
+type MockSender struct {
```
NyaaaWhatsUpDoc, line 29: i think this should be moved to the testrig if it's going to be exported

internal/webpush/noopsender.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+
+	"github.com/superseriousbusiness/gotosocial/internal/filter/usermute"
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+)
+
+// noopSender drops anything sent to it.
+// This should only be used in tests.
+type noopSender struct{}
```
NyaaaWhatsUpDoc, line 29: again if this is only used in tests (and not also used for say no push server configured), then it should be moved to the testrig so we don't have test code exported and visible in non-test code

internal/webpush/realsender.go
VyrCossont: Oh, good idea.

internal/webpush/noopsender.go
VyrCossont: Clients configure push servers, not GtS, so it's not used for that.

Moving to testrig as with above.

PR #3597, opened 2024-12-03
[feature] add support for receiving federated status edits

NyaaaWhatsUpDoc:
# Description

Adds support for GoToSocial properly handling federated status edits. When we receive a status update activity we now correctly store the current state as a historical edit in a new database table, and update the state to the newly received edit. This also handles the case of re-fetching a status we haven't received an update for, but then notice a change in.

Some other changes of note:
- adds getBunColumnDef() migration utliity function which generates the database column definition that bun would normally generate for a given struct field, which greatly simplifies manually adding new columns not having to write the type manually
- drops the unused mentions.updated_at database column
- drops the unused media_attachments.updated_at database column
- adds compile-time type checking of a bunch of our AP interface-able types
- removes the error signature from id.NewULIDFromTime() as we handle the end result in exactly the same manner all over the codebase
- tweaks federated status updates to now only update the necessary columns, possibly now a little cheaper computationally

## Checklist


cmd/gen-ulid/main.go
```diff
@@ -0,0 +1,22 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package main
+
+import "github.com/superseriousbusiness/gotosocial/internal/id"
+
+func main() { println(id.NewULID()) }
```
NyaaaWhatsUpDoc, line 22: this is just a tiny useful utility for generating new testmodel ulids using exactly the version / library from gotosocial

internal/processing/timeline/public_test.go
```diff
@@ -102,8 +102,8 @@ func (suite *PublicTestSuite) TestPublicTimelineGetHideFiltered() {
 		requester           = suite.testAccounts["local_account_1"]
 		maxID               = ""
 		sinceID             = ""
-		minID               = "01F8MHAAY43M6RJ473VQFCVH36" // 1 before filteredStatus
```
NyaaaWhatsUpDoc, line 105: i removed this as the test has nothing to do with paging, and designed this way it requires you to update it every time you add new statuses which is a pain in the ass :p

tsmethurst:
Big juicy PR! Gonna review it now.

internal/ap/interfaces_test.go
```diff
@@ -0,0 +1,93 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+package ap_test
+
+import (
+	"github.com/superseriousbusiness/activity/streams/vocab"
+	"github.com/superseriousbusiness/gotosocial/internal/ap"
+)
+
+var (
```
tsmethurst, line 25: Nice one, good idea.

internal/federation/dereferencing/util.go
```diff
@@ -70,3 +70,11 @@ func pollUpdated(existing, latest *gtsmodel.Poll) bool {
 func pollJustClosed(existing, latest *gtsmodel.Poll) bool {
 	return existing.ClosedAt.IsZero() && latest.Closed()
 }
+
+// statusChanged returns whether a status has changed in a way that
+// indicates that existing should be snapshotted for version history.
+func statusChanged(existing, latest *gtsmodel.Status) bool {
+	return existing.Content != latest.Content ||
+		existing.ContentWarning != latest.ContentWarning ||
+		!slices.Equal(existing.AttachmentIDs, latest.AttachmentIDs)
```
tsmethurst, line 79: Is it possible to also edit image alt text and have that appear as an update? I'm not sure how it works exactly.

internal/db/bundb/migrations/util.go
```diff
@@ -19,11 +19,209 @@ package migrations
...
+	return nil
+}
+
+// getBunColumnDef generates a column definition string for the SQL table represented by
+// Go type, with the SQL column represented by the given Go field name. This ensures when
+// adding a new column for table by migration that it will end up as bun would create it.
+//
+// NOTE: this function must stay in sync with (*bun.CreateTableQuery{}).AppendQuery(),
+// specifically where it loops over table fields appending each column definition.
+func getBunColumnDef(db bun.IDB, rtype reflect.Type, fieldName string) (string, error) {
```
tsmethurst, line 135: Handy :)

internal/db/bundb/migrations/20241113152126_add_status_edits.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+
+			// Create the main StatusEdits table.
+			_, err = tx.NewCreateTable().
+				IfNotExists().
+				Model((*gtsmodel.StatusEdit)(nil)).
+				Exec(ctx)
+			return err
+		})
+	}
```
tsmethurst, line 56: Probably a good idea to also add an index on status_id here, since we'll be selecting using that a lot.

tsmethurst: Oh wait no the status itself stores a list of edit IDs doesn't it? Never mind.

tsmethurst:
Looking great :)

internal/db/bundb/migrations/20241113152126_add_status_edits.go
NyaaaWhatsUpDoc: indeed. though it's useful to keep this back reference to the status_id around ready for when we implement cleaning up of old edits and old statuses, and at that point it might be useful to add an index :)

internal/federation/dereferencing/util.go
NyaaaWhatsUpDoc: we already handle that, though currently in the fetchStatusAttachments() function so tbh i might do some fiddling and also see if mastodon handles media description changes as historic status edits

tsmethurst:
When you're happy then squerge away pal :)

internal/federation/dereferencing/util.go
NyaaaWhatsUpDoc: there we go, tweaked! 534792467d85427cfc52ab0b33c181b3a4c81076

PR #3601, opened 2024-12-05
[bugfix] Load instance-wide custom css in page stylesheets template

### LINKED ISSUE #3600: [bug/frontend] Instance-wide custom css doesn't apply to some pages

Following #3594, I just found more page that the instance-wide custom css doesn't apply to:

- `/auth/sign_in`
- OAuth Authorization
- `/signup`
- the 404 page

The following pages might also be affected, but I haven't tested them specifically yet:
- OAuth Finalization
- the Error page

v0.17.4-SNAPSHOT+git-3e18d97

### Browser version

Firefox 132.0 (64-bit)

### What happened?

`/custom.css` does not load.

`/custom.css` loaded successfully.

### How to reproduce it?

1. Apply instance-wide custom css.
2. Navigate to pages listed above

_No response_

CDN18:
# Description

closes #3600

After making this modification to `page_stylesheets.tmpl`, all pages mentioned in #3600 can now load `/custom.css`. On the profile page, `/custom.css` still loads before `/@username/custom.css`, and it seems there’s no redundant import for pages like `/about` where `instanceCustomCSSPath` was loaded in `internal/web` beforehand. So I’m not entirely sure if we should remove the reference to `instanceCustomCSSPath` in `internal/web`.

Additionally, the changes in this PR use a hardcoded path for `/custom.css`. When updating `instanceCustomCSSPath` in the future, we'll also need to update the content in `page_stylesheets.tmpl`. Alternatively, we could consider changing `customCSS` to `customCSSPath` in the instance model, which would allow for unified referencing in the Go template imo.

Edit: sorry, pushed wrong commit

## Checklist

### DIFF
--- a/internal/web/about.go
+++ b/internal/web/about.go
@@ -54,7 +54,7 @@ func (m *Module) aboutGETHandler(c *gin.Context) {
 		Template:    "about.tmpl",
 		Instance:    instance,
 		OGMeta:      apiutil.OGBase(instance),
-		Stylesheets: []string{cssAbout, instanceCustomCSSPath},
+		Stylesheets: []string{cssAbout},
 		Extra: map[string]any{
 			"showStrap":        true,
 			"blocklistExposed": config.GetInstanceExposeSuspendedWeb(),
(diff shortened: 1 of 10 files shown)


tsmethurst:
I think with this change, we should probably remove all the references to custom CSS in the web handlers, as they're no longer necessary. This way of handling it seems like a more elegant solution to make sure it shows up on all pages, thanks for poking at it :)

tsmethurst:
Thank you very much!

PR #3604, opened 2024-12-07
[feature] add support for sending headers to tracing system

joschi:
# Description

Add support for sending additional HTTP or gRPC headers which can be used for authentication or other additional information for the tracing system without having to set up a local instance of the OpenTelemetry Collector to add these headers.

Example with [Dash0](https://www.dash0.com/):

```yaml
tracing-enabled: false
tracing-transport: "grpc"
tracing-endpoint: "ingress.eu-west-1.aws.dash0.com:4317"
tracing-headers:
  "Authorization": "Bearer DASH0_AUTH_TOKEN"
  "Dash0-Dataset": "gotosocial"
```

Example with [Honeycomb](https://www.honeycomb.io/):

```yaml
tracing-enabled: false
tracing-transport: "grpc"
tracing-endpoint: "api.honeycomb.io:443"
tracing-headers:
  "x-honeycomb-team": "YOUR_API_KEY"
  "x-honeycomb-dataset": "YOUR_DATASET"
```

refs #1230

See also https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp@v1.32.0#WithHeaders.

## Checklist

### DIFF
--- a/docs/locales/zh/configuration/observability.md
+++ b/docs/locales/zh/configuration/observability.md
@@ -31,6 +31,11 @@ tracing-transport: "grpc"
 # 默认值: ""
 tracing-endpoint: ""
 
+# TODO
+# 示例: {"Authorization": "Bearer super-secret-token", "Dataset": "gotosocial"}
+# 默认值: {}
+tracing-headers: {}
+
 # 布尔值。禁用gRPC和HTTP传输协议的TLS。
 # 默认值: false
 tracing-insecure-transport: false
--- a/internal/config/helpers.gen.go
+++ b/internal/config/helpers.gen.go
@@ -2191,6 +2191,22 @@ func (st *ConfigState) SetTracingEndpoint(v string) {
 	st.reloadToViper()
 }
 
+// GetTracingHeaders safely fetches the Configuration value for state's 'TracingHeaders' field
+func (st *ConfigState) GetTracingHeaders() (v map[string]string) {
+	st.mutex.RLock()
+	v = st.config.TracingHeaders
+	st.mutex.RUnlock()
+	return
+}
+
+// SetTracingEndpoint safely sets the Configuration value for state's 'TracingHeaders' field
+func (st *ConfigState) SetTracingHeaders(v map[string]string) {
+	st.mutex.Lock()
+	defer st.mutex.Unlock()
+	st.config.TracingHeaders = v
+	st.reloadToViper()
+}
+
 // TracingEndpointFlag returns the flag name for the 'TracingEndpoint' field
 func TracingEndpointFlag() string { return "tracing-endpoint" }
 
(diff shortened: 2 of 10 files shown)


docs/locales/zh/configuration/observability.md
```diff
@@ -31,6 +31,11 @@ tracing-transport: "grpc"
 # 默认值: ""
 tracing-endpoint: ""
 
+# TODO
```
joschi, line 34: note: help would be appreciated here for a proper translation.

tsmethurst:
Oh this looks interesting! :)

internal/config/helpers.gen.go
```diff
@@ -2191,6 +2191,22 @@ func (st *ConfigState) SetTracingEndpoint(v string) {
 	st.reloadToViper()
 }
 
+// GetTracingHeaders safely fetches the Configuration value for state's 'TracingHeaders' field
+func (st *ConfigState) GetTracingHeaders() (v map[string]string) {
```
tsmethurst, line 2195: Make sure you've regenerated this file using `go run ./internal/config/gen/ -out ./internal/config/helpers.gen.go` rather than editing it manually (I think that's what you did here?).

joschi: Yes, you're correct. I've regenerated the file and pushed the commit. 😄

tsmethurst: Thanks :) Could you also run `go fmt ./internal/config/helpers.gen.go`? It looks like some things got pushed around in the file.

internal/tracing/tracing.go
```diff
@@ -60,6 +60,7 @@ func Initialize() error {
 	case "grpc":
 		opts := []otlptracegrpc.Option{
 			otlptracegrpc.WithEndpoint(config.GetTracingEndpoint()),
+			otlptracegrpc.WithHeaders(config.NewState().GetTracingHeaders()),
```
daenney, line 63: I don't think this should be calling `NewState()`.

daenney:
I actually wonder if we shouldn't just tell people to use the OTLP environment variables instead. Otherwise we're going to have to add configuration options over time for anything the SDK already supports.

For example, you could set those values already with `OTEL_EXPORTER_OTLP_TRACES_HEADERS` and they should get picked up. By explicitly using `WithHeaders` we actually disable that facility.

I'm not entirely sure where to draw the line with this. Realistically, we can remove the `tracing-*` config options entirely since everything we have can already be done using the `OTLP_*` environment variables.

tsmethurst:
Thanks for the poke! We've been busy with the 0.18.0 release candidates so (i personally at least) haven't had much time to review other PRs. Since @daenney and @theSuess have both worked with the OTEL stuff and I haven't at all, I defer to them!

daenney:
We've migrated to Codeberg. Your PR has been imported as: https://codeberg.org/superseriousbusiness/gotosocial/pulls/3604.

PR #3613, opened 2024-12-09
[chore] Log during potentially long migrations

tsmethurst:
Log step-by-step during new migrations so admins know what's going on.


internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
```diff
@@ -81,14 +84,17 @@ func init() {
 
 			// Convert all visibility tables.
 			for _, table := range visTables {
+				log.Info(ctx, "remapping %s.%s to save space, please wait...", table.Table, table.Column)
```
NyaaaWhatsUpDoc, line 87: the convertEnums() function already adds logging

tsmethurst: Ah thanks! I'll drop those.

PR #3631, opened 2024-12-24
[bugfix] small editing tweaks

NyaaaWhatsUpDoc:
# Description

- ensures that edited_at is never set for boosts
- adds better handling of the updated_at field for remote statuses

sometime soon i'll make a PR that just drops the updated_at column and instead has an optional edited_at column. though for now this should make it behave a bit better for users currently on main, and the PR for dropping that column and adding a new one will be a monstrous migration that requires iterating the entire statuses table :')

## Checklist


NyaaaWhatsUpDoc:
actually, got one last tweak to this before it's ready to squerge

Issue #3531, opened 2024-11-09
[bug] MP3 upload support not advertised using standard mime type


graue:
### Describe the bug with a clear and concise description of what the bug is.

The standard mime type for MP3 files is `audio/mpeg`, not `audio/mp3` which is [deprecated](https://mimetype.io/audio/mp3). However, the instance configuration endpoint in GTS says only that `audio/mp3` is supported, and does not include `audio/mpeg` under `supported_mime_types`.

This causes at least one client (Phanpy) to [incorrectly report that MP3 uploads are not supported on GTS servers](https://github.com/cheeaun/phanpy/issues/898).

Although clients _could_ write custom code to handle this, it's easier if GTS just advertises support for the standard mime type, `audio/mpeg`.  (For what it's worth, Mastodon currently advertises support for both `audio/mpeg` and `audio/mp3`.)

### What's your GoToSocial Version?

0.17.3+git-6f4cb2f

### What happened?

Phanpy errored, saying mp3 uploads were not supported by GTS

### What you expected to happen?

The mp3 successfully uploads

### How to reproduce it?

Using Phanpy, attempt to upload an mp3 file to a GTS server

tsmethurst:
Ah thanks!

PR #3632, opened 2024-12-27
[bugfix] Advertise audio/mpeg as supported

daenney:
# Description

This will be either an mp1, mp2 or mp3 file. In practice it'll probably be mp3, but this handles mp1 too for good measure. We don't advertise audio/mp1 as a supported media type since best I can tell that was never a MIME type that's been used.

This also changes the returned MIME-type for mp2 and mp3 to audio/mpeg, to match what's expected and supported by most things nowadays.

Fixes: #3531

## Checklist


tsmethurst:
Nice!

PR #3636, opened 2025-01-06
[chore] replace statuses.updated_at column with statuses.edited_at

NyaaaWhatsUpDoc:
# Description

- replaces the `statuses.updated_at` column with `statuses.edited_at`, a new nullable column, and carefully only copying over this value for existing statuses if we have edits stored for it.
- fixes status.EditIDs being wiped-out on every dereference of a remote status

## Checklist

- [x] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [x] I/we have discussed the proposed changes already, either in an issue on the repository, or in the Matrix chat.
- [x] I/we have not leveraged AI to create the proposed changes.
- [x] I/we have performed a self-review of added code.
- [x] I/we have written code that is legible and maintainable by others.
- [x] I/we have commented the added code, particularly in hard-to-understand areas.
- [ ] I/we have made any necessary changes to documentation.
- [ ] I/we have added tests that cover new code.
- [x] I/we have run tests and they pass locally with the changes.
- [x] I/we have run `go fmt ./...` and `golangci-lint run`.

PR #3637, opened 2025-01-08
[bugfix] migration to cleanup dropped status edits

NyaaaWhatsUpDoc:
# Description

A migration that fixes the resulting data of a bug introduced with status edits, but fixed in #3636 

This finds all the dropped status edits with a status to relink to and reinserts it back into the status' list of edits.

This shouldn't do anything for anyone that has this migration run on the final release (as they won't have had the bug!), but it will fix up the databases of those running the snapshot currently :)

## Checklist


NyaaaWhatsUpDoc:
marking as draft as postgres doesn't seem to like this for reasons

NyaaaWhatsUpDoc:
confirmed working now on a live instance

Issue #1891, opened 2023-06-14
[chore/bugfix] Implement `federator.FederatingCallbacks` and `federatingDB.Exists` properly


tsmethurst:
Right now, due to an unfortunate fuck-up by *YOURS TRULY* years ago, our go-fed FederatingCallbacks and Exists functions aren't implemented properly:

- Exists always returns `false` without doing any checks.
- FederatingCallbacks uses default behavior for too many things (blocks, for example).

The upshot of these misimplementations is that many activities are not handled properly in the initial call to resolve the federating callbacks, and so they're only handled properly *by accident* when the InboxForwarding function calls Create on an activity (when Exists returns false). This causes annoyances like https://github.com/superseriousbusiness/gotosocial/issues/1744 and other things.

Steps to resolve this:

1. ~~Implement Exists properly to check for (at least) statuses and accounts.~~ #3647
2. ~~Take that switch statement out of federatingDB.Create so that it's not doing all the heavy lifting, and move the other functionality into separate federatingDB functions.~~ #3697
3. ~~Override other federating callbacks to call those new functions created in step 2.~~

These fixes will *also* pave the way to potentially implementing inbox forwarding (https://github.com/superseriousbusiness/gotosocial/issues/1890).

NyaaaWhatsUpDoc:
Important comment from the above PR:

> Unfortunately we can never return a definite answer in Exists(), since the activity JSONLD-ID doesn't necessarily match the stored object's JSONLD-ID, which is the only part that we actually store in a lot of cases. So this provides a best-effort implementation to at least reduce unnecessary calls for recently handled activities.

PR #3647, opened 2025-01-15
[performance] reduce InboxForward->Create calls by partially implementing Exists()

NyaaaWhatsUpDoc:
# Description

This adds a simple in-memory map that stores the IDs of activities that we have recently created, which allows us to partially implement Exists() and reduce the number of unnecessary calls to Create() during inbox forwarding. Unfortunately we can never return a definite answer in Exists(), since the **activity** JSONLD-ID doesn't necessarily match the stored **object**'s JSONLD-ID, which is the only part that we actually store in a lot of cases. So this provides a best-effort implementation to at least reduce unnecessary calls for recently handled activities.

In the future, I don't think we should strive to "fully" implement Exists(), instead focusing on removing our reliance on the go-fed code for federation logic.

Partially handles #1891 

## Checklist

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -40,17 +40,17 @@ type Caches struct {
 	// the block []headerfilter.Filter cache.
 	BlockHeaderFilters headerfilter.Cache
 
+	// TTL cache of statuses -> filterable text fields.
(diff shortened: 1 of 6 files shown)


tsmethurst:
Nice little optimization! :)

PR #3668, opened 2025-01-20
[feature] Implement `deliveryRecipientPreSort` to prioritize delivery to mentioned accounts

tsmethurst:
# Description

>

Because of the way Activity expands addressees into inboxes, when writing a post mentioning an account that doesn't follow you, it's possible to attempt delivery to all of your followers first, and only then to attempt a delivery to the mentioned account. This can lead to laggy-feeling conversations.

This pull request implements the DeliveryRecipientPreSort function added here -- https://github.com/superseriousbusiness/activity/pull/26 -- to make it so that GtS prioritizes delivery of activities to direct conversation participants first, and only then to followers.

## Checklist

- [ ] I/we have read the [GoToSocial contribution guidelines](https://github.com/superseriousbusiness/gotosocial/blob/main/CONTRIBUTING.md).
- [ ] I/we have not leveraged AI to create the proposed changes.


NyaaaWhatsUpDoc:
this is a very good idea!! though i think implementation-wise this can be greatly simplified. since the "raw" JSON is passed in to our BatchDeliver() function, you can grab the mentions property array from the object (if it exists, ofc), and grab a list of the mentionee URLs there. then essentially perform your sort function in BatchDeliver(). it saves needing to modify the go-fed/activity library

tsmethurst:
Discussed on Matrix!

> If you don't mind I would quite like to do it in the activity library, for two reasons: 1. Other people can use it then. 2. It gets sorted before expansion into inboxes, so it's a much smaller slice to sort. We're usually only sorting like three entries at most that way, compared to hundreds or thousands if we sort after expansion into inbox URIs

PR #3688, opened 2025-01-26
[feature] Add database migration rollback command

DysphoricUnicorn:
# Description

This pull request adds a CLI command that allows administrators to rollback the last migration that was run.  
When I updated my instance over the holidays, I ran into an issue because the PV containing my database had run out of space. This left the database in a completely broken status and, being a bit too trusting since all previous updates completed without any issues, I did not have a backup handy.  
Since bun has a rollback command, I thought it would be useful to be able to trigger a migration rollback using the GtS CLI, and it actually was able to resurrect my instance without any data loss.  

I am creating a merge request for this in the hope that it will be useful to others. I am willing to update the code for the CLI command if needed in the future, so even though this does not implement a previously existing feature request, it is not a drive-by MR.

## Checklist


daenney:
I don't think this can work. The feature itself is fine, however rolling back requires the `down` function in migrations to be specified. None of our migrations do, so executing rollback is a no-op currently. It won't actually restore the database to the previous, pre-migration state.

tsmethurst:
Mmm indeed this won't *undo* migrations because we don't fill out those "down" commands, but it will let you roll back the bun migrations table so that it tries migrations again. Which might be useful in some cases :thinking:

daenney:
Ah yeah, it didn't roll it back so much as that it gave the database some light amnesia. As long as migrations are idempotent that works, but it might be dangerous to expose that if we don't programmatically guarantee that somehow.

DysphoricUnicorn:
I agree that it is not optimal, but I do think there is some value in this feature. Should I maybe add a warning that it could mess up a broken database even more?

daenney:
We've migrated to Codeberg. Your PR has been imported as: https://codeberg.org/superseriousbusiness/gotosocial/pulls/3688.

Issue #3690, opened 2025-01-26
[bug] trusted-proxies UI recommendation gets confused within a LAN


milas:
### Describe the bug with a clear and concise description of what the bug is.

The `trusted-proxies` configuration gets confused if you're able to access the trusted/reverse proxy by private IP.

My setup:
 * GoToSocial running on K8s cluster (on my LAN) w/ trusted proxy set appropriately
 * ingress-nginx with correct client IP reporting

As a result, if I hit my instance (https://social.notaphish.fyi) from outside my network, no banner! Everything works right, correct client IPs are detected, etc.

However, when I'm on my Wi-Fi, GTS thinks that my proxy is misconfigured because my real IP is (correctly) reported as a "private" LAN IP.

That is, in all cases, the GTS logs show `clientIP` correctly (a remote WAN IP for external requests, my laptop's LAN IP when I'm home). The "problem" is that this is basically a split horizon DNS set up -- when I'm home all my traffic stays LAN-bound.

### What's your GoToSocial Version?

2a46681 🦥  (snapshot)

### GoToSocial Arch

arm64 Docker

### What happened?

![Image](https://github.com/user-attachments/assets/83257378-6b8a-461c-a166-bb400e3da018)

>  Warning! It looks like `trusted-proxies` is not set correctly in this instance's configuration. This may cause rate-limiting issues and, by extension, federation issues.
> If you are the instance admin, you should fix this by adding `192.168.1.205/32` to your `trusted-proxies`. 

[My laptop's LAN IP is `192.168.1.205`.]

### What you expected to happen?

* Trusted-proxies warning not shown 

### How to reproduce it?

I believe this should be minimally reproducible with a config like:

 * Create a network (e.g. Docker Compose) for nginx/GTS & use as reverse proxy in front of GTS
    * e.g. `10.244.0.1` for nginx & `10.244.0.2` for GTS 
    * nginx should also serve traffic on a LAN IP, e.g. `--net=host` or `-p 80:80`
 * Add `"10.244.0.0/16"` to `trusted-proxies` in GTS
 * Access GTS instance via nginx from `192.168.1.x` IP

### Anything else we need to know?

This is different than:
* #3680 

I could probably resolve it by completely disabling rate limiting:
* #3686

AFAIK this is purely a "visual" issue -- my config is correct and rate-limiting should be applied properly, it's just GTS UI giving a flawed recommendation.

SadmL:
[...]

PR #3699, opened 2025-01-27
[chore] skip `trusted-proxies` warning if ip excepted from rate limiting

tsmethurst:
# Description

>

This pull request adds a check to skip injecting the trusted-proxies warning if the clientIP address is in `advanced-rate-limit-exceptions`, to encourage people to add rate limit exceptions for their home networks, and to prevent annoying warnings when accessing one's own instance on one's own network, even though rate limiting is otherwise working correctly.

closes https://github.com/superseriousbusiness/gotosocial/issues/3690

Also updates config to store a parsed slice of exceptions, to avoid re-parsing them multiple times.

## Checklist

### DIFF
--- a/cmd/gotosocial/action/server/server.go
+++ b/cmd/gotosocial/action/server/server.go
@@ -455,11 +455,10 @@ var Start action.GTSAction = func(ctx context.Context) error {
 	// create required middleware
 	// rate limiting
 	rlLimit := config.GetAdvancedRateLimitRequests()
-	rlExceptions := config.GetAdvancedRateLimitExceptions()
-	clLimit := middleware.RateLimit(rlLimit, rlExceptions)        // client api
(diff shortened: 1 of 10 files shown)


NyaaaWhatsUpDoc:
i think in terms of getting v0.18.0 out, this looks good to go :+1: . but ideally instead of storing an extra `___Parsed` field type we should instead just have a type that instead supports automatic unparsing otherwise it's just adding to our already cluttered config struct. i think wrapping the []netip.Prefix type to specifically support some kind of text unmarshaler interface will probably do the trick.

but before squerging this would you be able to make an issue to clean that up, and then once v0.18.0 is out get on it just so it doesn't get burried under our large tech-debt pile? :P

tsmethurst:
Yes good idea! I'll just do it now, I don't think it'll take that long.

Issue #776, opened 2022-08-29
[feature] Make `robots.txt` and `noindex` customizable


tsmethurst:
Right now, all GtS instances serve a simple hardcoded `robots.txt` that disallows all crawling:

```
User-agent: *
Disallow: /
```

The code for this is here: https://github.com/superseriousbusiness/gotosocial/blob/main/internal/api/security/robots.go

There are a couple problems with this though.

Firstly, this isn't actually enough to prevent sites from appearing in Google results, it just means that Google shows no information for that site. For example:

![Screenshot from 2022-08-29 12-44-32](https://user-images.githubusercontent.com/31960611/187184172-8b3c8350-f5f6-4ff1-b32f-2803d792f406.png)

Here, Google still has the site indexed, it just hasn't crawled the page to gather information, leading to this 'stub' search result entry which is not particularly useful.

Secondly, some users and instances might actually want their profile or instance to be indexed by search engines, and by hardcoding this blanket rejection robots.txt, we're not allowing them that option.

Instead of serving this hardcoded robots.txt, we should allow instance admins and users to choose whether their stuff is indexable (and retain 'no indexing' as the default).

To do this, we should use targeted `noindex` meta tags instead: https://developers.google.com/search/docs/advanced/crawling/block-indexing. For users, we can use the 'discoverable' field of their account to decide whether to inject this header or not in web views of their pages and statuses.

For instance pages, we'll have to think of something else.

tsmethurst:
Partially resolved by https://github.com/superseriousbusiness/gotosocial/pull/842 but we need a way for instance admins to set discoverable on the instance as a whole: using the Discoverable field of the instance account perhaps?

daenney:
We've migrated to Codeberg. Your issue has been imported as https://codeberg.org/superseriousbusiness/gotosocial/issues/776.

PR #3718, opened 2025-01-31
[feature] Add `instance-stats-randomize` config option

tsmethurst:
# Description

>

Adds a config option `instance-stats-randomize` to randomize user + post stats at `/api/v1|v2/instance` and `/nodeinfo/2.0` endpoints. Does not affect web views. Stats are re-randomized every hour thanks to the ttlcache.

## Checklist

### DIFF
--- a/internal/api/model/instancev1.go
+++ b/internal/api/model/instancev1.go
@@ -110,6 +110,13 @@ type InstanceV1 struct {
 	Terms string `json:"terms,omitempty"`
 	// Raw (unparsed) version of terms.
 	TermsRaw string `json:"terms_text,omitempty"`
+
+	// Random stats generated for the instance.
+	// Only used if `instance-stats-randomize` is true.
+	// Not serialized to the frontend.
+	//
+	// swagger:ignore
+	RandomStats `json:"-"`
 }
 
 // InstanceV1URLs models instance-relevant URLs for client application consumption.
--- a/internal/typeutils/converter.go
+++ b/internal/typeutils/converter.go
@@ -18,10 +18,18 @@
 package typeutils
 
 import (
+	crand "crypto/rand"
+	"math/big"
+	"math/rand"
 	"sync"
+	"time"
 
+	"codeberg.org/gruf/go-cache/v3"
+	apimodel "github.com/superseriousbusiness/gotosocial/internal/api/model"
+	"github.com/superseriousbusiness/gotosocial/internal/config"
 	"github.com/superseriousbusiness/gotosocial/internal/filter/interaction"
 	"github.com/superseriousbusiness/gotosocial/internal/filter/visibility"
+	"github.com/superseriousbusiness/gotosocial/internal/log"
 	"github.com/superseriousbusiness/gotosocial/internal/state"
 )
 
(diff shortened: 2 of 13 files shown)


internal/typeutils/converter.go
```diff
@@ -31,13 +39,64 @@ type Converter struct {
 	randAvatars    sync.Map
 	visFilter      *visibility.Filter
 	intFilter      *interaction.Filter
+	randStatsCache cache.TTLCache[struct{}, apimodel.RandomStats]
```
NyaaaWhatsUpDoc, line 42: i think having an entire cache for this is probably unnecessary. i'd maybe make it an `atomic.Pointer[SomeTypeThatHoldsStats]` (obviously generating once on init, to make sure it's always set) and add a scheduler task once every 24hrs to regenerate it. saves needing an entire cache but does the same thing

internal/api/model/instancev1.go
```diff
@@ -110,6 +110,13 @@ type InstanceV1 struct {
 	Terms string `json:"terms,omitempty"`
 	// Raw (unparsed) version of terms.
 	TermsRaw string `json:"terms_text,omitempty"`
+
+	// Random stats generated for the instance.
+	// Only used if `instance-stats-randomize` is true.
+	// Not serialized to the frontend.
+	//
+	// swagger:ignore
+	RandomStats `json:"-"`
```
NyaaaWhatsUpDoc, line 119: why do we need to store these RandomStats{} on the instance models separately instead of just directly setting the user and statuses counts? i get having a model to store a combined bunch of stats for caching to reduce need for re-generation, (though depending on the answer to this, maybe doesn't need to be stored in apimodels)

tsmethurst: Basically because we pass those apimodels into the HTML templating functions, where we want the true stats. So this is a workaround to only replace them at the last minute (if necessary) when serializing via the API.

internal/typeutils/converter.go
tsmethurst: One thing I like about not doing this with a scheduler is that it naturally introduces a bit of jitter to the regeneration, cuz it'll be regenerated after an hour + however it long it takes for the next instance / nodeinfo request to come in. Agree that using a whole cache is maybe a bit silly though. I'll poke at it. Would like to avoid using a scheduler though as it might be overkill.

NyaaaWhatsUpDoc: i mean the TTL cache uses the scheduler in the background anywho :p

NyaaaWhatsUpDoc: it just has one goroutine that regularly iterates over all the background tasks it's given, and depending on when the next scheduled task will be it just goes to sleep. so adding more tasks to the scheduler shouldn't be a performance hit unless we're adding them all over the shop (if that's what you were worrying about)

NyaaaWhatsUpDoc: if anything will be cheaper (and simpler) than calculating the stats age on every request and regenerating them if over a certain age, you can just always trust the latest atomic.Pointer value

NyaaaWhatsUpDoc: resolved this on matrix :)

tsmethurst: Changed! https://github.com/superseriousbusiness/gotosocial/pull/3718/commits/bde5520cc001e54c0b252e752c1f933613c222f1

internal/api/model/instancev1.go
tsmethurst: Ah sort of yes. The API modules don't have access to a typeconverter, as all the type conversion is expected to be done in the processors (which do have access to type converters).

NyaaaWhatsUpDoc (approved):

rimu:
Tools like [FediDB](https://fedidb.org/) rely on these statistics to make pretty graphs. If you want to hide the number of users, consider setting it to zero instead?

martijndeb:
> Tools like [FediDB](https://fedidb.org/) rely on these statistics to make pretty graphs. If you want to hide the number of users, consider setting it to zero instead?

I support this, providing faked statistics creates noise in the counting. Some promotion of the fediverse relies on those numbers. One can differ on opinion if that's a good thing, but intentionally corrupting data is never a great idea.

Whilst I do support the idea, I think that providing 0 instead of random data is the better option, as it basically creates an opt-out situation

shleeable:
Shouldn't the randomise number be between 0 and the actual number? for this to be acting as a bad actor?

tsmethurst:
> Between 0 and 9223372036854775807 is a bit of a stretch?

The number is limited between 0 and 1,000,000 (for users), and 0 and 10,000,000 (for statuses):

https://github.com/superseriousbusiness/gotosocial/pull/3718/files#diff-7b929d08989a3f96f39b47a817eff54d4f0a0d3e1df866144bd20f652d73b2b7R73-R76

Still ridiculous, but not enough to break json parsing etc.

Re: other concerns about graphs etc: GtS has set a restrictive robots.txt file disallowing crawling of `/.well-known` and `/api` endpoints for years now. The fediverse.observer dev [agreed three years ago](https://gitlab.com/diasporg/poduptime/-/issues/10) that this should be taken account of. Crawlers can easily avoid putting wonky data in their data sets by respecting robots.txt. If they want accurate user + status counts, they can parse the html of `/` if they want, once we add a way for instance admins to mark the whole instance as discoverable + thereby remove the robots `noindex` meta tag from the homepage (see https://github.com/superseriousbusiness/gotosocial/issues/776#issuecomment-1269768458). Crawlers can also just avoid crawling GoToSocial instances entirely, which is now what fedidb and fediverse.observer both do.

Will consider changing the setting so that admins can choose between serving actual numbers, serving 0, and serving random numbers though, that's a nice idea :)

tsmethurst:
Anyway I'm going to lock this discussion since a now-closed PR is not really the place for it, but I'll open a new issue for the more granular setting :)

PR #3722, opened 2025-02-01
[bugfix] Swagger: fix media_ids[] param for creating statuses

VyrCossont:
# Description

This pull request makes it possible to post media with a generated Swagger client by fixing the param name and format.

It also fixes a similar issue with poll options. There may be other places in the codebase where we forgot to apply `collectionFormat: multi`, so keep an eye out. Without it, with multiple values for a param, you get `foo[]=AAA,BBB` when what a Mastodon-compatible API generally expects is instead `foo[]=AAA&foo[]=BBB`.

`uniqueItems: true` is useful as documentation but I'm not sure that generators actually care about it.

Discovered while working on `slurp`.

## Checklist

### DIFF
--- a/docs/api/swagger.yaml
+++ b/docs/api/swagger.yaml
@@ -10261,7 +10261,8 @@ paths:
                   name: status
                   type: string
                   x-go-name: Status
-                - description: |-
+                - collectionFormat: multi
+                  description: |-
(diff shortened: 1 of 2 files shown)


VyrCossont:
I'm just gonna merge this since it fixes `slurp` and does not affect actual GTS code.

PR #3782, opened 2025-02-11
[performance] improved enum migrations

NyaaaWhatsUpDoc:
# Description

This improves our shared enum migration function to use a singular update query that handles all possible variations using a CASE statement (thank you for the suggestion @mirabilos !). I probably should have gone with this approach first but our ORM doesn't expose CASE logic so it hadn't initially jumped to mind. The ups-and-downs of ORMs, eh!

I ran this on our test suite (which isn't always the best test of migrations) with both SQLite and Postgres just fine. And I had an old backup of my own database which I was able to use to confirm in a production-like environment that this definitely works on SQLite. Obviously if anyone has an old backup of some Postgres data to test this against feel free to let us know and we can wait for a test on that!

## Checklist

### DIFF
--- a/internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
+++ b/internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
@@ -71,7 +71,7 @@ func init() {
 			// we must drop all indices that rely on it.
 			log.Info(ctx, "dropping old visibility indexes...")
 			for _, index := range visIndices {
-				log.Info(ctx, "dropping old index %s...", index.name)
+				log.Infof(ctx, "dropping old index %s...", index.name)
 				if _, err := tx.NewDropIndex().
 					Index(index.name).
 					Exec(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


tsmethurst:
Nice one :)

mirabilos:
On Tue, 11 Feb 2025, kim wrote:

>This improves our shared enum migration function to use a singular
>update query that handles all possible variations using a CASE
>statement (thank you for the suggestion @mirabilos !). I probably

Cool!

>should have gone with this approach first but our ORM doesn&#39;t
>expose CASE logic so it hadn&#39;t initially jumped to mind. The
>ups-and-downs of ORMs, eh!

Yeah… it took me a while, and learning SQL *properly*, but now I’m
on the “not using ORM other than for trivial CRUD queries” side,
with “SQL-first” schema design.

(A while is… I think roughly ten years here.)

An ORM is really helpful in the beginning, if you have an OOP
approach and your queries do mostly resemble CRUD, but as soon as
you need to write slightly complex analysis queries, or, as you’ve
noticed, migrations, and still want to be performant, it reaches
its limits. I’ve seen others use the ORM for the simple queries and
the ORMs’ ability to forward and run native SQL (if not HQL or so)
for more complex queries, too.

Cool that you had the backup around and could verify it, so that
everyone who hasn’t yet had that migration can benefit from it.

Thanks,
//mirabilos

PR #3795, opened 2025-02-14
[bugfix] Drop status indices AFTER updating visibility

tsmethurst:
Performance fix to drop status visibility index *AFTER* using it to select into the new visibility column.

### DIFF
--- a/internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
+++ b/internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
@@ -32,8 +32,8 @@ func init() {
 	up := func(ctx context.Context, db *bun.DB) error {
 		return db.RunInTx(ctx, nil, func(ctx context.Context, tx bun.Tx) error {
 
-			// Visibility type indices.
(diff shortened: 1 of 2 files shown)


internal/db/bundb/migrations/20241121121623_enum_strings_to_ints.go
```diff
@@ -67,16 +55,43 @@ func init() {
...
+				IndexCleanupCallback func(ctx context.Context, tx bun.Tx) error
+			}{
+				{
+					Table:  "statuses",
+					Column: "visibility",
+					IndexCleanupCallback: func(ctx context.Context, tx bun.Tx) error {
+						// After new column has been
+						// created and populated, drop
+						// indices relying on old column.
+						for _, index := range visIndices {
```
NyaaaWhatsUpDoc, line 72: given the indices are only used in this one callback now, it probably makes more sense to just define them in that function. otherwise if this gets edited again and the same visIndices variable gets reused in this loop it might get quite confusing tracking bugs down :p

PR #3796, opened 2025-02-14
[performance] separate enum migrations into their own individual transactions

NyaaaWhatsUpDoc:
Should help a bit with performance not performing things in one *massive* transaction, while still using them enough to not leave a database in too-horrific a state should it break part-way through.


tsmethurst:
Gonna close this since it's included in #3798

PR #3807, opened 2025-02-19
[feature] Forward-compatibility with Approval objects

tsmethurst:
This PR adds forward-compatibility with Approval objects that will be referenced in the `result` field of `Accept` activities for interactions in GtS v0.19.0 onwards. The PR allows LikeApproval, ReplyApproval, and AnnounceApproval to be dereferenced and validated when a uri set in a Like or Statusable's `approvedBy` points to one of these objects. PR also adds code to use the `result` uri of an `Accept` instead of the Accept's ID, and to use that as the `approvedBy` for incoming statuses + likes.

https://github.com/superseriousbusiness/activity/pull/27 <-- this PR should be merged first, and then *this* PR should be updated to activity `v1.11.0-gts`.


internal/federation/dereferencing/status_permitted.go
```diff
@@ -674,51 +685,52 @@ func (d *Dereferencer) isValidAccept(
...
+	rsp, err := tsport.Dereference(ctx, approvedByIRI)
 	if err != nil {
-		l.Errorf("error dereferencing Accept: %v", err)
+		l.Errorf("error dereferencing approvedByIRI: %v", err)
 		return false, nil
 	}
 
-	acceptable, err := ap.ResolveAcceptable(ctx, rsp.Body)
+	// Try to parse response as an AP type.
+	t, err := ap.DecodeType(ctx, rsp.Body)
```
tsmethurst, line 698: Main logic changes for this part of the PR are here downwards in this file. Basically instead of just dereffing and checking an Accept, this now derefs to a type, checks whether that type is an Accept or some kind of Approval, and does checking based on the type.

internal/federation/dereferencing/status_permitted.go
```diff
@@ -727,6 +739,52 @@ func (d *Dereferencer) isValidAccept(
...
+	// Type wasn't something we
+	// could do anything with!
+	l.Errorf(
+		"%T at %s not approvable or acceptable",
+		t, approvedByIRIStr,
+	)
+	return false, nil
+}
+
+func isValidAcceptable(
```
tsmethurst, line 776: Flow for checking an Accept is basically the same as before, but now expected target optionally gets checked as well.

internal/federation/dereferencing/status_permitted.go
```diff
@@ -775,6 +833,113 @@ func (d *Dereferencer) isValidAccept(
...
+			targetIRIStr, expectTargetURIStr,
+		)
+		return false, nil
+	}
+
+	// Everything looks OK.
+	return true, nil
+}
+
+func isValidApprovable(
```
tsmethurst, line 853: V. similar to isValidAcceptable, but handles `*Approval` type instead

internal/federation/federatingdb/accept.go
```diff
@@ -102,6 +102,51 @@ func (f *federatingDB) Accept(ctx context.Context, accept vocab.ActivityStreamsA
...
+				); err != nil {
+					return err
+				}
+
+			// ACCEPT TYPE-HINTED REPLY OR ANNOUNCE.
+			//
+			// ie., a statusable or Announce with
+			// just `id` and `type` properties set.
+			case name == ap.ActivityAnnounce,
+				ap.IsStatusable(name):
```
NyaaaWhatsUpDoc, line 132: i think this might be a bit easier to see if you explicitly make it `name == ... || ap.IsStatusable()` instead of using the comma which usually gets used in const value switches (imo)

NyaaaWhatsUpDoc:
once my one above comment is resolved feel free to squerge :)

internal/federation/federatingdb/accept.go
tsmethurst: Ack! Thanks pal

PR #3809, opened 2025-02-19
[bugfix] update fedi api to support multiple separate votes in same multi-choice poll

NyaaaWhatsUpDoc:
Post v0.18.0, we need to handle poll vote deletes (just realised we don't have those...), and handle this same behaviour of multiple separate votes in a multi-choice poll via the client API too.

### DIFF
--- a/internal/federation/federatingdb/create.go
+++ b/internal/federation/federatingdb/create.go
@@ -20,6 +20,7 @@ package federatingdb
 import (
 	"context"
 	"errors"
+	"slices"
 
 	"github.com/superseriousbusiness/activity/streams/vocab"
 	"github.com/superseriousbusiness/gotosocial/internal/ap"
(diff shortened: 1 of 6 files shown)


internal/federation/federatingdb/create.go
```diff
@@ -169,27 +174,24 @@ func (f *federatingDB) createPollOptionables(
 			case inReplyTo.PollID == "":
 				return gtserror.Newf("poll vote in status %s without poll", statusURI)
 
-			// We don't own the poll ...
-			case !*inReplyTo.Local:
-				return gtserror.Newf("poll vote in remote status %s", statusURI)
+			// Ensure poll isn't closed.
+			case inReplyTo.Poll.Closed():
```
NyaaaWhatsUpDoc, line 178: 😬

internal/federation/federatingdb/create.go
```diff
@@ -210,21 +212,60 @@ func (f *federatingDB) createPollOptionables(
...
+		// vote, then sort them by index size.
+		vote.Choices = append(vote.Choices, choices...)
+		slices.Sort(vote.Choices)
+
+		// Populate the poll vote,
+		// later used in fedi worker.
+		vote.Poll = inReplyTo.Poll
+		vote.Account = requester
+
+		// TODO: would it be useful to store an UpdatedAt field
```
tsmethurst, line 241: Ehhh

tsmethurst:
Good job chief!

PR #3821, opened 2025-02-23
[chore/frontend] use localized time string in status & poll info page

### LINKED ISSUE #3819: [feature/frontend] Display user's local time on the post's details page

Currently, when accessing user posts through the website, we see the server's local time.

If the user's time zone differs from the server's, this can be confusing. I want the webpage to display the current time based on the user's time zone.

Display user's local time on the post's details page, instead of always showing server time.

Provide a globe icon that users can spin to select their time zone. 😈

### Additional context.

Some browsers offer privacy protection features that hide the user's time zone. In such cases, we won't know which time to display anyway, but since the user chooses to hide it, I believe this won't affect the feature's implementation.

cdn0x12:
# Description

>

This PR makes a workaround to display timestamps in the user's local timezone:

Precise timestamps (containing hours and minutes) now converted to user's local time using browser's toLocaleString(), while vague timestamps (e.g. "Dec, 2024") are unchanged

Currently, the `timestampPrecise` in the Template has been removed because there's no need to return the server's local time right now. However, the corresponding field in `template.go` is still retained because I'm not sure if we might need it elsewhere.

~~The time format that users actually use might differ from the originally defined format. For example, it might now display as `12/31/2024, 11:11:11 PM` instead of `Dec 31, 2024, 23:11`.~~ It seems that some browsers convert it to a format like `2024-06-06T00:33:54.427Z`, so I'm explicitly specifying it to match the `timestampPrecise` format.

Other parts of the settings panel already use `toLocalString`, so I didn't make any changes there.

closes #3819

## Checklist

### DIFF
--- a/web/source/frontend/index.js
+++ b/web/source/frontend/index.js
@@ -181,3 +181,23 @@ Array.from(document.getElementsByClassName("plyr-video")).forEach((video) => {
 	video._player = player;
 	video._plyrContainer = player.elements.container;
 });
+
+document.addEventListener('DOMContentLoaded', () => {
+	const timeTags = document.getElementsByTagName('time');
+	Array.from(timeTags).forEach(timeTag => {
+		const datetime = timeTag.getAttribute('datetime');
+		const currentText = timeTag.textContent.trim();
+		// Only format if current text contains precise time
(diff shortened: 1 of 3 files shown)


tsmethurst:
Oh nice one, thank you for this :)

PR #3832, opened 2025-02-24
[docs] correct docker version format

elliotblackburn:
# Description

The documentation explains that the release tag is in the format of `vX.Y.Z` for container images. This does not appear to be correct when looking at the image tags in dockerhub. It seems the `v` prefix has been dropped at some point, or was never in place.

## Checklist


tsmethurst:
Ah well spotted, thank you :)

elliotblackburn:
@tsmethurst Thanks for the quick merge! Sorry for not saying so sooner, this got lost in my notifications 😅 And thank you for GoToSocial!

PR #3843, opened 2025-02-27
[frontend] Add an extra CSS class name to the About This Instance and Register an Account sections in the root index template to allow for easier targeting with custom CSS

VirtualWolf:
# Description
This is a teeeensy little change that just adds two CSS classes to the main index page. I was mucking around with some custom CSS with 0.18 and realised that while the "What is this?" section has `class="about-section what-is-this"`, the other two sections ("About this instance" and "Register an account") don't have that extra class for distinguishing them more easily.

So this adds them!

## Checklist

### DIFF
--- a/web/template/index.tmpl
+++ b/web/template/index.tmpl
@@ -27,7 +27,7 @@
 
 {{- with . }}
 <main class="about">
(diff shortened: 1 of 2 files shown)


tsmethurst:
Thanks :)

Issue #2959, opened 2024-06-03
[frontend/feature] Add Microformats to frontend HTML pages


tsmethurst:
To make GtS posts etc work in things like Firefox Reader mode, we could consider adding some Microformats classes where relevant: https://developer.mozilla.org/en-US/docs/Web/HTML/microformats

Some possibilities:

- `h-card` for the info at the top of a user's profile page 
- `h-feed` for list of posts on a user's profile page
- `h-entry` for each post within an `h-feed`
- nested `h-card` for author info on each post

Extracted from https://github.com/superseriousbusiness/gotosocial/issues/1020

tsmethurst:
Closed by https://github.com/superseriousbusiness/gotosocial/pull/3848

PR #3848, opened 2025-02-28
[feature] add microformat support

AnInternetTroll:
Signed-off-by: Luca Matei Pintilie <luca@lucamatei.com>

# Description

Adds microfrontends documentation, as described in #2959

Currently running here as well for easy inspection

https://activitypub.lucamatei.com/@luca

Some notes:
- the main h-card contains the h-feed(s). Maybe unfortunate, but it's
  the easy way to get the bio in that h-card
- First gotosocial PR, so I may have missed some stuff.
- I do not believe tests or documentation is necessary for this change,
  as it's only markup and hidden from users.

## Checklist

### DIFF
--- a/web/template/profile.tmpl
+++ b/web/template/profile.tmpl
@@ -106,7 +106,7 @@
             />
             {{- end }}
             <img
-                class="avatar"
+                class="avatar u-photo"
                 src="{{- .account.Avatar -}}"
                 alt="{{- template "avatarAlt" . -}}"
                 title="{{- template "avatarAlt" . -}}"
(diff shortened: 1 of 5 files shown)


tantek (approved):

tsmethurst:
This looks great to me, thank you for doing this! Only one thing: could you remove the newlines you've added at the end of the template files? Normally I'm all for ending files with a newline but in this case the lack of a newline is significant iirc.

AnInternetTroll:
Thanks for the review! Seems like vim messes with newlines when saving, but emacs let me delete those. Today I learned!

tsmethurst:
Wahoo, thank you very much! :)

PR #3849, opened 2025-03-01
[feature] Refactor tokens, allow multiple app redirect_uris

tsmethurst:
# Description

>

This pull request:

1. Drops the daft `clients` table from the database. Applications now fulfil the oauth2.ClientInfo interface, we don't need both.
2. Permits, stores, and validates multiple redirect_uri entries for an application, as a newline-separated list.
3. Allows a client to request a token with redirect to any of the application redirect URIs.
4. Updates the application model to add a `managed_by_user_id` field, which will be used in an upcoming PR.
5. Updates the token model to add a `last_used` field, which is set on a cooldown of 1hr, used in an upcoming PR.
6. Does some tweaking, bugfixing, and tidying up in the oauth package.

Closes https://github.com/superseriousbusiness/gotosocial/issues/962 (multiple redirect URIs)

## Checklist


internal/api/model/application.go
```diff
@@ -33,12 +33,17 @@ type Application struct {
 	// Post-authorization redirect URI for the application (OAuth2).
 	// example: https://example.org/callback?some=query
 	RedirectURI string `json:"redirect_uri,omitempty"`
+	// Post-authorization redirect URIs for the application (OAuth2).
+	// example: [https://example.org/callback?some=query]
+	RedirectURIs []string `json:"redirect_uris,omitempty"`
```
tsmethurst, line 38: Returning both redirect_uri and redirect_uris is documented by Mastodon here: https://docs.joinmastodon.org/methods/apps/

internal/api/model/application.go
```diff
@@ -33,12 +33,17 @@ type Application struct {
...
 	// Client ID associated with this application.
 	ClientID string `json:"client_id,omitempty"`
 	// Client secret associated with this application.
 	ClientSecret string `json:"client_secret,omitempty"`
 	// Push API key for this application.
 	VapidKey string `json:"vapid_key,omitempty"`
+	// OAuth scopes for this application.
+	Scopes []string `json:"scopes,omitempty"`
```
tsmethurst, line 46: Whoops, we didn't have this before!

internal/api/model/application.go
```diff
@@ -50,14 +55,15 @@ type ApplicationCreateRequest struct {
...
 	//
 	// To display the authorization code to the user instead of redirecting to a web page, use `urn:ietf:wg:oauth:2.0:oob` in this parameter.
 	//
+	// If no redirect URIs are provided, defaults to `urn:ietf:wg:oauth:2.0:oob`.
+	//
 	// in: formData
-	// required: true
-	RedirectURIs string `form:"redirect_uris" json:"redirect_uris" xml:"redirect_uris" binding:"required"`
-	// Space separated list of scopes.
+	RedirectURIs string `form:"redirect_uris" json:"redirect_uris" xml:"redirect_uris"`
```
tsmethurst, line 65: No longer required as it now defaults to oob.

internal/api/util/auth.go
```diff
@@ -150,3 +156,51 @@ func TokenAuth(
 
 	return a, nil
 }
+
+// GetClientScopeHandler returns a handler for testing scope on a TokenGenerateRequest.
```
tsmethurst, line 160: This is here to avoid import cycles, rather than in the oauth package, but it's a bit annoying. Maybe there's somewhere better to put it? Fine for now though I think.

internal/db/bundb/admin.go
```diff
@@ -369,38 +370,26 @@ func (a *adminDB) CreateInstanceApplication(ctx context.Context) error {
...
-	appID, err := id.NewRandomULID()
-	if err != nil {
-		return err
-	}
 
 	// Generate the application
 	// to put in the database.
 	app := &gtsmodel.Application{
-		ID:           appID,
+		ID:           id.NewULID(),
```
tsmethurst, line 377: App IDs are now just created using time.Now() rather than random time, will make it easier to page through them in upcoming PR...

internal/oauth/clientstore.go
```diff
@@ -20,46 +20,30 @@ package oauth
...
-	return pts
+// NewClientStore returns a minimal implementation of
+// oauth2.ClientStore interface, using state as storage.
+//
+// Only GetByID is implemented, Set and Delete are stubs.
+func NewClientStore(state *state.State) oauth2.ClientStore {
+	return &clientStore{state: state}
 }
 
 func (cs *clientStore) GetByID(ctx context.Context, clientID string) (oauth2.ClientInfo, error) {
```
tsmethurst, line 39: Set and Delete are never actually called afaict so they're just stubbed.

internal/processing/app.go
```diff
@@ -19,66 +19,70 @@ package processing
...
 	} else {
 		scopes = form.Scopes
 	}
 
-	// generate new IDs for this application and its associated client
-	clientID, err := id.NewRandomULID()
-	if err != nil {
-		return nil, gtserror.NewErrorInternalError(err)
+	// Normalize + parse requested redirect URIs.
+	form.RedirectURIs = strings.TrimSpace(form.RedirectURIs)
```
tsmethurst, line 46: Tried to avoid some potential issues by normalizing these before they get put in the db :+1:

internal/oauth/tokenstore.go
```diff
@@ -99,55 +123,97 @@ func (ts *tokenStore) Create(ctx context.Context, info oauth2.TokenInfo) error {
...
+	return ts.getUpdateToken(
+		ctx,
+		ts.state.DB.GetTokenByRefresh,
+		refresh,
+	)
+}
+
+// package-internal function for getting a token
+// and potentially updating its last_used value.
+func (ts *tokenStore) getUpdateToken(
```
tsmethurst, line 188: This is where token last_used gets updated. The field isn't used yet in this PR but will be in the token review / auditing one.

internal/oauth/server.go
```diff
@@ -75,17 +80,58 @@ type s struct {
...
-		AccessTokenExp:    0,     // access tokens don't expire -- they must be revoked
-		IsGenerateRefresh: false, // don't use refresh tokens
+		// Following the Mastodon API,
+		// access tokens don't expire.
+		AccessTokenExp: 0,
+		// Don't use refresh tokens.
+		IsGenerateRefresh: false,
 	})
+
+	manager.SetValidateURIHandler(func(hasRedirectList, wantsRedirect string) error {
```
tsmethurst, line 102: Previously we were using the default validate URI handler from the oauth package, which only did partial validation of URIs, not the best! This one is more in line with oauth2 standards.

internal/db/bundb/application.go
```diff
@@ -233,6 +198,21 @@ func (a *applicationDB) PutToken(ctx context.Context, token *gtsmodel.Token) err
 	})
 }
 
+func (a *applicationDB) UpdateToken(ctx context.Context, token *gtsmodel.Token, columns ...string) error {
+	_, err := a.db.
```
NyaaaWhatsUpDoc, line 202: if you wrap this whole update operation in a.state.caches.DB.Token.Store(...) function it will invalidate the old token and store the latest model in the cache :)

tsmethurst: Ohhh right thank you, forgot about that.

internal/api/model/application.go
NyaaaWhatsUpDoc: oob, hehe

internal/oauth/clientstore.go
NyaaaWhatsUpDoc: i'd maybe make them panic, or return error("oauth2.ClientStore{}.Set() not implemented") type thing, just so we find out pretty fast if they do ever happen to start getting used without initially realising

internal/db/bundb/migrations/20250224105654_token_app_client_refactor.go
```diff
@@ -0,0 +1,200 @@
...
+				Exec(ctx); err != nil {
+				return err
+			}
+
+			if len(oldApps) != 0 {
+				// Convert all the old model applications into new ones.
+				newApps := make([]*newmodel.Application, 0, len(oldApps))
+				for _, oldApp := range oldApps {
+					newApps = append(newApps, &newmodel.Application{
+						ID:           id.NewULIDFromTime(oldApp.CreatedAt),
```
NyaaaWhatsUpDoc, line 96: ohhhh this is a nice way of getting ordered ULIDs from the existing models, nice stuff

internal/oauth/tokenstore.go
NyaaaWhatsUpDoc: i'm not sure we actually need the lastUsedLocks here. instead of adding another mutexes map into memory you could simply have a `DB.UpdateTokenLastUsed(ctx, token)` that specifically only updates the last_used column *if* it's greater than the previous value. oh and possibly returning the `last_used` column value just to make sure the passed token model ptr gets its last used value updated to the latest accepted time.

internal/oauth/tokenstore.go
```diff
@@ -99,55 +123,97 @@ func (ts *tokenStore) Create(ctx context.Context, info oauth2.TokenInfo) error {
...
 	if err != nil {
+		// Unlock on error.
+		unlock()
 		return nil, err
 	}
+
+	// If token was last used more than
+	// an hour ago, update this in the db.
+	wasLastUsed := token.LastUsed
+	if time.Since(wasLastUsed) > 1*time.Hour {
```
NyaaaWhatsUpDoc, line 207: `time.Since()` actually calls `time.Now()` under the hood, so you could reduce time calls here by doing something like:
```golang
if now := time.Now(); now.Sub(wasLastUsed) > 1*time.Hour {
    token.LastUsed = now
    
    // ...
}
```

internal/oauth/server.go
```diff
@@ -75,17 +80,58 @@ type s struct {
...
+			func(hasRedirect string) bool {
+				hasRedirectURI, err := url.Parse(hasRedirect)
+				if err != nil {
+					log.Errorf(nil, "error parsing hasRedirect: %v", err)
+					return false
+				}
+
+				// Want an exact match.
+				// See: https://www.oauth.com/oauth2-servers/redirect-uris/redirect-uri-validation/
+				return wantsRedirectURI.String() == hasRedirectURI.String()
```
NyaaaWhatsUpDoc, line 126: it would be a little more performant here if you calculated the wantsRedirectURI string just once outside this embedded function loop, and just compared to the string here

internal/oauth/server.go
```diff
@@ -231,13 +301,28 @@ func (s *s) HandleAuthorizeRequest(w http.ResponseWriter, r *http.Request) gtser
...
+			return gtserror.NewErrorInternalError(err)
+		}
+
+		if util.IsNil(client) {
+			// Application just not found.
 			return gtserror.NewErrorUnauthorized(err, HelpfulAdvice)
 		}
-		req.RedirectURI = client.GetDomain()
+
+		app, ok := client.(*gtsmodel.Application)
```
NyaaaWhatsUpDoc, line 319: given the default behaviour for a failed cast is to panic, it would be simpler here to just do `app := client.(*gtsmodel.Application)`

internal/oauth/server.go
```diff
@@ -143,30 +202,41 @@ func (s *s) HandleTokenRequest(r *http.Request) (map[string]interface{}, gtserro
...
-			// remove this key from the returned map
-			// if the value is 0 or less, so that clients
-			// don't interpret the token as already expired
-			if expiresIn <= 0 {
-				delete(data, "expires_in")
-			}
-		default:
-			err := errors.New("expires_in was set on token response, but was not an int64")
-			return nil, gtserror.NewErrorInternalError(err, HelpfulAdvice)
+		expiresIn, ok := expiresInI.(int64)
```
NyaaaWhatsUpDoc, line 230: given the default behaviour for a failed cast is to panic, it would be simpler here to just do `expiresIn := expiresInI.(int64)`

internal/api/util/auth.go
NyaaaWhatsUpDoc: it would be nicer if we could put it somewhere else... would something like `internal/oauth/handlers` or something of the like work? :thinking:

NyaaaWhatsUpDoc: (i.e. some kind of subpkg of our oauth type, so it's specifically different to just `internal/oauth`, or whatever the cycling package in particular is)

internal/oauth/server.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/3849/commits/469bd759995520e2bb7440dc7bbd8ffd7fc554b2

internal/oauth/clientstore.go
tsmethurst: Done! https://github.com/superseriousbusiness/gotosocial/pull/3849/commits/a676bd3602a6e062ae494e3fa39c490beb2935dc

internal/db/bundb/application.go
tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3849/commits/13c03a89096ae81f171689e4fdb3c3a6f373d6af

internal/oauth/tokenstore.go
tsmethurst: Oh nice, that's cheeky.

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3849/commits/d5f425486273cf67802b099d343403c02da3d37d

PR #3867, opened 2025-03-03
[bugfix] fix refreshed additional media info being ignored

NyaaaWhatsUpDoc:
# Description

Previously when additional info was provided, this was skipped being checked over if the force flag was already set, but we actually need to iterate over this information since the additional info isn't actually later passed anywhere.

This should (hopefully!) fix https://github.com/superseriousbusiness/gotosocial/issues/2851

## Checklist

### DIFF
--- a/internal/federation/dereferencing/emoji.go
+++ b/internal/federation/dereferencing/emoji.go
@@ -134,19 +134,23 @@ func (d *Dereferencer) RefreshEmoji(
 	*gtsmodel.Emoji,
 	error,
 ) {
-	// Check emoji is up-to-date
-	// with provided extra info.
-	switch {
-	case info.URI != nil &&
-		*info.URI != emoji.URI:
+	// Check uri up-to-date.
+	if info.URI != nil &&
+		*info.URI != emoji.URI {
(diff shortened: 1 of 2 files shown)


internal/federation/dereferencing/emoji.go
```diff
@@ -134,19 +134,23 @@ func (d *Dereferencer) RefreshEmoji(
 	*gtsmodel.Emoji,
 	error,
 ) {
-	// Check emoji is up-to-date
-	// with provided extra info.
-	switch {
-	case info.URI != nil &&
-		*info.URI != emoji.URI:
+	// Check uri up-to-date.
```
NyaaaWhatsUpDoc, line 137: this is mainly just to keep the emoji code up-to-date with the media code below, it's not necessary in the same way since for emoji the additional info is later passed to another function.

NyaaaWhatsUpDoc: at some point i'll add an UpdateMedia() function with a similar signature to the emoji one, just so there's a little more parity between those paths of logic in the media manager

PR #3870, opened 2025-03-03
[bugfix] Store and expose status content type

### LINKED ISSUE #3836: [bug] Editing a status can result in unexpected changes to content type

`/api/v1/statuses/:id/source` does not report a status's current `content_type`. This means that when a user tells a client to edit a post, the client has no information about what the original content type was, and submitting an edit sometimes results in unexpected changes.

If a client sends no `content_type` back when editing a message, GtS will use the user's "default post format" preference as the new content type for the post. If the post's format was originally something else, it will be changed when the user didn't expect it to be. The client could instead send an explicit `content_type` with the edit to ensure that the format isn't changed, except that it isn't told what the correct format is, so the bug will manifest no matter what clients try to do.

The only workaround is for users to manually confirm the correct content type is selected every time they edit a post, which is annoying and easy to forget.

v0.18.0

### GoToSocial Arch

amd64 Docker

### What happened?

User-level: I edited a markdown post with [Enafore](https://enafore.social), and in the compose popup, the selected content type defaulted to plain text rather than markdown.

API-level: Client fetched `/api/v1/statuses/:id/source` on the post and received no `content_type` in the response. (Enafore assumes `text/plain` in the absence of the field.)

User-level: The selected content type should default to the content type that was used when I initially posted the status (in this case, markdown).

API-level: Client should have received `"content_type": "text/markdown"` as part of the response.

### How to reproduce it?

Make a Markdown post however you like, then request its source:

```
$ curl -H "Authorization: Bearer $token" "https://my.instance/api/v1/statuses/01JMRB1FB8G01G0W0543DFM7FV/source
```
```json
{"id":"01JMRB1FB8G01G0W0543DFM7FV","text":"a","spoiler_text":""}
```

Note the lack of `content_type` field in the response.

GtS doesn't currently store a post's content type in the database; in [`internal/api/model/status.go`](https://github.com/superseriousbusiness/gotosocial/blob/f734a94c1ca9bb732b2ae818098b48794e374c8a/internal/api/model/status.go), it's part of `StatusCreateRequest` and `StatusEditRequest`, but not `Status` or `StatusEdit` (nor is it part of the analogous structs in [`internal/gtsmodel/status.go`](https://github.com/superseriousbusiness/gotosocial/blob/f734a94c1ca9bb732b2ae818098b48794e374c8a/internal/gtsmodel/status.go#L28)). It should store this on the status, and the stored value should be used as the default `content_type` for edit requests instead of the user preference (which should only apply to *newly-created* posts).

Prior art: Glitch [stores `content_type` on statuses in the database](https://github.com/glitch-soc/mastodon/blob/6fd1930912237f2f237f9ac39c02c00a69362b2a/db/migrate/20190512200918_add_content_type_to_statuses.rb) and includes it in responses to `/api/v1/statuses/:id/source`.[^1] It [also stores it on status edits](https://github.com/glitch-soc/mastodon/blob/6fd1930912237f2f237f9ac39c02c00a69362b2a/db/migrate/20220209175231_add_content_type_to_status_edits.rb), which seems to make sense, since content type can be changed by editing, and the point of a `StatusEdit` is to record such changes. However I'm not sure whether anything actually relies on it being present on edits in practice.

I don't know how this is handled by other servers, though Enafore appears to have [some Akkoma-specific logic](https://github.com/enafore/enafore/blob/e35c1fed04a9bdb07fa680afe4137fa6aa049b8a/src/routes/_actions/edit.js#L20) to check a different field entirely for this information.

In summary:

- Content type should be stored on statuses (and potentially status edits).
- `/api/v1/statuses/:id/source` should return the `content_type` of the status.
- When `PUT /api/v1/statuses/:id` doesn't receive a new `content_type`, the status's current format should be assumed instead of the "default post format" preference.

[^1]:  I'm pretty sure the response to the `/statuses/:id/source` request is generated by [`Statuses::SourcesController`](https://github.com/glitch-soc/mastodon/blob/6fd1930912237f2f237f9ac39c02c00a69362b2a/app/controllers/api/v1/statuses/sources_controller.rb#L6-L8), which invokes [`StatusSourceSerializer`](https://github.com/glitch-soc/mastodon/blob/main/app/serializers/rest/status_source_serializer.rb), which includes the status's `content_type` from [the `Status` model](https://github.com/glitch-soc/mastodon/blob/6fd1930912237f2f237f9ac39c02c00a69362b2a/app/models/status.rb#L102).

ewwwin:
# Description

Fixes #3836.

- Adds a `ContentType` field to the internal `Status` and `StatusEdit` models, and exposes it in the `Status` and `StatusSource` API models.
- Updates status create processing to always set this field on created statuses, using either the value provided in the request, the user's "default post format" setting, or the hardcoded default value (which is `text/plain`).
- Updates status edit processing to maintain a status's existing content type if there's no updated value in the request. For old statuses with no stored content type, the user's default setting or the hardcoded default is used and saved back.
- Across the codebase, in locations where these models are instantiated and given a value for `Text`, updates those places to also set `ContentType` appropriately. This applies to:
	- Creating `StatusEdit`s in response to federated edits
	- Sizing helpers in `internal/cache/size.go`
	- The testrig and test models
	- One location is left out; I gave the field explicitly unset on the test status `local_account_1_status_2` so it can be used to test the behavior of editing old statuses.

Changes to tests:

- Adds a test case to `processing/status/create_test.go` to verify that creating a status without providing a content type will use the user's setting/global default.
- Adds test cases in `processing/status/edit_test.go` to verify that changes to content type are saved, and that editing an old status will always assign it a content type even if the request doesn't include one (again using the user's setting/global default).
- Updates a test case in `api/client/statuses/statusdelete_test.go` to verify that deleting a status returns the text and content type fields to ensure "delete and redraft" client features can work the same as editing.
- Updated existing test cases that compare JSON-serialized `Status`es to include the appropriate `content_type` fields.
- Manually tested the migration code and the behavior of edits to old statuses by building and running from `main`, then building and running from this branch.

## Checklist


internal/processing/status/edit.go
```diff
@@ -342,6 +348,37 @@ func (p *Processor) Edit(
...
+
+	// No content type on the form, get the status's current content type and
+	// set it back on the form for later use.
+	case status.ContentType != 0:
+		form.ContentType = p.converter.ContentTypeToAPIContentType(ctx, status.ContentType)
+
+	// Old statuses may not have a saved content type; update the status to the
+	// user's preference and set this back on the form for later use.
+	case accountDefaultContentType != "":
+		status.ContentType = typeutils.APIContentTypeToContentType(apimodel.StatusContentType(accountDefaultContentType))
```
ewwwin, line 371: The fact that `gtsmodel.AccountSettings.StatusContentType` is a `string` rather than using the `gtsmodel.StatusContentType` enum type makes this conversion pretty awkward. It works, but the setting should really be converted to the enum type at some point. Would that make sense to do as part of this PR, or would a separate one be better?

tsmethurst:
Heya, let me and/or @NyaaaWhatsUpDoc know when you're happy with this so we can take a look, or let us know if you need any help or have any questions :)

ewwwin:
@tsmethurst @NyaaaWhatsUpDoc This should be ready for review now. CI has seemed unhappy for unrelated reasons, not sure if the current run will pass or not, but I've run tests locally and everything looks okay on my end. Did my best to follow surrounding style when it comes to comment formatting, etc.

I kinda just threw my migration code together based on other migrations in the repo. It looks like the idea is for migrations to include duplicates of the models they update, presumably to ensure that if the canonical model changes later the migration's effect will still be the same? It definitely worked for me but let me know if it looks okay, I don't feel confident about it.

I tried to be consistent about setting `ContentType` anywhere `Text` was already being set, since the two are so related. This results in `gtsmodel.StatusEdit` getting the new field, but not to the corresponding API model. This also means federation is virtually untouched, since only local statuses need to have `text` and it doesn't seem like the federation code even checks for it on incoming statuses (unless I'm missing something!).

The fact that `gtsmodel.AccountSettings.StatusContentType` is just a `string` rather than using the `StatusContentType` enum makes certain type conversions a bit verbose and awkward. Let me know if you'd accept a future PR to change that. I see there's already a helper for writing string-column-to-enum-column migrations, so it doesn't seem difficult, but it feels outside the scope of this PR to me.

uhh i've been staring at my screen for too long and i'm sure there's something i'm forgetting to mention or ask about but it's probably fine. Looking forward to your comments ^^

internal/processing/status/create_test.go
```diff
@@ -238,6 +238,36 @@ func (suite *StatusCreateTestSuite) TestProcessReplyToUnthreadedRemoteStatus() {
 	suite.NotEmpty(dbStatus.ThreadID)
 }
 
+func (suite *StatusCreateTestSuite) TestProcessNoContentTypeUsesDefault() {
+	ctx := context.Background()
+	creatingAccount := suite.testAccounts["local_account_1"]
+	creatingApplication := suite.testApplications["application_1"]
+
+	statusCreateForm := &apimodel.StatusCreateRequest{
+		Status:      "poopoo peepee",
```
NyaaaWhatsUpDoc, line 247: good stuff :ok_hand:

NyaaaWhatsUpDoc:
i'm *most* of the way through reviewing this right now and just wanted to pause to say this is a really well put together and thorough PR. really great, thank you :)

internal/typeutils/internaltofrontend.go
```diff
@@ -1626,6 +1627,17 @@ func (c *Converter) VisToAPIVis(ctx context.Context, m gtsmodel.Visibility) apim
 	return ""
 }
 
+// Converts a gts status content type into its api equivalent
+func (c *Converter) ContentTypeToAPIContentType(ctx context.Context, m gtsmodel.StatusContentType) apimodel.StatusContentType {
```
NyaaaWhatsUpDoc, line 1631: given here we're not actually using any fields of the Converter{} method receiver type, or the context, you can probably just drop those parts of the function signature. i.e. `func ContentTypeToAPIContentType(m gtsmodel.StatusContentType) apimodel.StatusContentType`

internal/processing/status/edit.go
```diff
@@ -342,6 +348,32 @@ func (p *Processor) Edit(
...
+	status *gtsmodel.Status,
+	accountDefaultContentType string,
+) gtsmodel.StatusContentType {
+	switch {
+	// Content type set on form, return the new value.
+	case form.ContentType != "":
+		return typeutils.APIContentTypeToContentType(form.ContentType)
+
+	// No content type on the form, return the status's current content type.
+	case status.ContentType != 0:
```
NyaaaWhatsUpDoc, line 363: i think if you update this case here to be `case status != nil && status.ContentType != 0:` (and the comment too obviously :p) you should be able to use this `processContentType()` function in the status create content-type handling pathway too :)

internal/typeutils/internaltofrontend.go
ewwwin: Yeah I was wondering about this - I started out by copying `VisToAPIVis` which has the same unused parameters, but I can remove them from the new function.

tsmethurst: Oh yeah the VisToAPIVis stuff is *really* old code :')

internal/processing/status/edit.go
ewwwin: https://github.com/superseriousbusiness/gotosocial/pull/3870/files/5ecbda32af7ef393a13ec26e84581488c17186e2..d83cf7ed9b983693fda42122d4ef0eaa17fd60b0

internal/typeutils/internaltofrontend.go
ewwwin: [`d71935d` (#3870)](https://github.com/superseriousbusiness/gotosocial/pull/3870/commits/d71935dde06f962f920ad5ac2783ca7233077580)

Issue #3875, opened 2025-03-04
[bug] posts from gancio aren't formated correctly


lukaprincic:
### Describe the bug with a clear and concise description of what the bug is.

when subscribed to a gancio (calendar, usually a @relay user) the new events don't appear the same when shown via mastodon as opposed to gotosocial instance. gotosocial posts show bare html tags.

tested with tusky and phanpy clients on vanilla mastodon v.4.3.4, gotosocial GoToSocial 0.18.0+git-c9de6c9 amd gancio 1.23.1

### What's your GoToSocial Version?

0.18.0

### GoToSocial Arch

amd64 vps via yunohost

### What happened?

post/event from Gancio shows html tags

### What you expected to happen?

clear post, identical or better formated to one via mastodon

### How to reproduce it?

use any client and subscribe to gancio calendar (@relay at domain - for example @relay@do.basspistol.org)

### Anything else we need to know?

gancio: https://gancio.org/

screenshots:

![Image](https://github.com/user-attachments/assets/f4a8dc04-d8f9-45ce-8816-f120e78bac05)
![Image](https://github.com/user-attachments/assets/ba9e416b-6b5e-4635-ac2b-b9c72c135ca4)
![Image](https://github.com/user-attachments/assets/becd7b7e-0673-4da4-b00e-f96cf014916e)
![Image](https://github.com/user-attachments/assets/f6e9b566-2fd0-49f5-8b32-90e9b36d9871)

tsmethurst:
Could you give a URL for one of those posts, so I can fetch the ActivityStreams representation of it & have a look?

tsmethurst:
Alright thanks!

This is the json it's returning:

```json
{
  "id": "https://koledar.kompot.si/federation/m/132",
  "name": "skupnostni brunč in delavnica // 26. Mednarodni feministični in kvirovski festival Rdeče zore",
  "url": "https://koledar.kompot.si/event/skupnostni-brunc-in-delavnica-26-mednarodni-feministicni-in-kvirovski-festival-rdece-zore",
  "type": "Event",
  "startTime": "2025-03-09T13:00:00.000+01:00",
  "endTime": "2025-03-09T17:00:00.000+01:00",
  "location": {
    "id": "https://koledar.kompot.si/federation/p/18",
    "type": "Place",
    "name": "PLAC, Linhartova 43",
    "address": "43 Linhartova cesta, 1000 Ljubljana, Slovenia"
  },
  "attachment": [
    {
[...]

Issue #3806, opened 2025-02-17
[bug] "<" characters in content warnings end up HTML-escaped as "&lt;"


ewwwin:
### Describe the bug with a clear and concise description of what the bug is.

When creating a status with a content warning that contains the `<` character, the character is replaced with the `&lt;` HTML entity. Other similar characters like `&` and `>` appear to be unaffected.

### What's your GoToSocial Version?

v0.18.0-rc.2

### GoToSocial Arch

amd64 Docker

### What happened?

All `<` characters in content warnings are replaced with `&lt;` in created statuses.

### What you expected to happen?

The character should be preserved.

### How to reproduce it?

```bash
$ curl \
    -H "Authorization: Bearer $ACCESS_TOKEN" \
    -F 'status=test' \
    -F 'spoiler_text=test < & >' \
    -F 'content_type=text/plain' \
    'https://my.instance.xyz/api/v1/statuses'
```

Resulting JSON status will have `spoiler_text` set to `"test &lt; & >"` instead of `"test < & >"`. I tested with a freshly created account, so all user settings should be at their defaults if that matters at all.

### Anything else we need to know?

Possibly related to #421?

The sanitization/formatting of the string appears to be happening somewhere in/around here: https://github.com/superseriousbusiness/gotosocial/blob/5dc8009e30004811e9d4d46c82a23219f533a3b5/internal/processing/status/common.go#L199-L205

tsmethurst:
Ah thanks! Thought we had an issue open for this 🤔 But apparently we didn't! Gotta do some tweaking to our HTML escaping I think!

PR #3876, opened 2025-03-04
[feature] Parse content warning to HTML, serialize via client API as plaintext

tsmethurst:
# Description

>

While ActivityStreams says that `summary` should be HTML (https://www.w3.org/TR/activitystreams-vocabulary/#dfn-summary), and we sanitize and then store it in the db as HTML when we get it from remote instances, the Mastodon API says that the `spoiler_text` property should be non-HTML string. However, previously we were serializing HTML there, leading to issues like https://github.com/superseriousbusiness/gotosocial/issues/3875.

This pull request fixes that by continuing to sanitize + store `summary` as HTML, but *serializing* it via the API as plaintext.

Also updates the way we parse content warning on new statuses created via the API. We now parse + store it without removing HTML, just sanitizing it instead. So people can put HTML in content warnings now and we will federate it as HTML. Other instances can then do whatever sanitization they want with it.

In the web frontend it looks like this (actual HTML):

Screenshot from 2025-03-05 19-33-13

Via the API it looks like this (converted to plaintext):

Screenshot from 2025-03-05 19-33-36

Closes https://github.com/superseriousbusiness/gotosocial/issues/3875
Closes https://github.com/superseriousbusiness/gotosocial/issues/3806

Sending `summary` as HTML seems to cause some issues in Mastodon: https://github.com/mastodon/mastodon/issues/34079 Haven't checked other servers yet.

## Checklist


tsmethurst:
Looks poo in iceshrimp as well:

image

tsmethurst:
And poo in Akkoma also apparently: https://dood.net/notice/Arl022cthtjSG29pmy

internal/text/markdown.go
```diff
@@ -20,37 +20,104 @@ package text
...
+	ctx context.Context,
+	basic bool,
+	parseMention gtsmodel.ParseMentionFunc,
+	authorID string,
+	statusID string,
+	input string,
+) *FormatResult {
+	var (
+		result          = new(FormatResult)
+		rendererOptions = []renderer.Option{
```
NyaaaWhatsUpDoc, line 100: given the simplicity of the logic here i think it might be a little easier to read and would ensure we only ever end up with the length slice we need if you just did:

```golang
result := new(FormatResult)

var (
    result = new(FormatResult)
    opts []renderer.Option
)

if basic {
    // Only allow most basic HTML
    opts = []renderer.Option{
        html.WithXHTML(),
        html.WithHardWraps(),    
    }
} else {
    opts = []renderer.Option{
        html.WithXHTML(),
        html.WithHardWraps(),
        
        // Allow raw HTML, we
        // sanitize at the end anyways.
        html.WithUnsafe(),
    }
}

```

internal/processing/status/edit.go
```diff
@@ -297,13 +297,22 @@ func (p *Processor) Edit(
...
 	status.AttachmentIDs = form.MediaIDs
 	status.Attachments = media
 	status.EditedAt = now
 
+	// Only store ContentWarningText if the parsed
+	// result is different from the given SpoilerText,
+	// otherwise skip to avoid duplicating db columns.
+	var contentWarningText string
+	if content.ContentWarning != form.SpoilerText {
+		contentWarningText = form.SpoilerText
```
NyaaaWhatsUpDoc, line 312: this would be shorter here if you just cut-out the `contentWarningText` variable and simply set it on the status model in the if clause :p

internal/webpush/realsender.go
```diff
@@ -361,10 +361,10 @@ func formatNotificationBody(apiNotification *apimodel.Notification) string {
 		if apiNotification.Status.SpoilerText != "" {
 			body = apiNotification.Status.SpoilerText
```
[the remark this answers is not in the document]
tsmethurst: Yes, and it's already done with StatusToAPIStatus :sunglasses:

internal/text/plain.go
```diff
@@ -181,8 +187,20 @@ func (f *Formatter) fromPlain(
...
-	result.HTML = SanitizeToHTML(result.HTML)
+	result.HTML = SanitizeHTML(result.HTML)
 	result.HTML = MinifyHTML(result.HTML)
 
 	return result
 }
+
+// HTMLToPlain parses the given HTML and then outputs
+// it to close-as-possible equivalent plaintext.
+func HTMLToPlain(html string) string {
```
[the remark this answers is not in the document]
tsmethurst: Tried to clarify this a bit here: https://github.com/superseriousbusiness/gotosocial/pull/3876/commits/c8556cfd9758d2e014fe6c7df8b80f7f532d383c

internal/text/markdown.go
tsmethurst: Ah nice one, changed! https://github.com/superseriousbusiness/gotosocial/pull/3876/commits/2497a4d5962651797b01eec2b4bc07b9db3fc1bb

internal/processing/status/edit.go
tsmethurst: Hah, so silly :') My god. Fixed https://github.com/superseriousbusiness/gotosocial/pull/3876/commits/1092bc4f5496667f7ad8f5a7f7b7057208a081d3

mirabilos:
Ah, this should improve the situation with a bug I reported about half-escaping in CWs.

However, this begs the question: if I have code like:

```
#!/usr/bin/python3
from mastodon import Mastodon

cw = 'somecwtext'
body = 'somebodytext'

fv = Mastodon(version_check_mode="none", api_base_url=…, client_id=…, client_secret=…, access_token=…)
fv.status_post(body, spoiler_text=cw)
```

What encoding do I need for `somecwtext`?

Currently, my account is set to use Markdown, so I’m using Markdown for both `somebodytext` and `somecwtext`, which leads to errors, though. (I had (before seeing this) come to the conclusions that CWs need to be unformatted plaintext instead.)

From the description above, I gather it needs to be HTML instead?

What does stock Mastodon do in such a scenario? That is, which encoding does it assume the user used for `somecwtext`?

tsmethurst:
Mastodon always assumes text/plain for `somecwtext` I believe

mirabilos:
On Sat, 22 Mar 2025, tobi wrote:

>Mastodon always assumes text/plain for `somecwtext` I believe

Then I’d argue GtS should do the same, so that when someone
posts <b> they get <b> and not bolding (federated) or the <b>
hidden (client API).

tsmethurst:
When someone posts `<b>` they should also just get `<b>` on GtS, as we don't allow unsafe (ie., raw) html elements in content warnings specifically. When they post something like `**test**` they will end up with `<b>test</b>` though, assuming they've set their parsing language to markdown. Tbh I think that's fine.

mirabilos:
On Sun, 23 Mar 2025, tobi wrote:

>When someone posts `<b>` they should also just get `<b>` on GtS, as we
>don't allow unsafe (ie., raw) html elements in content warnings
>specifically. When they post something like `**test**` they will end up
>with `<b>test</b>` though, assuming they've set their parsing language
>to markdown. Tbh I think that's fine.

This is even more confusing to me now.

Does this mean GtS currently parses CWs as Markdown if the profile is
set to Markdown, but with different rules for parsing than for the body?

This is even more confusing.

I would have the expectation that either CWs are always plaintext
(preferred), or that they are handled the same as the post body.

That is, for the input string “**test** &amp; & a<b>c” I would expect…

① always plaintext, or same rules as for the body and the user
  did not enable Markdown in their profile or the content type
  option in the post (once GtS supports that) is set to plaintext:

  – federate “**test** &amp;amp; &amp; a&lt;b&gt;c”
  – clients show “**test** &amp; & a<b>c”

② same rules as for the body and the user enabled Markdown in
  their profile or the post content type option is set to it:

  – federate “<b>test</b> &amp; &amp; a<b>c</b>”
  – clients show “test & & ac”

(Here you can see why I’d prefer the first option.)

Arguably, CWs are short titles akin to the Subject line of
eMail so a case could be made for them to please not contain
any formatting (the client will normally format them different
to the main post body anyway).

Issue #3874, opened 2025-03-04
[bug] Server runtime panics and crashes


zxzzxv:
### Describe the bug with a clear and concise description of what the bug is.

When pulling the latest changes, the code built but when I run it I get a segfault

````
gotosocial@gts1:/gotosocial$ GTS_WAZERO_COMPILATION_CACHE=/gotosocial/.cache /gotosocial/gotosocial.new --config-path config.yaml server start
timestamp="04/03/2025 14:08:59.869" func=cache.(*Caches).Stop level=INFO msg="stop: 0x4002300008"                                                                                                                                                                              
panic: runtime error: invalid memory address or nil pointer dereference                                                                                                                                                                                                        
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13bb818]                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                               
goroutine 1 [running]:                                                                                                                                                                                                                                                         
codeberg.org/gruf/go-cache/v3/ttl.(*Cache[...]).Stop(0x10?)                                                                                                                                                                                                                    
        codeberg.org/gruf/go-cache/v3@v3.6.1/ttl/ttl.go:85 +0x18                                                                                                                                                                                                               
github.com/superseriousbusiness/gotosocial/internal/cache.tryUntil({0x2365ae6, 0x18}, 0x5, 0x400049f468)                                                                                                                                                                       
        github.com/superseriousbusiness/gotosocial/internal/cache/util.go:58 +0x78
github.com/superseriousbusiness/gotosocial/internal/cache.(*Caches).Stop(0x4002300008)
        github.com/superseriousbusiness/gotosocial/internal/cache/cache.go:144 +0x88
github.com/superseriousbusiness/gotosocial/cmd/gotosocial/action/server.init.func2.1()
        github.com/superseriousbusiness/gotosocial/cmd/gotosocial/action/server/server.go:122 +0x54
github.com/superseriousbusiness/gotosocial/cmd/gotosocial/action/server.init.func2({0x2a881e0, 0x49de0e0})
        github.com/superseriousbusiness/gotosocial/cmd/gotosocial/action/server/server.go:185 +0x130
main.run({0x2a881e0, 0x49de0e0}, 0x277b010)
        github.com/superseriousbusiness/gotosocial/cmd/gotosocial/common.go:82 +0x1b8
[...]

PR #3879, opened 2025-03-05
[bugfix] Return useful err on `server start` failure

tsmethurst:
Some tweaks to `server start` to try to avoid panicking when returning an error. This should make it easier to debug stuff like https://github.com/superseriousbusiness/gotosocial/issues/3874


internal/scheduler/scheduler.go
```diff
@@ -44,6 +44,11 @@ func (sch *Scheduler) Start() bool {
 	return false
 }
 
+// Started returns true if the scheduler has been started.
+func (sch *Scheduler) Started() bool {
+	return sch.ts != nil
```
NyaaaWhatsUpDoc, line 49: this isn't needed for the scheduler, running Stop() if the scheduler is already stopped is a no-op

tsmethurst: https://github.com/superseriousbusiness/gotosocial/pull/3879/commits/4580ae194cafb81f4ca65bb97c806484c9338f64

internal/cache/cache.go
```diff
@@ -136,6 +136,13 @@ func (c *Caches) Start() {
 	})
 }
 
+// Inited returns true if the
+// caches have been initialized.
+func (c *Caches) Inited() bool {
+	// Use nilness of *ttl.Cache pointers as heuristic.
+	return c.Webfinger != nil && c.StatusesFilterableFields != nil
```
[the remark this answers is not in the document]
tsmethurst: Got ya! Changed it here: https://github.com/superseriousbusiness/gotosocial/pull/3879/commits/be85a5b64245dd8c492abe2015518c2003fb5da2 And returned an error message if stuff can't be started, just because it was simple enough to do. You never know if it'll catch some weird bug further down the line :P

Issue #2523, opened 2024-01-13
[feature] List / delete applications


drym3r:
### Is your feature request related to a problem ?

I'd like to list all the apps that I've created and even delete them.

### Describe the solution you'd like.

I'd like more endpoints under `/api/v1/apps` that let list, describe, edit and delete any app you may have.

### Describe alternatives you've considered.

Not sure there's any. 

### Additional context.

My assumption is that's there's no way to do what I say after reading the OAS, which only shows the `/api/v1/apps/` endpoint accepting the POST verb. If that's not the case, sorry for the noise and please point me in the right direction.

tsmethurst:
I've added this to the OAuth improvements issue.

jcarnat:
Hi,

AFAIK this is not available in the APIs yet.

In the meantime, what is the proper way to list and delete applications?
Shall one simply browse the `applications` table to get the list of authorized apps and delete records from it to remove authorization?

(I'm using sqlite if that matters)

alimoeeny:
is there any potential issues with deleting a row from the applications table?

tsmethurst:
> is there any potential issues with deleting a row from the applications table?

Yes, it will probably cause issues with tokens created by that application. You can do it, but delete the tokens the application created as well.

PR #3906, opened 2025-03-13
[feature] Application creation + management via API + settings panel

tsmethurst:
# Description

>

This pull request adds application creation + management via the API and settings panel. In this context, a "managed" application is one which the user created by using the existing `/api/v1/apps` endpoint **and passing an access token in with that request**. So basically, only apps created via the settings panel will be managed applications. Also adds a nice flow for users to get an access token via the settings panel, so they don't have to fanny about with the command line if they don't want to!

Screenshot from 2025-03-13 13-32-35
Screenshot from 2025-03-13 12-23-41

closes https://github.com/superseriousbusiness/gotosocial/issues/2523

## Checklist


internal/db/bundb/admin.go
```diff
@@ -194,6 +194,17 @@ func (a *adminDB) NewSignup(ctx context.Context, newSignup gtsmodel.NewSignup) (
 		return nil, err
 	}
 
+	// If no app ID was set,
+	// use the instance app ID.
+	if newSignup.AppID == "" {
```
tsmethurst, line 199: I whacked this in here because I noticed that accounts created via the CLI had no CreatedByApplicationID set, and it makes sense to fall back to the instance account (as though they'd been created by using the sign-up page). We don't really use that CreatedByApplicationID for anything yet, but it seems sensible to have a fallback/default anyways.

internal/db/bundb/migrations/20250310144102_application_management.go
```diff
@@ -0,0 +1,105 @@
...
+				return err
+			}
+
+			if usersLen == 0 {
+				// Nothing to do.
+				return nil
+			}
+
+			// Get Instance account ID.
+			var instanceAcctID string
```
tsmethurst, line 62: This just corrects the users table to use the default CreatedByApplicationID as mentioned above.

internal/db/bundb/status.go
```diff
@@ -299,11 +299,12 @@ func (s *statusDB) PopulateStatus(ctx context.Context, status *gtsmodel.Status)
 
 	if status.CreatedWithApplicationID != "" && status.CreatedWithApplication == nil {
 		// Populate the status' expected CreatedWithApplication (not always set).
+		// Don't error on ErrNoEntries, as the application may have been cleaned up.
 		status.CreatedWithApplication, err = s.state.DB.GetApplicationByID(
 			gtscontext.SetBarebones(ctx),
 			status.CreatedWithApplicationID,
 		)
-		if err != nil {
+		if err != nil && !errors.Is(err, db.ErrNoEntries) {
```
tsmethurst, line 307: This is sort of worth noting: previously we were just failing to return a status if the application wasn't found. But now that it's possible to create a managed application, post with it, and then delete the application, it doesn't make sense to fail if the app can't be found. We could add some logic to remove the application ID from old statuses that used it, when the app is deleted, but that's pretty damn expensive given the size of the statuses table and the fact we don't have an index on application ID right now. Seems easier to just look for it and fail gracefully if it's not in there.

internal/processing/account/delete.go
```diff
@@ -107,19 +107,33 @@ func (p *Processor) deleteUserAndTokensForAccount(ctx context.Context, account *
 		return gtserror.Newf("db error getting user: %w", err)
 	}
 
-	tokens := []*gtsmodel.Token{}
-	if err := p.state.DB.GetWhere(ctx, []db.Where{{Key: "user_id", Value: user.ID}}, &tokens); err != nil {
-		return gtserror.Newf("db error getting tokens: %w", err)
+	// Get all applications owned by user.
+	apps, err := p.state.DB.GetApplicationsManagedByUserID(ctx, user.ID, nil)
```
tsmethurst, line 111: I couldn't believe there were still paths using `[]db.Where`, what a blast from the past!!!

internal/processing/account/delete.go
```diff
@@ -107,19 +107,33 @@ func (p *Processor) deleteUserAndTokensForAccount(ctx context.Context, account *
...
+		return gtserror.Newf("db error getting apps: %w", err)
 	}
 
-	for _, t := range tokens {
-		// Delete any OAuth applications associated with this token.
-		if err := p.state.DB.DeleteApplicationByClientID(ctx, t.ClientID); err != nil {
-			return gtserror.Newf("db error deleting application: %w", err)
+	// Delete each app and any tokens it had created
+	// (not necessarily owned by deleted account).
+	for _, a := range apps {
```
tsmethurst, line 118: If a different account (account a) had used the managed application created by another account (account b), then tokens of account a will also be removed here. In practice this probably won't happen much but it's worth mentioning.

internal/processing/application/create.go
```diff
@@ -49,13 +53,32 @@ func (p *Processor) AppCreate(ctx context.Context, authed *apiutil.Auth, form *a
 		// Redirect URIs can be just one value, or can be passed
 		// as a newline-separated list of strings. Ensure each URI
 		// is parseable + normalize it by reconstructing from *url.URL.
-		for _, redirectStr := range strings.Split(form.RedirectURIs, "\n") {
+		// Also ensure we don't add multiple copies of the same URI.
+		redirectStrs := strings.Split(form.RedirectURIs, "\n")
+		added := make(map[string]struct{}, len(redirectStrs))
+
+		for _, redirectStr := range redirectStrs {
```
tsmethurst, line 60: Gentle bit of touching up here to trim any included `\r` characters from redirect URIs, and to avoid storing duplicates pointlessly.

internal/typeutils/internaltofrontend.go
```diff
@@ -623,8 +623,14 @@ func (c *Converter) AppToAPIAppSensitive(ctx context.Context, a *gtsmodel.Applic
 		return nil, gtserror.Newf("error getting VAPID public key: %w", err)
 	}
 
+	createdAt, err := id.TimeFromULID(a.ID)
```
tsmethurst, line 626: Honestly love this pattern of getting `created_at`.

internal/typeutils/internaltofrontend.go
```diff
@@ -1412,14 +1418,28 @@ func (c *Converter) baseStatusToFrontend(
 	apiStatus.InReplyToAccountID = util.PtrIf(s.InReplyToAccountID)
 	apiStatus.Language = util.PtrIf(s.Language)
 
-	if app := s.CreatedWithApplication; app != nil {
-		apiStatus.Application, err = c.AppToAPIAppPublic(ctx, app)
+	switch {
+	case s.CreatedWithApplication != nil:
```
tsmethurst, line 1422: This change is the counterpart of the change in the db functions where the call to get a status doesn't fail anymore if the application can't be found. Now we can use the presence of CreatedWithApplicationID but the absence of CreatedWithApplication to determine that the application a status was created with has been deleted, and indicate this to the user via the dummy application `Unknown/deleted application`. This is a lot neater than just failing imo.

web/source/settings/components/authorization/index.tsx
```diff
@@ -17,7 +17,7 @@
 	along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
tsmethurst, line 17: It's all frontend stuff below this point, avert your dainty eyes from the horrors kim

docs/api/authentication.md
```diff
@@ -19,18 +22,15 @@ curl \
...
-
-- `read`
-- `write`
-- `admin`
+!!! tip "Scopes"
+    It is always good practice to grant your application the lowest tier permissions it needs to do its job. e.g. If your application won't be making posts, use `scope=read` or even a subscope of that.
+    
+    In this spirit, "read" is used in the example above, which means that the application will be restricted to only being able to do "read" actions.
+    
+    For a list of available scopes, see https://docs.gotosocial.org/en/latest/api/swagger/
```
dmoonfire, line 30: This doesn't show as a link in the [rendered link](https://gotosocial--3906.org.readthedocs.build/en/3906/user_guide/settings/#applications) you posted, only text. It may need to be converted into a Markdown link. Maybe?

```
For a list of available scopes, see the [Swagger documentation](https://docs.gotosocial.org/en/latest/api/swagger/).
```

docs/user_guide/settings.md
```diff
@@ -287,3 +287,70 @@ Logging out of an application does not necessarily remove the token from the GoT
...
 !!! note
     Token "Last used" time is approximate and may be off by an hour in either direction.
+
+## Applications
+
+In the applications section, you can create a new managed OAuth client application, and search through applications that you've created.
+
+### What is an OAuth client application?
+
+A GoToSocial OAuth client application is equivalent to an OAuth 2.0 client as described here: https://auth0.com/intro-to-iam/what-is-oauth-2#oauth20-roles.
```
dmoonfire, line 297: This link also didn't render a link in https://gotosocial--3906.org.readthedocs.build/en/3906/user_guide/settings/#applications.

docs/api/authentication.md
tsmethurst: Oh thanks, I'll poke it!

docs/user_guide/settings.md
```diff
@@ -287,3 +287,70 @@ Logging out of an application does not necessarily remove the token from the GoT
 
 !!! note
     Token "Last used" time is approximate and may be off by an hour in either direction.
+
+## Applications
```
NyaaaWhatsUpDoc, line 291: damn these are some great docs, nice work!

internal/api/client/apps/appcreate.go
```diff
@@ -81,46 +86,80 @@ func (m *Module) AppsPOSTHandler(c *gin.Context) {
 		return
 	}
 
+	if authed.Token != nil {
+		// If a token has been passed, user
+		// needs write perm on applications.
+		if !slices.ContainsFunc(
```
NyaaaWhatsUpDoc, line 92: given this seems like code that might often be re-used, i think in the future it would be nice to have a function that does this and returns the appropriate error. Something like a method receiver on the Token{}, or an apiutil function that accepts a token, and just does something like `CheckScope(requires Scope) gtserror.WithCode`

internal/api/client/apps/appcreate.go
```diff
@@ -81,46 +86,80 @@ func (m *Module) AppsPOSTHandler(c *gin.Context) {
...
+			},
+		) {
+			const errText = "token has insufficient scope permission"
+			errWithCode := gtserror.NewErrorForbidden(errors.New(errText), errText)
+			apiutil.ErrorHandler(c, errWithCode, m.processor.InstanceGetV1)
+			return
+		}
+	}
+
+	if authed.Account != nil && authed.Account.IsMoving() {
```
NyaaaWhatsUpDoc, line 105: can the account here actually be nil?

tsmethurst: Yeah! This is the only endpoint you can POST to without being authed, so that client apps can register themselves.

internal/db/bundb/admin.go
NyaaaWhatsUpDoc: when we create them via the CLI, do we call a processor method? or just directly call db methods? if it's the former i think it might make a bit more sense to move this up to the processor method, otherwise the db method is doing a lot more application related logic

internal/db/bundb/application.go
```diff
@@ -69,34 +136,54 @@ func (a *applicationDB) getApplication(ctx context.Context, lookup string, dbQue
...
 		Table("applications").
-		Where("? = ?", bun.Ident("client_id"), clientID).
-		Exec(ctx); err != nil {
+		Where("? = ?", bun.Ident("id"), id).
+		Exec(ctx)
+	if err != nil {
 		return err
 	}
 
-	// NOTE about further side effects:
```
NyaaaWhatsUpDoc, line 88: i think it might still be worth leaving a note of some kind here, maybe just updated to indicate that applications can now be deleted elsewhere but still handle necessary changes

internal/db/bundb/application.go
```diff
@@ -363,3 +450,27 @@ func (a *applicationDB) DeleteTokenByRefresh(ctx context.Context, refresh string
...
+
+func (a *applicationDB) DeleteTokensByClientID(ctx context.Context, clientID string) error {
+	// Delete tokens owned by
+	// clientID and gather token IDs.
+	var tokenIDs []string
+	if _, err := a.db.
+		NewDelete().
+		Table("tokens").
+		Where("? = ?", bun.Ident("client_id"), clientID).
+		Returning("id").
```
NyaaaWhatsUpDoc, line 462: if you're using returning in a delete statement do you need to check for errnorows here?

internal/processing/account/delete.go
NyaaaWhatsUpDoc: god damn i thought we'd removed that everywhere :O that is some old-ass code

internal/typeutils/internaltofrontend.go
```diff
@@ -1412,14 +1418,28 @@ func (c *Converter) baseStatusToFrontend(
...
 				s.CreatedWithApplicationID, err,
 			)
 		}
+
+	case s.CreatedWithApplicationID != "":
+		// App existed for this status but not
+		// anymore, it's probably been cleaned up.
+		// Set a dummy application.
+		apiStatus.Application = &apimodel.Application{
+			Name: "Unknown/deleted application",
```
NyaaaWhatsUpDoc, line 1437: i think this might read a bit nicer in clients if it was just "uknown application", a bit shorter and without capitalisation as it's not an application name or anything

web/source/settings/components/authorization/index.tsx
NyaaaWhatsUpDoc: 💅

internal/db/bundb/application.go
NyaaaWhatsUpDoc: or not even, now i've read your other comment. i think it's still useful to have a code comment here though explaining why we don't specifically handle side-effects

NyaaaWhatsUpDoc:
a few comments, but otherwise looks great :)

internal/db/bundb/admin.go
tsmethurst: We just go straight to db methods. So not for this PR I'd say but probably a good idea if we refactor that a little into a processor thingy.

internal/db/bundb/application.go
tsmethurst: Good question. Just checked and apparently not :) https://github.com/superseriousbusiness/gotosocial/pull/3906/commits/0db3fe3ba67d5596a3a20043286945a3fe147931

internal/typeutils/internaltofrontend.go
tsmethurst: Good idea :) https://github.com/superseriousbusiness/gotosocial/pull/3906/commits/c96ec969de8db94748afd3cfec5f957f731d1fec

PR #3917, opened 2025-03-17
[feature] Allow user to choose "gallery" style layout for web view of profile

tsmethurst:
# Description

>

This pull request adds a "gallery" layout for the web view of a user's profile, which can be selected in the settings panel. This should be helpful for accounts that post mainly media attachments, and don't care so much about post content.

Screenshot from 2025-03-17 12-26-51

Rewritten index:

sqlite:

```
sqlite> .expert
sqlite> SELECT "status"."id" FROM "statuses" AS "status"
WHERE ("status"."account_id" = '016T5Q3SQKBT337DAKVSKNXXW1')
AND ("status"."visibility" = 2)
AND ("status"."in_reply_to_uri" IS NULL)
AND ("status"."boost_of_id" IS NULL)
AND ("status"."federated" = TRUE)
AND (("status"."attachments" IS NOT NULL) AND ("status"."attachments" != 'null') AND ("status"."attachments" != '[]'))
AND ("status"."id" < '01HF3XRR3X64J4VQ7YZ815DA54')
ORDER BY "status"."id" DESC LIMIT 20;
   ...>    ...>    ...>    ...>    ...>    ...>    ...>    ...> (no new indexes)

SEARCH status USING INDEX statuses_profile_web_view_idx (account_id=? AND visibility=? AND in_reply_to_uri=? AND boost_of_id=? AND federated=? AND id<?)

sqlite> explain query plan SELECT "status"."id" FROM "statuses" AS "status"
WHERE ("status"."account_id" = '016T5Q3SQKBT337DAKVSKNXXW1')
AND ("status"."visibility" = 2)
AND ("status"."in_reply_to_uri" IS NULL)
AND ("status"."boost_of_id" IS NULL)
AND ("status"."federated" = TRUE)
AND (("status"."attachments" IS NOT NULL) AND ("status"."attachments" != 'null') AND ("status"."attachments" != '[]'))
AND ("status"."id" < '01HF3XRR3X64J4VQ7YZ815DA54')
ORDER BY "status"."id" DESC LIMIT 20;
   ...>    ...>    ...>    ...>    ...>    ...>    ...>    ...> QUERY PLAN
`--SEARCH status USING INDEX statuses_profile_web_view_idx (account_id=? AND visibility=? AND in_reply_to_uri=? AND boost_of_id=? AND federated=? AND id<?)
sqlite> 
```

## Checklist


tsmethurst:
Marking this a draft until I figure out which indexes need to change (if any) for this.

internal/db/bundb/account.go
```diff
@@ -878,6 +878,30 @@ func (a *accountDB) GetAccountFaves(ctx context.Context, accountID string) ([]*g
...
+	return q.WhereGroup(" AND ", func(q *bun.SelectQuery) *bun.SelectQuery {
+		switch d := q.Dialect().Name(); d {
+		case dialect.PG:
+			return q.
+				Where("? IS NOT NULL", bun.Ident("status.attachments")).
+				Where("? != '{}'", bun.Ident("status.attachments"))
+
+		case dialect.SQLite:
+			return q.
+				Where("? IS NOT NULL", bun.Ident("status.attachments")).
```
tsmethurst, line 894: Removed some of the clauses in here as I checked in my db and there were no entries where `attachments = ''` for example.

internal/db/bundb/account.go
```diff
@@ -878,6 +878,30 @@ func (a *accountDB) GetAccountFaves(ctx context.Context, accountID string) ([]*g
...
+				Where("? != 'null'", bun.Ident("status.attachments")).
+				Where("? != '[]'", bun.Ident("status.attachments"))
+
+		default:
+			log.Panicf(ctx, "dialect %d was neither pg nor sqlite", d)
```
NyaaaWhatsUpDoc, line 899: honestly i would just make this a regular panic, then no need to take an otherwise useless ctx argument, and with direct calls to "panic()" the compiler can infer more easily that *that* particular codepath is a panic.

internal/processing/account/rss.go
```diff
@@ -115,8 +115,20 @@ func (p *Processor) GetRSSFeedForUsername(ctx context.Context, username string)
...
 		feed.Updated = lastPostAt
 
-		// Retrieve latest statuses as they'd be shown on the web view of the account profile.
-		statuses, err := p.state.DB.GetAccountWebStatuses(ctx, account, rssFeedLength, "")
+		// Retrieve latest statuses as they'd be shown
+		// on the web view of the account profile.
+		//
+		// Take into account whether the user wants
+		// their web view laid out in gallery mode.
+		mediaOnly := account.Settings != nil &&
```
NyaaaWhatsUpDoc, line 123: my only worry with this is it feels like we're starting to leak a lot of the frontend logic into the backend. i'm not saying that this needs to be changed, and this comment is more a general comment of this PR than this particular rss feed function change, but just something to be wary of in future feature additions to the frontend.

tsmethurst: Yeah true... we might want to eventually separate off the web model stuff and the api model stuff more cleanly :thinking: And treat the web stuff more as frontend stuff, rather than the sort of half backend half frontend stuff it is now.

PR #3925, opened 2025-03-20
feat: Relax URL matching

daenney:
# Description

Instead of only linkifying things with an explicit http or https scheme, the xurls.Relaxed also matches links with known TLDs. This means that text like 'banana.com' will also be matched, despite the missing http/https scheme. This also works to linkify email addresses, which is handy.

This should also ensure we catch links without a scheme for the purpose of spam checking.

## Checklist

### DIFF
--- a/internal/regexes/regexes.go
+++ b/internal/regexes/regexes.go
@@ -40,7 +40,6 @@ const (
 	reports   = "reports"
 	accepts   = "accepts"
 
-	schemes                  = `(http|https)://`                                         // Allowed URI protocols for parsing links in text.
 	alphaNumeric             = `\p{L}\p{M}*|\p{N}`                                       // A single number or script character in any language, including chars with accents.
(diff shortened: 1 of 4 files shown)


internal/regexes/regexes.go
```diff
@@ -79,14 +78,9 @@ const (
...
-	LinkScheme = func() *regexp.Regexp {
-		rgx, err := xurls.StrictMatchingScheme(schemes)
-		if err != nil {
-			panic(err)
-		}
-		return rgx
-	}()
+	// URL captures anything that looks like a URL. This includes
+	// URLs without a scheme, based on a built-in list of TLDs.
+	URL = xurls.Relaxed()
```
NyaaaWhatsUpDoc, line 83: i think we should maybe name this `URLlike` just so we don't accidentally end up using this somewhere important and end up marking things as URLs we shouldn't do.

NyaaaWhatsUpDoc:
one quick comment but otherwise good to go. once you've addressed it either by commenting back with good reasoning / whatever or fixing it up feel free to squerge yourself :)

Issue #3381, opened 2024-10-02
[bug] Error dereferencing account https://fed.brid.gy/fed.brid.gy


suflors:
### Describe the bug with a clear and concise description of what the bug is.

When attempting to follow and be followed back by @bsky.brid.gy@bsky.brid.gy, I get an error with status code 500.

### What's your GoToSocial Version?

v0.16.0+git-f1cbf6f

### GoToSocial Arch

amd64 Binary

### What happened?

The log reported the following:
`Oct 02 18:53:05 Marinette gotosocial[892]: timestamp="02/10/2024 18:53:05.983" func=server.init.func1.Logger.14.1 level=ERROR latency="562.026329ms" userAgent="Bridgy Fed (https://fed.brid.gy/)" method=GET statusCode=500 path=/users/kelsey clientIP=34.42.221.223 pubKeyID=https://fed.brid.gy/fed.brid.gy#key errors="Error #01: AuthenticateFederatedRequest: error dereferencing account https://fed.brid.gy/fed.brid.gy: enrichAccountSafely: error getting account https://fed.brid.gy/fed.brid.gy from database after race: sql: no rows in result set\n" requestID=1mm98kmj04001hp5trs0 msg="Internal Server Error: wrote 910B"`

### What you expected to happen?

The bot to be able to accept my follow request and follow me back.

### How to reproduce it?

Send a follow request to @bsky.brid.gy@bsky.brid.gy

### Anything else we need to know?

I had first reported this to the bridgyfed repo in the issue [here](https://github.com/snarfed/bridgy-fed/issues/1356).

tsmethurst:
Thanks for reporting. Sounds like it might still be related to the unique url constraint issue but we'll have to investigate properly after 0.17.0 is out. I'm a bit puzzled that you try to follow bsky.brid.gy but end up at fed.brid.gy, we'll have to have a look in their docs and see what's up there, probably some redirect stuff to make the bridge work. Will come back to this :)

snarfed:
Hi! Bridgy Fed developer here. Thanks for looking at this!

[...]

Issue #2082, opened 2023-08-07
[bug] ActivityPub collection URL:s should not be required to be unique


voxpelli:
### Describe the bug with a clear and concise description of what the bug is.

Nowhere in the ActivityPub is it mandated that two actors cannot share the same collection for eg an outbox, but GoToSocial seems to model these collections as having a unique URL per actor, something that is not spec compliant and can cause interoperability problems.

Probably defined here: https://github.com/superseriousbusiness/gotosocial/blob/be3718f6e4c7bb229f9eb78b44c52d14399977cc/internal/gtsmodel/account.go#L70

### What's your GoToSocial Version?

latest

### What happened?

Two actors with the same outbox URL is reported as not working in GoToSocial: https://github.com/astro/buzzrelay/pull/9

### What you expected to happen?

That no additional limitations outside of those defined in 4. And 5. in the ActivityPub spec are enforced: https://www.w3.org/TR/activitypub/#actors

(Unless there is a `MUST` or at least a `SHOULD` for such uniqueness elsewhere in the spec which I have missed)

NyaaaWhatsUpDoc:
GoToSocial currently doesn't support federation with relays, so broken behaviour is to be expected https://github.com/superseriousbusiness/gotosocial/issues/1123. Implementation specifics on our end will likely happen at a later discussion, but until then isn't necessarily a bug.

voxpelli:
@NyaaaWhatsUpDoc The bug isn’t about relays, it’s about lack of spec compliance

tsmethurst:
We'll investigate and have a think about this. This is the first service we've seen that uses outboxes this way.

While what you're doing *is* strictly spec compliant according to the spec as written, to me the explicit case of shared inboxes existing in the protocol sort of suggests that inbox and outbox URIs which are not shared should be unique to the actor.

But yes, perhaps we can relax our database constraint. Did you have trouble with any other fedi server based on this issue?

voxpelli:
I'm neither the creator of the relay nor the user of GoToSocial, I'm merely an independent implementor of Fediverse specifications who prefer spec compliance over potential surprises, and who thought it would be good to notify you of this since the original author of that other issue didn't feel compelled to do so :)

Issue #1820, opened 2023-05-23
[bug] Relax unique `URL` requirement for accounts


natsukagami:
### Describe the bug with a clear and concise description of what the bug is.

Writefreely instances with `single_user` mode enabled will WebFinger the only user and the internal instance account with the same URL. 

This is giving GtS a hard time trying to follow the blog, since it cannot create a new account: GtS assumes and enforces that each account has an unique URL field.

Writefreely's webfinger output, from the discussion on the Matrix channel (on the now deceased Writefreely instance https://write.nkagami.me) 
```json
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    "https://w3id.org/security/v1"
  ],
  "type": "Person",
  "id": "https://write.nkagami.me/api/collections/write.nkagami.me",
  "inbox": "https://write.nkagami.me/api/collections/write.nkagami.me/inbox",
  "outbox": "https://write.nkagami.me/api/collections/write.nkagami.me/outbox",
  "preferredUsername": "[write.nkagami.me](http://write.nkagami.me/)",
  "url": "https://write.nkagami.me/",
  "name": "[write.nkagami.me](http://write.nkagami.me/)",
  "icon": {
    "type": "Image",
    "mediaType": "image/png",
    "url": "https://write.nkagami.me/img/avatars/w.png"
  },
  "following": "https://write.nkagami.me/api/collections/write.nkagami.me/following",
  "followers": "https://write.nkagami.me/api/collections/write.nkagami.me/followers",
  "summary": "",
  "publicKey": {
    "id": "https://write.nkagami.me/api/collections/write.nkagami.me#main-key",
    "owner": "https://write.nkagami.me/api/collections/write.nkagami.me",
[...]

PR #3928, opened 2025-03-22
[chore] Migrate accounts to new table, relax uniqueness constraint of actor `url` and collections

tsmethurst:
# Description

>

Previously we were treating account `url`, `followers`, `following`, `inbox`, `outbox`, and `featured` as though the spec says they have to be unique, like `id`. However, they don't!

This PR migrates our accounts to a new table without those uniqueness constraints, and enforces some uniqueness at an application-logic level instead with new `GetOneAccountByXYZ` functions.

Along the way, removes the `bot` field from accounts and instead relies on the actor type to determine if an actor is a bot, and changes the actor type to an enum type instead of a string (should save a bit of space). 

closes https://github.com/superseriousbusiness/gotosocial/issues/2082
closes https://github.com/superseriousbusiness/gotosocial/issues/1820
closes https://github.com/superseriousbusiness/gotosocial/issues/3381

Tested the migration on a backed-up SQLite DB and it went fine. Tested on an empty PG db and that was also fine. ~~But this will remain a draft until I've tested it on a backed-up PG db, just to be sure everything works.~~ Tested on backed-up PG db and that was fine.

## Checklist


internal/db/bundb/account.go
```diff
@@ -46,6 +46,72 @@ type accountDB struct {
 	state *state.State
 }
 
+func (a *accountDB) getAccountsBy(
```
NyaaaWhatsUpDoc, line 49: unfortunately we can't really make this generic, unless it was selecting by another unique field type, like IDs. for a unique field, the cache knows if it doesn't a value for that particular field then it needs to pass it to the "load" callback. but for non-unique fields if it say has 2 values stored for a particular URL, but 3 actually exist in the database, it's impossible for it to know that there are 3 in the database and so it will return only the 2 values. it's impossible for it to know without it doing a database call every single time (at which point, the value of the cache is severely diminished).

you could possibly have an intermediary SliceCache that stores a lookup of URLs to lists of IDs accounts with that URL, but given we *always* fallback on URL if a URI fails, and the SliceCache doesn't handle storing negative results (i.e. errors), there's potentially a lot of times it will end up hitting the database.

this would be much simpler if we took a stricter tolerance on dereferencing, that unless it's coming from the search endpoint (and so can be either URL or URI), they must always provide a URI when they should be providing one. in that case we could have a separate GetAccountsByURL() function that calls the appropriate database function and returns either one or many accounts, and not have the potential performance issues. what do you think of this as a potential solution? because honestly i'm not sure how i feel about keeping our hazy on URI vs URL when it comes to dereferencing of things both from the performance aspect as mentioned previously, and because it feels like it's a possible easy mistake to accidentally end up returning one account for linking to something instead of another.

tsmethurst: Got ya, thank you! This should be resolved in this commit:

https://github.com/superseriousbusiness/gotosocial/pull/3928/commits/1bfc1aa7d178ccbb19cb70c52141831b9b07e810

I added an option for GetAccountByURI to try by URL, and it's only set to true when we're doing stuff with user input involved, not when we're doing stuff with automated / federation thingies where we should strictly enforce that things happen using URI not URL.

Also updated the database stuff to select IDs and use the ID cache function. It doesn't avoid database calls completely but I think it's probably good enough for now, as those selects should be very quick since everything is appropriately indexed.

NyaaaWhatsUpDoc:
good stuff comrade :sunglasses:

NyaaaWhatsUpDoc (approved):

PR #3932, opened 2025-03-23
Fix Swagger URL for the "edit status" operation

### LINKED ISSUE #3910: [docs] Status edit API request URL requires status ID
In the Swagger docs for the stus edit PUT request, there is no mention currently of requiring a status ID:

Image

However, in the [Mastodon docs](https://docs.joinmastodon.org/methods/statuses/#edit), it is stated that the status ID is a part of the request URL:

Image

After testing a status edit in the Elk client with my GtS server and recording the network requests using Firefox's Network Monitor, the status edit is using the status ID in the request URL:

Image

Shouldn't this mean that the URL in the Swagger docs should be `PUT /api/v1/statuses/{id}` like the others?

leonid-shevtsov:
# Description

The "Edit status" operation is incorrectly listed under `/api/v1/statuses` whereas it actually requires an `{id}` route param.

closes #3910

## Checklist

### DIFF
--- a/docs/api/swagger.yaml
+++ b/docs/api/swagger.yaml
@@ -10718,6 +10718,76 @@ paths:
             summary: Create a new status using the given form field parameters.
             tags:
                 - statuses
+    /api/v1/statuses/{id}:
+        delete:
+            description: |-
+                The deleted status will be returned in the response. The `text` field will contain the original text of the status as it was submitted.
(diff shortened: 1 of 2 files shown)


leonid-shevtsov:
Yay I beat the linter (and updated the doc properly in the process)

tsmethurst:
Aha, thank you for this! :)

PR #3941, opened 2025-03-25
[performance] rewrite timelines to rely on new timeline cache type

NyaaaWhatsUpDoc:
# description

this is a rewrite of our caching technique for timelines. our wonderful code in `internal/timeline` has served us well for years now, but unfortunately doesn't play too nicely with attempts to cache status filtering or user mute results (hence why this ends up handled in the typeconverter). this rewrite moves the timeline caching core logic into [go-structr](https://codeberg.org/gruf/go-structr) (maintained by myself) in order to take advantage of the automated indexing logic, which should make all timeline operations other than direct selects significantly faster since it will be hashmap lookups instead of linked-list-iterations. this code also allows sharing memory pools for a bunch of commonly used internal objects with the other "go-structr" cache types we have which also *might* help improve memory usage.

there is then a wrapping package under `internal/cache/timeline` that wraps the new `structr.Timeline{}` type with our necessary status loading and frontend preparation logic. this makes it very easy to add any new status timeline types, e.g. tags, local, public in the future, and should be reasonably easy to copy and modify for other types e.g. favourites and notifications.

there has also been an important change to the philosophy of exactly *what* gets cached in each status timeline, best summarised by the `StatusTimeline{}` type description:

> // StatusTimeline provides a concurrency-safe sliding-window
// cache of the freshest statuses in a timeline. Internally,
// only StatusMeta{} objects themselves are stored, loading
// the actual statuses when necessary, but caching prepared
// frontend API models where possible.
//
// Notes on design:
//
// Previously, and initially when designing this newer type,
// we had status timeline caches that would dynamically fill
// themselves with statuses on call to Load() with statuses
// at *any* location in the timeline, while simultaneously
// accepting new input of statuses from the background workers.
// This unfortunately can lead to situations where posts need
// to be fetched from the database, but the cache isn't aware
// they exist and instead returns an incomplete selection.
// This problem is best outlined by the follow simple example:
//
// "what if my timeline cache contains posts 0-to-6 and 8-to-12,
// and i make a request for posts between 4-and-10 with no limit,
// how is it to know that it's missing post 7?"
//
// The solution is to unfortunately remove a lot of the caching
// of "older areas" of the timeline, and instead just have it
// be a sliding window of the freshest posts of that timeline.
// It gets preloaded initially on start / first-call, and kept
// up-to-date with new posts by streamed inserts from background
// workers. Any requests for posts outside this we know therefore
// must hit the database, (which we then *don't* cache).

under the manual testing of this PR, i also discovered missing and ended up adding appropriate hooks for received federated undo events not necessarily invalidating timelines. and then more generally just modifying a bunch of existing timeline related logic (that will be later updated to use caches) to share as much code as possible.

## todo

- ~~fix remaining `processing/timeline` test~~
- ~~finish missing code comments in places~~
- ~~add test for list timeline in `processing/timeline`~~ out-of-scope, also PR is getting long in the tooth :p 
- ~~fix / squash linter warnings~~
- ~~add check before adding new boost of existing status~~


NyaaaWhatsUpDoc:
another note on this: given it's such a large change it needs some *serious* testing before we consider releasing it. it may be worth hurrying v0.19.0 out the door then squerging this.

internal/cache/db.go
```diff
@@ -1329,7 +1329,6 @@ func (c *Caches) initStatus() {
 		s2.Mentions = nil
 		s2.Emojis = nil
 		s2.CreatedWithApplication = nil
-		s2.Edits = nil
```
tsmethurst, line 1332: Should this not also be nilled?

internal/cache/timeline/preload.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+// preloader provides a means of synchronising the
+// initial fill, or "preload", of a timeline cache.
+// it has 4 possible states in the atomic pointer:
+// - preloading    = &(interface{}(*sync.WaitGroup))
+// - preloaded     = &(interface{}(nil))
+// - needs preload = &(interface{}(false))
+// - brand-new     = nil (functionally same as 'needs preload')
+type preloader struct{ p atomic.Pointer[any] }
```
tsmethurst, line 34: Oh my god :') But also, yes, hahaha, YES!

internal/cache/timeline/preload.go
```diff
@@ -0,0 +1,152 @@
...
+
+// CheckPreload will safely check the preload state,
+// and if needed call the provided function. if a
+// preload is in progress, it will wait until complete.
+func (p *preloader) CheckPreload(preload func(*any)) {
+	for {
+		// Get state ptr.
+		ptr := p.p.Load()
+
+		if ptr == nil || *ptr == false {
```
tsmethurst, line 68: Is there a chance of a race here if something else changes the ptr value? Or no?

internal/cache/timeline/status.go
```diff
@@ -0,0 +1,842 @@
...
+// how is it to know that it's missing post 7?"
+//
+// The solution is to unfortunately remove a lot of the caching
+// of "older areas" of the timeline, and instead just have it
+// be a sliding window of the freshest posts of that timeline.
+// It gets preloaded initially on start / first-call, and kept
+// up-to-date with new posts by streamed inserts from background
+// workers. Any requests for posts outside this we know therefore
+// must hit the database, (which we then *don't* cache).
+type StatusTimeline struct {
```
tsmethurst, line 104: This seems like a reasonable tradeoff!

internal/cache/timeline/status.go
```diff
@@ -0,0 +1,842 @@
...
+	// underlying timeline cache of *StatusMeta{},
+	// primary-keyed by ID, with extra indices below.
+	cache structr.Timeline[*StatusMeta, string]
+
+	// preloader synchronizes preload
+	// state of the timeline cache.
+	preloader preloader
+
+	// fast-access cache indices.
+	idx_ID               *structr.Index //nolint:revive
```
tsmethurst, line 115: Is revive complaining about the pascal case? :thinking:

internal/cache/timeline/status.go
```diff
@@ -0,0 +1,842 @@
...
+	// Clear timeline
+	// before preload.
+	t.cache.Clear()
+
+	// Our starting, page at the top
+	// of the possible timeline.
+	page := new(paging.Page)
+	order := paging.OrderDescending
+	page.Max.Order = order
+	page.Max.Value = plus24hULID()
```
tsmethurst, line 236: Plus 24h is pretty generous, we should probably limit this to plus an hour or something

internal/paging/page.go
```diff
@@ -78,6 +79,27 @@ func (p *Page) order() Order {
 	}
 }
 
+// GetBounds is a small helper function to return low and high page bound ptrs (checking for nil page).
+func (p *Page) GetBounds() (lo *Boundary, hi *Boundary) {
```
NyaaaWhatsUpDoc, line 83: these were used in an old version of this PR, i'll remove them as they're now unused :)

internal/processing/timeline/home.go
```diff
@@ -19,132 +19,79 @@ package timeline
...
+
+		// Database load function.
+		func(pg *paging.Page) (statuses []*gtsmodel.Status, err error) {
+			return p.state.DB.GetHomeTimeline(ctx, requester.ID, pg)
+		},
+
+		// Pre-filtering function,
+		// i.e. filter before caching.
+		func(s *gtsmodel.Status) bool {
+
```
NyaaaWhatsUpDoc, line 88: just realized this is now missing the final filtering stage of local-only statuses. will fix :)

internal/db/bundb/timeline.go
```diff
@@ -38,346 +36,147 @@ type timelineDB struct {
...
+		page,
 
-	if frontToBack {
-		// Page down.
-		q = q.Order("status.id DESC")
-	} else {
-		// Page up.
-		q = q.Order("status.id ASC")
-	}
+		func(q *bun.SelectQuery) (*bun.SelectQuery, error) {
```
tsmethurst, line 158: same question here re: query parameter ordering

internal/federation/federatingdb/undo.go
```diff
@@ -178,7 +180,16 @@ func (f *federatingDB) undoFollow(
 		return err
 	}
 
-	log.Debug(ctx, "Follow undone")
+	// Send the deleted follow through to
+	// the fedi worker to process side effects.
+	f.state.Workers.Federator.Queue.Push(&messages.FromFediAPI{
```
tsmethurst, line 185: Hmm can you explain these additions? Were there side effects that weren't being processed properly before? Not a criticism, just want to understand.

internal/processing/timeline/public.go
```diff
@@ -19,152 +19,135 @@ package timeline
...
+	page *paging.Page,
+) (
+	*apimodel.PageableResponse,
+	gtserror.WithCode,
+) {
+	return p.getStatusTimeline(ctx,
+
+		// Auth'd
+		// account.
+		requester,
```
tsmethurst, line 63: Don't forget requester account can be nil, probably worth mentioning it here.

internal/processing/timeline/public.go
```diff
@@ -19,152 +19,135 @@ package timeline
...
```
tsmethurst, line 115: Same as above, think requester can also be nil here.

internal/processing/timeline/tag.go
```diff
@@ -40,37 +38,73 @@ import (
...
+		func(pg *paging.Page) (statuses []*gtsmodel.Status, err error) {
+			return p.state.DB.GetTagTimeline(ctx, tag.ID, pg)
+		},
+
+		// Filtering function,
+		// i.e. filter before caching.
+		func(s *gtsmodel.Status) bool {
+
+			// Check the visibility of passed status to requesting user.
+			ok, err := p.visFilter.StatusHomeTimelineable(ctx, requester, s)
```
tsmethurst, line 102: This should be status public timelineable i believe

tsmethurst (commented):
Great work!!!

internal/cache/db.go
NyaaaWhatsUpDoc: it should be!

internal/cache/timeline/preload.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		// Anything else
+		// means success.
+		return
+	}
+}
+
+// start attempts to start the given preload function, by
+// performing a CAS operation with 'old'. return is success.
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: it stands for CompareAndSwap, and is often referred to as CAS, specifically in atomic operations. though i updated the code here to use the full version (no accronym) for easier understanding: 6e26972fcd5862f2e0650f4cbba4e384cce8e4ac

internal/cache/db.go
NyaaaWhatsUpDoc: fixed that here: dd0f51ed1e45ec42c74ce5778cb3f5f4a43ff5fc

internal/cache/timeline/status.go
```diff
@@ -0,0 +1,842 @@
...
+// preload as being required, call .Clear().
+func (t *StatusTimeline) Preload(
+
+	// loadPage should load the timeline of given page for cache hydration.
+	loadPage func(page *paging.Page) (statuses []*gtsmodel.Status, err error),
+
+	// filter can be used to perform filtering of returned
+	// statuses BEFORE insert into cache. i.e. this will effect
+	// what actually gets stored in the timeline cache.
+	filter func(each *gtsmodel.Status) (delete bool),
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: thank you, thank you :sunglasses:

internal/cache/timeline/status.go
NyaaaWhatsUpDoc: yeah that's a fair point, have updated that here: 8911c9194

internal/paging/page.go
NyaaaWhatsUpDoc: removed here: 03fdc2704

internal/processing/timeline/public.go
NyaaaWhatsUpDoc: and again, here: 8a5fcd7db

internal/processing/timeline/tag.go
NyaaaWhatsUpDoc: oh nice catch! have fixed that here: 7cd2269cd

internal/processing/timeline/timeline.go
```diff
@@ -18,9 +18,33 @@
...
+	"github.com/superseriousbusiness/gotosocial/internal/gtsmodel"
+	"github.com/superseriousbusiness/gotosocial/internal/id"
+	"github.com/superseriousbusiness/gotosocial/internal/paging"
 	"github.com/superseriousbusiness/gotosocial/internal/state"
 	"github.com/superseriousbusiness/gotosocial/internal/typeutils"
+	"github.com/superseriousbusiness/gotosocial/internal/util/xslices"
+)
+
+var (
+	// pre-prepared URL values to be passed in to
```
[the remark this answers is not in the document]
NyaaaWhatsUpDoc: :sunglasses:

internal/cache/timeline/preload.go
NyaaaWhatsUpDoc: No there shouldn't be. We never directly set the memory value at the location provided by the ptr, we only ever set the memory location pointer itself with a new value each time (which is protected by atomic loads / CompareAndSwap operations)

internal/federation/federatingdb/undo.go
NyaaaWhatsUpDoc: specifically we weren't handling timeline side-effects for undo fave or undo follows, so now we do correctly handle removal of an accounts posts from your timelines on unfollow, and invalidating cached status interactions on unfave.

internal/db/bundb/timeline.go
NyaaaWhatsUpDoc: just confirmed it here:

```
kim @ ~/Projects/main/gts.5
--> go clean -testcache; GTS_LOG_LEVEL=trace test/run-sqlite.sh ./internal/db/bundb/... -v -run=TestTimelineTestSuite/TestGetPublicTimelineLocal | grep 'SELECT.*statuses.*local.*visibility.*pending_approval.*boost_of_id'
timestamp="25/04/2025 16:56:36.621" func=bundb.queryHook.AfterQuery level=TRACE duration=87.555µs query=SELECT "id" FROM "statuses" WHERE ("local" = TRUE) AND ("visibility" = 2) AND ("pending_approval" = FALSE) AND ("boost_of_id" IS NULL) ORDER BY "id" DESC LIMIT 20
```

internal/cache/timeline/status.go
NyaaaWhatsUpDoc: it is indeed, though i have tried removing the underscore and it does make it a bit harder to read in this case. what do you think??

tsmethurst: Well personally I'd prefer to keep consistency with the rest of the codebase in terms of variable names, but I am actually not really that bothered :') Feel free to add a linting rule to `.golangcilint` or whatever it's called if you want though, so you don't have to comment these things individually.

tsmethurst:
Looks good to me, squerge if you're happy :)

internal/cache/timeline/status.go
NyaaaWhatsUpDoc: I'll have a play around with it in a follow-up PR, maybe there's another alternative i can find that still reads well :)

PR #3964, opened 2025-04-03
Fix Atkinson Hyperlegible font embedding on Ecks Pee theme.

### LINKED ISSUE #3946: [bug/frontend] No visible text on Ecks Pee theme viewed on Tor Browser

OK, this is a weird edge case, but for some reason none of the text is visible on my GTS profile with the XP theme when I view it with Tor on Linux. I've spot-checked about half the other themes and they all display OK.

I remember running into a similar issue with another weird browser combo (something running on Haiku IIRC), but this one's based on Firefox so it ought to be able to handle it.

I suspect it's probably related to the anti-fingerprinting changes, maybe in relation to media queries or fonts not being downloaded. But at the moment that's only a suspicion. I'll dig into it with the dev tools when I get a chance and hopefully add some more solid info

0.18.3

### Browser version

Tor Browser 14.0.7

### What happened?

Viewing my profile shows icons and layout, with post boundaries sized correctly to fit the content in them...but no letters, numbers or symbols. 

Image

Actually the Show More/Show Less buttons show up correctly and even function, but the post content being revealed is still empty.

I was expecting to see letters.

### How to reproduce it?

Set a GTS profile to use the EcksPee theme. Visit the profile with Tor Browser.

When I get a chance I'll see if the same thing happens on Windows too, or if it's specific to Tor on Linux.

kvibber:
Most browsers just take the second src line and they're fine, but Tor has trouble displaying the woff version on Linux. With two separate lines it doesn't fall back correctly.

closes #3946

## Checklist

Not sure how to add a test to an issue that only shows up in Tor on Linux.

It's CSS only, so these aren't really applicable

### DIFF
--- a/web/assets/themes/ecks-pee.css
+++ b/web/assets/themes/ecks-pee.css
@@ -93,29 +93,29 @@
   font-family: "Atkinson Hyperlegible";
   font-weight: normal;
   font-style: normal;
-  src: url(/assets/fonts/Atkinson-Hyperlegible-Regular-102a.woff2) format('woff2');
-  src: url(/assets/fonts/Atkinson-Hyperlegible-Regular-102.woff) format('woff');
+  src: url(/assets/fonts/Atkinson-Hyperlegible-Regular-102a.woff2) format('woff2'),
+       url(/assets/fonts/Atkinson-Hyperlegible-Regular-102.woff) format('woff');
 }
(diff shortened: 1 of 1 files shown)


tsmethurst:
Ohhhh great, thank you very much :)

PR #3980, opened 2025-04-08
[docs] add a command to test the server compatibility with GTS

OniriCorpe:
# Description

I think an easy way to know if we have t use the standard or ‘nowasm’ builds would be handy and help people to not use the standard builds on uncompatible hardware

## Checklist


tsmethurst:
That's handy, thanks! ^_^

Issue #3996, opened 2025-04-13
[bug] WebSockets don't work when using through masto-fe-standalone in Chrome-based browsers


iliazeus:
### Describe the bug with a clear and concise description of what the bug is.

It looks like some web clients, [like `masto-fe-standalone`](https://codeberg.org/superseriousbusiness/masto-fe-standalone/src/commit/370a666d27638530fbf51007d1defc2f883a0b62/app/javascript/flavours/glitch/stream.js#L238), pass the token _both_ in query and in `Sec-WebSocket-Protocol`.

As was discovered in https://github.com/superseriousbusiness/gotosocial/pull/3169, Chrome-based browsers expect the response to have the same value in this header that the request had.

But [the way this code is currently written](https://github.com/superseriousbusiness/gotosocial/blob/c8a780e12a8112a4d4cbe8ed72d241a382f73903/internal/api/client/streaming/stream.go#L162), it first checks the query, and if the token is there, it does not send the `Sec-WebSocket-Protocol` back, regardless of whether the client passed it. Because of that, `masto-fe-standalone` + GoToSocial does not work in Chrome-based browsers.

Error text from Chromium console:

```
WebSocket connection to 'wss://gts.iliazeus.lol/api/v1/streaming?access_token=!!REDACTED!!' failed: Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' header but no response was received
```

### What's your GoToSocial Version?

v0.18.3

### GoToSocial Arch

x86_64 binary

### Anything else we need to know?

<details>
<summary>Request headers example - taken from Firefox</summary>

```
ET /api/v1/streaming?access_token=!!REDACTED!! HTTP/1.1
Host: gts.iliazeus.lol
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:136.0) Gecko/20100101 Firefox/136.0
Accept: */*
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate, br, zstd
Sec-WebSocket-Version: 13
Origin: https://web.gts.iliazeus.lol
Sec-WebSocket-Protocol: !!REDACTED!!
Sec-WebSocket-Extensions: permessage-deflate
Sec-WebSocket-Key: lA0isNfhh2OVn/o904V+Tw==
DNT: 1
Sec-GPC: 1
Connection: keep-alive, Upgrade
Sec-Fetch-Dest: empty
Sec-Fetch-Mode: websocket
[...]

PR #4009, opened 2025-04-14
[bugfix] websocket header token not always returned

NyaaaWhatsUpDoc:
Should fix: https://github.com/superseriousbusiness/gotosocial/issues/3996

### DIFF
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -163,8 +163,8 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
 	// correct, but some client do it since Mastodon allows it, so
 	// we must allow it for compatibility.
 	//
-	// Chrome browser also *always* expects the "Sec-Websocket-Protocol"
-	// response value to match input, so we always have to check it.
+	// Chrome also *always* expects the "Sec-Websocket-Protocol"
+	// response value to match input, so it must always be checked.
 	queryToken := c.Query(AccessTokenQueryKey)
 	headerToken := c.Query(AccessTokenHeader)
 
--- a/internal/api/client/streaming/stream.go
+++ b/internal/api/client/streaming/stream.go
@@ -18,6 +18,7 @@
 package streaming
 
 import (
+	"cmp"
 	"context"
 	"net/http"
 	"slices"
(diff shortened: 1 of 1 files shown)


internal/api/client/streaming/stream.go
```diff
@@ -153,25 +154,22 @@ var pingMsg = []byte("ping!")
...
-	}
+	// Check both query parameter AND header "Sec-Websocket-Protocol"
+	// value for a token. The latter is hacky and not technically
+	// correct, but some client do it since Mastodon allows it, so
+	// we must allow it for compatibility.
+	//
+	// Chrome also *always* expects the "Sec-Websocket-Protocol"
+	// response value to match input, so it must always be checked.
+	queryToken := c.Query(AccessTokenQueryKey)
+	headerToken := c.Query(AccessTokenHeader)
```
iliazeus, line 169: Should it be `GetHeader()`?
```suggestion
	headerToken := c.GetHeader(AccessTokenHeader)
```

NyaaaWhatsUpDoc: indeed, thank you :sweat_smile:

internal/api/client/streaming/stream.go
```diff
@@ -241,12 +238,14 @@ func (m *Module) StreamGETHandler(c *gin.Context) {
...
-	if tokenInHeader {
-		// Return the token in the response,
-		// else Chrome fails to connect.
+	if headerToken != "" {
+
+		// Chrome always requires the input header
+		// token to be provided in response, even
+		// if it also provided via query... *shrugs*
 		//
 		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#sec-websocket-protocol
```
iliazeus, line 247: The MDN link should be enough to document this. Or, at least, better to remove the `*shrug*`, since (according to the linked MDN article) Chrome _is_ technically more correct in requiring the header here 🤓 

> The Sec-WebSocket-Protocol header specifies one or more WebSocket protocols that you wish to use, in order of preference. The first one that is supported by the server will be selected and returned by the server in a Sec-WebSocket-Protocol header included in the response.

tsmethurst:
It works!

PR #4043, opened 2025-04-22
[feature] proof of work scraper deterrence

NyaaaWhatsUpDoc:
This adds a proof-of-work based scraper deterrence to GoToSocial's middleware stack on profile and status web pages. Heavily inspired by https://github.com/TecharoHQ/anubis, but massively stripped back for our own usecase.

Todo:
- ~~add configuration option so this is disabled by default~~
- ~~fix whatever weirdness is preventing this working with CSP (even in debug)~~
- ~~use our standard templating mechanism going through apiutil helper func~~
- probably some absurdly small performance improvements to be made in pooling re-used hex encode / hash encode buffers
- verify the cryptographic assumptions re: using a portion of token as challenge data
- ~~theme / make look nice??~~
- add a spinner
- ~~add entry in example configuration~~
- ~~add documentation~~

Verification page originally based on https://github.com/LucienV1/powtect

### DIFF
--- a/web/template/nollamas.tmpl
+++ b/web/template/nollamas.tmpl
@@ -23,9 +23,9 @@
         data-nollamas-challenge="{{ .challenge }}"
         data-nollamas-difficulty="{{ .difficulty }}"
     >
-        <p>One moment while we verify your connection...</p>
+        <p>One moment please, we are verifying you are not a scraper 😎</p>
         <noscript>
-            <p>Javascript must be enabled to verify your browser!</p>
+            <p>Javascript must be enabled to verify your browser 👿</p>
         </noscript>
     </section>
 </main>
--- a/web/template/nollamas.tmpl
+++ b/web/template/nollamas.tmpl
@@ -5,43 +5,6 @@
     <title>Verifying...</title>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <style>
-        @import url('https://fonts.googleapis.com/css2?family=IBM+Plex+Sans:wght@480&display=swap');
-        @media (prefers-color-scheme: light) {
-            :root {
-                --color-1: #f9fafb;
-                --color-2: #2563eb;
-            }
-        }
-        @media (prefers-color-scheme: dark) {
-            :root {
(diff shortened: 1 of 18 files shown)


Fastidious:
Why not use Anubis instead? Is there a true compelling reason to built this into GtS?

tsmethurst:
@Fastidious Ease of deployment :) It's a lot easier for people who aren't very familiar with networking small services together to just turn on a settings option, rather than deploying anubis (or whatever other thing) in front of GtS and handling the extra config of doing that. Also we only need a subset of anubis's functionality afaik. The alternative we were considering was putting this in a separate library and importing it into GtS, but we'll see.

NyaaaWhatsUpDoc:
also to follow-on from what @tsmethurst said, it's easier for us to get more granular about when we apply the middleware, e.g. if there's a valid OAuth token (and not just provided / unchecked in header)

NyaaaWhatsUpDoc:
ooooh a web linter failure, that's a new one for me :D

web/template/nollamas.tmpl
```diff
@@ -0,0 +1,32 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/ -}}
+
+{{- with . }}
+<main>
+    <section class="nollamas"
+        data-nollamas-challenge="{{ .challenge }}"
+        data-nollamas-difficulty="{{ .difficulty }}"
+    >
+        <p>One moment please, we are verifying you are not a scraper 😎</p>
```
tsmethurst, line 26: I thought about this a bit and I think we should use javascript to inject this `<p>` into the `nollamas` container so that the "please wait" message doesn't appear if javascript is not enabled, otherwise it gives confusing mixed messages. Happy to update the PR to do this for you if you want :)

NyaaaWhatsUpDoc: yeah go for it! i'm just writing some example config docs at the moment :)

daenney:
We've migrated to Codeberg. Your PR has been imported as: https://codeberg.org/superseriousbusiness/gotosocial/pulls/4043.

PR #4054, opened 2025-04-25
[chore] Move deps to `code.superseriousbusiness.org`

tsmethurst:
We now have module aliases for our codeberg stuff set up at `code.superseriousbusiness.org`, so this just points GtS towards those instead of pointing to codeberg directly. Full credit to @daenney for setting up the aliases <3


daenney:
:shipit:

daenney:
The build completed successfully for the tests to even be able to run, so it's not caused by the import path renames.

tsmethurst:
Ah yeah unrelated flaky test I think, I'm kicking the ci/cd box!

